Issue #1563, opened 2015-06-12
Layer Search doesn't accept mouse clicks


DesignSpaceBot:
Open a couple multi-layered docs, cmd/ctrl + f, click on search drop down, scroll down to a layer or doc and click on it with mouse.
Result: search is dismissed and layer/doc is not selected
workaround: use arrow keys and once target layer/doc has focus, hit Enter

-- @chadrolfs at [2015-06-11T22:26:52Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1563)

Issue #1547, opened 2015-06-12
Layer Search dialog should be dismissed on canvas clicks


DesignSpaceBot:
Added layers don't show up in Layer Search (while it's open)
new doc, select Rect tool, hit cmd+f or Layer > Search, drag out rects on canvas (or do the same with the Type or Ellipse tool)

Result: new layers are not populated in the list

Not sure what desired workflow should be here.  Should we not allow tools to work while Search is open?

-- @chadrolfs at [2015-06-09T20:24:39Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1547)

PR #1580, opened 2015-06-15
Search bar click handling

hamburge:
Addresses #1563, #1547

### DIFF
--- a/src/style/search/search-bar.less
+++ b/src/style/search/search-bar.less
@@ -50,7 +50,7 @@
     left: 0;
     right: 0;
     bottom: 0;
-    background-color: rgba(0,0,0,0);
+    background-color: rgba(0,0,0,0.01);
 }
 
 .search-bar__dialog svg {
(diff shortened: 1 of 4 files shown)


hamburge:
@iwehrman @volfied I narrowed down what was causing the problem in TextInput's `_commit()` It shouldn't be a problem to remove the call to the change handler because all other DataLists are live and are calling the change handler beforehand. 

I also fixed a little bug that I noticed when you opened the dialog and then hit Esc, so I added a check for that in TextInput as well.

src/style/search/search-bar.less
```diff
@@ -50,7 +50,7 @@
     left: 0;
     right: 0;
     bottom: 0;
-    background-color: rgba(0,0,0,0);
+    background-color: rgba(0,0,0,0.01);
```
volfied: Honestly I'm all for increasing the transparency so we make it clear to the user that they're in a modal state. With huge monitors, it's easy to miss the search bar in the middle while looking at the panels in the right.

volfied:
LGTM

Issue #1443, opened 2015-06-12
Background layer not always deleted after unlocking


DesignSpaceBot:
1. Create a new document with a background layer.
2. Add a rectangle.
3. Deselect all layers.
4. Click the background layers's lock icon.

Result in DS:
![ds](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

The truth from standard PS:
![ps](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

I suspect that the heuristic in `LayerStructure.prototype.addLayer` needs to be refined further.

-- @iwehrman at [2015-05-17T01:58:25Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1443)

volfied:
FBNC to @chadrolfs , can you verify the fix?

chadrolfs:
Looks fixed to me, CNR - closing

PR #1582, opened 2015-06-15
Unlock background layer: correct history states and layer replacement

mcilroyc:
This PR provides proper background layer unlocking wrt to both history store, *and* the "replace layer" logic.

1. The replace layer logic now handles the special case of unlock background layer by allowing an explicit layer ID parameter of that which is to be replaced.  The recent simplifications to replace layers based on selection state does not handle this situation gracefully otherwise.
2. The unlock background layer action no longer emits two separate history-changing flux events

addresses #1443

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -277,12 +277,14 @@ define(function (require, exports) {
      * Emit an ADD_LAYER event with the layer ID, descriptor, index, whether
      * it should be selected, and whether the existing layer should be replaced.
      *
+     * If `replace` is  unspecified, an existing single selected layer will only be replaced if it is an empty
+     * non-background layer.  If a number is specified, that layer ID will be replace.
+     * If false, no replacement will take place
+     *
      * @param {Document} document
      * @param {number|Array.<number>} layerSpec
      * @param {boolean=} selected Default is true
-     * @param {boolean=} replace Whether to replace the layer at the given index.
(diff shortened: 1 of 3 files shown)


src/js/actions/layers.js
```diff
@@ -892,7 +894,7 @@ define(function (require, exports) {
             .bind(this)
             .then(function (event) {
                 var layerID = event.layerID;
-                return this.transfer(addLayers, document, layerID, true, true);
+                return this.transfer(addLayers, document, layerID, true, layer.id);
```
volfied: addLayersCommand's JSDoc suggests that replace takes an index, yet you're passing the layer ID here.

src/js/models/layerstructure.js
```diff
@@ -713,28 +713,31 @@ define(function (require, exports, module) {
      * @param {Array.<object>} descriptors Photoshop layer descriptors
      * @param {boolean} selected Whether the new layer should be selected. If
      *  so, the existing selection is cleared.
-     * @param {boolean} replace Whether to replace the first existing layer model at
-     *  the given index
+     * @param {boolean= || number=} replace can be explicitly false, undefined, or a layer ID
```
volfied: I see the correct JSDoc here.

volfied:
Got one JSDoc nit, otherwise changes look good to me @mcilroyc

mcilroyc:
JSDoc fixed.  nice catch.  back to you @volfied

PR #1590, opened 2015-06-16
Search by layer ancestry and for recent documents in search dialog

hamburge:
Only displays the ancestry if the layer name is not unique, but you can always search using the ancestry (even if it isn't displayed).

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -71,8 +71,8 @@ define(function (require, exports, module) {
             }
 
             var idArray = id.split("_"),
-                type = idArray.length === 2 ? idArray[0] : "",
-                idInt = idArray.length === 2 ? parseInt(idArray[1]) : parseInt(id),
+                type = idArray.length > 0 ? idArray[0] : "",
+                idInt = idArray.length > 1 ? parseInt(idArray[1]) : parseInt(id),
                 flux = this.getFlux();
 
             switch (type) {
@@ -90,6 +90,16 @@ define(function (require, exports, module) {
                 if (selectedDoc) {
                     flux.actions.documents.selectDocument(selectedDoc);
                 }
+                break;
+            case "recent-doc":
+                var fileName = idArray[1];
+                
+                if (idArray.length > 2) {
+                    // There are underscores in the file name, so calling split separated the file name
+                    fileName = id.substring(id.indexOf("_") + 1);
+                }
+
+                flux.actions.documents.open(fileName);
                 break;
             }
             this.props.dismissDialog();
@@ -171,7 +181,7 @@ define(function (require, exports, module) {
                 if (layerItem.title === newInfo) {
                     layerItem.displayInfo = "";
                 } else {
-                    layerItem.displayInfo = shortenedPaths[index];
+                    layerItem.displayInfo = newInfo;
                 }
             });
 
@@ -217,14 +227,52 @@ define(function (require, exports, module) {
         },
 
         /**
+         * Make list of recent documents to be used as dropdown options
+         * 
+         * @return {Array.<Object>}
+         */
+        _getRecentDocOptions: function () {
+            var appStore = this.getFlux().store("application"),
+                recentFiles = appStore.getRecentFiles(),
+                recentDocMap = recentFiles.map(function (doc) {
+                    return {
+                        id: "recent-doc_" + doc,
+                        title: pathUtil.getShortestUniquePaths([doc])[0],
+                        type: "item",
+                        info: doc,
+                        displayInfo: doc
+                    };
+                });
+            
+            // Get shortest unique file path
+            var paths = collection.pluck(recentDocMap, "info"),
+                shortenedPaths = pathUtil.getShortestUniquePaths(paths);
+
+            recentDocMap.map(function (docItem, index) {
+                docItem.displayInfo = shortenedPaths[index];
+            });
+
+            var recentDocLabel = {
+                    id: "recent-doc_header",
+                    title: "Recent Documents",
+                    type: "header"
+                };
+
+            recentDocMap.unshift(recentDocLabel);
+
+            return recentDocMap;
+        },
+
+        /**
          * Make list of items and headers to be used as dropdown options
          * @return {Array.<Object>}
          */
         _getSelectOptions: function () {
             var layerOptions = this._getLayerOptions(),
-                docOptions = this._getCurrDocOptions();
+                currentDocOptions = this._getCurrDocOptions(),
+                recentDocOptions = this._getRecentDocOptions();
            
-            return layerOptions.concat(docOptions);
+            return layerOptions.concat(currentDocOptions).concat(recentDocOptions);
         },
 
         render: function () {
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -30,6 +30,9 @@ define(function (require, exports, module) {
         _ = require("lodash"),
         Datalist = require("jsx!js/jsx/shared/Datalist"),
         Immutable = require("immutable");
+
+    var pathUtil = require("js/util/path"),
+        collection = require("js/util/collection");
         
     var SearchBar = React.createClass({
         mixins: [FluxMixin],
@@ -138,7 +141,7 @@ define(function (require, exports, module) {
          * @return {Array.<Object>}
          */
         _getLayerOptions: function () {
-            // get list of layers
+            // Get list of layers
             var appStore = this.getFlux().store("application"),
                 document = appStore.getCurrentDocument(),
                 layers = document.layers.allVisibleReversed,
@@ -151,17 +154,32 @@ define(function (require, exports, module) {
                         id: "layer_" + layer.id.toString(),
                         title: layer.name,
                         info: ancestry,
+                        displayInfo: ancestry,
                         svgType: iconID,
                         type: "item"
                     };
                 }.bind(this)),
 
-                layerLabel = {
+                // Get shortest unique layer ancestry
+                ancestors = collection.pluck(layerMap, "info"),
+                shortenedPaths = pathUtil.getShortestUniquePaths(ancestors.toArray());
+
+            layerMap.map(function (layerItem, index) {
+                var newInfo = shortenedPaths[index];
+                
+                // Only show ancestry if other layers have the same name
+                if (layerItem.title === newInfo) {
+                    layerItem.displayInfo = "";
+                } else {
+                    layerItem.displayInfo = shortenedPaths[index];
+                }
+            });
+
+            var layerLabel = {
                     id: "layer_header",
                     title: "Layers",
                     type: "header"
                 },
-
                 layerOptions = layerMap.unshift(layerLabel);
 
             return layerOptions;
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -63,7 +63,6 @@ define(function (require, exports, module) {
          *
          * @param {string} id
          */
-
         _handleChange: function (id) {
             if (id === null) {
                 this.props.dismissDialog();
@@ -138,17 +137,16 @@ define(function (require, exports, module) {
             var layerTree = this.getFlux().store("application").getCurrentDocument().layers,
                 ancestors = layerTree.ancestors(layer),
                 ancestorNames = ancestors.first().name;
-
-            ancestors.skip(1).forEach(function (ancestor) {
-                ancestorNames += ("/" + ancestor.name);
-            });
-            return ancestorNames;
+                
+            return ancestors.skip(1).reduce(function (ancestors, ancestor) {
+                return ancestorNames += ("/" + ancestor.name);
+            }, ancestorNames);
         },
 
         /**
          * Make list of layers in the current document to be used as dropdown options
          * 
-         * @return {Array.<Object>}
+         * @return {Array.<object>}
          */
         _getLayerOptions: function () {
             // Get list of layers
@@ -172,9 +170,9 @@ define(function (require, exports, module) {
 
                 // Get shortest unique layer ancestry
                 ancestors = collection.pluck(layerMap, "info"),
-                shortenedPaths = pathUtil.getShortestUniquePaths(ancestors.toArray());
+                shortenedPaths = pathUtil.getShortestUniquePaths(ancestors).toJS();
 
-            layerMap.map(function (layerItem, index) {
+            layerMap.forEach(function (layerItem, index) {
                 var newInfo = shortenedPaths[index];
                 
                 // Only show ancestry if other layers have the same name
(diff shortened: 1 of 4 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -92,45 +95,135 @@ define(function (require, exports, module) {
...
+                        info: ancestry,
+                        displayInfo: ancestry,
+                        svgType: iconID,
+                        type: "item"
+                    };
+                }.bind(this)),
+
+                // Get shortest unique layer ancestry
+                ancestors = collection.pluck(layerMap, "info"),
+                shortenedPaths = pathUtil.getShortestUniquePaths(ancestors.toArray());
```
volfied: Nice re-use of existing util.

volfied:
I'm not a fan of showing the layer name, and then ancestry next to it. There is now TMI on the search list. It's also something we'd have to get rid of once we start showing layer previews etc. on the same `<li>`

iwehrman:
Calling our UX expert, @placegraphichere! Do you have any suggestions about how to present this additional information? Or are you not a fan of this? I am guilty of making this suggestion, originally, thinking that there would need to be some way to disambiguate layers, kind of like how full paths are shown as auxiliary information here:
alfred

iwehrman:
Assigning this to @placegraphichere, since this is blocked on him and he is the decider. @hamburge: if he doesn't respond, leave an appropriate gif.

hamburge:
Also wondering if it should search by ancestry (and if so, should it search by total ancestry path or just the shortest unique path)?

iwehrman:
Brackets' QuickOpen and Sublime Text 2's Goto Anything both seem to search the full paths of the files that are in the range of the search. That doesn't really answer the question, but.. hey.
quickopen

iwehrman:
In any case, it's time to set a gif fire in here to get @placegraphichere's attention :fire:

placegraphichere:
I think shortest is best as a first start. Any user input is meant to trim results to the smallest set possible.

I think things like a single string search should only match layers that directly apply to the layer name. If we start working on multi-string searches, then perhaps we need some more rules to govern it.

So for instance, if I searched "artboard1 button", should I get all matches for layers underneath "artboard1" that also match "button"? How would we know which string is the parent or child of the search result? For instance, if we had a parent called "button" with a child "artboard1"? Or does the user intend that their result match both "artboard1" and "button"? I think that could be solved in time, but perhaps that is too crazy to deal with? 

For names and types, we should probably just use && matching. partial_layer_name && type = match.

I'm not sure if that helps what you are asking or not.

hamburge:
Right now how I have it (in this branch) is that you have to search for the words consecutively with either a space or a '/' to search by layer ancestry. So in the case of artboard1>button, you could get it by typing "artboard1 button" or "artboard1/button" but it would also show up in the list if you type "button". I'll change it to just search for the shortest path instead of the full ancestry like it does now.

I'm thinking it would all make better sense if we did some sort of sorting of the options so that the "best" or most likely matches are on top. Right now the order is just the ones that match in order of the layer panel's list.

The other part of the question was if the layer ancestry should show up visually in the dropdown.

iwehrman:
@hamburge have you made the changes here that you wanted to make? I think we should get this merged and continue the discussion with the changes in master. I'm going to do a quick low-level review, but let me know if you're still interested in making more changes before merging.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -68,8 +71,8 @@ define(function (require, exports, module) {
             }
 
             var idArray = id.split("_"),
-                type = idArray.length === 2 ? idArray[0] : "",
-                idInt = idArray.length === 2 ? parseInt(idArray[1]) : parseInt(id),
+                type = idArray.length > 0 ? idArray[0] : "",
+                idInt = idArray.length > 1 ? parseInt(idArray[1]) : parseInt(id),
```
iwehrman: I gather that, in some cases, we're encoding an array of integers as a string? We should make whatever changes are necessary to `Datalist` to not do that. In particular, it would be fine if the `Datalist` change handler was applied to either a single number or an Array of numbers. It would be better for the handler to do this test than to bother with encoding and decoding an array of numbers as a string.

And, although this will be obviated by the above change, we prefer to use `parseNumber` from `js/util/math` instead of `window.parseInt` because the latter fails in weird ways. In particular, if the former fails it just returns null, whereas the latter returns `NaN`. (Fun JavaScript fact: `NaN !== NaN`)

hamburge:
Yes, I think I've made all the changes I wanted to do.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -68,8 +71,8 @@ define(function (require, exports, module) {
             }
 
             var idArray = id.split("_"),
-                type = idArray.length === 2 ? idArray[0] : "",
```
iwehrman: Also, there's an extra new-line between the JSDoc annotation above this function and the function declaration ^^^

src/js/jsx/search/SearchBar.jsx
iwehrman: (A number or array of numbers might not actually be appropriate above, but then we should still figure out a data structure that is better than a string encoding.)

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
 
+        /**
+         * Get the class name for the layer face icon for the layer
+         *
+         * @private
+         * @param {Layer} layer
+         * @return {string}
+         */
+        _getSVGInfo: function (layer) {
+            var iconID = "layer-";
```
iwehrman: Is this function basically copied from somewhere within the layers panel? If so, should we move both copies to a utility function somewhere else? Or maybe even have a higher level abstraction like "the SVGIcon for a particular layer"?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+         * @private
+         * @param {Layer} layer
+         * @return {string}
+         */
+        _formatLayerAncestry: function (layer) {
+            var layerTree = this.getFlux().store("application").getCurrentDocument().layers,
+                ancestors = layerTree.ancestors(layer),
+                ancestorNames = ancestors.first().name;
+
+            ancestors.skip(1).forEach(function (ancestor) {
```
iwehrman: It would be slightly nicer to use `reduce` here.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+                        svgType: iconID,
+                        type: "item"
+                    };
+                }.bind(this)),
+
+                // Get shortest unique layer ancestry
+                ancestors = collection.pluck(layerMap, "info"),
+                shortenedPaths = pathUtil.getShortestUniquePaths(ancestors.toArray());
+
+            layerMap.map(function (layerItem, index) {
```
iwehrman: Use `forEach` for side effects instead of `map`. Use `map` if you're mapping from one array/iterable to another.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+                if (layerItem.title === newInfo) {
+                    layerItem.displayInfo = "";
+                } else {
+                    layerItem.displayInfo = newInfo;
+                }
+            });
+
+            var layerLabel = {
+                    id: "layer_header",
+                    title: "Layers",
```
iwehrman: We can do this is in a separate PR, but strings like this should be localized. It's not hard, but we can talk about how to do this separately.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
 
-            // get list of currently open documents
-            var docStore = this.getFlux().store("document"),
+            return layerOptions;
+        },
+
+        /**
+         * Make list of currently open documents to be used as dropdown options
+         * 
+         * @return {Array.<Object>}
```
iwehrman: This should be `@return {Array.<object>}`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+         */
+        _getCurrDocOptions: function () {
+            var appStore = this.getFlux().store("application"),
+                docStore = this.getFlux().store("document"),
+                document = appStore.getCurrentDocument(),
                 openDocs = Immutable.fromJS(appStore.getOpenDocumentIDs()).filterNot(function (doc) {
```
iwehrman: The open document IDs from the application store are now an immutable list, so this immutable conversion is unnecessary.

iwehrman: See here: https://github.com/adobe-photoshop/spaces-design/blob/955e1995d80c81c0c8591f3dcbb8146303f7b219/src/js/stores/application.js#L119-L126

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
                 docOptions = docMap.unshift(docLabel);
 
-            return layerOptions.concat(docOptions);
+            return docOptions;
+        },
+
+        /**
+         * Make list of recent documents to be used as dropdown options
+         * 
+         * @return {Array.<Object>}
```
iwehrman: This should be `Array.<object>`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+                        info: doc,
+                        displayInfo: doc
+                    };
+                });
+            
+            // Get shortest unique file path
+            var paths = collection.pluck(recentDocMap, "info"),
+                shortenedPaths = pathUtil.getShortestUniquePaths(paths);
+
+            recentDocMap.map(function (docItem, index) {
```
iwehrman: Use `.forEach` for side effects instead of `.map`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -88,50 +91,188 @@ define(function (require, exports, module) {
...
+                docItem.displayInfo = shortenedPaths[index];
+            });
+
+            var recentDocLabel = {
+                    id: "recent-doc_header",
```
iwehrman: One-too-many tab stops in this object literal.

src/js/jsx/search/SearchBar.jsx
iwehrman: Actually, maybe this logic should just be pushed into the first `.map` above?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -306,8 +306,14 @@ define(function (require, exports, module) {
...
-                    return option.title.toLowerCase().indexOf(searchableFilter) > -1 &&
-                        option.hidden !== true;
+
+                    // Search for title and if option has info, search for that as well, with and without '/' characters
+                    var info = option.displayInfo ? option.displayInfo.toLowerCase() : "",
+                        searchableInfo = info.concat(info.replace(/\//g, " ")),
+                        searchableTitle = option.title.toLowerCase();
+
+                    return (searchableTitle.indexOf(searchableFilter) > -1 ||
+                        searchableInfo.indexOf(searchableFilter) > -1) && option.hidden !== true;
```
iwehrman: Nit: this can just be `!option.hidden`

src/js/jsx/shared/Select.jsx
```diff
@@ -47,12 +47,13 @@ define(function (require, exports, module) {
             value: React.PropTypes.shape({
                 id: React.PropTypes.string.isRequired,
                 title: React.PropTypes.string.isRequired,
-                style: React.PropTypes.object
+                info: React.PropTypes.string,
+                style: React.PropTypes.object,
+                svgType: React.PropTypes.string
```
iwehrman: Re: my comment above about a utility function for icons, let's keep thinking about whether this is the right way to expose that functionality. Other options: 1) parametrize values with an optional `icon` ReactElement; 2) don't parametrize them at all, but just include this as part of the `style`.

iwehrman: I actually wonder if all these values shouldn't just be ReactElements...

src/style/shared/select.less
```diff
@@ -42,3 +42,9 @@
     border-top: 1px solid @warm-gray;
     margin: 0 0.4rem;
 }
+
+.select__option__info {
+    margin-left: 1rem;
+    color: @warm-gray;
+    font-size: 12px;
```
iwehrman: Although there are some exceptions, in general we should never style UI with pixel values. Instead, we want to always use relative units in order to make the UI completely scalable. (Read this for a concise overview: http://snook.ca/archives/html_and_css/font-size-with-rem). Instead, we want something like this: https://github.com/adobe-photoshop/spaces-design/blob/955e1995d80c81c0c8591f3dcbb8146303f7b219/src/style/sections/pages/face.less#L104-L106

iwehrman:
OK, I'm done with my review pass. You don't necessarily need to address the more architectural questions I raised before we merge this. But, ideally, take a look at all my comments and for each one, let me know if a) my comment is crazy and wrong and you aren't going to change it; b) you already fixed it; or b) you will fix it later.

Sorry for the long delay on this PR!

src/js/jsx/search/SearchBar.jsx
hamburge: I can change it to `.forEach` but I don't think it can go into the mapping above because it needs to first look have the info of all the layers to find the shortest unique path.

src/js/jsx/search/SearchBar.jsx
hamburge: Yes, moving this to a utility function would probably be a good idea. I messed with this more in the branch `allison/search_filters` though when putting the svgs next to the text box, so it would probably be easier to change it there.

src/js/jsx/shared/Select.jsx
hamburge: I don't think it can be added as part of `style` because I think that's just CSS being passed in.

Before I got to this, it had `id`, `value`, and `style` grouped under `value`. I'm not sure what makes those values different from the other 3 that an `Option` has (`selected`, `next`, and `prev`), so I don't know what the point would be of making them into a separate ReactElement.

src/js/jsx/search/SearchBar.jsx
hamburge: Ok, not entirely sure what that means but sounds good!

iwehrman: Sorry for the jargon, what I meant there was that the string "Layers" eventually needs to be specified in a way that will make it possible for it to be translated to other languages (i.e., for use with other locales). We do that in two parts 1) user-visible strings are added to a dictionary here https://github.com/adobe-photoshop/spaces-design/blob/955e1995d80c81c0c8591f3dcbb8146303f7b219/src/nls/root/strings.js#L354 and 2) we refer to those strings using keys into the dictionary, as in here: https://github.com/adobe-photoshop/spaces-design/blob/955e1995d80c81c0c8591f3dcbb8146303f7b219/src/js/jsx/sections/pages/PagesPanel.jsx#L493

hamburge:
@iwehrman I fixed the less invasive things to be fixed and commented on what I didn't fix yet. I think it makes more sense to work on the ID parsing/SVG refactoring off of my other branch `allison/search_filters` since it is a branch off of this one and touches a lot of the same code.

src/js/jsx/search/SearchBar.jsx
hamburge: I know this doesn't fully address your concern, but (for now) I changed this so that all IDs are now "<type>_<number>" whereas before the recent documents would have a file path string instead of a number, so at least now it's a bit more consistent in format. The Select component requires a string ID, so the other ideas about fixing this would require a lot of changes in shared components.

iwehrman:
Looks great, thanks!

PR #1591, opened 2015-06-16
quick fix for reoder bug

pineapplespatula:


mcilroyc:
this doesn't seem to work?  Either way I think it is preferred to use `transfer` instead of calling a *Command directly, and this is addressed in my forthcoming PR.

mcilroyc:
closer.  still doesn't work, nor does it grunt correctly.

pineapplespatula:
its working for me. do you still get the same error in this branch?

mcilroyc:
I think you just needed: `this.transfer(getLayerOrderCommand, document)`  ... but anyway see my PR 1593

Issue #1489, opened 2015-06-12
Deleting artboard or group causes an uncaught error (postcondition)


DesignSpaceBot:
![screen shot 2015-05-29 at 2 39 16 pm](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

Select Group 1 and ungroup, then delete Rectangle 1.

Verification of postcondition 1 failed for action layers.deleteSelected
fluxcontroller.js:382 Action layers.deleteSelected failed: Error: Incorrect selected layer count: 0 instead of 1

![screen shot 2015-05-29 at 2 36 02 pm](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @ktaki at [2015-05-29T21:41:28Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1489)

ktaki:
Verified the fix.

PR #1593, opened 2015-06-16
Layer actions fixes: post-deletion selection reset, and minor resetSelection refactoring.

mcilroyc:
Addresses #1489 

Two commits:

1. After deleting a layer, we now query photoshop for the current selection state.  We previously were making no attempt to update the selection state, and that was bad
1. Attempt to make more consistent use of "resetSelection" within the layer actions.  Note that this affects three "other" code paths: reordering layers, deleting the last anchor of a shape, certain types of "ungrouping".

Note: this "introduces" a post condition error when doing some ungrouping (such as ungrouping an empty group) but I think it is a separate, existing issue that is just being surfaced more correctly now.  I didn't want to hold up this PR for that fix.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -316,15 +316,34 @@ define(function (require, exports) {
             });
     };
 
+
+    /**
+     * Get a list of selected layer indexes from photoshop, based on the provided document
+     *
+     * @private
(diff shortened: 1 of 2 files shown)


volfied:
That's a PR I don't mind merging

PR #1596, opened 2015-06-16
Set cloaking overlay on window resize

volfied:
This is an API Jesper developed for us, which allows us to set rectangles so certain events to immediately clear out regions of the HTML page. We use this during `scroll` events that come from Photoshop so when user starts panning / scrolling, we can immediately erase our layer bounds.

It's a big improvement to one of the oldest issues.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -21,8 +21,6 @@
  * 
  */
 
-/* global _spaces */
-
 define(function (require, exports) {
     "use strict";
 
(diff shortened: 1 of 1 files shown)


mcilroyc:
This looks fine to my eye, except that the direct `_spaces` reference seems like a slippery slope.  

I'm having a hard time detecting the improvement, so I'll defer to @iwehrman and @chadrolfs who you called out on slack as being particularly interested in this.

volfied:
I don't want to expose rest of the `_spaces.window` functions, but I will expose `setOverlayCloaking` inside adapter/ps/ui.

volfied:
@mcilroyc Added a new PR so we don't sully our code with `_spaces`

iwehrman:
I heartily approve. I might the only one who notices, but I think this is a definite improvement.

Issue #1227, opened 2015-06-12
React transition warning in first-launch dialog


DesignSpaceBot:
I've seen this a few times when skipping through the first-launch slides quickly. It doesn't happen every time.
```
transition(): tried to perform an animation without an animationend or transitionend event
  after timeout (5000ms). You should either disable this transition in JS or add a CSS 
  animation/transition.
```
I didn't notice any particular problems as a result of the warning. 

-- @iwehrman at [2015-04-18T22:53:51Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1227)

iwehrman:
_dusts off hands_

PR #1601, opened 2015-06-17
Jecohen/1227

pineapplespatula:
fix for issue #1227. we remove transitions, but we decided that is ok for now


designedbyscience:
Looks good

Issue #108, opened 2015-06-12
Don't change caret while stepping in NumberInputs


DesignSpaceBot:
@iwehrman Thanks for the reference! I've also fixed the _playground issue.

-- @volfied at [2014-11-05T23:22:07Z](https://github.com/adobe-photoshop/spaces-design-archive/pull/108)

Issue #1367, opened 2015-06-12
document.layers.byIndex(0) refers to the top layer


DesignSpaceBot:
Given a Document model `document` with a background, `document.layers.byIndex(0)` is the top-most layer, but `document.layers.byIndex(1)` is the background. This is extremely confusing. It results from the fact that, for an `Immutable.IndexedIterable` `list`, `list.get(-N)` is equivalent to `list.reverse().get(N)`.

-- @iwehrman at [2015-04-30T18:01:20Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1367)

PR #1603, opened 2015-06-17
Use a multiget reference to load a document's layer descriptors

iwehrman:
Relies on adobe-photoshop/spaces-design#108, which adds a `getPropertiesRange` call to efficiently get a list of properties of a range of references (e.g., all layer references on a document). This adds a new private helper function, `layers._getLayersForDocumentRef` which uses `getPropertiesRange` to more efficiently get all the layer descriptor objects when loading documents (by, e.g., `initActiveDocument` and `initInactiveDocuments`).  This provides a significant performance improvement at document-load time for large documents. The mega-artboards document previously spent ~7s waiting on Photoshop for these descriptors and now only waits for ~4s. 

In a follow-up PR, I'll apply similar new functionality in `batchPlay` to the existing `_getLayersByRef` function to go from O(mn) play commands, for m layers with n properties, to O(m).

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -157,20 +157,14 @@ define(function (require, exports) {
      * @return {Promise.<{document: object, layers: Array.<object>}>}
      */
     var _getLayersForDocument = function (doc) {
-        var layerCount = doc.numberOfLayers,
-            startIndex = (doc.hasBackgroundLayer ? 0 : 1),
-            layerRefs = Immutable.Range(layerCount, startIndex - 1, -1).map(function (i) {
-                return [
-                    documentLib.referenceBy.id(doc.documentID),
-                    layerLib.referenceBy.index(i)
-                ];
-            });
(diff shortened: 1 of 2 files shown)


src/js/actions/documents.js
```diff
@@ -157,20 +157,14 @@ define(function (require, exports) {
...
+        var docRef = documentLib.referenceBy.id(doc.documentID),
+            startIndex = (doc.hasBackgroundLayer ? 0 : 1);
 
-        return layerActions._getLayersByRef(layerRefs)
+        return layerActions._getLayersForDocumentRef(docRef, startIndex)
             .then(function (layers) {
                 return {
                     document: doc,
-                    layers: layers
+                    layers: layers.reverse()
```
iwehrman: Strangely, callers of this function seem to expect layers to be in reverse (descending) order. That seems wrong. Eventually we should push this `reverse` call out further to the callers. Possibly in conjunction with fixing the layer indexing weirdness noted in #1367.

Issue #1564, opened 2015-06-12
Shapes sometimes switch to grey default when switching to shape tool


DesignSpaceBot:
Open a doc with a shape that is not default fill/stroke color, select it, hit shortcut for shape tool (R, E, P)

Result: shape turns grey

-- @chadrolfs at [2015-06-11T22:36:45Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1564)

mcilroyc:
fbnc @chadrolfs

PR #1606, opened 2015-06-17
fix shape defaults to not overwrite selected shape properties

pineapplespatula:
fix for issue #1564

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -28,6 +28,8 @@ define(function (require, exports) {
 
     var descriptor = require("adapter/ps/descriptor"),
         toolLib = require("adapter/lib/tool"),
+        layerLib = require("adapter/lib/layer"),
+        documentLib = require("adapter/lib/document"),
         adapterOS = require("adapter/os"),
         adapterPS = require("adapter/ps");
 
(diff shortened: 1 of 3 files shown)


src/js/actions/tools.js
```diff
@@ -36,6 +38,41 @@ define(function (require, exports) {
...
+    var installShapeDefaultsCommand = function (tool, strokeColor, strokeSize, strokeOpactiy, fillColor) {
+        var document = this.flux.store("application").getCurrentDocument();
+
+        var defaultObj = toolLib.defaultShapeTool(tool, strokeColor, strokeSize, strokeOpactiy, fillColor);
+        // If document doesn't exist, or is a flat document
+        if (!document || document.unsupported || document.layers.all.size === 1 &&
+            document.layers.all.first().isBackground) {
+            return descriptor.PlayObject(toolLib.resetShapeTool());
+        }
+        var layerSpec = document.layers.allSelected.toList();
```
mcilroyc: nit: if you're going to short-circuit above, then might as well move `defaultObj` dow to here.  In general, this function seems to to be heavy on individual `var` statements, which differs from our usual pattern.

src/js/tools/ellipse.js
```diff
@@ -57,6 +60,7 @@ define(function (require, exports, module) {
             }
         };
 
+
```
mcilroyc: :space_invader:

pineapplespatula:
@mcilroyc

src/js/actions/tools.js
```diff
@@ -338,8 +375,15 @@ define(function (require, exports) {
 
     var resetSuperselect = {
         command: resetSuperselectCommand,
-        reads: [locks.JS_APP, locks.JS_TOOL, locks.JS_SHORTCUT],
-        writes: [locks.PS_APP, locks.JS_POLICY, locks.PS_TOOL, locks.JS_TOOL, locks.JS_SHORTCUT]
+        reads: [locks.JS_APP, locks.JS_TOOL, locks.JS_SHORTCUT, locks.PS_DOC, locks.JS_DOC],
+        writes: [locks.PS_APP, locks.JS_POLICY, locks.PS_TOOL, locks.JS_TOOL, locks.JS_SHORTCUT,
+            locks.PS_DOC, locks.JS_DOC]
```
mcilroyc: note to self: sweet glory, this is a lot of locks.  if all these write locks are really needed, then maybe that means this action has too much going on.  Or maybe this just really is the intersection of a lot of moving parts.  cc @iwehrman

src/js/actions/tools.js
```diff
@@ -36,6 +38,40 @@ define(function (require, exports) {
...
+            layerRef = layerSpec
+                .map(function (layer) {
+                    return layerLib.referenceBy.id(layer.id);
+                })
+                .unshift(documentLib.referenceBy.id(document.id))
+                .toArray();
+
+        var selectObj = layerLib.select(layerRef, false);
+
+        return descriptor.batchPlayObjects([layerLib.deselectAll(), defaultObj, selectObj]);
```
mcilroyc: Idle comment, in case you have to shove more selection logic here in the near future ... 
since this is basically another flavor of the "selection dance", it might make sense to factor this previous few lines of code into `util/layeractions.js`.

mcilroyc:
I'm mildly worried that there by dragons in the edge cases of this new selection dance ... but I can't actually come up with anything.  merging!

Issue #1477, opened 2015-06-12
Pasted layers does not retain selection


DesignSpaceBot:
Open a multi-layered doc, select 2+ layers, cmd/ctrl + c, cmd/ctrl + n, cmd/ctrl + v

Result: only one layer is selected
Expected: layers that are pasted should be selected.

-- @chadrolfs at [2015-05-26T22:08:45Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1477)

mcilroyc:
For the record, it seems that we have to implement this as a two-stage process (which is lame):  Duplicate the layers, *and then* select them.  This needs to be handled delicately such that the history changing event knows about the new selection and can capture the state correctly.

mcilroyc:
This is resolved now.  @chadrolfs note that you will see a harmless console warning in this situation.  We hope to fix that more properly in the future.

chadrolfs:
re-verified bug still occurs with latest jenkins and master (mac dev-381, win dev-379). master at bae8eb5

pineapplespatula:
i see the fix go into master at 4a6ac5e9583aa78fd3e96cdf53d3e29952832be1

chadrolfs:
my bad, i was a `git pull` behind.

PR #1611, opened 2015-06-18
add selection after duplication

pineapplespatula:
Cory, if you have time can you take a look at this validation error and see if you have any thoughts

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1374,13 +1374,13 @@ define(function (require, exports) {
                         return allLayerIDs.concat(layerIDs);
                     }, []),
                     allLayers = allLayerIDs.map(function (currentVal) {
-                        return { id: currentVal }
+                        return { id: currentVal };
                     });
 
                 return this.transfer(addLayers, document, allLayerIDs, undefined, false)
                     .bind(this)
-                    .then( function () {
-                        return this.transfer(selectLayer, document,  Immutable.List(allLayers));
+                    .then(function () {
+                        return this.transfer(select, document, Immutable.List(allLayers));
                     });
             });
     };
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1372,9 +1372,16 @@ define(function (require, exports) {
                 var allLayerIDs = collection.pluck(results, "ID")
                     .reduce(function (allLayerIDs, layerIDs) {
                         return allLayerIDs.concat(layerIDs);
-                    }, []);
+                    }, []),
+                    allLayers = allLayerIDs.map(function (currentVal) {
+                        return { id: currentVal }
+                    });
 
-                return this.transfer(addLayers, document, allLayerIDs, undefined, false);
+                return this.transfer(addLayers, document, allLayerIDs, undefined, false)
+                    .bind(this)
+                    .then( function () {
+                        return this.transfer(selectLayer, document,  Immutable.List(allLayers));
+                    });
             });
     };
     duplicate.reads = [locks.PS_DOC, locks.JS_DOC];


mcilroyc:
Welp, I can't actually make the validation error occur.  Can you give me more detailed repro steps?  

Either way, I *think* your implementation is not "history safe", so... this may be a slightly bigger can of worms than you thought.

Perhaps more importantly... maybe this is a core bug?  I mean, after the paste we are correctly representing photoshop's selection state.  So, maybe core should select both layers?  Or maybe we can do something upstream that will cause it to select both layers?  The last resort should be doing an explicit "post paste" selection command.  But, if that is necessary, I can help you do this in a history safe way.

volfied:
@pineapplespatula Is this fix for #1477? Can you please clarify the review only flag, or remove it and mark it ready otherwise?

pineapplespatula:
This change is not history safe. there is currently an investigation as to if we can do this in a history safe way. or if core changes are needed

iwehrman:
What do the post-validation errors have to do with the history state? I thought the history was oblivious to layer selection.

mcilroyc:
@pineapplespatula - upon further review, your solution does not cause any noticeable issues except a warning from the history system.  When you do the "post selection" it is not history safe (as I mentioned earlier) because it is updating the selection **after** we've already captured the history state.  That's why you get  a warning message in the console: it recognizes that the document model is not 100% consistent.  BUT, selection state is a magical property that is actually blown away during undo/redo anyway, so the warning can be ignored in this case.  Depending on how hard it is to make this post-selection truly history safe (as we discussed on slack), it might be OK to just merge this as-is and capture a low-pri bug to get rid of the warning.   Hopefully that makes some sense.

pineapplespatula:
That sounds like a good plan, Because we have to wait for the layer adding to finish before calling select. adding the correct state to our model at the add step has proved difficult ( since we're out of sync with PS until the selection action is played) 

can you go ahead an merge this PR, and when you do i'll write up an issue about the warning?

src/js/actions/layers.js
```diff
@@ -1296,9 +1296,16 @@ define(function (require, exports) {
                 var allLayerIDs = collection.pluck(results, "ID")
                     .reduce(function (allLayerIDs, layerIDs) {
                         return allLayerIDs.concat(layerIDs);
-                    }, []);
+                    }, []),
+                    allLayers = allLayerIDs.map(function (currentVal) {
+                        return { id: currentVal }
```
mcilroyc: perhaps I'm splitting hairs, but you are creating this list of objects (with a single "id" property) even though the selectLayers command is expecting `Layer|Immutable.Iterable.<Layer>`

src/js/actions/layers.js
```diff
@@ -1296,9 +1296,16 @@ define(function (require, exports) {
...
+                    });
 
-                return this.transfer(addLayers, document, allLayerIDs, undefined, false);
+                return this.transfer(addLayers, document, allLayerIDs, undefined, false)
+                    .bind(this)
+                    .then( function () {
+                        return this.transfer(selectLayer, document,  Immutable.List(allLayers));
```
mcilroyc: best rebase first, since AFAICT this action has since changed names (now: 'select')

src/js/actions/layers.js
pineapplespatula: That is true, sadly the "result" that we have here is not directly compatible to `Layer|Immutable.Iterable.<Layer>` so i made a structure that matches Layer for the purposes of select. If you think it's important i could make a more fully formed layer structure

mcilroyc:
ok!  don't forget to log a cleanup task for this warning we get now.  I guess.

src/js/actions/layers.js
```diff
@@ -1372,9 +1372,16 @@ define(function (require, exports) {
                 var allLayerIDs = collection.pluck(results, "ID")
                     .reduce(function (allLayerIDs, layerIDs) {
                         return allLayerIDs.concat(layerIDs);
-                    }, []);
+                    }, []),
+                    allLayers = allLayerIDs.map(function (currentVal) {
+                        return { id: currentVal };
```
iwehrman: It would be much better to just fetch the layer for the given ID. Not only would that be less surprising to the next person who reads this, but it would also save an object allocation per ID.

Issue #1614, opened 2015-06-18
Moving a layer from an artboard to another in layers panel will cause a separation of bounding box


ktaki:
Create two artboards. Add a rectangle to Artboard 1. In layers panel, drag the rectangle to Artboard 2. The bounding box remains in Artboard 1 while the shape is drawn in Artboard 2.

![screen shot 2015-06-18 at 1 12 12 pm](https://cloud.githubusercontent.com/assets/2029597/8241126/58f809e2-15bc-11e5-90ce-bd29c8cdba05.png)

chadrolfs:
regress against PR #1617 once merged.

chadrolfs:
#1617 doesn't fix this.

volfied:
This is related to auto-position, we can get away with calling reset bounds on layers after reorder. If you want to be fancy and optimized, layers will be placed in the same x,y location in relation to the top left of the artboard.

pineapplespatula:
can you provide new steps for this. as it is written i dont understand the bug

volfied:
Re-order of the layer does not update it's bounds, even though auto-position causes it to be moved. 

1 - Create two artboards
2 - Draw a rectangle in the selected artboard
3 - Select the rectangle, and keep track of the X/Y location
4 - Re-order the rectangle to the other artboard through layers panel
5 - Check the bounds again, they're not updated, This breaks our border policies for on canvas transform.

Solution is to call resetBounds on the re-ordered layers.

chadrolfs:
Look at Ian's video in #1690 as well.

iwehrman:
FBNC to @chadrolfs.

Issue #1587, opened 2015-06-16
Auto-nesting: Shapes created outside of artboards won't be nested when they are moved on to artboards


ktaki:
The layers panel is not updated till a shape is moved from another artboard or new artboard is created.

Create a new artboard (Artboard 1)
Create a rectangle outside of the Artboard 1.
Move the shape on to the Artboard 1.
Notice that the Rectangle is not auto-nested in the Layers panel.
Create another rectangle on the Artboard 1.
Notice that the Rectangle is also not auto-nested in the Layers panel.

Create another artboard (Artboard 2) in the same doc.
Now the layers panel is updated and has two rectangle under the Artboard 1.

Continue:
Create a rectangle outside of the two artboards.
Move the rectangle on to Artboard 2.
Notice that the rectangle is not auto-nested in the Layers panel.
Move a rectangle from Artboard 1 to Artboard 2.
Now the layers panel is updated and has two rectangle under the Artboard 2.

pineapplespatula:
we get move and movetoArtboard in oposite orders depending on if the shape is outside an artboard to start. Our history model doesn't support this so i'm throwing it over to cory to extend our history model

mcilroyc:
I created a watson bug for this: 4011496  assigned to @isonno

mcilroyc:
I believe this has been addressed both by the above referenced watson bug, and by PR #1696.

Issue #1583, opened 2015-06-16
Shapes don't snap to document edge.


CristianMoisei:
In the regular version a vector shape or an element you are transforming snaps to the document edge even after having been created.

https://www.dropbox.com/s/vnwp00y0qj95x89/clip.mov?dl=0

iwehrman:
Thanks for the report! This is a known issue that, actually, @volfied is working on right now. Assigning to him so that he can reference this in his PR and close the issue once the fix lands.

PR #1617, opened 2015-06-18
Switch back to Photoshop transform controls

volfied:
Main change is getting rid of our HTML rotate / resize controls. Pour one out for them :beer:

This change brings couple hacks with it.

 - We now have a "border policy". It's not an immigration thing, but alongside transform controls we have to let Photoshop have the mouse events and inside the selection we need to handle clicks so we can select layers inside/behind a selection.

 - There is a new policy in town, and it's name is `ALPHA_PROPAGATE_WITH_NOTIFY`. This policy acts like alpha propagate, sending mouse events to either Photoshop or us, depending on if we have a non-transparent pixel underneath the cursor. But, if Photoshop gets the event, we get an `external_mouse_move` event from Photoshop OS notifiers. We use these events to let PS drive all cursors and mouse moves, but still be able to highlight our own superselect overlays.

- I stole the "move" handler from superselect dragging, and installed it for `transform` events, which happen when a selection gets resized/rotated through Photoshop.

This change is also quite likely to bring a lot of bugs with it, @chadrolfs @ktaki, if you guys can do some runs before merging and report back on the PR with bugs, it would help us land it smoothly.

Transform Scrim, good night sweet prince. You were my favorite feature.

----

## BUGS

- [x] Drag moving an artboard switches to artboard tool.
- [x] Show artboard borders (Required a PS change - 1028016)
- [x] Unable to draw shape on 2nd AB if 1st AB has focus in layer in layers panel (PS steals the mouse events from us while we’re on the panel)
- [ ] If you select a layer from the panel, transform controls don't redraw right away, this is a core issue. (#4012858)
- [x] Starting a resize on artboards send us no event, so we're not able to hide the plus signs on time. Another core issue (Fixed by CL 1028074)


volfied:
This PR requires latest Playground-dev build!

src/js/actions/transform.js
```diff
@@ -1045,6 +1045,26 @@ define(function (require, exports) {
         return Promise.join(positionPromise, dispatchPromise);
     };
 
+    var beforeStartupCommand = function () {
+        var _transformHandler = function () {
```
mcilroyc: This is just a drive-by comment, but since I've been looking at the (former) "move" handler: JC made some artboard nesting changes, which I recently tweaked to not fail on option-drag.  Is that logic obsoleted in the new world, or do you  need to implement it in _transformHandler now?  See: https://github.com/adobe-photoshop/spaces-design/pull/1618/files#diff-1bd8d606e837ab0e4ebf7c4eb81890aaR636

Two things for sure: 

1. issue 1613 is not fixed in this branch
1. Dragging across artboards causes history store re-init

iwehrman:
Does this address #1583?

chadrolfs:
If I select an AB and then move my mouse away, I now have the AB tool.

chadrolfs:
@iwehrman yes, it addresses #1583

chadrolfs:
looks like it addresses #1587 as well.

volfied:
#1614 will be fixed by this.

volfied:
@chadrolfs Can you please see if artboard tool change is still happening? 845d84d should address it.

chadrolfs:
checked, looks good.

chadrolfs:
you can resize AB on canvas but don't see the transform controls until after you start resizing which seems a little odd from a user perspective but it works.

iwehrman:
Bug:

1. position mouse away from a corner so that you see the rotation cursor. 
2. hold shift, and attempt to rotate

Result: marquee select starts instead of rotation.

src/js/actions/tools.js
```diff
@@ -31,12 +31,17 @@ define(function (require, exports) {
...
 
     var events = require("../events"),
         locks = require("js/locks"),
         policy = require("./policy"),
-        shortcuts = require("./shortcuts");
+        shortcuts = require("./shortcuts"),
+        EventPolicy = require("js/models/eventpolicy"),
+        PointerEventPolicy = EventPolicy.PointerEventPolicy;
+
+    var _currentTransformPolicyID = null;
```
iwehrman: JSDoc me.

chadrolfs:
Yeah, that's different from standard PS.  The workaround is hold down Shift after you start dragging.

src/js/actions/tools.js
```diff
@@ -200,6 +205,89 @@ define(function (require, exports) {
...
+    var resetBorderPoliciesCommand = function () {
+        var toolStore = this.flux.store("tool"),
+            appStore = this.flux.store("application"),
+            uiStore = this.flux.store("ui"),
+            currentDocument = appStore.getCurrentDocument(),
+            currentTool = toolStore.getCurrentTool();
+
+        // We only want to reset superselect tool
+        if (!currentDocument || !currentTool || currentTool.id !== "newSelect") {
+            return Promise.resolve();
```
iwehrman: Only vaguely know what's going on here, but: does this handle the case where we're in superselect-type-edit mode?

src/js/actions/tools.js
```diff
@@ -200,6 +205,89 @@ define(function (require, exports) {
...
+
+        var psSelectionTL = uiStore.transformCanvasToWindow(
+                selection.left, selection.top
+            ),
+            psSelectionBR = uiStore.transformCanvasToWindow(
+                selection.right, selection.bottom
+            ),
+            psSelectionWidth = psSelectionBR.x - psSelectionTL.x,
+            psSelectionHeight = psSelectionBR.y - psSelectionTL.y,
+            // The resize rectangles are roughly 10 points radius
```
iwehrman: Points/pixels?

src/js/actions/tools.js
```diff
@@ -200,6 +205,89 @@ define(function (require, exports) {
...
+            inset = 10,
+            // In case of artboards, we have no rotate, so we can stay within the border
+            outset = artboards ? inset : 25;
+
+        var insidePolicy = new PointerEventPolicy(adapterUI.policyAction.NEVER_PROPAGATE,
```
iwehrman: A little comment about what you're doing here geometrically might be useful to the next schmuck to come across this.

src/js/actions/transform.js
```diff
@@ -1045,6 +1045,27 @@ define(function (require, exports) {
...
+                nextDoc = appStore.getCurrentDocument();
+
+            // FIXME: We used to listen to "move" event's translation and optimistically update
+            // all selected layers, but due to a recent bug, "move" event sends us the displacement
+            // of layers from the changing (0,0) coordinates, which causes bugs like
+            // getting (650,0) when the move was actually (-100, 0) for a 750 px wide layer
+            this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
+        }.bind(this);
+
+        descriptor.addListener("transform", _transformHandler);
```
iwehrman: Any event handlers added in `beforeStartup` need to be removed in `onReset` to avoid leaks and duplicate handling.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -153,6 +157,17 @@ define(function (require, exports, module) {
             }
         },
 
+        mouseMoveHandler: function (event) {
```
iwehrman: JSDoc plz.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -388,6 +387,58 @@ define(function (require, exports, module) {
             this._currentMouseDown = true;
         },
 
+        updateMouseOverHighlights: function () {
```
iwehrman: JSDoc 4 lyfe.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -388,6 +387,58 @@ define(function (require, exports, module) {
...
+                uiStore = this.getFlux().store("ui"),
+                mouseX = this._currentMouseX,
+                mouseY = this._currentMouseY,
+                canvasMouse = uiStore.transformWindowToCanvas(mouseX, mouseY),
+                highlightFound = false;
+
+            // Yuck, we gotta traverse the list backwards, and D3 doesn't offer reverse iteration
+            _.forEachRight(d3.selectAll(".layer-bounds")[0], function (element) {
+                var layer = d3.select(element),
+                    layerLeft = parseInt(layer.attr("x")),
```
iwehrman: Don't use `window.parseInt`. It's weird. You have to supply a radix, and if it fails you get NaN which is strangely hard to test for. Instead use `parseNumber` from `js/util/math`, which returns null when it fails to parse.

src/js/scrim/TransformScrim.js
```diff
@@ -1,1050 +0,0 @@
...
-    /**
-     * Flag to tell whether we're currently dragging or not
-     *
-     * @type {boolean}
-     */
-    TransformScrim.prototype._dragging = null;
-
-
-    module.exports = TransformScrim;
-});
```
iwehrman: :(

src/js/tools/superselect.js
```diff
@@ -64,13 +64,13 @@ define(function (require, exports, module) {
             var toolOptions = {
                 "$AtSl": false, // Auto select on drag
                 "$ASGr": false, // Auto select Groups,
-                "$Abbx": false // Don't show transform controls
+                "$Abbx": true // Don't show transform controls
```
iwehrman: Stale comment.

iwehrman:
Found a few nits in the code, but mainly the above bug. If you want to merge this now (after fixing the nits), we can do so and file a separate issue for that. Back to @volfied for now.

chadrolfs:
Running into an issue where I can't draw a shape on canvas after selecting a layer/group/AB in layers panel.

src/js/actions/tools.js
volfied: superselect-type-edit mode is it's own tool, and is completely driven by Photoshop as it is modal.

src/js/actions/tools.js
volfied: I was going to say pixels, but I think it's points, not sure. @designedbyscience what would you call this unit of measure (I used OSX screen shot marquee to measure the distance)

volfied:
@iwehrman Responded to your review, also added shift key policy and a clean up for tool changes.

Last one should fix the issue where if you drew a rectangle over another artboard, you wouldn't be able to draw rectangles in the original artboard. We never cleaned up the policy on tool change so it was just bad.

chadrolfs:
glad I could be of assistance.  :grinning:

iwehrman:
New bug: sometimes, but not always, I can't drag the the side (not a corner) of a rectangle. Instead, it seems to go right into marquee selection.

iwehrman:
Well, now I can't repro that. But I swear I saw it...

iwehrman:
I also saw (one time, and can't repro) the "can't use the move tool because no layers are selected" dialog.

iwehrman:
OK, I figured out how to repro the drag/marquee problem. The problem is that if you put your cursor on the edge of the rectangle, you see the horizontal or vertical drag cursor, but it doesn't actually start a drag.

volfied:
@iwehrman The can't use the move tool... error was probably due to fact that I didn't clean up my border policies correctly, and oyu were sending a click to photoshop when you shouldn't. 

I think the inset/outset need a little bit tweaking.

volfied:
With 1376638, this PR will require the latest Jenkins build. Fitz has renamed the propagation mode.

ktaki:
Found an issue with the branch. Cannot use Type tool. Cannot go into text editing mode after clicking on Canvas with Text tool selected.

volfied:
@ktaki Great catch! Can you check with the latest commit to see if this is fixed?

Cloak command was causing PS to exit modal tool state, which was immediately committing the type layer.

ktaki:
Now text editing works. Wow, you can transform text layers. Cool. I see some issues here, but I assume text transformation will not be a part of this PR. Am I correct?

ktaki:
Issue: When a rectangle is created, changing the tool to super-select won't show the transform controls. This works with ellipse, pen, and text. Steps: Draw a rectangle, then switch the tool to super-select. Result: No controls shown.

ktaki:
Issue: Once text layer is rotated, it becomes hard to rotate it again.

ktaki:
Issue: Ps dialog "Could not use the move tool because no layers are selected." This happens when trying to move rotated text layers. Clicking off the text but still inside of the blue bounding box will cause this. Can you suppress the dialog?

ktaki:
Issue: Blue bounding box for objects sometimes shown and sometimes not. And sometimes it flashes momentarily when mousing over. What is expected behavior of the blue box here?

ktaki:
Issue: Resizing and moving artboard resulted in a separation of bounding box.
screen shot 2015-06-23 at 11 37 17 am

volfied:
Thanks @ktaki, here is my responses to your bug reports:

1 - Text transformation should work like any other layer transformation. From my testing, it looks like there is a small outset/inset difference when it comes to type layers, and I'll work on figuring out the pixel difference and set the area correctly.

2 - Transform controls not showing all the time: This is a core issue, the transform controls do not immediately redraw, I will work with a PS engineer to make sure I force redraw safely on this.

3 - Type layer re-rotate difficult: I don't see this, but related to (1), it's probably because type layers have a larger area than the one text dictates and our policies don't fit it correctly.

4 - Same reason as (1)

5 - Blue bounding box will show for highlighting, so you can still see the layer you're about to select. It will also stay on screen for a brief period after selecting a layer. This is a lower priority bug which I don't mind going to master with.

6 - Artboard moving: This is because you can move artboards by slightly clicking outside their transform controls, and we don't handle the move events. I will handle them.

ktaki:
The separation of transform control can easily be reproduced with an Apple Watch 38 mm artboard. Just create it and move it.

volfied:
@ktaki I think I've addressed all the issues. One more pass would be very appreciated.

@iwehrman Unless taki finds a big breaking bug, I'd like you to hit the merge button.

chadrolfs:
After I've rotated text, 2nd attempt on bottom right corner fails and then I can hit a model dialog: "Could not use the move tool because no layers are selected."

chadrolfs:
I also just hit this when trying to move an AB while text was selected:
`Action tools.resetSuperselect failed: Error: Host is in a modal state and the request cannot be executed.`

ktaki:
Yes, I still see that issue also.

volfied:
D'oh! I realized that I don't remove the last border policy I've installed when no layers are selected. So your clicks (near the type layer) were still going to Photoshop. 37e31a5 should address this issue.

ktaki:
The issue of getting a model dialog: "Could not use the move tool because no layers are selected" is fixed, but the text layer is still hard to rotate for the second time if the first rotation is not a multiple of 90 degrees.

iwehrman:
There are still bugs here, some of which will require core changes to fix, and some of which are severe. But, the only way out is through.

Issue #1613, opened 2015-06-18
Option copy of layer causes error "Too many layers in layer index"


ktaki:
Create a doc and place a rectangle. With the select tool, drag the rectangle while pressing Option key to create a copy of the rectangle.

Action layers.getLayerOrder failed: Error: Too many layers in layer index
    at Record.LayerStructure.updateOrder 

![screen shot 2015-06-18 at 10 55 56 am](https://cloud.githubusercontent.com/assets/2029597/8238150/bb2efc4c-15a8-11e5-8015-f673bc1ff7db.png)

mcilroyc:
git bisect, you my only friend.
first bad commit: 9cdf11a356e524b406642ca5ebbd5e93b5db9435
I'll continue investigating...

ktaki:
Verified the fix.

chadrolfs:
Looks like we have a regression here.  Ran into this duplicating Artboards

chadrolfs:
![screen shot 2015-09-29 at 2 37 11 pm](https://cloud.githubusercontent.com/assets/8442631/10178923/935bcca8-66b7-11e5-9e72-26b1a13f9098.png)

mcilroyc:
When duplicating a pair of artboards, *sometimes* I can repro this error.  My hunch is that it occurs when we get a "moveToArtboard" event before the "move" event.  It can't reliably get it to happen.  Any parting thoughts on this @volfied since you have some recent familiarity here, I think.

iwehrman:
When I select two artboards (each with two rectangles inside) and option-drag copy them, I get:
`Verification of postcondition 1 failed for action layers.addLayers`, which is `_verifyLayerSelection`. This might be spurious however (since we're processing the events out of sync with the action?) because the selection looks OK to me.

iwehrman:
Other than that, I can't break it. So, besides some requests for additional in-line documentation and a question about a different debounced handler, I'm basically satisfied for the time being. Also, to reiterate our discussion off-line: what need for a longer-term solution is to articulate a core request for how all these notifications ought to work, or at least how they could be changed to be significantly less broken.

mcilroyc:
@iwehrman I think your last two comments were supposed to be in the PR... but anyway:  At first glance I do think the verifyLayerSelection failure is spurious and not easily avoidable, but I'll look more closely.

iwehrman:
(Derp. Moved them over for later readability.)

iwehrman:
[...]

PR #1618, opened 2015-06-18
Fixes for option-dragging a new layer

mcilroyc:
This PR adds a new type of REORDER_LAYERS event that does *not* increment history.  This is to support the recently added logic that, after moving a layer between Artboards, resets the entire layer order by asking photoshop.  Additionally, this step is *no longer* called in the "copyDrag" case, because that ends up doing a full updateDocument anyway.

addresses #1613


mcilroyc:
This one has @pineapplespatula 's name all over it

pineapplespatula:
LGTM

Issue #765, opened 2015-06-12
Native menu commands harmful


DesignSpaceBot:
We should put some effort into reducing our dependency on native menu commands (i.e., `_playground.ps.performMenuCommand`). These commands all subvert our concurrency control mechanisms entirely because we have no general way to know when the commands have finished execution. All we know is that Photoshop has finished processing the menu command, but that may result in _its_ command being handled asynchronously. For example, when we execute the "revert" menu command, the actual document reversion happens some time after the callback supplied to `performMenuCommand` is processed. 

Alternatively, we should start a discussion about a core change that would prevent the callback from being applied until the underlying async command has finished.

-- @iwehrman at [2015-03-10T18:54:34Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/765)

PR #1619, opened 2015-06-18
Replace performMenuCommand(1016) with an API

volfied:
Helps alleviate #765 symptoms, but not fully.

Probably shouldn't have had trim_trailing_whitespace_on_save

Requires Adapter#109


mcilroyc:
:bowling: nice one

PR #1621, opened 2015-06-19
reorder layers when non artboard layer is dragged onto on artboard

pineapplespatula:
fix for issue #1587  we get the event after we finish the move.

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -622,8 +622,12 @@ define(function (require, exports) {
                         }
 
                         var artboardNested = false;
+                        var moveHandled  = false;
                         _moveToArtboardListener = _.once(function () {
                             artboardNested = true;
+                            if (moveHandled) {
+                                this.flux.actions.layers.getLayerOrder(doc, true);
+                            }
                         }.bind(this));
 
                         descriptor.addListener("moveToArtboard", _moveToArtboardListener);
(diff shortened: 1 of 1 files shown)


mcilroyc:
does this work for copyDrag?

Have we asked any core engineers if there is a way to guarantee the order of these two events from photoshop (or, better yet, to combine them)?  This seems brittle as it stand now.

chadrolfs:
reoder - needs to smell better

pineapplespatula:
Yes this works on copyDrag.

mcilroyc:
Oh, yes, my bad.  I see it works on copyDrag.  BUT, there is another issue: After moving an off-AB shape on to the AB, performing an undo will not update the bounds.

Also cc @volfied because this is related to that same code that I mentioned to him might need to be ported to his no-transform-scrim branch.

src/js/actions/superselect.js
```diff
@@ -622,8 +622,12 @@ define(function (require, exports) {
                         }
 
                         var artboardNested = false;
+                        var moveHandled  = false;
```
volfied: These two flags seem redundant. If `artboardNested` is true, we `getLayerOrder` and if `moveHandled` is true, we `getLayerOrder` ?

There has to be a cleaner way of doing this, or at least a comment explaining the need for two flags.

mcilroyc: noted.  FYI barkin, I'm going to mess with this because I need to make some historyStore changes to workaround this great photoshop quirk.

mcilroyc:
closing this in favor of #1696 which takes a different approach based on new updates to Photoshop which guarantee the order of the two notifications.

Issue #1597, opened 2015-06-16
Layer Search the order of the initial list could be reversed


ktaki:
...to match what you see in the layers panel. It may be more natural in that way. Especially when you have multiple layers of the same name, you want to get to the first occurrence, then do you count from the top or bottom?

PR #1622, opened 2015-06-19
Reverse layer order in search

hamburge:
Addresses issue #1597


iwehrman:
Well that's easy enough.

Issue #1577, opened 2015-06-15
Dragging layers in layer panel draws layer in upper left of app window


chadrolfs:
https://www.dropbox.com/s/1bdsp32wnz30z6p/LayersDrag.mov?dl=0

iwehrman:
Not addressed by #1625.

designedbyscience:
This should also be fixed by  #1673

chadrolfs:
I can still repro.

designedbyscience:
@chadrolfs I can't reproduce this, is there anything in particular that seems to trigger it?

designedbyscience:
Okay, this is related to the fact that the initial positions (state.startX and state.startY) are only applied when the layer is clicked on. If you select a bunch of layers with shift, you should be able to replicate this

chadrolfs:
Yep, you got it.  If you select more that one layer, you'll see this.

iwehrman:
Addressed by #1700. FBNC to @chadrolfs.

chadrolfs:
doh!  Still there.  Try these steps.  
New default doc (AB)
Select Rect tool and drag out 3 shapes (Rect 3 should be at top of AB stack and selected)
Hold down Shift and select Rect 1 (all 3 shape layers selected), then drag down

designedbyscience:
Hmm... I don't see it when I try these steps. @iwehrman do the repro steps work for you?

chadrolfs:
something was off with my environment, i refreshed everything and now CNR

Issue #1579, opened 2015-06-15
Layers jump momentarily at drag start


iwehrman:
![drag](https://s3.amazonaws.com/f.cl.ly/items/3e002P0L2P3x0H350h3S/Screen%20Recording%202015-06-15%20at%2002.18%20PM.gif)

iwehrman:
Not addressed by #1625:
![jump](https://s3.amazonaws.com/f.cl.ly/items/412S17462E162B1X3u2A/Screen%20Recording%202015-06-26%20at%2003.21%20PM.gif)

iwehrman:
Fixed by #1700.

Issue #1574, opened 2015-06-12
Opening a large file (lots of layers) results in lots of registerDroppable and Dialog actions


chadrolfs:
Open VermilionArtboards.psd and watch the console.  Here's a couple:
`Executing action dialog.registerDialog after waiting 5205ms; 8/739`
`Finished action draganddrop.registerDroppable in 50ms with RTT 6422ms; 8/13`

designedbyscience:
1. We can avoid most of the overhead by batch registering items.
2. The impact of this is definitely increased by debug mode (registration is somewhere around 10x slower by a quick test).

chadrolfs:
Switching between open docs becomes horribly painful and REALLY slow.

designedbyscience:
Because you are getting hit twice (register and deregister). Working on it

chadrolfs:
:hurtrealbad:

Issue #1570, opened 2015-06-12
Selected layer drop targets sometimes lose border radius


iwehrman:
![border](https://s3.amazonaws.com/f.cl.ly/items/1E422l0q1f3X1T2v2N1B/Screen%20Recording%202015-06-12%20at%2011.18%20AM.gif)

iwehrman:
Not addressed by #1625:
![radius](https://s3.amazonaws.com/f.cl.ly/items/3j3G2K3514341G3f3P1q/Screen%20Recording%202015-06-26%20at%2003.23%20PM.gif)

designedbyscience:
Okay, this partially fixed in that the selected layer should lose radius for the top/bottom (matched to drop above/below), otherwise, how would you tell where the dragged later is supposed to go? However, there is a weird issue that inset causes the top border to have a slightly different color.

iwehrman:
Given your explanation, and given that the layer and drag styling is going to change anyway, I'm just going to close this. Mission accomplished! :us: :us:

PR #1625, opened 2015-06-19
Flattened layer hierarchy

designedbyscience:
Cleaned up a bunch of bugs and CSS related to dragging and dropping, and layer selection in groups

On mount of Layers panel, we now batch the registration of the droppables
Similarly, we also batch deregister layer items when the set changes

Addresses:
#1579 
#1577
#1574
#1570
#1569
#1568
#1404
#1398


chadrolfs:
:tada:

designedbyscience:
@chadrolfs a) By itself, that little emoji somehow seems sarcastic b) Sorry it took so long to get right

chadrolfs:
@designedbyscience sorry, didn't at all mean to come off that way.  I know you've been tearing your hair out over this and I'm really excited to see it come in and fix so many issues!  Thanks for your diligence on this!

iwehrman:
If he wanted a sarcastic emoji, he would have gone with :balloon:

src/js/actions/draganddrop.js
```diff
@@ -50,6 +50,16 @@ define(function (require, exports) {
     };
 
     /**
+    * Add many droppables at once
+    *
+    * @param {Iterable.List} list of droppable registration information
```
iwehrman: `Immutable.Iterable.List<something here>`

src/js/actions/draganddrop.js
```diff
@@ -60,6 +70,16 @@ define(function (require, exports) {
     };
 
     /**
+    * Remove many drop targets by a list of keys a drop target by key
+    *
+    * @param {Iterable.List} keys List of keys to remove
```
iwehrman: Ditto throughout this file.

src/js/actions/draganddrop.js
iwehrman: Should we just rename these and get rid of the existing `register` and `deregister` functions? I.e., only have code that deals with lists of droppables?

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -193,15 +217,54 @@ define(function (require, exports, module) {
                 isDropTargetBelow = true;
             }
 
+            var depth = layerStructure.depth(layer);
+            var isLastInGroup = false;
```
iwehrman: Collapse these declarations.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -193,15 +217,54 @@ define(function (require, exports, module) {
...
+
+            if (layerStructure.byIndex(layerIndex - depth)) {
+                isLastInGroup = isChildOfSelected &&
+                    layerStructure.byIndex(layerIndex - 1).kind === layer.layerKinds.GROUPEND;
+            }
+
+            // Check to see if this layer is right before the end of nested group
+            var endOfGroupStructure;
+            
+            if ((layerIndex - 1) - (layerIndex - depth) > 0) {
```
iwehrman: This is a weird condition. I'm not exactly sure what this is testing, but it looks like... 
```
(layerIndex - 1) - (layerIndex - depth) > 0
iff
(layerIndex - 1) - layerIndex + depth > 0
iff 
depth - 1 > 0
iff
depth > 1
```

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -255,10 +318,12 @@ define(function (require, exports, module) {
             );
 
             return (
+                <div className={classnames(layerClasses)} data-kind={layer.kind}>
```
iwehrman: Presumably we should not need this layer kind information in here twice?

iwehrman: Also, the inner `<div>` should be pushed in one tabstop.

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -379,35 +428,23 @@ define(function (require, exports, module) {
                 layerCount = null;
                 childComponents = null;
             } else {
-                layerComponents = doc.layers.top
+                layerComponents = doc.layers.allVisible.reverse()
```
iwehrman: If I had a nickle for every time we reversed a layer list, I'd have about 25c.

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -113,6 +140,27 @@ define(function (require, exports, module) {
             if (!newSelection.isEmpty()) {
                 this._updateLowestNode();
             }
+            
+            if (prevProps.document.id !== this.props.document.id) {
+                // For all layer refs, ask for their registration info and add to list
+                var batchRegistrationInformation = [];
+
+                this.props.document.layers.allVisible.forEach(function (i) {
```
iwehrman: This should just be a `map` instead of a `forEach`. If you need it to be an array at the end, call `.toArray`, but even better if you can just send the `Immutable.Iterable` around. (That is, don't convert to an array unless it's actually necessary.)

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -379,35 +428,23 @@ define(function (require, exports, module) {
...
+                        var isDragTarget = !!(dragTarget && dragTarget.indexOf(layer) !== -1),
+                            isDropTarget = !!(this.props.dropTarget && this.props.dropTarget.keyObject === layer);
 
                         return (
                             <li key={layer.key}>
-                                <Layer
+                                <LayerFace
+                                    ref={layer.key}
                                     disabled={this.props.disabled}
+                                    registerOnMount={!this.state.batchRegister}
```
iwehrman: In a perfect world, all the state elements in this component would be documented with a `getInitialState` function. Or at least the new ones...

src/js/jsx/shared/Droppable.jsx
```diff
@@ -59,12 +59,28 @@ define(function (require, exports, module) {
                 this.getFlux().actions.draganddrop.registerDroppable(node, key, validateDrop, handleDrop, keyObject);
             },
 
+            getRegistration: function () {
```
iwehrman: Hit me with some JSDoc with a short note about what each property means?

src/js/actions/draganddrop.js
iwehrman: Above is just a question, not a hard ask. Sometimes it's nice to overload "outer" functions so that they can take 1 or many, but then use many consistently internally. But, it's up to you.

src/js/stores/draganddrop.js
```diff
@@ -119,12 +122,49 @@ define(function (require, exports, module) {
...
         /**
+         * Removes all current drop targets and adds a batch of new ones
+         *
+         * @param {object} payload list of registration information
+         */
+        _handleResetDroppables: function (payload) {
+            this.dropTargets = new Immutable.OrderedMap();
+            payload.forEach(function (p) {
+                this._handleRegisterDroppable(p);
+            }.bind(this));
```
iwehrman: Using the second `this` argument is preferred over an explicit bind.

src/style/sections/pages/face.less
```diff
@@ -24,7 +24,38 @@
...
+
+.layer__select_descendant {
+    background-color: @selected-layer-group-background;    
+}
+
+.layer__select_child {
+    background-color: @selected-layer-group-background;
+}
+
+.layer__group_end{
```
iwehrman: :space_invader:

src/style/sections/pages/face.less
```diff
@@ -113,6 +142,10 @@ input[type="text"].face__name {
     z-index: 99;
 }
 
+.layer-list__dragging .face:hover{
```
iwehrman: :space_invader:

src/style/sections/pages/face.less
```diff
@@ -139,46 +176,20 @@ input[type="text"].face__name {
...
-    padding-top: 0.3rem;
+.face__select_child {
+    background-color: @selected-layer-group-background;
 }
 
-.face__select_immediate + ul {
-     position:relative;
-     border-bottom-right-radius: @face-border-radius;
-     border-bottom-left-radius: @face-border-radius;
+.face__group_lastchild{
```
iwehrman: :space_invader:

src/style/sections/pages/face.less
```diff
@@ -139,46 +176,20 @@ input[type="text"].face__name {
...
-    border-bottom-left-radius: @face-border-radius;
-    content: " ";
-    left:0;
-    height:100%;
-    margin: 0 2rem 0 2rem;
-    position:absolute;
-    top:calc(@face-border-radius * -1);
-    width:calc(@properties-width - 4rem);
-    z-index:-1;
+.face__group_lastchildgroup{
```
iwehrman: :space_invader:

src/js/jsx/sections/pages/LayerFace.jsx
designedbyscience: Yea...MATH! And I don't need it.

iwehrman:
Bad news: this seems to cause significant performance regression in `layers.select`. In a document with ~340 layers, I see a median/mean time on master of 210/240ms but 392/398ms on your branch for that action. Most likely this is due to a change in React rendering (and whatever our React components call). See for yourself by clicking around and then doing: `require("js/util/performance").printExecTimes()`

iwehrman:
Done with a first pass. 

I'm also worried about the redraw speed of while dragging. I'm not sure if it's regressed though or if it was always this way.

chadrolfs:
took a quick peek and with the Vermilions doc, re-ordering layers in the layers panel performance is pretty bad, although I am seeing issues fixed such as dragging above artboards (to top of layer panel).

iwehrman:
Get back in there!
salt mines

designedbyscience:
Now salty as a potato chip, I'm back.

Changes to make rendering faster on selection
Changes to make dragging faster (lots faster)

iwehrman:
Bad news: there's still a significant performance regression with layer selection in big documents. On VermillionArtboards.psd, I see 350-400ms times on master for `layers.select` and 500-550ms on this branch. We're going down to 100ms for the next release (:bullettrain_front::dash:) and this gets us further from the promised land!

get back in there

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,16 +40,48 @@ define(function (require, exports, module) {
...
+                      
+        // Deeper selection changes
+        var document = this.props.document,
+            allSelected = document.layers.allSelected,
+            childOfSelection = allSelected.includes(this.props.layer),
+            parentSelectedChanged = false,
+            nextAllSelected = nextProps.document.layers.allSelected;
+
+        if (childOfSelection) {
+            parentSelectedChanged = !Immutable.is(document.layers.selected, nextProps.document.layers.selected);
```
iwehrman: Can we get away with just comparing their faces?

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,16 +40,48 @@ define(function (require, exports, module) {
...
+        }
+
+        return allSelected.includes(this.props.layer) !== nextAllSelected.includes(nextProps.layer) ||
```
iwehrman: Reverse the disjuncts to save some executions of the slightly more expensive-to-compute one.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,16 +40,48 @@ define(function (require, exports, module) {
...
         strings = require("i18n!nls/strings");
 
+    /*
+     * Function for checking whether React component should update
+     * Passed to Droppable composed component in order to save on extraneous renders
+     *
+     * @param {Object} nextProps - Next set of properties for this component
+     * @return {boolean}
+     */
+    var shouldComponentUpdate = function (nextProps) {
```
iwehrman: Is there some reason why this isn't a component method? If not, the bind-and-call below is unnecessarily expensive.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,16 +40,48 @@ define(function (require, exports, module) {
...
         mixins: [FluxMixin],
 
         shouldComponentUpdate: function (nextProps) {
-            return !Immutable.is(this.props.layer.face, nextProps.layer.face) ||
-                this.props.dragTarget !== nextProps.dragTarget ||
-                this.props.dropAbove !== nextProps.dropAbove ||
-                this.props.dragPosition !== nextProps.dragPosition ||
-                this.props.dragStyle !== nextProps.dragStyle ||
-                this.props.dropTarget !== nextProps.dropTarget;
+            return shouldComponentUpdate.bind(this, nextProps)();
```
iwehrman: (This one.)

src/js/jsx/shared/Droppable.jsx
```diff
@@ -45,7 +44,7 @@ define(function (require, exports, module) {
      * @param {function} getProps function to return an object with the props required by Droppable
      * @return {ReactComponent}
      */
-    var createWithComponent = function (Component, getProps) {
+    var createWithComponent = function (Component, getProps, isEqual, shouldUpdate) {
         var Droppable = React.createClass({
```
iwehrman: It might make more sense for `Droppable` to be an "abstract class" that is "subclassed" by concrete implementations, and `React.createClass` would be applied to those prototype-derived objects? Seems plausible that I'm missing something though which would prevent that...

src/js/stores/draganddrop.js
```diff
@@ -169,7 +207,7 @@ define(function (require, exports, module) {
                 }.bind(this));
             }
 
-            if ((potentialDropTarget && potentialDropTarget.validate(this._dragTarget, point, potentialDropTarget.b))) {
+            if ((potentialDropTarget && potentialDropTarget.validate(this._dragTarget, point, this._currentBounds))) {
```
iwehrman: This is a fairly hot function as a mousemove handler, and I can imagine a number of ways to speed it up:

1. Sort `_dropTargets` (say, by their y-coordinate) and search for matches using a binary instead of a linear search. Currently this is an O(n^2) algorithm, since you're doing an O(n) lookup on O(n) drop targets.
2. Memoize `potentialDropTarget.validate`.
3. Possibly consider caching the result of `getBoundingRect` since it's so expensive. Could we demand that the bounding rects of potential _drop targets_ must not change during the drag operation?
4. This is minor, but you should move the search procedure to a private method so you can avoid the binding step. (This would probably be insignificant in most cases, but perhaps not in a mousemove handler.)

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -193,15 +229,61 @@ define(function (require, exports, module) {
...
+            if (isDragTarget) {
+                dragStyle = this.props.dragStyle;
+            } else {
+                // We can skip some rendering calculations if dragging
+                if (layerStructure.byIndex(layerIndex - depth)) {
+                    isLastInGroup = isChildOfSelected &&
+                        layerStructure.byIndex(layerIndex - 1).kind === layer.layerKinds.GROUPEND;
+                }
+
+                // Check to see if this layer is right before the end of nested group
```
iwehrman: I don't quite understand what this is supposed to represent. It seems like you just want to know if the visible layer underneath this one in the panel is at the top level? (Or if there is no visible layer below.) Maybe there is a cheaper way to compute this?

iwehrman:
Done with this pass. Suggested a few opportunities for optimization, but I haven't looked at the profiler yet so I don't know how valid they are :sunglasses:

src/js/jsx/sections/pages/LayerFace.jsx
designedbyscience: Nope. The faces alone don't change if the parent selection changes

iwehrman: I meant "can we get away with just comparing the faces of the selected layers". This would be a slower check, because it would require `collection.pluck`, but it would be more precise. So, it's not clear that it would be preferable.

src/js/jsx/sections/pages/LayerFace.jsx
designedbyscience: Actually...I can probably kill this call all together (since we have already used the same thing on Droppable)

src/js/stores/draganddrop.js
designedbyscience: Initially, I had done the cache bounding rects thing because it seemed to be a clear win speed-wise. However, those bounding rects can change pretty easily (e.g. the layers panel scrolls, layers get reordered). If I calculated them all up front (expensive and that goes right into startup time), there'd be lots of cases where I'd need to invalidate those (because not only do absolute positions change, but relative positions can change), and once invalidated, recalculation is expensive. It seemed overall that calling it once here would minimize the overall number of calls.

I could cache it for the length of the dragging session (and invalidate if there is a scroll event)

Seems like 2. is the biggest target here

iwehrman: You don't think number 1 will be worthwhile?

Also, for the record, I'm somewhat _more_ concerned about the regression in `layers.select` than in the drag-and-drop redraw speed because the former is one of the most frequently used operations in the application. Of course, ideally both operations would get faster :)

designedbyscience:
@iwehrman  Back at you. My time has been focused on the selection speed, so checkBounds still hasn't been optimized further. These changes help drag/drop as they increase speed of render of the PagesPanel

iwehrman:
I'm getting this when loading Vermillion:
```
Action documents.open failed: TypeError: Cannot read property 'id' of undefined
    at Record.LayerStructure.parent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/models/layerstructure.js:470:40)
    at Record.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/models/layerstructure.js:648:27)
    at Record.value (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/object.js:145:41)
    at React.createClass.render (file://./js/jsx/sections/pages/LayerFace.jsx:243:37)
    at ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:7445:34)
    at ReactCompositeComponentMixin._renderValidatedComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:7472:14)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:10068:19)
    at wrapper [as _renderValidatedComponent] (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13968:31)
    at ReactCompositeComponentMixin.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6893:30)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:10068:19)
    at wrapper [as mountComponent] (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13968:31)
    at Object.ReactReconciler.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:14781:35)
    at ReactCompositeComponentMixin.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6903:34)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:10068:19)
    at wrapper [as mountComponent] (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13968:31)
    at Object.ReactReconciler.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:14781:35)
From previous event:
    at Object.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/documents.js:384:40)
From previous event:
    at Object.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/documents.js:378:22)
From previous event:
    at Object.initActiveDocument (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/documents.js:367:14)
    at FluxController._applyAction (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:290:36)
    at Object.create.transfer.value (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:227:33)
    at Object.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/documents.js:275:40)
From previous event:
    at Object.open (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/documents.js:274:14)
    at FluxController._applyAction (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:290:36)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:382:37)
```
This is on 01de5761dd183f1dcf2d087cd509b132cc57419d, master on spaces-adapter, plugin version 1.0.24, Jenkins build Mac cev 371.

designedbyscience:
I think something is wrong on master...

designedbyscience:
@iwehrman Okay, should work now

iwehrman:
thumbs up

src/js/actions/draganddrop.js
```diff
@@ -89,7 +113,7 @@ define(function (require, exports) {
     /**
     * Check the intersection of the current dragTarget and available drop targets
     *
-    * @param {{x: number, y: number}} point Point from event
+    * @param {Object.<{x: number, y: number}>} point - Point from event
```
iwehrman: The previous type annotation was correct: `@param {{x: number, y: number}}`. There are three different notations for objects used in different cases. It's weird, but consistent.

src/js/actions/draganddrop.js
```diff
@@ -64,9 +76,21 @@ define(function (require, exports) {
...
+        return this.dispatchAsync(events.droppable.DEREGISTER_DROPPABLE, keys);
+    };
+    batchDeregisterDroppables.reads = [];
+    batchDeregisterDroppables.writes = [];
+
+    /**
     * Fire event that dragging started
     *
-    * @param {Immutable.List} dragTarget List of currently dragging items
+    * @param {Immutable.Iterable.List<Object>} dragTarget - List of currently dragging items
```
iwehrman: Here, it should just be `@param {Immutable.Iterable.<object>}` or, alternatively, `@param {Immutable.List.<object>}`.

src/js/actions/draganddrop.js
```diff
@@ -64,9 +76,21 @@ define(function (require, exports) {
     deregisterDroppable.writes = [];
 
     /**
+    * Remove many drop targets by a list of keys a drop target by key
+    *
+    * @param {Immutable.Iterable.List<Object>} keys - List of keys to remove
```
iwehrman: Ditto.

src/js/actions/draganddrop.js
```diff
@@ -52,9 +52,21 @@ define(function (require, exports) {
     registerDroppable.writes = [];
 
     /**
+    * Add many droppables at once
+    *
+    * @param {Immutable.Iterable.List<Object>} list - List of droppable registration information
```
iwehrman: Ditto. Also, kill all the dashes. Just start 'splainin' after the variable name.

src/js/actions/draganddrop.js
designedbyscience: I think it yelled at me for that previous one

src/js/models/layerstructure.js
```diff
@@ -203,6 +203,14 @@ define(function (require, exports, module) {
                     return layer.kind === layer.layerKinds.GROUPEND;
                 });
         },
+        
+        /**
+        * All non-endgroup layers, in reverse order
+        * @type {Immutable.List.<Layer>}
+        */
+        "allVisibleReversed": function () {
+            return this.allVisible.reverse();
```
iwehrman: Any idea if `.reverse` is `O(1)` or `O(n)`? If it's the former, it might be better to just call `.reverse()` at the point of use. If it's the latter and it's actually expensive then instead it might be better to make this a derived (cached) property.

src/js/jsx/sections/pages/LayerFace.jsx
iwehrman: Now you're using this as a component method, but it's defined separately for no apparent reason. Unless there _is_ some reason, it should be moved inline with all the other methods for consistency.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,18 +40,43 @@ define(function (require, exports, module) {
         system = require("js/util/system"),
         strings = require("i18n!nls/strings");
 
-    var LayerFace = React.createClass({
-        mixins: [FluxMixin],
+    /*
+     * Function for checking whether React component should update
+     * Passed to Droppable composed component in order to save on extraneous renders
+     *
+     * @param {Object} nextProps - Next set of properties for this component
```
iwehrman: `@param {object}`

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -40,18 +40,43 @@ define(function (require, exports, module) {
...
-                this.props.dragPosition !== nextProps.dragPosition ||
-                this.props.dragStyle !== nextProps.dragStyle ||
-                this.props.dropTarget !== nextProps.dropTarget;
-        },
+        // Deeper selection changes
+        var document = this.props.document,
+            childOfSelection = document.layers.hasSelectedAncestor(this.props.layer),
+            nextChildOfSelection = nextProps.document.layers.hasSelectedAncestor(nextProps.layer);
+            
+        if (childOfSelection || nextChildOfSelection) {
```
iwehrman: It's awkward to do this with JavaScript, but there _is_ an opportunity here to potentially avoid one of the above method calls be rewriting this guard.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -180,54 +205,82 @@ define(function (require, exports, module) {
...
-                })
-                .value();
-
-            var dragStyle;
-            if (isDragTarget) {
-                dragStyle = this.props.dragStyle;
-            } else {
-                dragStyle = {};
-            }
+            faceClasses["depth-" + depth] = true;
```
iwehrman: These `depth-` class names should be prefixed.

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -180,54 +205,82 @@ define(function (require, exports, module) {
...
 
             // Super Hack: If two tooltip regions are flush and have the same title,
             // the plugin does not invalidate the tooltip when moving the mouse from
             // one region to the other. This is used to make the titles to be different,
             // and hence to force the tooltip to be invalidated.
+            // var tooltipPadding = _.repeat("\u200b", layerIndex);
```
iwehrman: Stale line comment here?

src/js/jsx/sections/pages/LayerFace.jsx
```diff
@@ -255,60 +308,66 @@ define(function (require, exports, module) {
...
+                            selected={layer.locked}
+                            onClick={this._handleLockToggle}>
+                        </ToggleButton>
+                    </div>
+                </li>
             );
         }
     });
 
     var draggedVersion = Draggable.createWithComponent(LayerFace, function (props) { return props.layer;}, "y"),
```
iwehrman: A little JSDoc on this guy would be helpful.

iwehrman:
This does not seem to address #1577. When I drag a range of layers in Vermillion, I still see a bunch of the layers up in the top-left corner of the window.

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -93,25 +101,45 @@ define(function (require, exports, module) {
...
 
             this._scrollToSelection(this.props.document.layers.selected);
-            this._updateLowestNode();
             this._bottomNodeBounds = 0;
+            
+            // For all layer refs, ask for their registration info and add to list
+            var batchRegistrationInformation = this.props.document.layers.allVisible.map(function (i) {
+                return this.refs[i.key].getRegistration();
+            }, this),
+                mappedBatchRegistrationInformation = new Immutable.OrderedMap(batchRegistrationInformation);
```
iwehrman: No `new` needed for the Immutable creation functions.

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -143,23 +171,37 @@ define(function (require, exports, module) {
...
 
+        /* Set initial state
+         * State variables are:
+         *  futureReorder {boolean} - Need this to prevent flash of dragged layer 
+         *     back to original positiion
+         *  batchRegister {boolean} - Should we register all dropppables at once 
+         *     (and prevent individual droppables from registering)
+         *  dropAbove {boolean} - Should the dragged layer drop above or below target
+         *
+         * @return {Object}
```
iwehrman: `@return {object}`

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -143,23 +171,37 @@ define(function (require, exports, module) {
...
+                batchRegister: true,
+                dropAbove: false
+            };
         },
 
         /**
          * Updates the lowest Node pointer to the current bottom of our list
          */
-        _updateLowestNode: function () {
+        _getLowestNode: function () {
```
iwehrman: `@return {?DOMElement}`, or something like that.

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -379,54 +417,47 @@ define(function (require, exports, module) {
...
-
-                        if (dragTarget) {
-                            // Should be false if the layer and the layers children do not appear in this list
-                            shouldPassDragTarget = dragTarget.some(function (dragT) {
-                                return layer && doc.layers.descendants(layer).indexOf(dragT) !== -1;
-                            });
-                        } else if (this.state.futureReorder) {
+                layerComponents = doc.layers.allVisibleReversed
+                    .map(function (layer, visibleIndex) {
+                        var dragTarget = this.props.dragTarget,
```
iwehrman: These two lookups/assignments can be hoisted out of the loop I think?

src/js/jsx/sections/pages/PagesPanel.jsx
```diff
@@ -379,54 +417,47 @@ define(function (require, exports, module) {
...
-                                    dropAbove={shouldPassDropTarget && this.props.dropAbove} />
-                            </li>
+                                    dropTarget={isDropTarget}
+                                    dropAbove={!!(isDropTarget && this.state.dropAbove)} />
                         );
                     }, this);
 
+                var layerListClasses = classnames({
+                    "layer-list": true,
+                    "layer-list__dragging": this.props.dragTarget
```
iwehrman: ... and re-used here.

iwehrman:
OK, done with this pass. Mostly just need some minor cleanups before merging. As noted, #1577 not yet resolved, but I'm OK with merging this before that's fixed. But invalid JSDoc is a bridge too far! Back to @designedbyscience for housecleaning.

designedbyscience:
Oh, I regressed that #1577 fix and I think I know what I did...

iwehrman:
Layer selection speed seems on par with standard PS, btw. Very nicely done. It's gonna be like greased lightning once we land all the other optimizations :sunglasses:

src/js/models/layerstructure.js
designedbyscience: Pretty sure `O(n)` based on my testing and (not particularly useful) reading 'round the Internet

designedbyscience:
1) Did not fix #1577 yet...some initial condition problem
2) For Draggable.shouldComponentUpdate/LayerFace.shouldComponentUpdate: I define it outside the createClass call because I want to use further up in the composition (and I actually don't need to call it on LayerFace at all. However, if I were to make another composed Droppable, I'd need a different function, so I define it in LayerFace. Moving to abstract class for Droppable and Draggable doesn't seem to make as much sense to me because of the multi-inheritance problem. (Draggable shouldn't inherit Droppable because that's not always the case, even those it is in this situation). 

On the flip side, I could try making Draggable the outer-most composed Component, make a Droppable abstract class of sorts and have LayerFace implement that. To me, it seems odd to have some abstract classes and some things being composed when they are all components... I realize that some of this confusion already exists because we are also using mixins still.

iwehrman:
1. ok
2. re: abstract class thing: that was definitely not a well reasoned suggestion. It seems perfectly fine as is.

Anyway, this is great. Bombs away!

PR #1629, opened 2015-06-21
Finish conversion to multiGetRefs 

iwehrman:
This PR makes full use of multiGetRefs, added in adobe-photoshop/spaces-adapter#110, for a variety of modest performance gains. `layers.addLayers`, for example, takes about 75% as long as on master. But, it all adds up!

Unless you're fetching an arbitrary set of properties from an arbitrary set of references, there isn't a good reason now to use `batchGet` and friends. Instead, choose something like `getPropertiesRange` if you're getting a fixed set of properties from a contiguous range of references, or `batchMultiGetProperties` to get a fixed set of properties from an arbitrary set of references.

Relies on adobe-photoshop/spaces-adapter#110.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -107,30 +107,16 @@ define(function (require, exports) {
      * @return {Promise.<object>}
      */
     var _getDocumentByRef = function (reference, properties, optionalProperties) {
-        var makeRefObj = function (property) {
-            return {
-                reference: reference,
-                property: property
-            };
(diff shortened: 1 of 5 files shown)


volfied:
This change looks good, but I have some review concerns on https://github.com/adobe-photoshop/spaces-adapter/pull/110

iwehrman:
Ready 2 rock :guitar: :rage1:

PR #1630, opened 2015-06-21
Remove artificial delays from zoom commands

iwehrman:
With the overlay cloaking functionality, bounds overlays are automagically cleared by the plugin on every "scroll" event. The zoom and centering commands always trigger a scroll event, so the current 50ms artificial delays in those commands (which were added to give the plugin time to repaint the CEF surface before changing the canvas) are no longer necessary.

It would be great to also remove these delays from the document open/close/selection actions, but those are not guaranteed to generate a scroll event. (They seem to do so only if the canvas is positioned or zoomed differently from one document to the next.) This will require some sort of core change, which I'll file a Watson bug for separately.

When testing, please confirm that there are no ghostly bounds when performing any of the zooming, centering or repositioning commands.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -103,17 +103,23 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var setOverlayCloakingCommand = function () {
-        var centerOffsets = this.flux.store("ui").getState().centerOffsets,
-            windowWidth = window.document.body.clientWidth,
-            windowHeight = window.document.body.clientHeight,
-            cloakRect = {
-                left: centerOffsets.left,
-                top: centerOffsets.top,
-                bottom: windowHeight - centerOffsets.bottom,
-                right: windowWidth - centerOffsets.right
-            };
+        var uiStore = this.flux.store("ui"),
+            cloakRect = uiStore.getCloakRect();
 
         return adapterUI.setOverlayCloaking(cloakRect, ["scroll"], "afterPaint");
+    };
+
+    /**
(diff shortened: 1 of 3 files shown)


iwehrman:
**UPDATE**: With 34013b8, this also eliminates the artificial delay from before the `close` and `select` (next/previous) document commands. Unlike the zoom and centering commands (and the open and new commands), these don't trigger events that we can listen to because they're played by the plugin. Instead, we solve the ghostly bounds problem by cloaking immediately before executing those commands.

This shaves ~50ms off the `close` and `select` (next/previous) commands.

Please also test the open, close and next/previous document commands w.r.t. to the bounds overlays when testing this PR.

src/js/actions/ui.js
```diff
@@ -97,26 +97,49 @@ define(function (require, exports) {
     };
 
     /**
+     * Get the current cloaking rectangle, which omits the static UI.
+     *
+     * @private
+     * @return {{top: number, right: number, left: number, bottom: number}}
+     */
+    var _getCloakRect = function () {
```
volfied: This function should be in the UI Store.

volfied:
One nit, changes look good.

src/js/actions/ui.js
iwehrman: OK! Will fix up ASAP.

iwehrman:
Second of two commits amended with the fix to move `getCloakRect` to the UI store. Ready for re-review.

volfied:
This willb e useful for my off-canvas-transform work!

PR #1635, opened 2015-06-22
Libraries Panel initial work

volfied:
This is for review only, it will not work on any machine (without putting the library in the right place / setting port correctly).

Final PR will include comments too, so please ignore lack of commentary.

@iwehrman @mcilroyc Can you guys do a general architecture review on this before I go further.

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -27,6 +27,7 @@ require.config({
     baseUrl: ".",
     packages: [{ name: "adapter", location: "../bower_components/spaces-adapter/src" }],
     paths: {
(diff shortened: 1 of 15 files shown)


src/js/config.js
```diff
@@ -27,6 +27,7 @@ require.config({
     baseUrl: ".",
     packages: [{ name: "adapter", location: "../bower_components/spaces-adapter/src" }],
     paths: {
+        "ccLibraries": "../bower_components/ccLibraries/cc-libraries-api.min",
```
volfied: This will be removed, and library will be loaded from `file://shared`, initial Websocket supporting CEF I had had an issue with file:// links.

Issue #1599, opened 2015-06-17
Type tool using modifier access standard PS move tool, results in bounds separation


chadrolfs:
create some shapes, select the Type tool, hold down cmd/ctrl key and move a shape

Result: shape and bounds are now off

PR #1639, opened 2015-06-23
Listen to move events during type tool

volfied:
Addresses #1599. 

I didn't want to disable this because I see use in it, so instead I'm handling it gracefully.

### DIFF
--- a/src/js/tools/type.js
+++ b/src/js/tools/type.js
@@ -44,6 +44,11 @@ define(function (require, exports, module) {
             firstLaunch = true;
             
         var selectHandler = function () {
+            // If this is set, means we didn't get to deselect the tool last time
+            if (_moveHandler) {
+                descriptor.removeListener("move", _moveHandler);
+            }
+
             _moveHandler = function () {
                 var documentStore = this.flux.store("application"),
                     currentDocument = documentStore.getCurrentDocument();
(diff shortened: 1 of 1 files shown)


src/js/tools/type.js
```diff
@@ -38,13 +44,26 @@ define(function (require, exports, module) {
...
+
             if (firstLaunch) {
                 firstLaunch = false;
                 return descriptor.batchPlayObjects([resetObj]);
             }
         };
 
-        Tool.call(this, "typeCreateOrEdit", "Type", "typeCreateOrEditTool", selectHandler);
+        var deselectHandler = function () {
+            descriptor.removeListener("move", _moveHandler);
```
iwehrman: This could get leaked on reset. Should we change `actions.tools.onReset` to call the active tool's deselect handler? Or does it already?

volfied: beforeStartup calls initTool which calls selectTool, which calls deselect on the tool. Right?

volfied: It does not. I'll see what I can do.

src/js/tools/type.js
```diff
@@ -38,13 +44,32 @@ define(function (require, exports, module) {
             firstLaunch = true;
             
         var selectHandler = function () {
+            // If this is set, means we didn't get to deselect the tool last time
+            if (_moveHandler) {
```
volfied: @iwehrman This is best I can do, during error caused reset, deselect handler does not get called.

iwehrman:
Looks great, thanks!

Issue #1641, opened 2015-06-23
Assertion failure in defaults


iwehrman:
```
Assertion failed: select is passed a non-layer referenceutil.js:52
assertlayer.js:206 
selecttools.js:70 
installShapeDefaultsCommandfluxcontroller.js:280 
FluxController._applyActionCommandfluxcontroller.js:216 
Object.create.transfer.valuerectangle.js:58 
selectHandlertools.js:170 
(anonymous function)bluebird.js:4550 
tryCatcherbluebird.js:2689 
Promise._settlePromiseFromHandlerbluebird.js:2765 
Promise._settlePromiseAtbluebird.js:207 (anonymous function)
```

The assertion was triggered because no layers were selected, which caused the reference used to construct the action descriptor to only target the document and not any layers.

iwehrman:
Fixed by #1642.

PR #1642, opened 2015-06-23
 fix for issue 1641 dont call select on empty layer set

pineapplespatula:
fix for issue #1641

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -57,10 +57,15 @@ define(function (require, exports) {
         // If document doesn't exist, or is a flat document
         if (!document || document.unsupported || document.layers.all.size === 1 &&
             document.layers.all.first().isBackground) {
-            return descriptor.PlayObject(defaultObj);
+            return descriptor.playObject(defaultObj);
         }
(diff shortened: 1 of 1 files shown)


src/js/actions/tools.js
```diff
@@ -59,8 +59,13 @@ define(function (require, exports) {
             document.layers.all.first().isBackground) {
             return descriptor.PlayObject(defaultObj);
         }
-        var layerSpec = document.layers.allSelected.toList(),
-            layerRef = layerSpec
+        var layerSpec = document.layers.allSelected.toList();
+
+        if (layerSpec.isEmpty()) {
+            return descriptor.PlayObject(defaultObj);
```
iwehrman: Shouldn't that be `descriptor.playObject`? Above too?

iwehrman: Also, why do we pass in layer references here if it suffices to just pass in the document reference?

pineapplespatula: where are we passing layer references?

iwehrman:
LGTM, thanks!

PR #1643, opened 2015-06-24
Get JSDoc to run alongside linter

volfied:
Does not fully create the documentation yet, but checks for valid JSDoc on `grunt`.


iwehrman:
Most excellent. Note to everyone: run `npm install` to make `grunt` work after this.

PR #1646, opened 2015-06-25
Unbreak the testrunner

iwehrman:
This unbreaks the testrunner. Not all tests pass, but at least the failures do not result in unhandled exceptions.

If it weren't for the vaguely educational value, I'd remove all the failing tests straightaway.

So much shame here :sheep:


volfied:
Ok!

PR #1647, opened 2015-06-25
Get rid of action objects, instead define properties of action functions

volfied:
While setting up JSDoc, I've realized that our way of defining actions is very incompatible with JSDoc. This PR gets rid of all the pesky action objects defined at the end of each action file, and instead relies on the "command" function definitions to be the actions. 

The reads/writes/modal etc. like properties are directly defined next to the function object.

This will require rebase on every other branch if it goes in first, and I hope it does, because getting rid of all those objects was a pain.

I also had to make couple extra changes to make sure that our actions were named same as their exported names. And I alphabetized the action index.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -202,7 +202,7 @@ define(function (require, exports, module) {
              */
             transfer: {
                 value: function (nextAction) {
-                    if (!nextAction) {
+                    if (!nextAction || (typeof nextAction !== "function")) {
                         throw new Error("Transfer passed an undefined action");
                     }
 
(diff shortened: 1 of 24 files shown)


src/js/fluxcontroller.js
```diff
@@ -202,8 +202,8 @@ define(function (require, exports, module) {
              */
             transfer: {
                 value: function (nextAction) {
-                    if (!nextAction || !nextAction.hasOwnProperty("command")) {
-                        throw new Error("Incorrect next action; passed command directly?");
+                    if (!nextAction) {
```
iwehrman: Maybe we should test to make sure that this is at least a function here?

iwehrman:
Regardless of JSDoc, I like this change because it reduces boilerplate and keeps the action properties and the command definition tied tightly together. Sorry for the trouble @designedbyscience, @hamburge, @pineapplespatula and @mcilroyc. Let one of us know if you have questions about how to update your branches.

PR #1651, opened 2015-06-25
Immutilate application state

iwehrman:
This replaces all the mutable `ApplicationStore` data structures with `Immutable` ones. This fixes one untracked [bug](https://github.com/adobe-photoshop/spaces-design/blob/063ed2575864b429e258e9e2ea753e78b9db2061/src/js/jsx/Properties.jsx#L94), and is necessary for an upcoming performance optimzation.

Please test things like switching among open documents and going to and from a no-doc state, and also the recent files list.


volfied:
LGTM

Issue #1650, opened 2015-06-25
Text Transform: Type layers don't report their actual bounding boxes, unable to rotate


volfied:
If a type layer is created by dragging a rectangle, we still only get the bounds of the text itself, but transform controls show over the entire bounding box.

chadrolfs:
does this need a watson bug?

volfied:
I'm fixing it.

volfied:
CL 1028819 , will be playground-dev-mac#370.

volfied:
Also requires #1652

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Maybe they do but there are still problems.  Here are results on Win (hiDPI) and 2nd rotation on mac fails.  https://www.dropbox.com/s/d2qhymeulqscm56/capture_Sat_Jun_27_19.02.32.mkv?dl=0

chadrolfs:
Looks like there's a conflict with mr. quee.  repro case may include, rotate, move, then rotate again (fail).

chadrolfs:
fixed by #1697

chadrolfs:
better but still not totally fixed.  Still get into cases where it fails, like this:
![screen shot 2015-07-10 at 2 32 22 pm](https://cloud.githubusercontent.com/assets/8442631/8629455/6f0fb6de-2710-11e5-8bd6-54a104de9ca3.png)

chadrolfs:
This is harder to hit now, lowering priority.  It also seems that switching to standard and back corrects this if you get in to this state.

chadrolfs:
This has gotten worse and not isolated to just text now.  seems like subsequent attempts after initial rotation/scale now drop selection and fail.

iwehrman:
Adding this to the S26 milestone, since this seems like something we should try to address sooner rather than later.

volfied:
@chadrolfs now that latest fixes are in. Can you review this bug? Please make sure you're on a release build as postconditions on layers.select hampers the policy reset.

chadrolfs:
Still fails after a few rotations.
https://www.dropbox.com/s/8c6ub6xp4tm8nfr/TextBoxRotateFail.mov?dl=0

volfied:
This happens due to superselect installing it's policies AFTER border policies. The fix will be either merging the two policies, or figuring out why that's happening and fixing it.

ktaki:
[...]

Issue #370, opened 2015-06-12
Type: unable to apply some fonts


DesignSpaceBot:
http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png

-- @chadrolfs at [2015-01-07T23:44:51Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/370)

PR #1652, opened 2015-06-25
Use boundingBox for type layers

volfied:
Will alleviate #1650 alonside with Jenkins build #370. We expose sheet bounds for type layers in layer descriptor in `boundingBox`


iwehrman:
Works great!

Issue #1255, opened 2015-06-12
Selecting another tool from Type edit mode fails back to super select


DesignSpaceBot:
Create a type layer.  With super select tool, double-click on on-canvas Type layer to enter edit mode, click on an alternate tool (e.g. Rectangle tool).

Result: you get the Rect tool for a split second and then kicked back to super select.
Expected: text is committed and the tool you selected is what you are left with.

-- @chadrolfs at [2015-04-21T22:33:38Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1255)

iwehrman:
FBNC to @chadrolfs.

PR #1656, opened 2015-06-26
Find functions for queue / prevent type=>select tool on user initiated tool change

volfied:
@iwehrman About f58a44e... I kept it in a separate commit so it's put under scrutiny. This is the only solution I could find to #1255.

The select caused by superselectTypeTool `deselectHandler` was being enqueued before rectangle tool switch happened, so it would enqueue a superselect switch as well, this prevents that.

Also renamed `tools.select` to `tools.selectTool` so exports matches function name

### DIFF
--- a/src/js/util/async-dependency-queue.js
+++ b/src/js/util/async-dependency-queue.js
@@ -38,8 +38,9 @@ define(function (require, exports, module) {
      * @param {function (): Promise} fn
      * @param {Array.<string>} reads
      * @param {Array.<string>} writes 
+     * @param {string} name
      */
-    var Job = function (fn, reads, writes) {
+    var Job = function (fn, reads, writes, name) {
         // Ensure that the read set subsumes the write set
         reads = _.union(reads, writes);
 
@@ -54,6 +55,7 @@ define(function (require, exports, module) {
         this.reads = reads;
         this.writes = writes;
         this.deferred = deferred;
+        this.name = name;
     };
 
     /**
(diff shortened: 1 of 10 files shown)


src/js/util/async-dependency-queue.js
```diff
@@ -358,5 +365,27 @@ define(function (require, exports, module) {
         return Object.keys(this._current).length;
     };
 
+    /**
+     * Find the first job with the given name in the pending queue
+     *
+     * @param {string} actionName module+action name e.g. "tools.selectTool"
+     *
+     * @return {Job}
```
iwehrman: Nit: should be `@return {Job=}` here and below.

src/js/util/async-dependency-queue.js
```diff
@@ -358,5 +365,27 @@ define(function (require, exports, module) {
         return Object.keys(this._current).length;
     };
 
+    /**
+     * Find the first job with the given name in the pending queue
+     *
+     * @param {string} actionName module+action name e.g. "tools.selectTool"
+     *
```
iwehrman: Also, we don't usually have extra NLs between these annotations.

iwehrman:
Well, I can't say I'm excited about this solution. I guess conceptually what would be cleaner would be to wait to select the rectangle tool until after some other event has fired which is triggered after both a) the modal tool state has changed, and b) the async handler for that event has finished resetting the tool. That would preserve all the abstractions, at least. But, I guess this is a simple enough workaround, so let's just stick with it. I'm happy once the JSDoc nits are fixed.

volfied:
Explicitly ending modal state is a winner

Issue #1191, opened 2015-06-12
Pen tool switches to superselect after deleting shape or point


DesignSpaceBot:
select the Pen tool, either single click on canvas to add an anchor point and then hit delete, or create a closed shape with the pen tool and hit delete.

Result: You switch to superselect tool.

-- @chadrolfs at [2015-04-16T18:32:59Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1191)

iwehrman:
FBNC to @chadrolfs.

PR #1657, opened 2015-06-26
Don't fail resetBounds if the batchGet fails

volfied:
During shape edit / pen tool, `resetBounds` would fail if you pressed delete fast enough to delete all the points that layer would get deleted. We would try to `batchGet` bounds of a layer that doesn't exist.

Reset bounds is an updater, and I think it's ok for it to be failsafe.

Addresses #1191

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -442,6 +442,10 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var resetBounds = function (document, layers, noHistory) {
+        if (!layers) {
+            throw new Error("Reset bounds passed invalid layers");
+        }
+
         var propertyRefs = layers.map(function (layer) {
             var property;
             if (layer.isArtboard) {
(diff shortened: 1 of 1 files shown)


iwehrman:
I don't think this is the right place for there to be a blind catch. Instead, we should put the catch in whatever action is transferring to `resetBounds`. It's conceivable that this change is appropriate for _all_ callers of `resetBounds`, but that isn't clear to me at all. So, instead, I would try to keep this as localized as possible.

It would be great if there were some way for us to distinguish among the different error messages, so that the caller of `resetBounds` could test for a specific error and ignore just that one...

volfied:
Problem is, this is not an action, it's an event handler (`"delete"`) that calls the action into the queue. So we can catch it around the action call, but controller also complains about an error and goes into the reset process.

iwehrman:
A dumb way to solve that is just to make another action which transfers to to reset bounds and handles the error if appropriate.

volfied:
@iwehrman Rebased with a new change, please review! I think it makes a lot of sense to call this function "quietly" during notification handlers, as they're not fast enough. After Shutdown Ian can look into making them more responsive :)

iwehrman:
LGTM, thanks!

PR #1658, opened 2015-06-26
End modal state explicitly on toolbar button click

volfied:
@iwehrman Based on our discussion, I found this solution to be simpler and less hacky.

Addresses #1255


iwehrman:
Let's give it a whirl!

PR #1660, opened 2015-06-26
CC Libraries v0.2

volfied:
Adds libraries panel to Design Space. This is going to require some one time changes for things to work and be tested.

MAC OS (make sure PS and other CC apps are not running):
 1. In your Photoshop.app/Contents/Required folder, create a file named `AMTInDebug.txt`, it can be empty. This will force debug builds to run with your Adobe ID.
 2. Install the new CC Library Process that supports web sockets
    1. Download the latest 2.1 CCLibrary process from Codex to local machine (SJ Store `/builds03/cclibs/localstorageservice/2.1.0-372/osx/`)
    2. Rename the `.pima` file to `.zip`
    3. Kill the CCLibrary and Creative Cloud processes (via Activity Monitor app)
    4. Extract the contents of the zip and `.pimx` file to `/Applications/Utilities/Adobe\ Creative\ Cloud/CCLibrary/` and double-click CCLibrary.app to launch
 3. Requires latest Jenkins build, as the CCLibraries API are in www-shared (playground-dev-mac#373)
 4. Requires https://github.com/adobe-photoshop/spaces-adapter/pull/116

Basic functionality is in, right now, only works with image layers, and does not link the layer after it's uploaded. This is a Photoshop bug, but this paves the path for future work.


volfied:
screen shot 2015-07-07 at 10 40 59 am copy

chadrolfs:
@volfied thanks for the quick guide, very helpful.  :+1:

iwehrman:
This isn't blocking the review, but there are some lint errors here. Also, we should squash these commits before merging.

volfied:
@iwehrman Please keep reviewing, right now, I have some front end bugs, namely the panel is not resizing and playing nicely with the other two panels (so for testing I always keep style and layers panel collapsed)

Also, I need feedback on Library.jsx and the way I handle multiple asset types.

src/js/actions/libraries.js
```diff
@@ -0,0 +1,272 @@
...
+                    .then(function (imsInfo) {
+                        var port = imsInfo.port;
+
+                        callback(JSON.stringify({ port: port }));
+                    });
+            }
+        };
+
+        // SHARED_LOCAL_STORAGE flag forces websocket use
+        CCLibraries.configure(dependencies, {
```
iwehrman: Is it safe to call this multiple times? Or do we need to "unconfigure" this on reset?

src/js/actions/libraries.js
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        var payload = {
+            id: currentLibrary.id
+        };
+
+        return Promise.fromNode(function (cb) {
+            libraryCollection.removeLibrary(currentLibrary, cb);
+        }).bind(this).then(function () {
+            this.dispatchAsync(events.libraries.LIBRARY_REMOVED, payload);
```
iwehrman: You probably want to return this.

src/js/actions/libraries.js
```diff
@@ -0,0 +1,272 @@
...
+                var newRepresentation = newElement.getPrimaryRepresentation();
+                return Promise.fromNode(function (cb) {
+                    newRepresentation.getContentPath(cb);
+                });
+            }).then(function (path) {
+                var createObj = libraryAdapter.createElement(currentDocument.id, currentLayer.id, newElement, path);
+                return descriptor.playObject(createObj);
+            }).then(function () {
+                // FIXME: Find a way around this update Document
+                this.flux.actions.documents.updateDocument();
```
iwehrman: Can you make a core Trello card for all the instances in which we need some kind of layer ID in the action descriptor result, or whatever it takes to get rid of the updateDocument calls? If you do that then we can try to find someone else on the core team to add those API features in parallel.

src/js/actions/libraries.js
```diff
@@ -0,0 +1,272 @@
...
+
+        var newElement = currentLibrary.createElement(currentLayer.name, IMAGE_ELEMENT_TYPE),
+            representation = newElement.createRepresentation(REPRESENTATION_TYPE, "primary"),
+            previewSize = {
+                w: 248,
+                h: 188
+            };
+
+        // FIXME: Mac/Win temporary locations!
+        var exportObj = libraryAdapter.exportLayer("/tmp/", "/tmp/preview.png",
```
iwehrman: Yikes. We'll need some way to get temporary paths... Might want to make a trello card for this too?

src/js/events.js
```diff
@@ -137,6 +137,21 @@ define(function (require, exports, module) {
             LOAD_HISTORY_STATE_REVERT: "loadHistoryStateRevert",
             ADJUST_HISTORY_STATE: "adjustHistoryState",
             DELETE_DOCUMENT_HISTORY: "deleteDocumentHistory"
+        },
+        droppable: {
```
iwehrman: I think this is a rebase error. All the droppable events have been removed in favor of direct store access.

src/js/jsx/Properties.jsx
```diff
@@ -115,6 +116,10 @@ define(function (require, exports, module) {
                     <TransformPanel
                         disabled={disabled}
                         document={document} />
+                    <LibrariesPanel
```
iwehrman: I might go so far as to ask that we find a way to disable this stuff by default unless you set a preference? Because, like, we might want to release the next sprint build on the pre-release forum, and although I want to get this merged so that we can get more eyes on it, I also don't want those prerelease builds to be completely broken.

src/style/sections/libraries/libraries-section.less
```diff
@@ -0,0 +1,55 @@
...
+.libraries .section-container__collapsed {
+    display: none;
+}
+
+.libraries.section.section__sibling-collapsed {
+    flex-grow: 999999;
+}
+
+.library-color-swatch {
+	width: 3rem;
```
iwehrman: Dear god is that a TAB CHARACTER!?

iwehrman:
Besides a few small bugs and requests for trello cards, this looks great. I do think I would like to hide this behind a preference before merging it though. It doesn't seem like that should be a lot of work. We'll just require that you manually set some preference before action and the relevant components are loaded. Back 2 U @volfied.

src/js/actions/libraries.js
volfied: Having looked at the source, it shuts down any existing interfaces if they're already up, so it's safe to call repeatedly.

src/js/actions/libraries.js
volfied: Is there a particular reason I need to return an internal promise result, despite having the return in the first promise? Do we care about the result of dispatch?

src/js/actions/libraries.js
volfied: This will be on our side, we need to check to see if the layer is a smart object, and convert it to cloud linked smart object.

src/js/actions/libraries.js
volfied: https://trello.com/c/lIuFmShz I can take care of this in core team's absence.

src/js/events.js
volfied: My bad.

src/js/jsx/Properties.jsx
volfied: https://gist.github.com/volfied/56f1c8c4ee82772d5c35

volfied:
Thanks for the feedback @iwehrman Made the changes. The gist for toggling the panel can be found here:
https://github.com/adobe-photoshop/spaces-design/pull/1660#discussion_r34214155

iwehrman:
Right on.

Issue #1663, opened 2015-06-27
Error loading current master on Windows


chadrolfs:
download and launch windows pg-dev.374 build and connect to master branch
Result: `Uncaught SyntaxError: Unexpected token }`

chadrolfs:
![screen shot 2015-06-27 at 11 19 29 am](https://cloud.githubusercontent.com/assets/8442631/8393286/6beb9d44-1cbe-11e5-944e-923a0a68f64e.png)

chadrolfs:
fixed by f3632ff, thanks @iwehrman

PR #1665, opened 2015-06-27
Lint JSON files with Grunt

iwehrman:
This would have prevented #1663.


volfied:
Only you can prevent lint fires.

iwehrman:
Btw, jshint ought to be sufficient for this, except: jshint/jshint#2488.

Issue #1667, opened 2015-06-29
Cloaking rectangle is out of position


iwehrman:
Any operation that cloaks the non-UI rectangle (panning the canvas, creating a new document, etc.) momentarily takes a chunk out of the toolbar:
![cloak](https://s3.amazonaws.com/f.cl.ly/items/2n171n0G3V1f0Y1u0x1w/Screen%20Recording%202015-06-29%20at%2010.15%20AM.gif)

iwehrman:
Fixed by #1669.

PR #1669, opened 2015-06-29
Correctly update the pinned toolbar width

volfied:
Addresses #1667


iwehrman:
Looks good. I pushed commit 8a5d0ef with a comment about this to @designedbyscience for when he embarks on the upcoming UI refactor.

Issue #1671, opened 2015-06-30
Dragged layers sometimes fall behind descendants


iwehrman:
![drag](https://s3.amazonaws.com/f.cl.ly/items/2W0s150f2q301225213w/Screen%20Recording%202015-06-30%20at%2003.52%20PM.gif)

Issue #1664, opened 2015-06-27
Expanding/collapsing layers panel is slow


iwehrman:
At least one reason why is O(n) calls to action `draganddrop.registerDroppable` for a document with n layers. This can takes >10s on a document with ~750 layers.

Issue #1413, opened 2015-06-12
Switching to big documents is slow.


DesignSpaceBot:
#790 is about initializing big documents from Photoshop state. This is about the performance of switching from one document to another. If the document you're switching to is big, this is slow. For example, switching from an empty document to one with ~750 layers takes 3300ms on a retina MBP. This is mostly time spent in React rendering and updating the DOM.

-- @iwehrman at [2015-05-09T14:31:52Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1413)

iwehrman:
Over to @chadrolfs for evaluation/verification.

PR #1673, opened 2015-07-02
Properties panel performance improvements and refactoring

iwehrman:
This is an unfortunately large pull request that touches a lot of the panel code. I tried but failed to tease it apart any further. A thousand pardons. Here's what it does:

1. `Properties` components are now lazily rendered per-document and left in the DOM for faster document switching. This results in a major performance improvement when switching between large documents. The `Properties` components are children of a new `Panel` component.
2. Many superfluous document checks were removed as a result of `Properties` components always being rendered with a document.
2. The vestigial "pages" terminology has been eliminated in favor of "layers", which is now more accurate and less confusing. In particular, the `PagesPanel` has been renamed to `LayersPanel`, and all the `.pages` CSS selectors and classes have been renamed to `.layers`.
3. The `LayersPanel` is now left in the DOM even when collapsed for faster expansion from a collapsed state with large documents.
4. The `draganddrop` store has been generalized to account for different drop zones, which currently correspond to different active documents.
5. The `draganddrop` actions have been removed in favor of direct store manipulation. I couldn't think of any reason to make those state updates should be asynchronous.
6. The drag handler has been optimized for improved performance in large documents. In particular, a) only one virtual render is required per `mousemove` event during the drag operation, b) calls to `getBoundingRect` are cached, and c) the drop targets in a zone are continually reordered by access recency to reduce the length of the search. 

Drag performance is still not good in big documents. On one huge document, I see 10-12fps (in debug mode) on compared to 7-8fps on master. Some notes on that:

1. Performance is almost completely constrained by React, and so it performs much more slowly in debug mode compared to production mode. I currently see a ~100% speedup in production mode compared debug mode.
2. React performance is still the bottleneck because every position update causes `LayersPanel` to re-render. That is slow because it has to virtually render O(N) `LayerFaces`. Even though those have `shouldComponentUpdate` implementations that are very efficient for this case (thanks @designedbyscience!) this is still the bottleneck, taking tens (>30) of milliseconds. Ideally, we only _only_ need to virtually render the dragging and drop target layers. Implementing could be tricky.
3. I haven't tried to optimize the drop (as opposed to the drag) at all, and that feels very slow to me. On a fast iMac in debug mode it seems to take ~700ms. I'll file a separate bug for that.

Addresses #1671, #1664, #1413.


src/js/jsx/shared/Draggable.jsx
```diff
@@ -182,14 +166,24 @@ define(function (require, exports, module) {
...
-                    this.getFlux().actions.draganddrop.registerDragging(this._getDragItems(getDragItem(this.props)));
+                    if (this.props.onDragStart) {
+                        this.props.onDragStart();
+                    }
+
+                    var dragItems = this.props.getDragItems(this);
+                    flux.store("draganddrop").startDrag(dragItems);
                 } else {
-                    this.getFlux().store("draganddrop").moveAndCheckBounds({ x: event.clientX, y: event.clientY });
+                    flux.store("draganddrop").updateDrag(this.props.zone, {
```
designedbyscience: Could you update getDefaultProps with zone?

src/style/sections/layers/face.less
```diff
@@ -215,6 +215,7 @@ input[type="text"].face__name {
     margin-left: 0;
     width: @properties-width - 4rem;
     border-radius: @face-border-radius;
+    z-index: 100;
```
designedbyscience: Why is this z-index needed?

designedbyscience:
Okay, only minor things

src/style/sections/layers/face.less
iwehrman: This seemed to fix #1671. Obviously, 100 may not be the right value here. Any suggestions?

src/js/jsx/shared/Draggable.jsx
iwehrman: I added a `isRequired` `propType` for `zone` to `Draggable` instead, since there isn't really a valid default here. See de1a254.

designedbyscience:
Oh, found a bug. After moving a layer out of a group (sometimes you need to do this twice), you can then attempt to drag a group into itself (which results in an error on drop)

https://www.dropbox.com/s/u5siwy156gx12w9/DragGroupIntoItself.mov?dl=0

src/style/sections/layers/face.less
designedbyscience: I'd knock it down to 1 (which seems to still work)

iwehrman:
Nice catch! Bug fixed (as well as the wacky z-index) in 636eb35. Back to you, @designedbyscience.

designedbyscience:
Okay, looks good

Issue #1532, opened 2015-06-12
Swap with Artboards returns unexpected results


DesignSpaceBot:
http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png

-- @chadrolfs at [2015-06-04T23:11:57Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1532)

chadrolfs:
Seeing wacky results:
https://www.dropbox.com/s/ob4rhtaewqfgcvu/AB-swap-fail.mov?dl=0
https://www.dropbox.com/s/9ryuas9pgytb5sy/AB-swap-fail2.mov?dl=0

iwehrman:
FBNC again to @chadrolfs. This time for real.

PR #1676, opened 2015-07-06
Separate flip/swap enable logic, and allow flip/swap for artboards

volfied:
Swap enable rules used to be reliant on flip rules, but with artboards correctly working, it's time to separate them. This change will allow artboards to be swapped, and flipped (with their contents)

Addresses #1532


iwehrman:
Okie doke!

PR #1678, opened 2015-07-06
Search using filters for layer or document type

hamburge:
Use tab or enter to select filter option. The filter appears as an svg when you confirm it.
Added autofill of text, other design changes.

Some filters don't have svgs associated with them currently so there is a rectangle placeholder.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -136,13 +136,13 @@ define(function (require, exports, module) {
         /**
          * Updates this.state.filter to be values contained in the filter id
          *
-         * @param {string} id Filter ID
+         * @param {string} id Filter ID. If null, then reset filter to no value
          */
         _updateFilter: function (id) {
-            var idArray = id.split("_"),
+            var idArray = id ? id.split("_") : [],
                 filterValues = _.drop(idArray),
-                updatedFilter = _.uniq(this.state.filter.concat(filterValues)),
-                filterIcons = this._getFilterIcons(updatedFilter);
+                updatedFilter = id ? _.uniq(this.state.filter.concat(filterValues)) : [],
+                filterIcons = svgUtil.getSVGClassesFromLayerTypes(updatedFilter);
                 
             this.setState({
                 filter: updatedFilter,
@@ -484,23 +484,9 @@ define(function (require, exports, module) {
                     break;
                 }
                 case "Backspace": {
-                    // TODO: should change this to check for if the cursor is at beginning of input,
-                    // not just if the input is empty
-                    if (!this.refs.datalist.hasNonEmptyInput() && this.state.filter.length > 0) {
-                        // For when want to have multiple SVGs, remove them one at a time:
-                        // var newFilter = this.state.filter,
-                        //     newIcons = this.state.icons;
-                        // newFilter.pop();
-                        // newIcons.pop();
-
+                    if (this.refs.datalist.cursorAtBeginning() && this.state.filter.length > 0) {
                         // Clear filter and icons
-                        var newFilter = [],
-                            newIcons = [];
-
-                        this.setState({
-                            filter: newFilter,
-                            icons: newIcons
-                        });
+                        this._updateFilter(null);
                     }
                     break;
                 }
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -31,10 +31,13 @@ define(function (require, exports, module) {
         Datalist = require("jsx!js/jsx/shared/Datalist"),
         Immutable = require("immutable");
 
-    var pathUtil = require("js/util/path"),
+    var layerLib = require("adapter/lib/layer"),
+        pathUtil = require("js/util/path"),
         mathUtil = require("js/util/math"),
         collection = require("js/util/collection");
-        
+    
+    var MAX_OPTIONS = 10;
+
     var SearchBar = React.createClass({
         mixins: [FluxMixin],
 
@@ -45,6 +48,13 @@ define(function (require, exports, module) {
         getDefaultProps: function () {
             return {
                 dismissDialog: _.identity
+            };
+        },
+
+        getInitialState: function () {
+            return {
+                filter: [],
+                icons: []
             };
         },
   
@@ -70,12 +80,16 @@ define(function (require, exports, module) {
                 return;
             }
 
+            // ID has type as first word, followed by the layer/document ID or file name
             var idArray = id.split("_"),
                 type = idArray[0],
                 idInt = mathUtil.parseNumber(idArray[1]),
                 flux = this.getFlux();
 
             switch (type) {
+            case "filter":
+                this._updateFilter(id);
+                return;
             case "layer":
                 var document = flux.store("application").getCurrentDocument(),
                     selected = document.layers.byID(idInt);
@@ -103,25 +117,62 @@ define(function (require, exports, module) {
         },
 
         /**
-         * Get the class name for the layer face icon for the layer
+         * Updates this.state.filter to be values contained in the filter id
+         *
+         * @param {string} id Filter ID
+         */
+        _updateFilter: function (id) {
+            var idArray = id.split("_"),
+                filterValues = _.drop(idArray),
+                updatedFilter = _.uniq(this.state.filter.concat(filterValues)),
+                filterIcons = this._getFilterIcons(updatedFilter);
+                
+            this.setState({
+                filter: updatedFilter,
+                icons: filterIcons
+            });
+
+            this.refs.datalist.resetInput(idArray, filterIcons.length);
+        },
+
+        /**
+         * Get the class name for the layer face icon for the layer.
+         * This is used both for finding the icons within the drop down and
+         * to find the icons for filters in the search bar.
+         *
+         * As the filters available become more complicated, might want to make
+         * getting the list icons a separate, simpler function.
          *
          * @private
-         * @param {Layer} layer
-         * @return {string}
+         * @param {string} layerKind
+         * @return {Array.<string>}
          */
-        _getSVGInfo: function (layer) {
-            var iconID = "layer-";
-            if (layer.isArtboard) {
-                iconID += "artboard";
-            } else if (layer.kind === layer.layerKinds.BACKGROUND) {
-                iconID += layer.layerKinds.PIXEL;
-            } else if (layer.kind === layer.layerKinds.SMARTOBJECT && layer.isLinked) {
-                iconID += layer.kind + "-linked";
-            } else {
-                iconID += layer.kind;
-            }
-            
-            return iconID;
+        _getSVGInfo: function (layerKind) {
+            var iconIDs = [],
+                isLinked = _.has(layerKind, "linked");
+
+            _.forEach(layerKind, function (kind) {
+                var iconID = "layer-";
+
+                // No svg for these?
+                if (kind === "solidcolor" || kind === "gradient" || kind === "pattern") {
+                    iconID = "tool-rectangle";
+                } else if (kind === "artboard") {
+                    iconID += "artboard";
+                } else if (kind === "background") {
+                    iconID += layerLib.layerKinds.PIXEL;
+                } else if (kind === "smartobject" && isLinked) {
+                    iconID += layerLib.layerKinds.SMARTOBJECT + "-linked";
+                } else if (kind !== "layer") {
+                    iconID += layerLib.layerKinds[kind.toUpperCase()];
+                }
+
+                if (kind !== "layer") {
+                    iconIDs.push(iconID);
+                }
+            });
+
+            return iconIDs;
         },
 
         /**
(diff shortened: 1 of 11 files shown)


iwehrman:
The first order of business is to squash your long list of commits into a single commit with a nice, direct, meaningful commit message. Here's an article that describes how to squash all the commits in this branch into a single commit: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

The second order of business is to fix the merge conflicts. Note Github's warning above: 
merge
This means that master has changed since you originally branched from it in such a way that git cannot figure out how to automatically apply your changes. In this case, the way to solve that problem is by rebasing your branch against the current HEAD of master (instead of off wherever was master when you originally branched). Do this roughly as follows:
```
# switch to master
$ git checkout master
# update your local copy of master w.r.t. the remote version
$ git fetch origin
$ git merge origin/master
# rebase your branch against your local copy of master
$ git checkout allison/search_filters
$ git rebase master
```
This last step will probably cause some conflicts that have to be fixed manually. Feel free to ask for help on this step if it's not clear how to do so. The basic idea will be to 1) edit the files to get rid of the >>> conflict markers, 2) `git add` each edited file, and once they're all fixed 3) do `git rebase --continue`. If something goes south you can always do `git rebase --abort` to get back to where you were.

At this point, you'll have rewritten the commits in your branch. When you try to push your branch to `origin`, it will complain because your local branch is not merely an extension of your remote branch. This is not actually surprising because you've intentionally rewritten your commits to squash them and resolve conflicts. Consequently, you'll tell git to just ignore the inconsistency and take your local version:
```$ git push origin +allison/search_filters```
Where the `+` means "force the remote to accept the local branch".

More details about rebasing can be found in this really great, free book:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

Good luck, and let me (or anyone else in Slack) know if you get stuck.

**Comment by hamburge** [comment]
@iwehrman These orders of business should be good to go now.

File: src/js/jsx/search/SearchBar.jsx [hunk 1]
```diff
@@ -31,10 +31,13 @@ define(function (require, exports, module) {
...
         Immutable = require("immutable");
 
-    var pathUtil = require("js/util/path"),
+    var layerLib = require("adapter/lib/layer"),
+        pathUtil = require("js/util/path"),
         mathUtil = require("js/util/math"),
         collection = require("js/util/collection");
-        
+    
+    var MAX_OPTIONS = 10;
```
iwehrman: JSDoc annotation please, including `@const` and `@type` annotations.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -142,6 +193,36 @@ define(function (require, exports, module) {
...
         /**
+         * Get the layer type
+         *
+         * @private
+         * @param {Layer} layer
+         * @return {Array.<string>}
+         */
+        _getLayerType: function (layer) {
+            var layerType = ["layer"];
+            _.forEach(Object.keys(layer.layerKinds), function (kind) {
```
iwehrman: The `layerKinds` property refers to the same object on every layer which describes ALL the available layer kinds. It's a convenience property that's kind of confusing. The `kind` property on layer models describes their individual kind.

iwehrman: That is to say, I'm either confused about what this function is for, or it should actually return just a single kind? If not, then this information ought to be static and computed only once?

hamburge:
@iwehrman Just pushed another commit that adds the JSDOC, refactors the svg code, localizes the strings, and also gets rid of a warning I was seeing. It happens to get rid of the loop through layer kinds as well. 
Let me know if I should squash that commit in as well. :jack_o_lantern:

hamburge:
...and now strings should actually be all localized.

src/js/jsx/shared/TextInput.jsx
```diff
@@ -272,8 +272,9 @@ define(function (require, exports, module) {
          * If the value is editable, goes into edit mode
          *
          * @private
+         * @param {bool} select Whether select should be enabled or not
          */
-        _beginEdit: function () {
+        _beginEdit: function (select) {
```
iwehrman: It would be better to flip this parameter so that "true" means select is disabled, and to make the boolean parameter optional: `@param {boolean=}`. Then all the current callers don't need to be changed, and the default remains sensible, but it's possible for your special case behavior to override the default.

src/js/util/svg.js
```diff
@@ -0,0 +1,89 @@
...
+        layerLib = require("adapter/lib/layer");
+   
+    /**
+     * Get the class name for the layer face icon for the given layer
+     *
+     * @private
+     * @param {Layer} layer
+     * @return {string}
+    */
+    var getSVGFromLayer = function (layer) {
```
iwehrman: I like that this is factored out into a helper module, but the naming is confusing. It seems like this is constructing class names. Is there a different name for the function/module that would make it more clear what this is used for?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -142,6 +170,29 @@ define(function (require, exports, module) {
         },
 
         /**
+         * Get the layer type and if it is linked or an artboard, as an array of strings
```
iwehrman: Can you elaborate here what you're returning? For example, I think these are user-visible strings that describe something, but I'm not sure how they might be used.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -250,41 +303,232 @@ define(function (require, exports, module) {
...
+         * 
+         * @param {string} header, either "layer" or "document"
+         * @return {Array.<object>}
+         */
+        _getFilterOptions: function (header) {
+            if (header !== "layer" && header !== "document") {
+                return;
+            }
+            var categoryList = header === "document" ? ["current", "recent"] : layerLib.layerKinds,
+                categories = Immutable.fromJS(categoryList).filterNot(function (kind) {
```
iwehrman: For clarity (and to avoid accidentally processing deep structures) we avoid using `Immutable.fromJS` and `toJS` in production code in favor of the more explicit constructor functions like `Immutable.List(categoryList)`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -250,41 +303,232 @@ define(function (require, exports, module) {
...
+
+        /**
+         * Make list of search category dropdown options based on header
+         * 
+         * @param {string} header, either "layer" or "document"
+         * @return {Array.<object>}
+         */
+        _getFilterOptions: function (header) {
+            if (header !== "layer" && header !== "document") {
+                return;
```
iwehrman: This should either throw an exception or just be removed. Returning here will just cause a failure elsewhere. An explicit exception is fine, but also for private functions we generally don't feel like it's necessary to do error checking like this up front unless the correctness conditions of the parameters are obscure.

src/js/jsx/search/SearchBar.jsx
iwehrman: Also, the tabbing is wonky here. For readability, it might be better to break this up into a few different variable declarations.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -250,41 +303,232 @@ define(function (require, exports, module) {
...
+                                   
+                filters = categories.map(function (kind) {
+                    var title,
+                        idType;
+
+                    if (header === "layer") {
+                        idType = CATEGORIES.LAYER_KIND[kind];
+                        title = CATEGORIES.LAYER_KIND[kind];
+                    } else {
+                        idType = CATEGORIES[kind.toUpperCase()];
```
iwehrman: Is there some way to get rid of all this case manipulation? Can the keys just all be made upper case, for example?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -250,41 +303,232 @@ define(function (require, exports, module) {
...
+                }
+
+                // Always add headers to list of searchable options
+                // The check to not render if there are no options below it is in Select.jsx
+                if (option.type === "header") {
+                    count = 0;
+                    return true;
+                }
+
+                if (count === MAX_OPTIONS) {
```
iwehrman: Since you're limiting to a relatively small `MAX_OPTIONS`, it might be worthwhile to short-circuit this search once the list is full. You could use `.some` instead of `.filter` for that.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -41,7 +43,8 @@ define(function (require, exports, module) {
             options: React.PropTypes.instanceOf(Immutable.Iterable),
             live: React.PropTypes.bool,
             startFocused: React.PropTypes.bool,
-            placeholderText: React.PropTypes.string
+            placeholderText: React.PropTypes.string,
+            useAutofill: React.PropTypes.bool
```
iwehrman: Can you document somewhere what this new property does? We should be better generally about document state and props in React components...

src/js/jsx/shared/Datalist.jsx
```diff
@@ -58,16 +62,34 @@ define(function (require, exports, module) {
             return {
                 active: false,
                 filter: null,
-                id: this.props.defaultSelected
+                id: this.props.defaultSelected,
+                suggestTitle: "",
+                width: 0
```
iwehrman: Same here. Just a little // comment on your new state elements would be great.

iwehrman: image

src/nls/root/strings.js
```diff
@@ -569,6 +569,39 @@ define(function (require, exports, module) {
...
+            CATEGORIES: {
+                DOCUMENT: "document",
+                CURRENT: "current",
+                RECENT: "recent",
+
+                LAYER: "layer",
+                ARTBOARD: "artboard",
+                LINKED: "linked",
+                LAYER_KIND: {
+                    0: "any",
```
volfied: these don't need to be localized, they're coming to us from PS and they're locale agnostic

src/js/jsx/shared/Datalist.jsx
```diff
@@ -281,65 +328,226 @@ define(function (require, exports, module) {
...
-                    // Always add headers to list of searchable options
-                    if (option.type && option.type === "header") {
-                        return true;
+                searchableOptions = this._filterOptions(searchableFilter);
+            
+            var autocomplete,
+                hiddenTI;
+
+            if (this.props.useAutofill) {
+                hiddenTI = (
```
iwehrman: Can you add a few // comments here about how this is changing the render method? I'm not exactly sure what all the string manipulation is doing below.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -281,65 +328,226 @@ define(function (require, exports, module) {
...
+                    <div ref="hiddenTextInput"
+                        className="hiddeninput">
+                    </div>
+                );
+                
+                var autocompStyle = { left: this.state.width + "px" },
```
iwehrman: Also, it _feels_ a little odd to have a `width` prop here. Is this something the caller really wants to change at runtime? Can you explain why this is the right place for this parameter?

iwehrman:
OK, done with a second pass! Mostly I have questions.

@chadrolfs: would you mind doing a testing pass on this branch, focusing on our _existing_ uses of the datalist component? So, the type faces and styles, blend modes, etc.

src/js/jsx/search/SearchBar.jsx
hamburge: I was using `MAX_OPTIONS` to mean the number of options per category. So if it was 3, you'd only see 3 layers, 3 recent documents, and 3 current documents. Otherwise when you haven't typed much, you'll just see options for layers because they come first.

iwehrman: Yes, but you're still traversing ALL the options. `.some` allows you to cut off the traversal early.

chadrolfs:
1. needs rebase. has close doc error 
`Action documents.close failed: TypeError: Cannot read property 'delete' of undefined`

2. Open a doc, cmd+f, hit Tab, then click on the search icon (magnifier glass next to search field) I get:
`Uncaught TypeError: Cannot read property 'indexOf' of null`
or 
`Uncaught TypeError: Cannot read property 'split' of null`

3. Typing in datalist fields (e.g. Typeface) doesn't seem to work and results in:
`Uncaught TypeError: Cannot read property 'length' of undefined`

src/js/jsx/shared/Datalist.jsx
hamburge: Yeah, so this is adjusting the positioning of the suggestion text. As the user types, the text needs to move over letter by letter.

Maybe calling it `position` makes mores sense?

src/nls/root/strings.js
hamburge: These ones get compared to user input to determine if an option should be included in the dropdown. Since that depends on user input, doesn't that mean it varies by location?

src/js/jsx/search/SearchBar.jsx
hamburge: But if it exits early, it would only get the 3 layers, not the other categories. It needs to keep going to find the start of the next category.

hamburge:
@iwehrman @chadrolfs back to you!

hamburge:
@chadrolfs Also I think this branch addresses issue #1682

iwehrman:
Needs another rebase against master to resolve merge conflicts.

iwehrman:
Bombs away!

Issue #1390, opened 2015-06-12
Toolbar icon underline is misaligned


DesignSpaceBot:
On a non-retina display:
![underline](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @iwehrman at [2015-05-06T15:50:57Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1390)

designedbyscience:
So in the UI Overhaul, this selected state changes to the highlight color, which I will shortly submit a PR for

designedbyscience:
Uh, so fixed in that the design is changing but not closed

chadrolfs:
@designedbyscience did you wanna leave this open or close it?  I'm in favor of closing it unless it'll come back.

designedbyscience:
I think close it

PR #1685, opened 2015-07-07
Update selected state for toolbar icons

designedbyscience:
This relates to #1390. As part of the UI update, the selection state for the tools will be the highlight blue. Rather than fix the underline, I just switched to the blue


mcilroyc:
:+1:

PR #1696, opened 2015-07-09
handle moveToArtboard events by reordering layers, and amending the current history state

mcilroyc:
we now can guarantee that moveToArtboard (MTA) event notifications will always come after the corresponding "move" event.  This is not ideal, but its at least consistent.  If we get an MTA, that means we need to fetch the layer order from photoshop.  This needs to be updated on the newly created history state, which is a new concept in the history store that I call "amendment".  

I've also consolidated this "move" listener logic in transform.js.  We might be able to take this further, but baby steps.  Please check this out @volfied since we've talked about it some, and you are working in the same circus ring.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1122,9 +1122,10 @@ define(function (require, exports) {
      *
      * @param {Document} document Document for which layers should be reordered
      * @param {boolean=} suppressHistory if truthy, dispatch a non-history-changing event.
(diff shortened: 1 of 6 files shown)


volfied:
Looks good, but I'd like to test this a bit more to remember why we get layer order on "move" events. That was not by accident IIRC. Will merge latest tomorrow and rebase my PR for marquee.

volfied:
:+1:

PR #1697, opened 2015-07-09
Marquee select performance pt. 1

volfied:
Addresses general marquee performance issues.

 1. Register/deregister dialog actions each take about 100 ms, which delays marquee by a lot, now they directly edit the store
 1. Transform handlers in transform actions are now debounced, so we don't constantly reset selection/bounds and layer order. This helps in cases where the user is quickly drag/dropping layers
 1. Resetborderpolicies is debounced, and it does not get called if we're in a marquee select mode
 1. Removed post condition for deselectAll, because it was adding a 50ms delay to marquee select starting code
 1. If an action is not modal, and we're trying to end modal tool state, we no longer completely fail when the API fails. This happens if you transform layers too fast and we're in a modal state during resizing/rotating.

@iwehrman please review.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -42,7 +42,8 @@ define(function (require, exports) {
         locking = require("js/util/locking"),
         layerActionsUtil = require("js/util/layeractions"),
         headlights = require("js/util/headlights"),
-        strings = require("i18n!nls/strings");
+        strings = require("i18n!nls/strings"),
+        synchronization = require("js/util/synchronization");
 
     /**
      * play/batchPlay options that allow the canvas to be continually updated.
@@ -1098,7 +1099,7 @@ define(function (require, exports) {
         _moveListener;
 
     var beforeStartup = function () {
-        _transformHandler = function () {
+        _transformHandler = synchronization.debounce(function () {
             this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
             
             // Since finishing the click, the selected layers may have changed, so we'll get
@@ -1110,7 +1111,7 @@ define(function (require, exports) {
             // all selected layers, but due to a recent bug, "move" event sends us the displacement
             // of layers from the changing (0,0) coordinates, which causes bugs like
             // getting (650,0) when the move was actually (-100, 0) for a 750 px wide layer
-            this.flux.actions.layers.getLayerOrder(nextDoc, true)
+            return this.flux.actions.layers.getLayerOrder(nextDoc, true)
                 .bind(this)
                 .then(function () {
                     nextDoc = appStore.getCurrentDocument();
@@ -1119,9 +1120,9 @@ define(function (require, exports) {
                     nextDoc = appStore.getCurrentDocument();
                     this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
                 });
-        }.bind(this);
+        }, this);
 
-        _moveListener = function () {
+        _moveListener = synchronization.debounce(function () {
             this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
             // Since finishing the click, the selected layers may have changed, so we'll get
             // the most current document model before proceeding.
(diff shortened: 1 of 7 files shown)


src/js/jsx/shared/Dialog.jsx
```diff
@@ -98,7 +98,7 @@ define(function (require, exports, module) {
         },
 
         componentWillMount: function () {
-            this.getFlux().actions.dialog.registerDialog(this.props.id, this._getDismissalPolicy());
+            this.getFlux().store("dialog").registerDialog(this.props.id, this._getDismissalPolicy());
```
mcilroyc: I feel this is not in line with the recommendations here http://fluxxor.com/documentation/stores.html which suggest that stores should only be updated via the dispatcher.    I do recognize that a) something needs to be done to prevent waiting on the *action* queue for these registrations, and b) that rules are made to be broken.  But there you have it.

iwehrman: He's got a point. But, does he have an alternate suggestion?

iwehrman:
I'm not sure I am in favor of removing the select postcondition. Yes, it slows things down in debug mode, but it also catches bugs...

volfied:
I'm blocking this PR in favor of Cory's PR as he started working on transform handlers before I did and I should rebase my code for his.

volfied:
Meanwhile, @mcilroyc let's continue the discussion on the above comment.

volfied:
Rebased, and good to go.

src/js/actions/transform.js
```diff
@@ -1119,9 +1120,9 @@ define(function (require, exports) {
                     nextDoc = appStore.getCurrentDocument();
                     this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
                 });
-        }.bind(this);
+        }.bind(this), this);
 
-        _moveListener = function () {
+        _moveListener = synchronization.debounce(function () {
```
mcilroyc: This is somewhat concerning to me on the history front.  I would have thought that each "move" event corresponds neatly to a history state.  This debouncing will collapse them into a single one, right?  This may get smoothed over by the fact that we are not debouncing the history state listener ... but still, food for thought.

volfied: Fair thought, and probably right.

BUT

Given the user is clicking and dragging like crazy, I'd argue that the correctness of the history is not the primary concern of the user.

mcilroyc: mkay.  I suppose the super-fancy way to do this would be to still treat each separately, but "debounce" the expensive part (namely: the call to Ps to resetBounds).  Another day.

src/js/actions/transform.js
```diff
@@ -1132,8 +1133,8 @@ define(function (require, exports) {
             // all selected layers, but due to a recent bug, "move" event sends us the displacement
             // of layers from the changing (0,0) coordinates, which causes bugs like
             // getting (650,0) when the move was actually (-100, 0) for a 750 px wide layer
-            this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
-        }.bind(this);
+            return this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
+        }.bind(this), this);
```
mcilroyc: I'm registering my hunch that `bind(this)` is not necessary (but harmless) now that debounce will bind it within

src/js/fluxcontroller.js
```diff
@@ -374,6 +381,9 @@ define(function (require, exports, module) {
                 }
 
                 return modalPromise
+                    .catch(function () {
+                        // If modal state already ended, quietly continue
```
mcilroyc: :ok_hand: :grey_question:

iwehrman: I think this is safe, though I really wish we had some better way of catching the expected errors only instead of all the errors.

volfied:
Rebased for cleanliness (and removed the redundant bind)

src/js/jsx/shared/Dialog.jsx
mcilroyc: I have no alternate suggestion I guess.  I find it a little disconcerting that the action queue adds enough overhead such that your "direct" solution is noticeably faster.

In this particular case, is it the DE-registering that is gumming up the works?  (otherwise, why does the dialog sub-system slow down the marquee select?)  If so... is there a way to make that completely asynchronous so that the de-registration does not block marquee selection.  Part of me thinks we don't need to de-register them at all... but I suppose that could result in extra memory usage over time.

I'm just saying words at this point.  This is certainly not enough to block this PR ... but something still seems off to me.

mcilroyc:
Merging.  My only misgiving is about the dialog registration which I believe is either not an actual issue, or could/should be solved another way.  For another day.

PR #1700, opened 2015-07-09
Fixes for misplacement of dragged layers

designedbyscience:
Couple of issues that arose:

1. 1 bug was caused by the fact that handleDragStart is where we defined the initial position of the dragged elements. However, if an element was never clicked on, this never ran. So moved it all into when the drag is actually moving
2. Similarly, the elements need to make sure to clear their state, otherwise there can be a slightly flicker in position when a second drag operation starts.

@chadrolfs : Could you give this a quick check for placement issues?
@iwehrman : Don't think I hurt performance with this, but would you give it a once over?

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -226,7 +226,7 @@ define(function (require, exports, module) {
                 isLastInGroup = false,
                 dragStyle;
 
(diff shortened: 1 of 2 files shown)


iwehrman:
The fix looks good, including performance. The compiled build renders at ~19fps with Vermillion on a MacBook Pro, which ain't buttery but ain't terrible either.

PR #1702, opened 2015-07-09
Refactor SearchBar component

hamburge:
Now we can search for ~\*~\*anything\*~\*~


src/js/jsx/Search.jsx
```diff
@@ -51,7 +55,57 @@ define(function (require, exports, module) {
             this.getFlux().actions.dialog.closeDialog(SEARCH_BAR_DIALOG_ID);
         },
 
+        _getOptions: function () {
```
volfied: Drive by comment, JSDocs for these functions plox.

iwehrman: A plox on both your houses.

hamburge: sry.
fixed now.

iwehrman:
I think @hamburge is going to re-think some of this. So, to avoid confusion, adding the "review only" flag for now.

chadrolfs:
Based on Seth's new layers panel design, is this going to move to the top of that?

iwehrman:
There will continue to be a stand-alone "app search" feature, which allows you to search for things other than just layers. The in-panel layer search feature will be built using the same infrastructure, but will be styled differently (in-panel) and will only search layers.

hamburge:
@iwehrman Ok gave this another shot. I wasn't sure where exactly to put handling the confirmation of a search item. I ended up making sort of a default handling in the search store (`_defaultHandleExecute()`), which can be used or overwritten in individual search modules.

Also fixed issues #1715 and #1716 in here.

iwehrman:
Going in the right direction here, but now all the result-specific information is hard-coded in the search store. Instead, anything related to layer results should be part of the layer actions (registered in `layers.afterStartup`), document results should be part of the document actions (registered in `documents.afterStartup`), etc. 

The rule of thumb is that I should be able to add another data provider without editing any of the search infrastructure, including `Search` component or the `search` store. To accomplish this, the information that each `afterStartup` function registers will probably have to include some callbacks, like a function `getResults` that returns the currently appropriate list of options for that provider.

hamburge:
@iwehrman I moved the search type specific logic to the `actions/layers.js` and `actions/documents.js` The layers action is now a bit messier, but I wasn't sure if it was messy enough to add a search utility.

Let me know what ya think!

src/js/events.js
```diff
@@ -146,6 +146,10 @@ define(function (require, exports, module) {
             LIBRARY_CREATED: "libraryCreated",
             LIBRARY_REMOVED: "libraryRemoved",
             LIBRARY_SELECTED: "librarySelected"
+        },
+        search: {
+            LAYER_SEARCH_OPTIONS: "layerSearchOptions",
```
iwehrman: There should just be one new event here. Something like `REGISTER_SEARCH_PROVIDER`. All the providers should emit the same event because they'll all be handled uniformly by the search store.

hamburge: @iwehrman :+1:

volfied:
Hey @hamburge, I think with the inclusion of these search functions, our action code is getting a bit cumbersome.

While I review your code, do you think you can refactor these search related actions in a new folder structure named: `js/actions/search/layers.js` / `js/actions/search/documents.js` etc...

Afterwards, include these files from their respective action files, then call/transfer into those functions from the afterStartup methods. This will help not have bunch of internal functions in the afterStartup that are only dealing with search, and let you document that code in a way JSDoc will be happily generated. Come ask me WTF if this is unclear.

src/js/stores/search.js
```diff
@@ -0,0 +1,212 @@
...
+            return this._registeredSearches[id].searchHeaders;
+        },
+
+        /**
+         * Perform action for item
+         *
+         * @param {string} id Search module to use
+         * @param {string} itemID Search item to handle
+         */
+        handleExecute: function (id, itemID) {
```
volfied: Not sure to the need for this wrapper.

volfied:
Also, I think it would be a great exercise for you to write a long block comment near the top of Search Store detailing it's mechanics. Describe the implementation details such as what actions call to subscribe to search results, how we display these results, how we handle a select of an option. 

This will make it easier for the rest of us to add our own modules to search later on, when we want to extend it.

src/js/actions/documents.js
```diff
@@ -856,12 +860,119 @@ define(function (require, exports) {
...
+                            }),
+                docMap = openDocs.map(function (doc) {
+                    return {
+                        id: "curr-doc_" + doc.toString(),
+                        title: docStore.getDocument(doc).name,
+                        type: "item",
+                        category: [strings.SEARCH.CATEGORIES.DOCUMENT, strings.SEARCH.CATEGORIES.CURRENT]
+                    };
+                }),
+                docLabel = {
```
volfied: It seems like this logic here (of mapping an object to a search item) is being duplicated across different actions where you're creating search results. 

Instead, consider registering with a result type (to use with IDs), header title, category in addition to things you're already registering with. That way, this function would be a really simple return of `openDocs.map(doc => docStore.getDocument(doc).name)` mapping.

src/js/actions/documents.js
```diff
@@ -856,12 +860,119 @@ define(function (require, exports) {
...
+                    id: "curr-doc_header",
+                    title: strings.SEARCH.HEADERS.CURRENT_DOC,
+                    type: "header"
+                },
+
+                docOptions = docMap.unshift(docLabel);
```
volfied: Also, what happens if one day we decide we don't want headers. If your function took care of this in one central location, we would have to change it in only one place, and clients of the Search API you're developing wouldn't need to worry about parity between different modules.

src/js/actions/documents.js
```diff
@@ -856,12 +860,119 @@ define(function (require, exports) {
...
+                id: "recent-doc_header",
+                title: strings.SEARCH.HEADERS.RECENT_DOC,
+                type: "header"
+            };
+
+            return recentDocMap.unshift(recentDocLabel);
+        };
+
+        var getDocumentSearchOptions = function () {
+            return recentDocSearchOptions().concat(currDocSearchOptions());
```
volfied: Consider registering twice, once for open docs and once for recent docs instead. Modularity is the name of the game.

src/js/actions/documents.js
```diff
@@ -856,12 +860,119 @@ define(function (require, exports) {
...
+        };
+
+        var confirmSearch = function (itemID) {
+            // ID has type as first word, followed by the layer/document ID
+            var idArray = itemID.split("_"),
```
volfied: I really like this logic of returning false if it's not being handled. But, I think it would be better if the id=>handleExecute mapping is taken care of by the store, so it can just pass by the actual 'idInt' in here where all you'd need to do is call the handling function.

src/js/actions/layers.js
```diff
@@ -1587,6 +1590,155 @@ define(function (require, exports) {
...
+                        iconID = svgUtil.getSVGClassFromLayer(layer);
+
+                    return {
+                        id: "layer_" + layer.id.toString(),
+                        title: layer.name,
+                        info: ancestry,
+                        displayInfo: ancestry,
+                        svgType: iconID,
+                        category: layerType,
+                        type: "item"
```
volfied: I now see that with layer results, you are maintaining a really fine balance of keeping some information in search store, and some in the providers. Still, I think it would be really valuable to document, in search store, the exact structure of the item objects list you're expecting from the providers.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -381,20 +209,15 @@ define(function (require, exports, module) {
          * @return {Array.<string>}
          */
         _getFilterIcons: function (filter) {
-            // currently only have icons for layers
-            if (filter.length > 1 && filter.join(" ").indexOf("layer") > -1) {
-                return svgUtil.getSVGClassesFromLayerTypes(filter);
-            } else {
-                return ["tool-rectangle"]; // standin for non-layers
-            }
+            return svgUtil.getSVGClassesFromLayerTypes(filter);
```
volfied: I see the word Layer here :)

src/js/stores/search.js
```diff
@@ -0,0 +1,212 @@
...
+            this._updateSearchItems(id);
+        },
+
+        /**
+         * Add search type
+         *
+         * @param {object} payload
+         */
+        _registerSearchType: function (payload) {
+            this._registeredSearchTypes[payload.type] = {
```
volfied: I think you can do a lot more weight lifting in this function here, per my comments above. 

Also, please expand on what's expected in the payload, similar to this:

```
* @param {object} descriptor
     * @param {number} descriptor.layerKind
     * @param {boolean?} descriptor.artboardEnabled If set, will parse artboard value
     * @param {object?} descriptor.artboard Contains the artboard bounds descriptor
     * @param {object?} descriptor.pathBounds If available, will be parsed as shape layer
     * @param {object?} descriptor.boundsNoEffects Bounds object available for all layers
```

src/js/stores/search.js
```diff
@@ -0,0 +1,212 @@
...
+            }.bind(this), []);
+        },
+
+        /**
+         * Perform default action for item
+         *
+         * @param {string} id Search module ID
+         * @param {string} itemID Selected search item ID
+         */
+        _defaultHandleExecute: function (id, itemID) {
```
volfied: Per my comments above, I think you should be able to skip this forEach search and immediately dive into the action registered as a "handler" for a menu item, given you provide it on registration.

volfied:
@hamburge I love this, you're building something incredible for us to use in any of the Spaces we'll implement. Just got couple more architectural concerns that'll make it easier to use later on. Back to you.

chadrolfs:
Looking good!  I did a front end pass and have some questions about the filtering.  The last 3 items, Gradient, Pattern and Solid Color.  I'm assuming these are Fill layers?  

If I create Fill layers in standard PS and jump into DS, they are filtered out as Adjustment layers.

src/js/actions/documents.js
hamburge: Are you suggesting that some of the information (like the ID type) would be set for the option in the search store while other information (like the name) would be set here in the document action?

hamburge:
@chadrolfs I'm actually not sure what those three items are supposed to be-- I was just working off of the list of layer types in the source. Who would know the answer to that?

hamburge:
@volfied Ready for :eyes:!

src/js/actions/documents.js
```diff
@@ -35,6 +35,7 @@ define(function (require, exports) {
 
     var historyActions = require("./history"),
         layerActions = require("./layers"),
+        searchActions = require("js/actions/search/documents"),
```
volfied: This can be `./search/documents`

src/js/actions/layers.js
```diff
@@ -38,6 +38,7 @@ define(function (require, exports) {
     var Layer = require("js/models/layer"),
         collection = require("js/util/collection"),
         documentActions = require("js/actions/documents"),
+        searchActions = require("js/actions/search/layers"),
```
volfied: Ditto

src/js/stores/search.js
```diff
@@ -0,0 +1,258 @@
...
+        registerSearch: function (id, types) {
+            this._registeredSearches[id] = { "searchHeaders": types };
+            this._updateSearchItems(id);
+        },
+
+        /**
+         * Add search type
+         *
+         * @param {object} payload
+            * {Immutable.List.<string>} filters Possible categories for items in this search type
```
volfied: The `*`s shouldn't be indented.

src/js/util/svg.js
```diff
@@ -50,25 +50,25 @@ define(function (require, exports) {
...
     */
-    var getSVGClassesFromLayerTypes = function (layerTypes) {
-        if (layerTypes.length === 0) {
+    var getSVGClassesFromFilter = function (filter) {
+        if (filter.length === 0) {
             return [];
         }
         // currently only have icons for layers
-        if (layerTypes.length < 2 || layerTypes.join(" ").indexOf("layer") === -1) {
+        if (filter.length < 2 || filter.join(" ").indexOf("layer") === -1) {
```
volfied: Not happy with this layer dependency here... but we can come back to it when we add more results with icons.

src/js/stores/search.js
volfied: this should also say `payload.filters`, `payload.handleExecute` etc.

src/js/stores/search.js
```diff
@@ -0,0 +1,258 @@
...
+            * {function} handleExecute Takes item ID (a number) as a parameter. Called when item is confirmed
+            * {function} getOptions Function that returns Immutable.List.<object> of option information 
+                * that can be mapped to a full search options list. List has objects of the form:
+                    * {string} id A number parsed as a string. ID used to perform action if item is selected
```
volfied: Since this is an object structure we really want to make sure is visible. 

Refer to http://usejsdoc.org/tags-typedef.html#examples second example.

volfied: And refer to http://usejsdoc.org/tags-callback.html for how to JSDoc callback functions.

volfied:
So much better in terms of cleanliness. Of course, I'm going to be annoying and nitpicky about the JSDoc, but you're almost there!

volfied:
@hamburge Back to you, don't forget to set the label to Ready and assign to me when you're done.

volfied:
:+1:

chadrolfs:
Here is distinction via menu in standard between Fill and Adjustment but I think both types of layers come into PS as "Adjustment": http://cl.ly/0u0z1e041y0z
You can probably remove Gradient, Pattern and Solid from the list since they get assumed under Adjustment, unless we want to specifically distinguish them.

hamburge:
@chadrolfs Ok great, thanks. I'll do that in a separate PR.

Issue #1705, opened 2015-07-09
Layer Search: clicking on datalist item results in error


chadrolfs:
new doc, drag out some shapes, cmd/ctrl + f, click on field to expand the list, click on Vector
Result:
`Uncaught TypeError: Cannot read property 'split' of null`

Issue #1704, opened 2015-07-09
Layer Search: hitting Tab with datalist exposed results in error


chadrolfs:
New doc, drag out some shapes, hit cmd/ctrl + f, click on field to expand datalist, hit Tab
Result:
`Uncaught TypeError: Cannot read property 'indexOf' of null`

Issue #1703, opened 2015-07-09
Tab away from datalist should collapse/hide it


chadrolfs:
new doc, new shape, click on Blend mode drop down, hit Tab a few times.

Result: Blend mode datalist says open/expanded.

Issue #1682, opened 2015-07-07
Tab will set focus to layers and arrow keys now nudge layers instead of scrolling the list


chadrolfs:
Open a doc with multiple layers, 
select a layer, 
cmd/ctrl + f and expand the list, 
use up/down arrow to nav list, 
hit Tab, 
arrow up/down

Result: arrow keys no longer scroll the list and instead nudge the layer

iwehrman:
Since the floating search component is modal, focus should **never** leave the modal.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
The layer is no longer nudged but arrow keys cease to work in datalist after hitting Tab.  Separate issue or blocked by #1704 or #1705?  If arrow keys aren't supposed to work after hitting Tab and this is as design, re-assign to me or just close.  thanks!

chadrolfs:
I can get into a state (on Windows only) where focus is not set on the Layer Search dialog on initial open (no active cursor in field).  Hitting Tab cycles through the layers, arrows nudge them, etc.

iwehrman:
FBNC to @chadrolfs.

PR #1708, opened 2015-07-09
Fix errors with Datalist tabbing/clicking

hamburge:
for issues #1703, #1704, #1705


hamburge:
Now also has a fix for #1682, so the search bar drop down keeps focus when you hit Tab on an unopened Datalist.

Issue #1710, opened 2015-07-09
Drag and drop multiple layers without reordering drops selection


chadrolfs:
https://www.dropbox.com/s/9tumh8ur7n6da1f/drag-dropSelection.mov?dl=0

PR #1723, opened 2015-07-13
Looking for the wrong class

designedbyscience:
Addresses #1710


volfied:
After extensive review, I think this PR is safe.

PR #1725, opened 2015-07-13
Switching back to state before #1723 because it broke everything.

designedbyscience:
And apparently the solution to #1710 is non-obvious


volfied:
After extensive review, I've decided this un-fix safe.

PR #1726, opened 2015-07-13
Suppress click after drag

iwehrman:
1. Internally suppress handlers for the click event that follows a drag operation. Draggables (like `LayerFace`) no longer need to provide special handling to ignore these clicks.
2. Add a proper `onDragStop` handler for draggables, renaming the existing handler to `onDrop`.

Addresses #1710.


chadrolfs:
quick look - looks good.

designedbyscience:
Yes looks good. Thanks @iwehrman

PR #1727, opened 2015-07-13
More CC Library - Design Space interactions

volfied:
- Create character style asset
 - Create layer style asset
 - Create color asset
 - Apply layer style to selected layer action
 - Apply text style to selected layer action

With these, all the low level interactions between cc-libraries-api <=> Design Space <=> Photoshop should be ready for hooking up.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -90,7 +90,7 @@ define(function (require, exports) {
                 return Promise.fromNode(function (cb) {
                     representation.updateContentFromPath(path, false, cb);
                 });
-            }).then(function () {
+            }).finally(function () {
                 currentLibrary.endOperation();
             }).then(function () {
                 var newRepresentation = newElement.getPrimaryRepresentation();
@@ -185,7 +185,7 @@ define(function (require, exports) {
                 newElement.setRenditionCache(104, filepath, function () {
                     // FIXME: In CEP Panel, they delete the temporary file afterwards
                 });
-            }).then(function () {
+            }).finally(function () {
                 currentLibrary.endOperation();
                 // FIXME: Do we need payload?
                 return this.dispatchAsync(events.libraries.ASSET_CREATED, {});
@@ -255,7 +255,7 @@ define(function (require, exports) {
                 newElement.setRenditionCache(108, thumbnailPath, function () {
                     // FIXME: In CEP panel, they delete the temporary thumbnail file afterwards
                 });
-            }).then(function () {
+            }).finally(function () {
                 currentLibrary.endOperation();
                 // FIXME: Do we need payload?
                 return this.dispatchAsync(events.libraries.ASSET_CREATED, {});
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -25,14 +25,19 @@ define(function (require, exports) {
     "use strict";
 
     var Promise = require("bluebird"),
+        tinycolor = require("tinycolor"),
         CCLibraries = require("file://shared/libs/cc-libraries-api.min.js");
 
     var descriptor = require("adapter/ps/descriptor"),
         docAdapter = require("adapter/lib/document"),
+        colorAdapter = require("adapter/lib/color"),
+        layerEffectAdapter = require("adapter/lib/layerEffect"),
+        textLayerAdapter = require("adapter/lib/textLayer"),
         libraryAdapter = require("adapter/lib/libraries");
 
     var events = require("../events"),
-        locks = require("../locks");
+        locks = require("../locks"),
+        layerActions = require("./layers");
 
     /**
      * Uploads the selected single layer to the current library
@@ -85,7 +90,7 @@ define(function (require, exports) {
                 return Promise.fromNode(function (cb) {
                     representation.updateContentFromPath(path, false, cb);
                 });
-            }).finally(function () {
+            }).then(function () {
                 currentLibrary.endOperation();
             }).then(function () {
                 var newRepresentation = newElement.getPrimaryRepresentation();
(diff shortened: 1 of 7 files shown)


src/js/actions/libraries.js
```diff
@@ -106,13 +111,210 @@ define(function (require, exports) {
...
+                return this.dispatchAsync(events.libraries.ASSET_CREATED, {});
+            });
+    };
+    createLayerStyleFromSelectedLayer.reads = [locks.JS_DOC, locks.JS_LIBRARIES, locks.CC_LIBRARIES];
+    createLayerStyleFromSelectedLayer.writes = [locks.JS_LIBRARIES];
+
+    /**
+     * Uploads the given color as a color asset
+     *
+     * @param {{r: number, g: number, b: number}} color Color to be added as an asset
```
iwehrman: If this param is a `Color` model, it might be nicer to just include a `toHexString()` method so that wraps the `tinycolor` dependency. Also, the `@param` should just be `{Color}`.

src/js/actions/libraries.js
```diff
@@ -85,7 +90,7 @@ define(function (require, exports) {
                 return Promise.fromNode(function (cb) {
                     representation.updateContentFromPath(path, false, cb);
                 });
-            }).finally(function () {
+            }).then(function () {
```
iwehrman: I guess this was intentional? Is the operation somehow ended if there is a failure?

src/js/actions/libraries.js
```diff
@@ -106,13 +111,210 @@ define(function (require, exports) {
...
+            filepath = "/tmp/textThumbnailPreview.png";
+
+        // Where magic happens
+        representation.setValue("characterstyle", "data", typeData);
+
+        // FIXME: Make sure this reflects the character style we're recreating
+        // check to see how CEP Panel does it.
+        var exportObj = libraryAdapter.createTextThumbnail(filepath,
+            typeData.adbeFont.postScriptName,
+            "Aa",
```
iwehrman: Localize this string?

src/js/actions/libraries.js
```diff
@@ -106,13 +111,210 @@ define(function (require, exports) {
...
+        return descriptor.playObject(exportObj)
+            .bind(this)
+            .then(function () {
+                return Promise.fromNode(function (cb) {
+                    imageRepresentation.updateContentFromPath(filepath, false, cb);
+                });
+            }).then(function () {
+                // FIXME: Constant here
+                newElement.setRenditionCache(104, filepath, function () {
+                    // FIXME: In CEP Panel, they delete the temporary file afterwards
```
iwehrman: What's our plan?

src/js/actions/libraries.js
```diff
@@ -106,13 +111,210 @@ define(function (require, exports) {
...
+                });
+            }).then(function () {
+                currentLibrary.endOperation();
+                // FIXME: Do we need payload?
+                return this.dispatchAsync(events.libraries.ASSET_CREATED, {});
```
iwehrman: For future reference, it's fine to just do `this.dispatch` instead of calling the asynchronous version and returning it's promise.

iwehrman: (But that's ONLY when you're in a `.then` handler.)

src/js/actions/libraries.js
```diff
@@ -106,13 +111,210 @@ define(function (require, exports) {
...
+    var createCharacterStyleFromSelectedLayer = function () {
+        var appStore = this.flux.store("application"),
+            libStore = this.flux.store("library"),
+            fontStore = this.flux.store("font"),
+            currentDocument = appStore.getCurrentDocument(),
+            currentLibrary = libStore.getCurrentLibrary(),
+            currentLayers = currentDocument.layers.selected,
+            currentLayer = currentLayers.first();
+
+        if (!currentLibrary || currentLayers.count() !== 1 ||
```
iwehrman: I'm pretty sure `.size` is appropriate here, but whatev.

src/js/actions/libraries.js
```diff
@@ -147,9 +349,72 @@ define(function (require, exports) {
...
+        return Promise.fromNode(function (cb) {
+            representation.getContentPath(cb);
+        }).bind(this).then(function (path) {
+            var layerRef = layerEffectAdapter.referenceBy.current,
+                placeObj = layerEffectAdapter.applyLayerStyleFile(layerRef, path);
+
+            return descriptor.playObject(placeObj);
+        }).then(function () {
+            // FIXME: This can be more optimistic
+            return this.transfer(layerActions.resetLayers, currentDocument, currentDocument.layers.selected);
```
iwehrman: Does this need to include the ancestors as well? I don't think so, but just making sure.

src/js/jsx/sections/libraries/LibraryBar.jsx
```diff
@@ -37,10 +37,25 @@ define(function (require, exports, module) {
     var LibraryBar = React.createClass({
         mixins: [FluxMixin],
 
-        createNewElement: function () {
+        addImageAsset: function () {
             this.getFlux().actions.libraries.createElementFromSelectedLayer();
         },
 
+        addCharacterStyle: function () {
```
iwehrman: These should probably be private methods. Their JSDoc annotations should have `@private`.

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
 
             return fontObj && fontObj.postScriptName;
         },
+
+        getTypeObjectFromLayer: function (layer) {
```
iwehrman: JSDoc, please.

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
...
+            if (layer.kind !== layer.layerKinds.TEXT) {
+                return null;
```
iwehrman: Do we actually expect this to happen? Should this just throw an exception instead? Will returning `null` just delay the inevitable error?

iwehrman: (Repeat for all similar returns below.)

iwehrman: If the caller always is prepared for `null` to be returned, then I'm not arguing that it should be an exception instead. Returning null is perfectly fine if the caller knows that it might get null. Just checking!

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
...
+
+            var obj = {},
+                textStyle = layer.text.characterStyles.first(),
+                psName = textStyle.postScriptName,
+                fontObj = this._postScriptMap.get(psName, null);
+                
+            if (!fontObj) {
+                return null;
+            }
+            var family = fontObj.family,
```
iwehrman: There should be a newline after the brace.

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
...
+                postScriptName: psObj.postScriptName,
+                style: psObj.style
+            };
+
+            obj.fontFamily = fontObj.family;
+
+            if (textStyle.textSize) {
+                obj.fontSize = {
+                    "type": "pt",
+                    "value": textStyle.textSize
```
iwehrman: Hmm... I thought all of our sizes were recorded in pixels, not points?

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
...
+                };
+            }
+
+            var style = psObj.style.toLowerCase();
```
iwehrman: I think all this logic already exists in `Type.jsx`. If so, it should be factored out into `js/util/type.js` and used both places.

src/js/stores/font.js
```diff
@@ -91,6 +91,98 @@ define(function (require, exports, module) {
...
+
+            if (textStyle.color) {
+                var color = textStyle.color;
+                obj.color = {
+                    mode: "RGB",
+                    value: color,
+                    type: "process"
+                };
+            } else {
+                obj.color = {
```
iwehrman: May as well just define one `LIBRARIES_COLOR_BLACK` variable outside this function scope instead of allocating and disposing of the equivalent object repeatedly...

iwehrman: Maybe that's a dumb idea. Maybe there is a better idea involving `Color` models. Who knows.

iwehrman:
Done with a first pass. Generally looks totally sane and readable. Nice work. Back to @volfied for nits and questions.

src/js/actions/libraries.js
volfied: Must have had a brain fart, supposed to be the other way. Thank you

src/js/actions/libraries.js
volfied: Not localized in the CEP panel / Photoshop as far as I can tell

src/js/actions/libraries.js
volfied: For PS parity, I'm removing the name completely (this is what the panel does as well)

src/js/actions/libraries.js
volfied: Once I work with Fitz on the temproary file API, I will delete them after these steps. For now, I let them stay in /tmp and be rewritten by us, or deleted by OS.

volfied: Keeping track of this in the clean up trello card

src/js/actions/libraries.js
volfied: It doesn't need to.

src/js/stores/font.js
volfied: Nope, this is on parity with PS.

src/js/stores/font.js
volfied: In this case, the object expected is different from what our renders expect. I suggest creating a code cleanup issue for this, instead of blocking the PR.

src/js/stores/font.js
volfied: Another code clean up. I wanted to get things up and running ASAP for hand off, and I'm well aware that there is A LOT of room for improvement when it comes to the type objects and color handling.

volfied:
@iwehrman Back to you. I agree there is some clean up to be done, but I'd like to not block @shaoshing and leave those for later when the requirements for color spaces etc. are clearer.

src/js/actions/libraries.js
iwehrman: :white_check_mark:

src/js/actions/libraries.js
```diff
@@ -106,13 +110,210 @@ define(function (require, exports) {
...
+                    imageRepresentation.updateContentFromPath(filepath, false, cb);
+                });
+            }).then(function () {
+                // FIXME: Constant here
+                newElement.setRenditionCache(104, filepath, function () {
+                    // FIXME: In CEP Panel, they delete the temporary file afterwards
+                });
+            }).finally(function () {
+                currentLibrary.endOperation();
+                // FIXME: Do we need payload?
```
iwehrman: For future reference: no, you don't need a payload.

iwehrman:
LGTM. Have at it, @shaoshing!

shaoshing:
[LGTM](http://lgtm.in/i/wKVeLKwgq)

Issue #1628, opened 2015-06-20
Zoom keyboard shortcut should match standard PS


iwehrman:
It's cmd-= in Design Space and cmd-+ in standard PS. I'm not that we did this for any good reason, though it's possible it had to do with some difficulty with menu shortcuts and shift modifiers. A friendly twitter user brought this to our attention here: https://twitter.com/ahoyscottbeaty/status/611919837250482176

volfied:
Please get in touch with Fitz to work on this issue and either add a secret keyboard shortcut for cmd+ = and display it as cmd++
Related to #1626

volfied:
FBNC to @chadrolfs

Issue #1626, opened 2015-06-19
CMD+ for numeric keyboard


placegraphichere:
CMD = (or plus) and CMD _ (minus) work, however for number pads, the same should be the case for CMD+ and CMD-. The + option is simply missing. We should add it.

Referenced from Scott Beaty@ahoyscottbeaty https://twitter.com/ahoyscottbeaty/status/611925737688305664

volfied:
This is either a one character fix, or is a core issue. @shaoshing please see if the obvious solution works, if it doesn't let's open a core bug.

volfied:
FBNC to @chadrolfs

PR #1728, opened 2015-07-14
zoom-in shortcut: match standard PS and support numeric keyboard

shaoshing:
This PR fixed issues #1628 and #1626.


volfied:
Looks great, confirmed with a numpad as well.

PR #1729, opened 2015-07-14
Layer Effects: Support deleting layer effects

shaoshing:
(This PR requires https://github.com/adobe-photoshop/spaces-adapter/pull/120 to work)

### DIFF
--- a/src/img/ico-layer-delete.svg
+++ b/src/img/ico-layer-delete.svg
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="layer-delete" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px"
+	 y="0px" viewBox="0 0 612 792" style="enable-background:new 0 0 612 792;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:none;stroke:#000000;stroke-width:80;stroke-miterlimit:10;}
+</style>
+<g transform="translate(1 1)">
+	<path class="st0" d="M408.6,291.4L201.4,498.6 M201.3,291.3l207.3,207.3"/>
+</g>
+</svg>
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -189,18 +189,11 @@ define(function (require, exports) {
      */
 
     var deleteShadow = function (document, layers, shadowIndex, type) {
-        var layerEffectIndexList = [];
-        layers.forEach(function (curLayer) {
-            var curLayerEffects = curLayer.getLayerEffectsByType(type);
-            var hasSelectedShadow = curLayerEffects && curLayerEffects.has(shadowIndex);
-            layerEffectIndexList.push(hasSelectedShadow ? shadowIndex : null);
-        });
-
         var payload = {
             documentID: document.id,
             layerIDs: collection.pluck(layers, "id"),
             layerEffectType: type,
-            layerEffectIndex: Immutable.List(layerEffectIndexList)
+            layerEffectIndex: shadowIndex
         };
         // Synchronously update the stores
         this.dispatch(events.document.history.optimistic.LAYER_EFFECT_DELETED, payload);
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -76,7 +76,7 @@ define(function (require, exports) {
                         return shadow.toAdapterObject();
                     }).toArray();
 
-            if (curLayer.hasLayerEffect) {
+            if (curLayer.usedToHaveLayerEffect) {
                 return {
                     layer: curLayer,
                     playObject: layerEffectLib.setExtendedLayerEffect(type, referenceID, shadowAdapterObject)
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -1,24 +1,24 @@
 /*
  * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
- *  
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a
- * copy of this software and associated documentation files (the "Software"), 
- * to deal in the Software without restriction, including without limitation 
- * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
- * and/or sell copies of the Software, and to permit persons to whom the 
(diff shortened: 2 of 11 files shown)


volfied:
I'll start reviewing this, while I'm doing that, can you rebase your changes to a concise commit?
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

src/img/ico-layer-delete.svg
```diff
@@ -0,0 +1,11 @@
...
+<!-- Generator: Adobe Illustrator 19.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="layer-delete" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px"
+	 y="0px" viewBox="0 0 612 792" style="enable-background:new 0 0 612 792;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:none;stroke:#000000;stroke-width:80;stroke-miterlimit:10;}
+</style>
+<g transform="translate(1 1)">
+	<path class="st0" d="M408.6,291.4L201.4,498.6 M201.3,291.3l207.3,207.3"/>
+</g>
+</svg>
```
volfied: If this is your addition, can you check with @placegraphichere to make sure you get the final icon?

volfied:
1 - Create a rectangle, add a drop shadow (X=5)
2 - Create another rectangle
3 - Select both rectangles, add a drop shadow (Y = 20)
4 - Delete either drop shadow
Result: All the shadows get deleted
Expected: Just the shadow I chose should be deleted

src/js/actions/layereffects.js
```diff
@@ -179,6 +179,39 @@ define(function (require, exports) {
     setShadowEnabled.writes = [locks.PS_DOC, locks.JS_DOC];
 
     /**
+     * Delete the selected Shadow for all selected layers
+     *
+     * @param {Document} document
+     * @param {Immutable.Iterable.<Layer>} layers list of layers to update
+     * @param {number} shadowIndex index of the Drop Shadow within the layer(s)
```
volfied: Is this index going to be same for all layers?

volfied: I see you `forEach` to calculate actual index list. Please refer to bug I added as a comment.

src/js/models/layer.js
```diff
@@ -134,12 +134,12 @@ define(function (require, exports, module) {
         /**
          * @type {Immutable.List.<Shadow>}
          */
```
volfied: Why the change from null to `Immutable.List()`? We have other Lists in this record that initialize as null.

src/js/models/layer.js
```diff
@@ -343,8 +357,7 @@ define(function (require, exports, module) {
             mode: "passThrough",
             proportionalScaling: false,
             isArtboard: false,
-            isLinked: false,
-            hasLayerEffect: false
```
volfied: Groups can have effects as well.. so I can reproduce some of the play errors you were getting by adding/removing effects from group layers.

src/js/models/layerstructure.js
```diff
@@ -1328,8 +1328,80 @@ define(function (require, exports, module) {
...
+        var nextLayers = Immutable.Map(layerIDs.reduce(function (map, layerID, index) {
+            var layer = this.byID(layerID),
+                layerEffects = layer.getLayerEffectsByType(layerEffectType) || Immutable.List(),
+                _layerEffectIndex,
+                layerEffect,
+                nextLayerEffect,
+                newProps,
+                nextLayer;
+
+            _layerEffectIndex = Immutable.List.isList(layerEffectIndex) ?
```
volfied: I think this style of declaring the var name, and then defining it below is not clean. Can you consolidate these?

volfied:
I like where these changes are going, but I still see some bugs, especially with group layers.

I'd like to also hear your reasoning for changing `hasLayerEffects` to `usedToHaveLayerEffects`. I understand the need to add the latter, but why remove the first one?

Back to you, @shaoshing

src/img/ico-layer-delete.svg
shaoshing: Got it from @designedbyscience as a temp icon, since we are moving toward UI redesign.

src/js/models/layer.js
shaoshing: I think a default value should be the value that works so that it doesn't rely on the users (us) to set a valid default value. In this case, I found I have to search over the files to make sure I can safely assume the dropShadows/innerShadows attr is always a list. Any thoughts?

src/js/models/layerstructure.js
shaoshing: These are already existed in the master, don't know why it is showing as new here. Do you want me to fix them here?

shaoshing:
```
1 - Create a rectangle, add a drop shadow (X=5)
2 - Create another rectangle
3 - Select both rectangles, add a drop shadow (Y = 20)
4 - Delete either drop shadow
Result: All the shadows get deleted
Expected: Just the shadow I chose should be deleted
```
Now deletion works correctly for selected:
* single / multiple layers 
* single / multiple groups
* layer + group

```
I'd like to also hear your reasoning for changing hasLayerEffects to usedToHaveLayerEffects. I understand the need to add the latter, but why remove the first one?
```

I am using *usedToHaveLayerEffects* to determine whether to use *setLayerEffect* or *setExtendedLayerEffect*. *setLayerEffect* is required only when the object is newly created and has no effect at all. After created its first effect, it will always has the layerEffect property, even after its layer effects are all deleted (remember the hidden layer effect). I am trying to replace a new variable name that represents its purpose correctly.

src/js/models/layer.js
iwehrman: I'm fine with initializing immutable record members like this to empty lists. Immutable is smart enough to only ever create a single empty list object, so there should be no performance impact. I would be _more_ concerned with making such a change consistently throughout our models, and also ensuring that client code no longer performs unnecessary falsey checks.

shaoshing: there are about 10 Immutable.List attributes defined with null default value after a quick search, shall I refactor them to keep consistency or wait until our major release?

iwehrman: I don't think you need to wait until after the release, though it might be better to submit a separate PR for that particular cleanup to make the review easier.

shaoshing: roger that

shaoshing:
back to you, Barkin

volfied:
Please ping me again after a rebase and addressing https://github.com/adobe-photoshop/spaces-design/pull/1729/files#r34593446

volfied:
:+1: Works great

Issue #1737, opened 2015-07-15
Layer Search: Background layer filter does not work


chadrolfs:
open a file containing a background layer. cmd/ctrl + f, select Background layer for filter

Result: Background (layer) does not appear in the filter list
If you select Pixel layer, it shows up there.

hamburge:
@chadrolfs @placegraphichere Do you think the fix for this should be not offering to filter by background layers (and just have them be considered pixel layers) or to have a background layer filter that actually lists the background layers?

chadrolfs:
I would lean towards not having a filter for background layers.  You can only have one per document anyway, users know where it is (always at the bottom) and technically it is a pixel layer.

iwehrman:
Yes, I would say just include it among the pixel layers.

Issue #1736, opened 2015-07-15
Layer Search: some filters result in missing .svg files


chadrolfs:
`GET file:///Users/crolfs/playground-repos/spaces-design/src/img/ico-layer-11.svg net::ERR_FILE_NOT_FOUND`

Issue #1735, opened 2015-07-15
Layer Search dismisses on 2nd attempt to type or expand list


chadrolfs:
cmd/ctrl + f, select a filter that is not associated with layers in current doc (Solid color layer), click in field to type or attempt to select a different filter

Result: 2nd click dismisses the search dialog

Issue #1719, opened 2015-07-10
Layer Search: arrow keys stop working once you select a Type


chadrolfs:
new doc with several layers
cmd/ctrl + f
click on field to expand list
click on Vector
attempt to use arrows to nav up/down list

result: they don't work, you have to use the mouse.

PR #1741, opened 2015-07-16
Lots of little searching bugs!

hamburge:
Addresses
1. Too many layer filters: issue #1737, #1736, and @chadrolfs comments about filters on PR #1702
2. #1735: for this I added an unselectable dropdown option that says there are no options available.
3. #1719

### DIFF
--- a/src/js/actions/search/layers.js
+++ b/src/js/actions/search/layers.js
@@ -27,8 +27,7 @@ define(function (require, exports) {
     var Immutable = require("immutable"),
         _ = require("lodash");
 
-    var layerLib = require("adapter/lib/layer"),
-        events = require("js/events"),
+    var events = require("js/events"),
         svgUtil = require("js/util/svg"),
         strings = require("i18n!nls/strings");
     
@@ -132,9 +131,7 @@ define(function (require, exports) {
      *
      */
     var registerLayerSearch = function () {
-        var filters = Immutable.List(Object.keys(layerLib.layerKinds)).filterNot(function (kind) {
-                        return (kind === "ANY" || kind === "GROUPEND" || kind === "3D" || kind === "VIDEO");
-                    });
+        var filters = Immutable.List(["PIXEL", "TEXT", "ARTBOARD", "ADJUSTMENT", "SMARTOBJECT", "GROUP", "VECTOR"]);
 
         var payload = {
             "type": "LAYER",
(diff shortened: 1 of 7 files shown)


chadrolfs:
I'm not sure how the Current document filter is supposed to work.  If I select it, there's nothing in the datalist.  Everything in the Type list above it imply current document so I might remove "Current Document"

hamburge:
@chadrolfs If you have more than one document open, it will list the other open documents and you can switch to them.

Maybe I should add something so that it only shows a filter if there are actual options that correspond to it?

chadrolfs:
@hamburge That would be beautiful!

src/js/actions/search/layers.js
```diff
@@ -132,9 +131,7 @@ define(function (require, exports) {
      *
      */
     var registerLayerSearch = function () {
-        var filters = Immutable.List(Object.keys(layerLib.layerKinds)).filterNot(function (kind) {
-                        return (kind === "ANY" || kind === "GROUPEND" || kind === "3D" || kind === "VIDEO");
-                    });
+        var filters = Immutable.List(["PIXEL", "TEXT", "ARTBOARD", "ADJUSTMENT", "SMARTOBJECT", "GROUP", "VECTOR"]);
```
volfied: `ARTBOARD` is not a layer kind, is this still going to work?

src/js/util/svg.js
```diff
@@ -71,10 +71,7 @@ define(function (require, exports) {
         _.forEach(filter, function (kind) {
             var iconID = "layer-";
 
-            // No svg for these
-            if (kind === "solid color" || kind === "gradient" || kind === "pattern") {
-                iconID = "tool-rectangle";
-            } else if (kind === "artboard") {
+            if (kind === "artboard") {
```
volfied: Same here, not sure if artboard is an acceptable kind, I ran a quick string search and don't see any `artboard` string being added anywhere. Can you correct it here, or correct me? :)

volfied:
@hamburge looks good, but got a question.

src/js/util/svg.js
hamburge: Yep, it works! When adding layer items, we add "artboard" as its category instead of group (`_getLayerCategory()` in actions/search/layers. 

This svg function uses the category list which is why the right icon shows up, and when the filter for artboards is on, it compares its value to the items' categories, so it finds the layers that are artboards as well.

volfied: I am enlightened.

volfied:
My questions have been answered!

Issue #1695, opened 2015-07-09
Unable to drag a layer below bottom empty AB without it nesting


chadrolfs:
new default doc, 
select rect tool and draw a shape outside the AB bounds (it will be above the AB in the layers panel), 
in the layers panel drag the Rectangle 1 layer below the AB without having it nest into the AB

Result: no matter how far down you drag, it always nests into the AB

iwehrman:
I think this happens for any group, not just artboards. So, removing that label.

chadrolfs:
Mostly fixed.  If you follow the steps above, it works.  If you first drag it into the AB and then try to drag it out and below, it takes several attempts before you're successful.

PR #1743, opened 2015-07-16
Expand/collapse groups

iwehrman:
1. Groups can be expanded or collapsed in the panel by single-clicking their icon using the new `layers.setGroupExpansion` action. This might not be the final interaction, but it works for now. Descendant layers are not mounted until they're first visible, and hidden layers are not rendered when models change until their visible again. (This is the same pattern we're using for the per-document `Properties` components.) This provides a significant load-time speed up for big documents when their top-level groups are collapsed.
2. When a layer is selected on canvas, it is revealed in the panel by expanding its ancestor groups using the new `layers.revealLayers` action.
2. It's now possible to drop dragged layers onto groups, instead of just above or below a particular layer.
3. This also fixes a regression my last panel work caused by suppressing the virtual render after the drop and before the model has updated.

Note that this does _not_ address #1695. I'll tackle that in a separate PR.

Depends on adobe-photoshop/spaces-adapter#121.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -538,6 +538,7 @@ define(function (require, exports) {
         // eventually remove SELECT_LAYERS_BY_INDEX.
         var dispatchPromise,
             revealPromise;
+            
         if (!modifier || modifier === "select") {
             payload.selectedIDs = collection.pluck(layerSpec, "id");
             dispatchPromise = this.dispatchAsync(events.document.SELECT_LAYERS_BY_ID, payload);
@@ -1231,14 +1232,14 @@ define(function (require, exports) {
 
         var dispatchPromise = this.dispatchAsync(events.document.history.optimistic.SET_LAYERS_PROPORTIONAL, payload),
             layerPlayObjects = layerSpec.map(function (layer) {
-            var layerRef = layerLib.referenceBy.id(layer.id),
-            proportionalObj = layerLib.setProportionalScaling(layerRef, proportional);
+                var layerRef = layerLib.referenceBy.id(layer.id),
+                proportionalObj = layerLib.setProportionalScaling(layerRef, proportional);
 
-            return {
-                layer: layer,
-                playObject: proportionalObj
-            };
-        }, this);
+                return {
+                    layer: layer,
+                    playObject: proportionalObj
+                };
+            }, this);
 
         var sizePromise = layerActionsUtil.playLayerActions(document, layerPlayObjects, true, options);
 
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -91,7 +91,8 @@ define(function (require, exports) {
         "artboardEnabled",
         "pathBounds",
         "smartObject",
-        "globalAngle"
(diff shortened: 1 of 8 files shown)


iwehrman:
@chadrolfs As with all my drag-and-drop-related PRs, this has a high likelihood of introducing regressions. Please test if you have a chance!

volfied:
Double clicking a layer expands and immediately collapses it. Breaks the use a bit.

volfied:
Scrolling doesn't work exactly right. In Vermillion, I was on the iPad artboard, and cmd+clicked the Art of Paint text in Desktop artboard, it scrolled to the top of it's group, not the layer.

src/js/actions/layers.js
```diff
@@ -535,10 +536,15 @@ define(function (require, exports) {
 
         // TODO: Dispatch optimistically here for the other modifiers, and
         // eventually remove SELECT_LAYERS_BY_INDEX.
-        var dispatchPromise = Promise.resolve();
+        var dispatchPromise,
+            revealPromise;
         if (!modifier || modifier === "select") {
```
volfied: newline here

src/js/actions/layers.js
```diff
@@ -1220,11 +1229,8 @@ define(function (require, exports) {
...
             };
 
-        var dispatchPromise = Promise.bind(this).then(function () {
-            this.dispatch(events.document.history.optimistic.SET_LAYERS_PROPORTIONAL, payload);
-        });
-
-        var layerPlayObjects = layerSpec.map(function (layer) {
+        var dispatchPromise = this.dispatchAsync(events.document.history.optimistic.SET_LAYERS_PROPORTIONAL, payload),
+            layerPlayObjects = layerSpec.map(function (layer) {
             var layerRef = layerLib.referenceBy.id(layer.id),
```
volfied: Function contents should be indented one more tab

src/js/actions/layers.js
```diff
@@ -1389,6 +1395,80 @@ define(function (require, exports) {
...
+            if (document.layers.hasCollapsedAncestor(layer)) {
+                document.layers.strictAncestors(layer).forEach(function (ancestor) {
+                    if (!ancestor.expanded) {
+                        collapsedAncestors.add(ancestor);
+                    }
+                });
+            }
+            return collapsedAncestors;
+        }, new Set(), this),
+        collapsedAncestors = Immutable.Set(collapsedAncestorSet).toList();
```
volfied: Wouldn't it be better to reduce layers to an Immutable.Set() and go from there?

iwehrman: That creates a lot of garbage.

iwehrman:
Re double-clicking: that only happens if you double-click the icon, AFAIK. I think the behavior is the same as standard PS.

iwehrman:
OK, I think I've fixed the scrolling issue by changing the way we calculate the layer to scroll to. It's a bit of a hack, but it seems to work. ¯\_(ツ)_/¯

Back to you, @volfied.

volfied:
:+1:

PR #1744, opened 2015-07-16
Cavalcade of transform fixes

volfied:
- Remove unnecessary plucks from SCUs of Fill, Opacity and Stroke.jsx
 - Clean up transform/move handlers to minimize time
 - Clean up overlay toggle (artifact)
 - Reset border policies immediately instead of as reaction to re-render

In addition, includes:

Policy visualizer: helps see where the mouse border policies are set

@chadrolfs Can you start testing and let me know if things are better in this branch? It would be very ideal to test these in a compiled build as post condition verifications etc. really slow things down.

@iwehrman Please review.

### DIFF
--- a/src/js/actions/history.js
+++ b/src/js/actions/history.js
@@ -31,6 +31,7 @@ define(function (require, exports) {
         documentLib = require("adapter/lib/document"),
         historyLib = require("adapter/lib/history"),
         layerActions = require("./layers"),
+        toolActions = require("./tools"),
         documentActions = require("./documents");
 
     var events = require("js/events"),
@@ -153,6 +154,9 @@ define(function (require, exports) {
                                 count: count,
                                 selectedIndices: selectedIndices
                             });
+                        })
+                        .then(function () {
+                            return this.transfer(toolActions.resetBorderPolicies);
                         });
                 });
         } else {
@@ -181,8 +185,8 @@ define(function (require, exports) {
     var incrementHistory = function (document) {
         return _navigateHistory.call(this, document, 1);
     };
-    incrementHistory.reads = [locks.PS_DOC];
-    incrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC];
+    incrementHistory.reads = [locks.PS_DOC, locks.JS_APP, locks.JS_TOOL];
+    incrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC, locks.PS_APP, locks.JS_POLICY];
 
     /**
      * Navigate to the previous history state
@@ -193,8 +197,8 @@ define(function (require, exports) {
     var decrementHistory = function (document) {
         return _navigateHistory.call(this, document, -1);
     };
-    decrementHistory.reads = [locks.PS_DOC];
-    decrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC];
+    decrementHistory.reads = [locks.PS_DOC, locks.JS_APP, locks.JS_TOOL];
+    decrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC, locks.PS_APP, locks.JS_POLICY];
 
     /**
      * Revert to the document's last saved state.
@@ -233,10 +237,13 @@ define(function (require, exports) {
             .bind(this)
             .then(function () {
                 return this.dispatchAsync(events.ui.TOGGLE_OVERLAYS, { enabled: true });
+            })
+            .then(function () {
+                return this.transfer(toolActions.resetBorderPolicies);
             });
     };
-    revertCurrentDocument.reads = [locks.PS_DOC];
-    revertCurrentDocument.writes = [locks.JS_HISTORY, locks.JS_DOC];
+    revertCurrentDocument.reads = [locks.PS_DOC, locks.JS_APP, locks.JS_TOOL];
+    revertCurrentDocument.writes = [locks.JS_HISTORY, locks.JS_DOC, locks.PS_APP, locks.JS_POLICY];
     revertCurrentDocument.post = [layerActions._verifyLayerIndex];
 
     /**
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -208,31 +208,12 @@ define(function (require, exports) {
             .then(function (result) {
                 // After setting everything, dispatch to stores
                 this.dispatch(events.tool.SELECT_TOOL, result);
+                
+                return this.transfer(resetBorderPolicies);
             });
     };
     selectTool.reads = [locks.JS_APP, locks.JS_TOOL, locks.JS_SHORTCUT, locks.PS_DOC, locks.JS_DOC];
     selectTool.writes = [locks.PS_APP, locks.JS_POLICY, locks.PS_TOOL, locks.JS_TOOL, locks.JS_SHORTCUT,
-            locks.PS_DOC, locks.JS_DOC];
-
-    /**
-     * If current tool is superselect, we reselect it to re-set it's scroll policies
-     * Currently this action is called by window "resize" listener
-     *
-     * @return {Promise} Resolves to superselect tool select
-     */
-    var resetSuperselect = function () {
-        var toolStore = this.flux.store("tool"),
-            currentTool = toolStore.getCurrentTool();
-
-        // We only want to reset superselect tool
-        if (!currentTool || currentTool.id !== "newSelect") {
-            return Promise.resolve();
-        }
-
-        return this.transfer(selectTool, toolStore.getToolByID("newSelect"));
-    };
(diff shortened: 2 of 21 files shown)


debug.json
```diff
@@ -0,0 +1,6 @@
+{
+	"postConditions": true,
```
iwehrman: Argh, tabs!!!

Also, I think we might want to name this something like `debug.example.json` and then add `debug.json` to the .gitignore file so that git isn't always squawking about your non-default debug config file.

chadrolfs:
If I create a text layer, commit, click in the Size field and Shift + up arrow, i get:
`Missing read locks: jsApp, jsTool, psApp, jsPolicy`
`Action type.setSize failed: Error: Next action requires additional read locks`

chadrolfs:
Looking much better.  Here's one: scale, undo, then attempt to rotate (selection drops)
http://cl.ly/2W2k062m1c3t

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -202,9 +202,9 @@ define(function (require, exports, module) {
                 return;
             }
 
-            if (!this.state.marqueeEnabled) {
-                this.getFlux().actions.tools.resetBorderPoliciesDebounced();
-            }
+            // if (!this.state.marqueeEnabled) {
```
iwehrman: :turtle:

src/js/jsx/tools/PolicyOverlay.jsx
```diff
@@ -0,0 +1,121 @@
...
+            var flux = this.getFlux(),
+                policyStore = flux.store("policy"),
+                pointerPolicies = policyStore.getMasterPointerPolicyList();
+
+            return {
+                policies: pointerPolicies
+            };
+        },
+
+        _scrimGroup: null,
```
iwehrman: Need a little JSDoc here.

src/js/jsx/tools/PolicyOverlay.jsx
```diff
@@ -0,0 +1,121 @@
...
+                        policyRect = this._scrimGroup
+                        .append("rect")
+                        .attr("x", area[0])
+                        .attr("y", area[1])
+                        .attr("width", area[2])
+                        .attr("height", area[3])
+                        .style("fill-opacity", "0.0")
+                        .style("stroke-opacity", "0.0");
+    
+                    if (policy.action === 1) {
```
iwehrman: It would be nicer to compare to `os.eventKind` (or whatever it is).

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -54,11 +54,7 @@ define(function (require, exports) {
         mixins: [FluxMixin],
 
         shouldComponentUpdate: function (nextProps) {
-            var sameLayerIDs = collection.pluck(this.props.layers, "id")
-                .equals(collection.pluck(nextProps.layers, "id"));
-
-            return !sameLayerIDs ||
-                !Immutable.is(this.props.strokes, nextProps.strokes) ||
+            return !Immutable.is(this.props.strokes, nextProps.strokes) ||
```
iwehrman: I bet this is going to turn out to break something :)

src/js/jsx/Scrim.jsx
```diff
@@ -246,7 +248,8 @@ define(function (require, exports, module) {
                 transform = this.state.transform,
                 overlays = !disabled && this.state.overlaysEnabled,
                 transformString = this._getTransformString(transform),
-                toolOverlay = (overlays && transform) ? this._renderToolOverlay(transformString) : null;
+                toolOverlay = (overlays && transform) ? this._renderToolOverlay(transformString) : null,
+                policyOverlay = false ? (<PolicyOverlay/>) : null;
```
iwehrman: Wut.

src/js/actions/ui.js
```diff
@@ -440,13 +443,16 @@ define(function (require, exports) {
                 .bind(this)
                 .then(function () {
                     return this.transfer(centerBounds, document.bounds, false);
+                })
+                .then(function () {
+                    return this.transfer(tools.resetBorderPolicies);
```
iwehrman: It seems like this does _not_ need to wait for `centerBounds` above? Could those be `Promise.join`'d in the same handler?

src/js/actions/ui.js
```diff
@@ -154,10 +158,12 @@ define(function (require, exports) {
                 };
 
                 this.dispatch(events.ui.TRANSFORM_UPDATED, payload);
+            }).then(function () {
+                this.transfer(tools.resetBorderPolicies);
```
iwehrman: This also can go in the previous handler, and you probably also want to return the promise.

src/js/actions/ui.js
```diff
@@ -99,6 +100,9 @@ define(function (require, exports) {
                 };
 
                 this.dispatch(events.ui.TRANSFORM_UPDATED, payload);
+            })
+            .then(function () {
+                this.flux.actions.tools.resetBorderPolicies();
```
iwehrman: Ditto: can be moved into the previous handler, and the promise should be returned.

src/js/actions/tools.js
```diff
@@ -422,29 +406,14 @@ define(function (require, exports) {
...
-                        .bind(this)
-                        .then(function () {
-                            this.dispatchAsync(events.ui.TOGGLE_OVERLAYS, { enabled: true });
-                        });
-                }
+            this.flux.actions.tools.changeModalState(modalState);
 
+            if (event.kind._value === "mouse") {
                 // During artboard transforms, PS switches to artboard tool, so switch back to superselect
                 if (event.tool && event.tool.ID === "ArtT") {
```
iwehrman: Collapse this into a single `if` statement?

src/js/actions/tools.js
```diff
@@ -208,6 +208,9 @@ define(function (require, exports) {
             .then(function (result) {
                 // After setting everything, dispatch to stores
                 this.dispatch(events.tool.SELECT_TOOL, result);
+            })
+            .then(function () {
+                this.transfer(resetBorderPolicies);
```
iwehrman: Ditto: move this into the previous handler and return the promise.

iwehrman:
Done with a first pass! Back to you, @volfied.

src/js/jsx/Scrim.jsx
volfied: Because I had to remove debug.json... Once I find a nicer solution, I'll switch to that.

src/js/actions/ui.js
volfied: I can move it to the previous handler, but if I return this promise, I end up enqueuing resetBorder policies and deadlocking the app.

Maybe this is a bug in our action queue, but it feels like if an action queues another action, it shouldn't wait for it to resolve to consider itself finishing.

src/js/jsx/sections/style/Stroke.jsx
volfied: :watch: :bomb:

volfied:
@iwehrman I added more actions, I think I'm going to work on the transfer simplification we talked about for a while and later on clean up all actions.

chadrolfs:
screen shot 2015-07-17 at 10 49 06 am

chadrolfs:
`Missing read locks: jsApp, jsTool, psApp, jsPolicy`
Stroke Size

iwehrman:
Unassigning myself since this doesn't seem ready. Re-assign me when appropriate.

volfied:
@iwehrman Back at you... I also sped up marquee select a bit by checking for clicks on our side, instead of round tripping to Photoshop select.

@chadrolfs I think I covered every action this time. 

Please test in compiled build, or at least turn off the post condition verifications.

src/js/actions/history.js
```diff
@@ -153,6 +154,9 @@ define(function (require, exports) {
                                 count: count,
                                 selectedIndices: selectedIndices
                             });
+                        })
+                        .then(function () {
+                            return this.transfer(toolActions.resetBorderPolicies);
```
iwehrman: Does this need to wait for the previous dispatch? Also, that previous dispatch doesn't need to be async.

iwehrman: If it does need to wait for the previous dispatch though, then this can just be left alone.

src/js/actions/history.js
```diff
@@ -233,10 +237,13 @@ define(function (require, exports) {
             .bind(this)
             .then(function () {
                 return this.dispatchAsync(events.ui.TOGGLE_OVERLAYS, { enabled: true });
+            })
+            .then(function () {
+                return this.transfer(toolActions.resetBorderPolicies);
```
iwehrman: Ditto.

iwehrman:
Bombs away!

iwehrman:
I'm not sure which bugs this PR addresses. @volfied: can you mark the right ones FBNC?

PR #1745, opened 2015-07-16
UI Revamp, part 1

designedbyscience:
Added 2 columns of panels

Move toolbar to left side of screen

File-wise:
PanelSet.jsx includes most of past Panel.jsx
Now, Panel.jsx is very cut down container for individual panels

There is a little wonkiness with the size (height) of the Layers and Libraries panels when there are a lot of layers. The Libraries panel needs some CSS love as well, so that will be an outstanding request


src/js/jsx/PanelSet.jsx
```diff
@@ -0,0 +1,265 @@
...
+
+                        var libraryPanel = this.state.librariesEnabled ? (
+                            <LibrariesPanel
+                                disabled={disabled}
+                                visible={this.state[UI.LIBRARY_PANEL]}
+                                onVisibilityToggle=
+                                    {this._handlePanelVisibilityToggle.bind(this, UI.LIBRARY_PANEL)} />
+                        ) : null;
+                        
+                        return (<div className={panelSetclassNames} key={documentID}>
```
iwehrman: Can you bump everything after the opening parenthesis to a new line, thereby saving one step of indentation for the large JSX block below?

src/js/jsx/PanelSet.jsx
```diff
@@ -0,0 +1,265 @@
...
+                                                viewBox="0 0 17 17"
+                                                CSSID="layers" />
+                                        </Button>
+                                    </div>
+                                </div>
+                        );
+                    }, this);
+                    
+                    
+                return (<div ref="panelSet" className="panel-set__container">
```
iwehrman: Ditto here. Also, remove one of the two successive newlines above.

src/js/jsx/PanelSet.jsx
```diff
@@ -0,0 +1,265 @@
...
+                        var libraryPanel = this.state.librariesEnabled ? (
+                            <LibrariesPanel
+                                disabled={disabled}
+                                visible={this.state[UI.LIBRARY_PANEL]}
+                                onVisibilityToggle=
+                                    {this._handlePanelVisibilityToggle.bind(this, UI.LIBRARY_PANEL)} />
+                        ) : null;
+                        
+                        return (<div className={panelSetclassNames} key={documentID}>
+                                    <Panel
```
iwehrman: Maybe we should rename `Panel` to `PanelColumn`?

iwehrman: (I realize this review is not currently what one would call "useful".)

src/style/properties.less
```diff
@@ -161,9 +230,12 @@
 }
 
 .layers {
-    flex-grow: 1000;
-    flex-basis: 5rem;
-    bottom: 0;
+    flex-grow: 1;
+    flex-shrink: 1;
+    // max-height: 100%;
```
iwehrman: Stale comment cleanup on aisle 9.

src/style/shared/title-header.less
```diff
@@ -68,7 +68,7 @@
 }
 
 .section-title {
-    border-top: @hairline solid @color-section-header;
+    // border-top: @hairline solid @color-section-header;
```
iwehrman: Ditto.

src/js/jsx/Panel.jsx
```diff
@@ -26,107 +26,37 @@ define(function (require, exports, module) {
...
+                    "panel": true,
+                    "panel__visible": this.props.visible
+                });
+
+            return (
+                <div className={className}>
+                    <div
+                        className="panel__hide"
+                        onClick={this.props.onVisibilityToggle}>
+                        ▸
```
iwehrman: In my experience, there is a high probability that this character will not render correctly on Windows. Also, I'm probably going to start a design argument about this particular UI signifier, but not for this PR.

iwehrman:
Well, this is remarkably painless so far. Great work! I'm contemplating holding up the merge until the beginning of the following sprint though, in case we decide that we want to release S26 on the pre-release forum.

Other things to confirm before the merge:

1. That this doesn't cause any performance regressions with big documents. In particular, please compare: document switch, panel expand/collapse, layer drag redraw, and layer targeting.
2. That the icon in the panel-visibility signifier works on Windows.

Also, the pinned toolbar looks screwy. And the branch needs to be rebased against master.

src/js/jsx/PanelSet.jsx
designedbyscience: Yea, I had called it PanelColumn at one point...changing it!

designedbyscience:
1. Checked performance, seems the same/comparable
2. It does work on Windows, but switched to an icon for consistency 

The toolbar looks less screwy. Added some other minor UI tweaks

chadrolfs:
for the record, i'd rather not hold up merging this for non-quality reasons.

iwehrman:
Here we go!

Issue #1707, opened 2015-07-09
Command click won't select artboards


iwehrman:
Neither clicking on the artboard background nor the artboard title selects the artboard. It seems like at least the latter should.

chadrolfs:
I think that is as design.  It's like trying to cmd + click on a group.

chadrolfs:
@iwehrman do you disagree?  ^^

placegraphichere:
CMD+click should always select an artboard (with move tool selected). Current release works that way

chadrolfs:
Current release is wrong tho.  You have to double-click to drill into the AB layer and select it's children and marquee select includes the AB which it shouldn't. 
My argument technically speaking is that AB layers under the hood are essentially group layers and behave similarly.  You can't cmd+click and select a group layer.  If we want to somehow special case ABs in this way, I think it will take up valuable dev time that is needed elsewhere.

chadrolfs:
I see what you mean about cmd+click on the title tho - that's how it works in standard PS

placegraphichere:
yeah, I'm only speaking about the artboard title for this whole thing. CMD+click should probably not select an artboard when clicking on the artboard area. 

This can create a lot of confusion for users, especially when zoomed-in where there may be no off-artboard place to de-select. Otherwise, accidental selecting and repositioning of the artboard can occur when it was probably not wanted. 

title CMD+click = yes
artboard area CMD+click = no

chadrolfs:
works if there are no underlying layers, related to #1751

PR #1746, opened 2015-07-16
Let artboard name badges be cmd+clickable, but not artboards

volfied:
Addresses #1707

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -188,15 +188,29 @@ define(function (require, exports) {
     };
 
     /**
-     * Gets the non group layers that have one of the passed in IDs
+     * Gets the artboards and leaf layers that have one of the passed in IDs
      * 
      * @param  {LayerStructure} layerTree
      * @param  {Object.<{number: boolean}>} layerMap       
      * @return {Immutable.Iterable.<Layer>}
      */
-    var _getLeafLayersWithID = function (layerTree, layerMap) {
-        return layerTree.leaves.filter(function (layer) {
+    var _getDirectAccessLayersWithID = function (layerTree, layerMap) {
+        return layerTree.artboards.concat(layerTree.leaves).filter(function (layer) {
             return layerMap.hasOwnProperty(layer.id);
+        });
+    };
+
+    /**
+     * Removes all artboards from the given ids
+     * We use this to prevent artboards from being cmd+clickable
+     *
+     * @param {LayerStructure} layerTree
+     * @param {Immutable.List.<number>} ids
+     * @return {Immutable.Iterable.<Layer>}
+     */
+    var _removeArtboardIDs = function (layerTree, ids) {
+        return ids.filterNot(function (id) {
+            return layerTree.layers.get(id).isArtboard;
         });
     };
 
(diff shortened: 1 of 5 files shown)


src/js/actions/superselect.js
```diff
@@ -188,19 +188,37 @@ define(function (require, exports) {
...
+     * Gets the artboards and leaf layers that have one of the passed in IDs
      * 
      * @param  {LayerStructure} layerTree
      * @param  {Object.<{number: boolean}>} layerMap       
      * @return {Immutable.Iterable.<Layer>}
      */
-    var _getLeafLayersWithID = function (layerTree, layerMap) {
-        return layerTree.leaves.filter(function (layer) {
+    var _getDirectAccessLayersWithID = function (layerTree, layerMap) {
+        var artboards = layerTree.top.filter(function (layer) {
```
mcilroyc: I guess maybe we've moved beyond the "cache everything" fad of the earlier this year, but seems like "allArtboards" would be a reasonable layerstructure property.

src/js/actions/superselect.js
```diff
@@ -188,19 +188,37 @@ define(function (require, exports) {
...
             return layerMap.hasOwnProperty(layer.id);
         });
     };
 
     /**
+     * Removes all artboards from the given ids
+     * We use this to prevent artboards from being cmd+clickable
+     *
+     * @param {LayerStructure} layerTree
+     * @param {Array.<number>} ids
```
mcilroyc: ids is an `Immutable.List`.  man I'm on fire this morning.

src/js/actions/superselect.js
```diff
@@ -353,20 +371,20 @@ define(function (require, exports) {
...
+                    coveredLayerIDs = collection.pluck(coveredLayers, "id").concat(hitLayerIDsSansArtboards);
 
                 coveredLayerIDs = coveredLayerIDs.sortBy(function (id) {
                     return hitLayerIDs.indexOf(id);
                 });
 
                 if (deep) {
-                    // Select any non-group layer
+                    // Select any non-group layer, and allow for artboard badges
                     var hitLayerMap = coveredLayerIDs.reduce(function (layerMap, id) {
```
mcilroyc: Seem like this doesn't really need to be a faux-map anymore.  It probably don't hurt anything, but might be more clear as a simple array of Ids, since all it is used for (I think) is testing for existence (?)

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -250,6 +250,12 @@ define(function (require, exports, module) {
 
             if (this._leafBounds) {
                 renderLayers = layerTree.leaves.sortBy(indexOf);
+
+                // We add artboards here, so they are shown selectable
+                renderLayers = renderLayers.concat(layerTree.top.filter(function (layer) {
+                    return layer.isArtboard;
```
mcilroyc: wouldn't it be great if this were a cached set of artboards?  :zap:

src/js/actions/superselect.js
volfied: Yeah it's only used for existence testing, but they're not sorted and I'd like it to be fast instead of memory efficient.

mcilroyc:
just a little JSDoc fix needed, and then it's mergable.

volfied:
@mcilroyc Thank you, it's ready.

chadrolfs:
@mcilroyc thanks for taking this.  @volfied I took a look too - solid

PR #1748, opened 2015-07-16
Only show search filter as an option when options with that category …

hamburge:
…are available

Now when you only have one document open, you won't be able to filter by Current Documents.
Or when you don't have any Smart Objects in a document, you won't be able to filter by smart objects.
ETC.

### DIFF
--- a/src/js/actions/search/documents.js
+++ b/src/js/actions/search/documents.js
@@ -27,8 +27,7 @@ define(function (require, exports) {
     var Immutable = require("immutable");
 
     var events = require("js/events"),
-        pathUtil = require("js/util/path"),
-        strings = require("i18n!nls/strings");
+        pathUtil = require("js/util/path");
 
     /**
      * Make list of current documents info so search store can create search options
@@ -47,7 +46,7 @@ define(function (require, exports) {
                 return {
                     id: doc.toString(),
                     name: docStore.getDocument(doc).name,
-                    category: [strings.SEARCH.CATEGORIES.DOCUMENT, strings.SEARCH.CATEGORIES.CURRENT],
+                    category: ["CURRENT_DOC"],
                     pathInfo: "",
                     iconID: "tool-rectangle"
                 };
@@ -69,7 +68,7 @@ define(function (require, exports) {
                 return {
                     id: index.toString(),
                     name: pathUtil.getShortestUniquePaths(Immutable.List.of(doc)).toJS()[0],
-                    category: [strings.SEARCH.CATEGORIES.DOCUMENT, strings.SEARCH.CATEGORIES.RECENT],
+                    category: ["RECENT_DOC"],
                     pathInfo: doc,
                     iconID: "tool-rectangle"
                 };
@@ -110,7 +109,7 @@ define(function (require, exports) {
         var currentDocPayload = {
             "type": "CURRENT_DOC",
             "getOptions": _currDocSearchOptions.bind(this),
-            "filters": Immutable.List(["CURRENT"]),
+            "filters": Immutable.List.of("CURRENT_DOC"),
             "handleExecute": _confirmCurrDocSearch.bind(this)
         };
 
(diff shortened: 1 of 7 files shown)


src/js/actions/search/documents.js
```diff
@@ -110,7 +109,7 @@ define(function (require, exports) {
         var currentDocPayload = {
             "type": "CURRENT_DOC",
             "getOptions": _currDocSearchOptions.bind(this),
-            "filters": Immutable.List(["CURRENT"]),
+            "filters": Immutable.List(["CURRENT_DOC"]),
```
volfied: The nit-picker 2000: You can use `Immutable.List.of("CURRENT_DOC")` here

src/js/stores/search.js
```diff
@@ -204,13 +205,16 @@ define(function (require, exports, module) {
         _updateSearchItems: function (id) {
             var search = this._registeredSearches[id],
                 types = search.searchHeaders;
-
+            
+            search.searchFilters = [];
             if (types) {
```
volfied: Newline here

src/js/stores/search.js
```diff
@@ -259,17 +264,23 @@ define(function (require, exports, module) {
...
-                if (this._registeredSearchTypes[header]) {
-                    return this._registeredSearchTypes[header].filters;
-                }
-            }.bind(this), []);
+        _getFiltersByItems: function (items, type) {
+            if (items.size === 0) {
+                return [];
+            }
+            var moduleFilters = this._registeredSearchTypes[type].filters.toJS(),
+                itemCategories = collection.pluck(items, "category").toJS(),
```
volfied: how about `collection.pluck(items, "category").flatten()`

volfied: Yeah I think you can get away with keeping this function fully immutable.

src/nls/root/strings.js
```diff
@@ -584,11 +584,9 @@ define(function (require, exports, module) {
                 RECENT_DOC: "Recent Documents"
             },
             CATEGORIES: {
-                CURRENT_DOC: "document",
-                RECENT_DOC: "document",
+                CURRENT_DOC: "current document",
```
volfied: I want to make sure, are these going to be localized? if not, they shouldn't be defined here. If so, they should be capitalized properly.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -223,7 +224,7 @@ define(function (require, exports, module) {
             case "Enter":
             case "Return":
                 if (!this.props.live && this.props.onKeyDown &&
-                        this.state.id.indexOf("filter") === 0) {
+                        this.state.id.indexOf("FILTER") === 0) {
```
volfied: `this.state.id.startsWith("FILTER")`

volfied: Also, don't you need to check for this.state.id here as well?

src/js/actions/search/layers.js
```diff
@@ -131,7 +129,8 @@ define(function (require, exports) {
      *
      */
     var registerLayerSearch = function () {
-        var filters = Immutable.List(["PIXEL", "TEXT", "ARTBOARD", "ADJUSTMENT", "SMARTOBJECT", "GROUP", "VECTOR"]);
+        var filters = Immutable.List(
+                ["LAYER", "PIXEL", "TEXT", "ARTBOARD", "ADJUSTMENT", "SMARTOBJECT", "GROUP", "VECTOR"]);
```
volfied: Would look better if the closing parentheses was in a new line with same indent as the var

volfied:
@hamburge did what I do best and nit picked. Also have couple potential bugs. Please take a look.

src/nls/root/strings.js
hamburge: They should be localized because they are visible as titles for the filter options and are compared to user input for autofilling and things. But yes, capitalization sounds great!

src/js/stores/search.js
hamburge: `flatten()` isn't flattening it, I think because the category lists are just arrays, not Immutable.

But I can make it more readable using `_.flatten()`

volfied:
I like it!

PR #1750, opened 2015-07-17
Basic Asset Export Feature

mcilroyc:
This provides a barebones UI for generating assets in Design Space.  You can configure layers, groups and artboards to generate assets, in several formats (PNG, JPG, etc), at any (or all) of 4 pre-defined scales, and then generate all configured assets for the document with one click.  

The UI is split in two places.

1. Assets are configured (one layer at a time) in a new EXPORT panel below the STYLE panel.  
1. Once configured, the assets are *generated* by opening the export modal dialog (currently using Cmd+e)

Dependencies:
- The necessary generator version (3.5) is now built-in to our playground-dev builds as of (mac) 423
- The necessary generator *plugin* is now built-in to our playground-dev builds as of mac 447 / win 449
- Generator Enabled via (classic UI) Preferences > Plugin Ins...
- Requires a `bower install`

If you want to run the generator plugin yourself:
- [generator-spaces](https://github.com/adobe-photoshop/generator-spaces) plugin (private repo)
  - download from github, run `npm install`, OR...
  - download zip file that I'll post to slack
  - Installed/unzipped as a subdirectory of `Adobe\ Photoshop\ CC\ 2015.app/Contents/Required/Plug-Ins/Generator/` 

usage/caveats
- ESC shortcut closes export dialog, but then it won't open back up properly
- open the export util with `Cmd+e`


src/js/actions/export.js
```diff
@@ -0,0 +1,222 @@
...
+        events = require("js/events"),
+        locks = require("js/locks"),
+        log = require("js/util/log"),
+        ExportService = require("js/util/exportservice");
+
+    var descriptor = require("adapter/ps/descriptor"),
+        documentLib = require("adapter/lib/document"),
+        layerLib = require("adapter/lib/layer");
+
+    var EXTENSION_DATA_NAMESPACE = "designSpace";
```
iwehrman: Elsewhere we use something like `com.adobe.photoshop.spaces.design` for a name in a global space.

src/js/actions/export.js
```diff
@@ -0,0 +1,222 @@
...
+        var documentID = document.id,
+            documentExports = this.flux.stores.export.getDocumentExports(documentID),
+            layerExportsMap = documentExports && documentExports.layerExportsMap,
+            exportArray = [];
+
+        if (!layerExportsMap || layerExportsMap.size < 1) {
+            return Promise.resolve("no assets to export");
+        }
+
+        layerExportsMap.forEach(function (layerExportAssets, layerID) {
```
iwehrman: Aside: the worst part of `window.Map` is the dearth of built-in iterators compared to those available in Immutable, like `.reduce`, `.map`, `.flatMap`, etc. (I think this would be nicer with the latter two.) We could always just write those helper functions ourselves though and stick them on `window.Map.prototype`. Food for thought :poultry_leg:

src/js/actions/export.js
```diff
@@ -0,0 +1,222 @@
...
+            var layer = document.layers.byID(layerID);
+            layerExportAssets.forEach(function (asset) {
+                exportArray.push({ layer: layer, asset: asset });
+            }, this);
+        }, this);
+
+        // TODO this is intentionally serial for now, because or WS is lame
```
iwehrman: Tell me more!

src/js/util/exportservice.js
```diff
@@ -0,0 +1,115 @@
...
+        this.ws = new window.WebSocket(_wsUrl);
+        this.deferred = null;
+    };
+
+    /**
+     * init
+     *
+     * @return {Promise}
+     */
+    ExportService.prototype.init = function () {
```
iwehrman: I'm curious why you didn't use `NodeConnection`/`NodeDomain` for this stuff?

src/js/actions/export.js
mcilroyc: Alas, It seems that we are not allowed to use `.` within this string.  "because of a program error"

src/js/actions/export.js
mcilroyc: Agreed, and noted

src/js/actions/export.js
mcilroyc: Just to be clear, this will *no longer be serial* now that we're using NodeDomain.  My original lame websocket implementation was braindead and single threaded.

mcilroyc:
I pushed a rebased and updated version, and updated the install instructions based on the new generator-spaces plugin

mcilroyc:
I pushed another heavily-squashed rebase.  

@chadrolfs this fixes the error I had mentioned earlier that was preventing the export feature from being used with multiple open docs.  This is now, in my opinion, worth a quick functional pass by you.  The UI is far from what Seth has spec'd.

@iwehrman there are still missing JSDoc, and other known code cleanliness things that I intend to address prior to submitting for full review.

chadrolfs:
ok, I :eyes:

chadrolfs:
I was able to get it going on Win but ran into some _Destination Path Too Long_ issues, specifically with ...Generator/node_modules/.../lodash

Also there's no way to access the export util on Win, ctrl + shift + J doesn't work

mcilroyc:
@iwehrman If you could take a pass through this again, please.  I think it is all-but-ready for merging as a basic feature.  I'm thinking I should forcibly disabled it on windows until I can figure out the plugin issues there, comments welcome.

I think DataList has rotted a bit in the last week.  The drop-downs in the Export Panel don't work as well as they did previously.   I'll look in to that.

chadrolfs:
Windows disabled?  :worried:

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+    var dialog = require("./dialog"),
+        events = require("js/events"),
+        locks = require("js/locks"),
+        collection = require("js/util/collection"),
+        log = require("js/util/log"),
+        ExportAsset = require("js/models/exportasset"),
+        ExportService = require("js/util/exportservice");
+
+    var descriptor = require("adapter/ps/descriptor"),
+        // documentLib = require("adapter/lib/document"),
```
iwehrman: Stale comment.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+
+    var dialog = require("./dialog"),
+        events = require("js/events"),
+        locks = require("js/locks"),
+        collection = require("js/util/collection"),
+        log = require("js/util/log"),
+        ExportAsset = require("js/models/exportasset"),
+        ExportService = require("js/util/exportservice");
+
+    var descriptor = require("adapter/ps/descriptor"),
```
iwehrman: Also, for consistency, we order these module import groups in the following order:

1. third-party imports
2. adapter imports
3. spaces-design imports

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+        // documentLib = require("adapter/lib/document"),
+        layerLib = require("adapter/lib/layer");
+
+    /**
+     * Export metadata is stored in the "extension" data structure in Ps, using this as the namespace
+     * @type {String}
```
iwehrman: Basic JavaScript types ("string", "boolean", "object", "number") should be lower case in JSDoc type expressions.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+     * @private
+     * @type {ExportService}
+     */
+    var _exportService;
+
+    /**
+     * before start up, initialize the export service
+     * @return {Promise}
+     */
+    var beforeStartup = function () {
```
iwehrman: All the lifecycle functions (`beforeStartup`, etc.) should go at the bottom of the file.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+        _exportService = new ExportService();
```
iwehrman: Also, if you create state in the module, you need to make sure to free it in `onReset`.

iwehrman: Well, you don't necessarily _have_ to, but keep in mind that, on error, the controller will call `beforeStartup` again. So most likely in `onReset` you want to do some kind of cleanup.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+
+        assetPropsArray[assetIndex] = props;
+
+        payload = {
+            documentID: documentID,
+            layerIDs: layer.id,
+            assetPropsArray: assetPropsArray
+        };
+
+        return this.dispatchAsync(events.export.ASSET_CHANGED, payload).bind(this).then(function () {
```
iwehrman: For consistency, all .binds, .thens, .catches and .finallys go on a new line.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+        return this.transfer(updateLayerExportAsset, document, layer, assetIndex, { format: format });
+    };
+    updateLayerAssetFormat.reads = [];
+    updateLayerAssetFormat.writes = [];
+    updateLayerAssetFormat.transfers = [updateLayerExportAsset];
+
+    /**
+     * Update the status of all assets that are being requested
+     *
+     * @param {Document} document [description]
```
iwehrman: `[description]`

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+                collection.pluck(layers, "id").toArray() :
+                [layers.id];
+
+        var payload = {
+                documentID: document.id,
+                layerIDs: layerIDs,
+                exportEnabled: exportEnabled
+            };
+
+        this.dispatch(events.document.LAYER_EXPORT_ENABLED_CHANGED, payload);
```
iwehrman: Dispatches not in a then handler should be async.

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+            documentExports = this.flux.stores.export.getDocumentExports(documentID),
+            layerExportsMap = documentExports && documentExports.layerExportsMap,
+            exportArray = [];
+
+        if (!layerExportsMap || layerExportsMap.size < 1) {
+            return Promise.resolve("no assets to export");
+        }
+
+        // Iterate over the exports map, find the associated layer, test of "exportEnabled"
+        layerExportsMap.forEach(function (layerExportAssets, layerID) {
```
iwehrman: This could probably be a `.flatMap`?

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+    updateLayerAssetFormat.writes = [];
+    updateLayerAssetFormat.transfers = [updateLayerExportAsset];
+
+    /**
+     * Update the status of all assets that are being requested
+     *
+     * @param {Document} document [description]
+     */
+    var setAllAssetsRequested = function (document) {
+        var layerIDs = document.layers.all.reduce(function (IDs, layer) {
```
iwehrman: If it suits your fancy, this could be more structured with `.toSeq`/`.filter`/`.map`/`.toJS`

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+     * the photoshop layer "extension data"
+     *
+     * TODO maybe this should handle a set of layers, for efficiency when called by layer actions?
+     *
+     * @private
+     * @param {number} documentID
+     * @param {number} layerID
+     * @return {Promise}
+     */
+    var _syncLayerExportMetadata = function (documentID, layerID) {
```
iwehrman: Can we roughly order the functions in this file in dependency order? Privates up top, publics down below, lifecycle all the way at the bottom?

src/js/actions/export.js
```diff
@@ -0,0 +1,365 @@
...
+            layer = document && document.layers.byID(layerID),
+            exportEnabled = document && layer && layer.exportEnabled;
+
+        var exportsMetadata = {
+            exportAssets: layerExportsArray,
+            exportEnabled: exportEnabled === undefined ? false : exportEnabled
+        };
+
+        var playObject = layerLib.setExtensionData(documentID,
+            layerID, EXTENSION_DATA_NAMESPACE, "exportsMetadata", exportsMetadata);
```
iwehrman: Maybe indent this line one more tab stop.

src/js/actions/layers.js
```diff
@@ -144,7 +146,19 @@ define(function (require, exports) {
         });
 
         return Promise.join(requiredPropertiesPromise, optionalPropertiesPromise, function (required, optional) {
-            return _.chain(required).zipWith(optional, _.merge).reverse().value();
+            var propArray = _.chain(required).zipWith(optional, _.merge).reverse().value();
+            
+            // TODO ideally we could perform this extension data fetch as part of (or at least parallel to)
```
iwehrman: Yikes? Let's talk about how we can make this better before merging. This is performance-sensitive.

iwehrman: If nothing else it seems like we should be able to `batchPlay` these?

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+        CheckBox = require("jsx!js/jsx/shared/CheckBox"),
+        TitleHeader = require("jsx!js/jsx/shared/TitleHeader");
+
+    var strings = require("i18n!nls/strings"),
+        ExportAsset = require("js/models/exportasset");
+
+    /**
+     * Private React component that represents one Layer and its configured export assets
+     */
+    var PanelLayer = React.createClass({
```
iwehrman: Bad variable name.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+            this.getFlux().actions.export.setLayerExportEnabled(this.props.document, this.props.layer, checked);
+        },
+
+        render: function () {
+            var layer = this.props.layer,
+                layerExports = this.props.layerExports,
+                freshState = this.props.freshState;
+
+            // The list of assets within a layer
+            var assetsComponent = layerExports.map(function (asset, key) {
```
iwehrman: Could exports have more useful (i.e., persistent) than just this index? It's faster for React to add/remove/reorder children if their keys are persistent.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+
+            return (
+                <div className="exports-panel__layer-wrapper" >
+                    <CheckBox
+                        checked={layer.exportEnabled}
+                        onChange={this._handleLayerSelectedChanged}
+                        size="column-2" />
+                    <div className="exports-panel__layer-info">
+                        <div className="exports-panel__layer__name">
+                            {layer.name}
```
iwehrman: Might want a `title` here for a tooltip in case the name is clipped.

src/js/jsx/sections/export/ExportAllPanel.jsx
iwehrman: Er, maybe not a bad variable name. But, still kind of confusing. Anyway, I was confused.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+            return {
+                dismissDialog: _.identity
+            };
+        },
+
+        getInitialState: function () {
+            // the fresh state property allows assets to be determined as "stale"
+            // meaning that they have been exported previously, but not during this modal session
+            return {
+                fresh: true
```
iwehrman: image

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+                });
+            }.bind(this));
+        },
+
+        render: function () {
+            var document = this.state.document,
+                layerExportsMap = this.state.documentExports && this.state.documentExports.layerExportsMap,
+                freshState = this.state.fresh;
+
+            if (!document || !layerExportsMap) {
```
iwehrman: There is no reason I can think of why there would not be a document at this point. The panel architecture guarantees this so that these checks don't need to be made repeatedly.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+
+            // Iterate over all configured assets, build the individual components,
+            // and separate into separate lists (artboards vs. non-artboards)
+            var layerExportComponents = [],
+                artboardExportComponents = [];
+            if (layerExportsMap && layerExportsMap.size > 0) {
+                layerExportsMap.forEach(function (layerExports, key) {
+                    var layer = document.layers.byID(key);
+
+                    if (layer && layerExports && layerExports.size > 0) {
```
iwehrman: Is there also a reason you know of that would cause `layer` to be null here?

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+            }
+
+            return (
+                <div className="exports-panel__container">
+                    <TitleHeader
+                        title={strings.TITLE_EXPORT} />
+                    <div className="exports-panel__two-column">
+                        <div className="exports-panel__asset-list__container">
+                            <TitleHeader
+                                title="Artboards" />
```
iwehrman: Should be localized.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+                            <div className="exports-panel__asset-list__list">
+                                {artboardExportComponents}
+                            </div>
+                        </div>
+                        <div className="exports-panel__asset-list__container">
+                            <TitleHeader
+                                title="Layers" />
```
iwehrman: Should be localized.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -0,0 +1,226 @@
...
+                            <div className="exports-panel__asset-list__list">
+                                {layerExportComponents}
+                            </div>
+                        </div>
+                    </div>
+                    <hr />
+                    <div className="exports-panel__button-group">
+                        <Button
+                            onClick={this.props.dismissDialog}>
+                            Cancel
```
iwehrman: Should be localized. (And, yes, we should do this ASAP so that the loc team can get started even before the UI is final.)

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -0,0 +1,122 @@
...
+        },
+
+        /**
+         * Workaround a CEF bug by clearing any active tooltips when scrolling.
+         * More details here: https://github.com/adobe-photoshop/spaces-design/issues/444
+         *
+         * @private
+         */
+        _handleScroll: function () {
+            this._setTooltipThrottled("");
```
iwehrman: Maybe we should turn this into a mixin?

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+        Label = require("jsx!js/jsx/shared/Label"),
+        Button = require("jsx!js/jsx/shared/Button"),
+        SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
+        Datalist = require("jsx!js/jsx/shared/Datalist"),
+        TextInput = require("jsx!js/jsx/shared/TextInput");
+
+    var collection = require("js/util/collection");
+
+
+    var scaleOptions = Immutable.OrderedMap({
```
iwehrman: JSDoc, `@private`, `_` prefix.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+            index: React.PropTypes.number.isRequired,
+            layer: React.PropTypes.object.isRequired,
+            exportAsset: React.PropTypes.object.isRequired
+        },
+
+        /**
+         * Delete this asset
+         *
+         * @private
+         * @return {Promise}
```
iwehrman: This does not actually return anything, and I don't think it needs to either.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                index = this.props.index;
+
+            this.getFlux().actions.export.deleteLayerExportAsset(document, layer, index);
+        },
+
+        /**
+         * Update this asset's scale
+         *
+         * @private
+         * @return {Promise}
```
iwehrman: Ditto, throughout.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+         */
+        _handleUpdateScale: function (scale) {
+            var scaleNum = Number.parseFloat(scale);
```
iwehrman: It's very easy to make `parseFloat` return a `NaN`. I think `math.parseNumber` does what you want here, but better. But, you still need to test somewhere for the result being `null`.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                exportAsset = this.props.exportAsset,
+                scale = exportAsset.scale || 1,
+                scaleOption = scaleOptions.has(scale.toString()) ?
+                    scaleOptions.get(scale.toString()) : scaleOptions.get("1"),
+                scaleListID = "layerExportAsset-scale" + layer.id + "-" + this.props.index,
+                formatListID = "layerExportAsset-format-" + layer.id + "-" + this.props.index;
+
+            return (
+                <div className="formline">
+                    <Datalist
```
iwehrman: So we aren't going to allow for arbitrary numbers to be entered here? If this is is because @placegraphichere's comps showed widget that allowed you to enter arbitrary numbers but also had a menu for default selections, we've decided elsewhere that for the first pass this should just be a number input. But, maybe in this case we _only_ want to allow the user to choose from a finite set of scale options. In any case, we should explicitly choose one or the other for the near term.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                        className="dialog-export-format"
+                        options={formatOptions.toList()}
+                        value={exportAsset.format.toUpperCase()}
+                        onChange={this._handleUpdateFormat}
+                        live={false}
+                        size="column-4" />
+                    <Gutter />
+                    <Button
+                        className="button-plus" // a bit of a hack
+                        title="remove asset configuration"
```
iwehrman: Requires localization.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                        onClick={this._handleDeleteClick}>
+                        <SVGIcon
+                            viewbox="0 0 12 14"
+                            CSSID="libraries-delete" />
```
iwehrman: Presumably the end-goal is to either have an export-specific icon, or to rename the libraries delete icon to be something more general? If nothing else, add a FIXME here about this.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+            var document = this.props.document,
+                documentExports = this.state.documentExports,
+                layerExports = documentExports && documentExports.layerExportsMap.get(layer.id);
+
+            // Determine the scale of the potential next asset
+            var remainingScales = _.difference(allScales, collection.pluck(layerExports, "scale").toArray()),
+                nextScale = remainingScales.length > 0 ? remainingScales[0] : null,
+                nextAssetIndex = (layerExports && layerExports.size) || 0;
+
+            return this.getFlux().actions.export.updateLayerAssetScale(document, layer, nextAssetIndex, nextScale);
```
iwehrman: Probably no need to return this?

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+        },
+
+        render: function () {
+            var document = this.props.document,
+                documentExports = this.state.documentExports;
+
+            if (!document || document.layers.selected.size !== 1) {
+                return (<div>Please select a single layer</div>);
```
iwehrman: Requires localization. Also, above, the redundant no-document checks.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+            }
+
+            var selectedLayer = document.layers.selected.first(),
+                layerExports = documentExports && documentExports.layerExportsMap.get(selectedLayer.id);
+
+            var exportComponents = [];
+            if (layerExports && layerExports.size > 0) {
+                layerExports.forEach(function (i, k) {
```
iwehrman: This would be nicer as `.map`/`.toArray`.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                    var x = (
+                        <LayerExportAsset
+                            document={this.props.document}
+                            index={k}
+                            key={k}
+                            layer={selectedLayer}
+                            exportAsset={i} />
+                    );
+                    exportComponents.push(x);
+                }.bind(this));
```
iwehrman: Also, we prefer the `this` argument to an explicit bind.

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -0,0 +1,297 @@
...
+                            className="button-plus"
+                            onClick={this._addAssetClickHandler.bind(this, selectedLayer)}>
+                            <SVGIcon
+                                viewbox="0 0 12 12"
+                                CSSID="plus" />
+                        </Button>
+                    </div>
+                    <div className="formline">
+                        <Label
+                            title="Scale"
```
iwehrman: These titles all require localization.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -44,7 +44,7 @@ define(function (require, exports, module) {
      */
     var Datalist = React.createClass({
         propTypes: {
-            options: React.PropTypes.instanceOf(Immutable.Iterable),
+            options: React.PropTypes.instanceOf(Immutable.Iterable), // FIXME should be LIST b/c findIndex() used
```
iwehrman: Just fix it?

src/js/locks.js
```diff
@@ -52,7 +52,8 @@ define(function (require, exports, module) {
         JS_STYLE: "jsStyle",
         JS_LIBRARIES: "jsLibraries",
         CC_LIBRARIES: "ccLibraries",
-        OS_CLIPBOARD: "osClipboard"
+        OS_CLIPBOARD: "osClipboard",
+        PS_GENERATOR: "psGenerator"
```
iwehrman: Probably no need for a `PS_` prefix here? That is, probably no need to add it to `ALL_PS_LOCKS`. But, might want to add it to `ALL_NATIVE_LOCKS`?

src/js/models/exportasset.js
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Merge in a set of properties, and additionally
+     * derive a new suffix IF the scale is changing AND the suffix was not previously customized
+     *
+     * @param {object} props ExportAsset-like properties
+     * @return {ExportAsset}
+     */
+    ExportAsset.prototype.mergeProps = function (props) {
+        if (!this.suffixCustomized && props.scale && (props.scale !== this.scale)) {
+            return this.merge(props).deriveSuffix();
```
iwehrman: Looks like the prop is called `derivedSuffix`?

iwehrman: Er, nm.

src/js/models/layer.js
```diff
@@ -436,7 +442,8 @@ define(function (require, exports, module) {
             text: Text.fromLayerDescriptor(resolution, layerDescriptor),
             proportionalScaling: layerDescriptor.proportionalScaling,
             isArtboard: layerDescriptor.artboardEnabled,
-            vectorMaskEnabled: layerDescriptor.vectorMaskEnabled
+            vectorMaskEnabled: layerDescriptor.vectorMaskEnabled,
+            exportEnabled: layerDescriptor.exportEnabled
```
iwehrman: What about in `resetFromDescriptor` below?

src/js/util/collection.js
```diff
@@ -149,7 +149,7 @@ define(function (require, exports) {
     };
 
     /**
-     * Calculate the difference of the intersection and collection.
+     * Calculate the difference of the iterable and collection.
```
iwehrman: :white_check_mark:

src/js/util/exportservice.js
```diff
@@ -0,0 +1,133 @@
...
+     * @type {number}
+     */
+    var GENERATOR_PORT = 59596;
+
+    /**
+     * Duration after which we should give up trying to connect to the export plugin
+     * @const
+     * @type {number}
+     */
+    var CONNECTION_TIMEOUT_MS = 2000;
```
iwehrman: Might want to be a little more generous here.

src/js/util/exportservice.js
```diff
@@ -0,0 +1,133 @@
...
+     */
+    var getRemotePort = function (callback) {
+        callback(null, GENERATOR_PORT);
+    };
+
+    /**
+     * @constructor
+     */
+    var ExportService = function () {
+        this.spacesDomain = null;
```
iwehrman: This property should be defined on the prototype, initialized to null there, and probably also marked as private?

src/nls/root/shortcuts-win.js
```diff
@@ -38,6 +38,7 @@ define(function (require, exports, module) {
                 SAVE: "s",
                 SAVE_AS: "s",
                 REVERT: "KEY_F12",
+                EXPORT_UTILITY: "e",
```
iwehrman: Are these keyboard shortcuts cleared with @placegraphichere? I just want to make sure this doesn't conflict with some important PS shortcut...

src/static/menu-win.json
```diff
@@ -81,6 +81,15 @@
                     "separator": true
                 },
                 {
+                    "id": "EXPORT_UTILITY",
+                    "shortcut": {
+                        "keyChar": true,
+                        "modifiers": {
+                            "command": true
```
iwehrman: `control`. Also, the menu infrastructure should really throw an error here!

iwehrman: Would you let me know if you have tested this on Windows? And, if so and if it does not throw an error, let me know I'll file a bug.

iwehrman:
Done with a first review pass. Looks stupendous. We'll get this merged post haste.

src/js/actions/export.js
mcilroyc: This prompted to find a more permanent solution to long-running annoyance.  DocBlockr pref:
`"jsdocs_lower_case_primitives": true`

src/js/actions/export.js
mcilroyc: "more structured" seems dubious therefore it does not, at present, suit my fancy.  Mentally noted.

src/js/actions/export.js
mcilroyc: In this case I am relying on the synchronicity of the dispatch.  I actually originally had `dispatchAsync(...).then()` ... but I felt it read better this way.  I've added a comment here, and added a further explanation in `_syncLayerExportMetadata` which I had meant to do anyway.  Let's discuss further if you're still not buying it :money_with_wings:

src/js/actions/export.js
mcilroyc: Seems like the intermediate `layers.byID(layerID)` and subsequent exportEnabled test makes it more efficient and clear as written, but I'm open to schoolin'

src/js/actions/layers.js
mcilroyc: captured as high pri perf bug 1988

src/js/jsx/sections/export/ExportAllPanel.jsx
mcilroyc: `<LayerExportsItem>` :grey_question:

src/js/jsx/sections/export/ExportAllPanel.jsx
mcilroyc: Originally I had a method to generate a composite key (`ExportAsset.getId()`) which used scale+suffix+format to build a persistent key... but that ended up seeming wrong and I wasn't using it.  I don't know, I always ended up back at the concept that these things are, in fact, defined by their position in this list of assets.  I'm not sure the existing implementation's performance would be improved (in term of React rendering) since the index order is stable (currently).  Can you convince me further?

src/js/jsx/sections/export/ExportAllPanel.jsx
mcilroyc: This currently lives outside the panel architecture.  It is like the <Help> component.  Attached at the top level.  I then watches Document and Export state via StoreWatchMixin.  In theory, it should not be possible to open the panel via the menu/shortcut without a document open... but it is not as strictly enforced (via the aforementioned panel arch) as you had maybe thought when you wrote this comment.

src/js/actions/export.js
iwehrman: Relying? In some cases performing a sync dispatch not in a then handler can cause an exception. The sync version definitely reads better, but in this case I think that's secondary. Hopefully post M3 we can untangle all this sync/async dispatch stuff..

src/js/actions/export.js
iwehrman: My general opinion:
for/while loops < .forEach < .reduce < .map/.filter/.flatMp, etc.
But, I'm not going to hold up progress for that.

src/js/jsx/sections/export/ExportAllPanel.jsx
mcilroyc: Currently, if a layer has assets, but then the layer itself is deleted and the associated assets are not purged from the export store.  And, as of right now, the export store is not listening for layer deletes ... so yes I guess this is a hack around bad programming!

src/js/jsx/sections/export/ExportPanel.jsx
mcilroyc: That thought occurred to me as I was blindly copying this stuff from another Panel component.  What do you think about me creating an issue, and tagging eric in it for his input?

src/js/jsx/sections/export/LayerExports.jsx
mcilroyc: My plan was to ONLY allow the user to choose from a finite set of scale options, and I wasn't even aware that it had been discussed that we may allow arbitrary numbers.  I'll ping seth.

src/js/jsx/sections/export/LayerExports.jsx
mcilroyc: changed to use mathUtil.  IIRC, the reason this function is receiving a string value is because that seemed to be the way datalist works ... or maybe because I don't know how to use it correctly ... at any rate I think this is fairly safe.  If `null` is parsed, then the action will resolve it to 1

iwehrman: Yep. But, also, see the other command about possibly switching away from `Datalist` to `NumberInput` in the short term.

src/js/jsx/sections/export/ExportPanel.jsx
iwehrman: #1995.

src/js/jsx/sections/export/LayerExports.jsx
mcilroyc: Confirmed with seth it should be generic, so I'm changing the name and both references (export and library)

src/js/jsx/sections/export/ExportAllPanel.jsx
iwehrman: Well, we should stop even bothering with this discussion since the performance issue is so minimal, but: if you deleted the 0th asset, then I believe React would then do a DOM-level update to assets 1...N instead of just removing the 0th one from the DOM. This could be solved by giving each asset a counter-based ID.

src/js/jsx/sections/export/LayerExports.jsx
mcilroyc: agreed, changing.

src/js/locks.js
mcilroyc: I actually am not at all sure it is necessary even to have a generator lock since.  I'd welcome your thoughts on this matter.  In the immediate term, I will apply your suggested changes.

src/static/menu-win.json
iwehrman: Good news: I believe the menu infrastructure should throw an error if you try to load a `command` shortcut on Windows.

src/js/locks.js
iwehrman: Assuming your plugin is written to be able to handle multiple concurrent requests, the lock is not necessary. If you're not sure if your plugin is written to handle multiple concurrent requests, we should chat about that offline. The only potential downside to using the lock is, potentially, somewhat worse performance.

src/js/models/layer.js
mcilroyc: yes, have some

src/js/util/exportservice.js
mcilroyc: What do you think about my removing the timeout completely from here, and simply relying on NodeConnection to timeout?  There's a fine line between belt&suspenders and competing/confusing safeguards.

src/nls/root/shortcuts-win.js
mcilroyc: Seth suggested Cmd/Ctrl+e, and I couldn't find any conflict with existing Ps shortcuts.

src/js/util/exportservice.js
mcilroyc: done.  I also moved `this._spacesDomain = new NodeDomain(GENERATOR_DOMAIN_NAME, getRemotePort)` to the constructor, and now the `init()` method simply returns the promise from the underlying promise from NodeDomain.  Does that seem right to you?

mcilroyc:
I pushed up a commit with *most* of the review comments addressed, as well as rebased to master (which included, notably, the CC libs UI).  I have not yet tackled #1988 but I will before re-assigning it back for further review.

mcilroyc:
@iwehrman I just pushed another commit that addresses 1988 (at least using batch play, for now) so I believe all your comments have been addressed.  I'll assign this back to you for re-review, but I want to *hold off for two things before merging* (in addition to anything else you may find in your second pass)

1. I'm going to squash the commits when it's ready
1. I would like to get the plugin into playground-dev build, which I'll work on now

mcilroyc:
Alright @iwehrman I've rebased again (fixed an ico conflict with seth's recent changes b/c I had renamed libraries-delete to simply delete). and I added a new commit that improves the layer extension data query.  Note there is some monkey business in there to handle docs that contain ONLY a background layer, because fetching a layer by index(0) FAILS in that case (because photoshop).

iwehrman:
thumbs up

PR #1752, opened 2015-07-17
Add priority values to search options so we get a better ordering of options

hamburge:
A simple priority system to better predict options. This can definitely be finessed in the future, but as a first pass, it gives better results.

Now we get:
screen shot 2015-07-17 at 12 40 53 pm
screen shot 2015-07-17 at 12 41 19 pm

Instead of:
screen shot 2015-07-17 at 12 41 40 pm
screen shot 2015-07-17 at 12 41 54 pm

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -40,7 +40,7 @@ define(function (require, exports, module) {
      * @const
      * @type {number} 
     */
-    var MAX_OPTIONS = 10;
+    var MAX_OPTIONS = 5;
 
     /**
      * Strings for labeling search options
@@ -70,8 +70,10 @@ define(function (require, exports, module) {
 
         getInitialState: function () {
             return {
-                // All possible search options
+                // All possible search options as a flat list
                 options: Immutable.List(),
+                // All possible search options, grouped in Immutable lists by search type
+                groupedOptions: Immutable.List(),
                 // Broad categories of what SearchBar has as options
                 searchTypes: [],
                 // List of more specific categories that correlate with searchTypes to be used as filters
@@ -91,7 +93,8 @@ define(function (require, exports, module) {
             this.setState({
                 options: searchStore.getSearchItems(this.props.searchID),
                 searchTypes: searchStore.getHeaders(this.props.searchID),
-                searchCategories: searchStore.getFilters(this.props.searchID)
+                searchCategories: searchStore.getFilters(this.props.searchID),
+                groupedOptions: searchStore.getGroupedSearchItems(this.props.searchID)
             });
         },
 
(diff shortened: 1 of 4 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -308,6 +360,7 @@ define(function (require, exports, module) {
                     if (this.refs.datalist.cursorAtBeginning() && this.state.filter.length > 0) {
                         // Clear filter and icons
                         this._updateFilter(null);
+                        this.refs.datalist.forceUpdate();
```
iwehrman: Bad indentation below in `render`. See other JSX files to see how we indent multi-line component expressions.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -383,12 +391,14 @@ define(function (require, exports, module) {
                 
                 width += 20 * iconCount; // 20 pixels is the computed width + padding of an svg icon
 
+                React.findDOMNode(this.refs.autocomplete).style.left = width + "px";
```
iwehrman: Yikes, this is pretty scary. It doesn't seem like this is something that should exist inside the Datalist component. Ideally the width of the menu would expand to accommodate the widths of all the entries. I'm sure there are lot of challenges with this in CSS though. Can you explain what's going on here?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -341,13 +348,14 @@ define(function (require, exports, module) {
          *
          * @private
          * @param {string} filter
+         * @param {bool} truncate Whether or not to restrict number of options
```
iwehrman: `@param {boolean}`

src/js/jsx/shared/Datalist.jsx
```diff
@@ -202,10 +207,12 @@ define(function (require, exports, module) {
             switch (event.key) {
             case "ArrowUp":
                 select.selectPrev();
+                event.preventDefault();
```
iwehrman: Why is this necessary?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -78,12 +77,18 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
+            if (this.props.live) {
```
iwehrman: Why do we update here even if `this.state.filter === nextState.filter`?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -91,7 +93,8 @@ define(function (require, exports, module) {
             this.setState({
```
iwehrman: Above, it worries me a little bit that this modifies the search store at mount time. Can you explain what that `_update` call accomplishes? Also, it seems like this should be a `StoreWatchMixin` so that it could pick up new search providers registered after this component's first mount time.

iwehrman: http://fluxxor.com/documentation/store-watch-mixin.html

src/js/jsx/search/SearchBar.jsx
```diff
@@ -213,82 +216,131 @@ define(function (require, exports, module) {
         /**
          * Find options to show in the Datalist drop down
          *
-         * @param {Immutable.List.<object>} options Full list of potential options
          * @param {string} searchTerm Term to filter by
+         * @param {string} autofillID ID of option that is currently being suggested
+         * @param {bool} truncate Whether or not to restrict number of options
```
iwehrman: `@param {boolean}`

src/js/jsx/search/SearchBar.jsx
```diff
@@ -213,82 +216,131 @@ define(function (require, exports, module) {
...
+                return Immutable.List();
+            }
 
-            return options && options.filter(function (option) {
-                if (option.hidden) {
-                    return false;
-                }
+            var filteredOptionGroups = optionGroups.map(function (options) {
+                return options && options.filter(function (option) {
+                    option.priority = 1; // Add to priority to indicate less important
```
iwehrman: It looks like you're mutating the shared options from the search store in this function by setting their priority. Is that right? If so, that feels a little risky, and it would be nicer to find a way to do this without mutating shared state.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -213,82 +216,131 @@ define(function (require, exports, module) {
...
+
+                    if (this.state.filter.length > 0) {
+                        // All terms in this.state.filter must be in the option's category
+                        _.forEach(this.state.filter, function (filterValue) {
+                            if (!_.contains(category, filterValue)) {
+                                useTerm = false;
+                            }
+                        });
+
+                        if (!useTerm) {
```
iwehrman: It seems like if you ever encounter the line `useTerm = false` above you just want to end this function immediately. If so, you should use `.some` instead of `.forEach` to exit the loop immediately so that you can return immediately.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -213,82 +216,131 @@ define(function (require, exports, module) {
...
+                    
+                    var searchTerms = searchTerm.split(" "),
+                        numTermsInTitle = 0;
+                    useTerm = false;
+                    // At least one term in the search box must be in the option's title
+                    // or path. Could add check for if term is somewhere in category list too
+                    _.forEach(searchTerms, function (term) {
+                        var titleContains = title.indexOf(term) > -1,
+                            pathContains = searchablePath.indexOf(term) > -1;
+                        if (term !== "" && (titleContains || pathContains)) {
```
iwehrman: Should each term just be trimmed of whitespace? `term = term.trim()`?

iwehrman: Maybe that's redundant w.r.t. `.split(" ")`. Just a question!

iwehrman: Also, you probably want a newline between the var declarations and the if statement above.

iwehrman:
OK, done with a first pass. Back to you @hamburge!

src/js/jsx/shared/Datalist.jsx
hamburge: Without it, the cursor in the text input jumps around while you are arrowing up/down.

src/js/jsx/shared/Datalist.jsx
hamburge: This is adjusting where to place the autocompletion word fragment, not changing the width of the input. It's in Datalist so that it can be up to date on what has been entered so far in the text box and update whenever that changes.

src/js/jsx/search/SearchBar.jsx
hamburge: Yep `split(" ")` should take care of that.

src/js/jsx/shared/Datalist.jsx
hamburge: With live Datalists, we have the value change as you hover over different options. I'll add in a check for `this.state.id !== nextState.id` (Before I added in the autosuggestions, `shouldCompontentUpdate` wasn't in Datalist at all)

src/js/jsx/search/SearchBar.jsx
hamburge: The call to `_update` gets the options for the search module. I added it in here because Recent Documents options weren't being added if I just got all the options when registering the module. It sounds like making it a `StoreWatchMixin` would take care of that so I'll try it out.

hamburge: @iwehrman I made it a StoreWatchMixin, which will be useful for picking up new providers, but there wasn't an apparent way to remove the call to update the store from component did mount. If it is called any earlier, then the options haven't loaded yet in the document/application stores, so we just get empty lists of layers and documents.

The only way around it that I can think of would be to add listeners to actions that indicate when the layers/documents are ready, but those would have to be specific based on the providers, so we'd have to modify the search store whenever we wanted to add a new type of provider.

hamburge:
@iwehrman Made some changes and answered your questions!

iwehrman:
@hamburge Next time, also re-assign the PR to me and fix up the labels as appropriate. It helps me with github triage :)

iwehrman:
Oh, wait, you did. Github just didn't update my page. SORRY.

iwehrman:
Okie doke!

Issue #1694, opened 2015-07-09
Undo re-ordering layers via layer panel results in error and layer panel misrepresentation


chadrolfs:
new default doc, 
select rect tool and draw a shape outside the AB bounds (it will be above the AB in the layers panel), 
in the layers panel drag the Rectangle 1 layer below the AB
cmd/ctrl + z to undo

Result: Rect layer is now still on canvas but not in layers panel and if you click on Layer 1 (the default empty layer), you'll get:
```Action layers.select failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Select” is not currently available.```
and then the layers panel will update.

chadrolfs:
You can also run into 
`Uncaught TypeError: Cannot read property 'id' fo null`
with multiple layers (reorder them, then undo)

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
I no longer get errors but the layers panel still does not update correctly.  related to or the same as #1769 
if same, mark 1769 as dupe and reassign to me.

https://www.dropbox.com/s/w8fi9x6nedjlwew/Undo-LPupdateFail.mov?dl=0

mcilroyc:
Hey @chadrolfs  - it seems to me like *this* bug is fixed.  The remaining visual indented-ness (1769) seems like a separate and UI-related bug.  Even your video linked above seems to correctly show that the behavior described in this bug is fixed (since the repro steps never involved dragging the rect *IN* to the artboard).  Sure, I'm splitting hairs :banana:, but also at the technical level these seem to be distinct from each other.

PR #1755, opened 2015-07-17
Explicitly provide false parameters when transferring to getLayerOrder

mcilroyc:
This addresses #1694.  We were previously passing only one argument when transferring to `getLayerOrder()` after reordering layers.  Because of a fluxor(?) quirk which passes its own second argument when none is supplied during transfer, getLayerOrder was misinterpreting its existence and fouling up the logic.  No history state was being captured, and the rest was...

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1121,8 +1121,9 @@ define(function (require, exports) {
             reorderObj = layerLib.reorder(layerRef, targetRef),
             reorderPromise = descriptor.playObject(reorderObj);
       
-        return reorderPromise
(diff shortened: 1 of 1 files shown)


iwehrman:
The problem is that bluebird is applying extra arguments to the function in the then handler. The correct way to solve this is to apply the then handler to a new anonymous function, and to return the result of a normal call to `this.transfer` from that anonymous function.

mcilroyc:
updated per your suggestion.  squashed commit.  belt and suspenders.

iwehrman:
:hamburger: :fries:

Issue #1754, opened 2015-07-17
Color picker field broken; won't accept new entries, tab won't apply, have to hit return 2x 


chadrolfs:

volfied:
@hamburge This one's for you, at least the "tab won't apply" part of the issue was caused by this commit. Please prioritize this higher than the search feature.

```
00e2c366af4d901fa1c15b8ac54dd92540fb6e42 is the first bad commit
commit 00e2c366af4d901fa1c15b8ac54dd92540fb6e42
Author: Allison Hamburger <hamburge@adobe.com>
Date:   Mon Jun 15 14:52:57 2015 -0700

    Click on search dialog option to select it
```

iwehrman:
FBNC to @chadrolfs.

PR #1757, opened 2015-07-17
Commit color picker values on tab/enter fix

hamburge:


iwehrman:
Does this have any other ramifications? I think you removed that if statement originally? Or am I confused?

iwehrman:
Also, is this for #1754?

chadrolfs:
first glance: fix looks good.
one thing that should also be fixed is after you type into the field and hit return it should apply AND select-all in the field (see other field behavior to compare).

chadrolfs:
@iwehrman yes, this is a fix for that.

hamburge:
Yes, #1754.

Originally I took out a call to _commit, which was bad. Then to fix that I undid it and added in `this.props.onChange(event, nextValue);` after `event.stopPropagation()` That broke layer renaming, so I added in the condition (53c1b). I don't really know why it works now (and doesn't break layer renaming)-- did someone change how that works?

hamburge:
@chadrolfs @iwehrman Fixed the highlighting as well!

chadrolfs:
yay!  :facepunch:

hamburge:
@chadrolfs @iwehrman Gah wait now I remember why that condition is in there. Without it, clicking on options in layer search only sometimes works.

chadrolfs:
d'oh!

hamburge:
OK I think this should be better. I removed the condition on if a TextInput is live in `_commit`, which was added as a fix for a bug in search a while back (#1563). That should fix the color picker. Then I dealt with the search bug for real in Datalist instead.

iwehrman:
Man you dealt with that bug _for real_. :bug: :gun:

Issue #1739, opened 2015-07-15
Clobbering Shadow color fails with mixed selection


chadrolfs:
create Shape1 with a default Drop and Inner Shadow
create Shape2 with no shadows
select them both
click on the color picker for one of the shadows (mixed) and set a color

Result:
`Action layerEffects.setShadowColor failed: TypeError: Cannot read property '1' of undefined`
`Uncaught TypeError: Cannot read property 'update' of undefined`
`Uncaught TypeError: Cannot read property '1' of undefined`
`Action shortcuts.removeShortcut failed: Error: Shortcut does not exist: colorpicker-shadow-dropShadow-027`

chadrolfs:
Here is a visual of my steps: https://www.dropbox.com/s/unx3kcaxoujdrcn/LS-Clobber.mov?dl=0

chadrolfs:
I initially wasn't able to repro on Win but looks like a timing issue and possibly the color picker being partially covered by the layers panel.

`Action dialog.openDialog failed: TypeError: Cannot read property 'getBoundingClientRect' of null`

chadrolfs:
when you do successfully open the color picker without error, clicking to set the shadow as red will have the target jump to black leaving both shapes with a shadow (one red and one black).

chadrolfs:
![screen shot 2015-07-16 at 2 56 46 pm](https://cloud.githubusercontent.com/assets/8442631/8736119/b050fc98-2bca-11e5-9706-4bf56df5f475.png)
![screen shot 2015-07-16 at 2 57 20 pm](https://cloud.githubusercontent.com/assets/8442631/8736120/b05196d0-2bca-11e5-91ce-7538ae789654.png)

shaoshing:
Thank you @chadrolfs

shaoshing:
Hey @chadrolfs , this kind of operations is disabled in PR #1759 so the issue no long exits. :)

chadrolfs:
I will be happy to close this once that PR gets merged.  :grinning:

shaoshing:
PR #1759 is merged, please try again @chadrolfs

PR #1759, opened 2015-07-18
Disable adding / editing shadow effects when selected layers have different shadow counts

shaoshing:
This PR is the result of our discussion on adding / editing effects for multiple layers. The conclusion is that adding / editing only works when selected layers have same shadow counts (see Seth's comments below for the cases to consider if want to support any mixed effects). When the counts don't match, the panel displays "Mixed Appearances". Example:

screen shot 2015-07-17 at 5 11 20 pm

Pasting Seth's comments for reference

```
Small scoping this would say if effects do not match, then we do to not allow editing of effects for that selection. That’s kind of annoying, but it is the smallest, simplest thing.

If we want to allow that kind of editing, here’s the  cases I’m thinking about:

1) user selects two layers for the purpose of adding effects to both at the same time.

2) user selects layers with mismatched effects, and changes an effect that is  shared.

3) user selects layers with mismatched effects and changes an effect that is NOT shared.

4) users selects layers with the same effects, but the effects are in different order on each layer. The user then changes an effect value.

5) user selects layers with different effects applied. Adds a new effect to both, and changes values

* I would think for the small scoping scenario: If the effect types match but not the effect's values —it would still work. For instance layers that have a shadow that have different values, but are the same effect order, would be allowed
```

### DIFF
--- a/src/js/jsx/sections/style/Shadow.jsx
+++ b/src/js/jsx/sections/style/Shadow.jsx
@@ -445,10 +445,20 @@ define(function (require, exports) {
             if (layers.isEmpty()) {
                 return null;
             }
-
-            // Group into arrays of dropShadows, by position in each layer
-            var dropShadowGroups = collection.zip(collection.pluck(layers, "dropShadows")),
-                dropShadowList = dropShadowGroups.map(function (dropShadows, index) {
+                
+            var shadowCountsUnmatch = !layers.every(function (l) {
+                    return l.dropShadows.size === layers.first().dropShadows.size;
+                }),
+                reachMaximumShadows = false,
+                shadowsContent;
+                
+            if (!shadowCountsUnmatch) {
+                // Group into arrays of dropShadows, by position in each layer
+                var shadowGroups = collection.zip(collection.pluck(layers, "dropShadows"));
+                
+                reachMaximumShadows = shadowGroups.size >= this.props.max;
+                
+                shadowsContent = shadowGroups.map(function (dropShadows, index) {
                     return (
                         <Shadow {...this.props}
                             layers={layers}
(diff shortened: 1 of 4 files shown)


src/js/jsx/sections/style/Shadow.jsx
```diff
@@ -459,6 +468,13 @@ define(function (require, exports) {
                             type="dropShadow" />
                     );
                 }, this).toList();
+            } else {
+                shadowsContent = (
+                    <div className="shadow-list__list-container__mixed">
+                        <i>Mixed Appearances</i>
```
volfied: This should be localized.

src/js/jsx/sections/style/Shadow.jsx
```diff
@@ -445,10 +445,19 @@ define(function (require, exports) {
             if (layers.isEmpty()) {
                 return null;
             }
-
-            // Group into arrays of dropShadows, by position in each layer
-            var dropShadowGroups = collection.zip(collection.pluck(layers, "dropShadows")),
-                dropShadowList = dropShadowGroups.map(function (dropShadows, index) {
+                
+            var shadowCountsUnmatch = layers.size > 1 &&
+                    Immutable.Set(layers.map(function (l) { return l.dropShadows.size; })).size !== 1,
```
volfied: agh... I mean... there has to be a simpler way for this... or at least can be broken into couple lines to be more readable.

Maybe `layers.some(function (l) { return l.dropShadows.size !== layers.first().dropShadows.size; })`

chadrolfs:
Will this fix #1739?

src/js/jsx/sections/style/Shadow.jsx
shaoshing: your suggestion looks great! BTW: it still needs the `layers.size > 1 ` to exclude single layer selection.

shaoshing:
thank you @volfied , it is ready for the second review.

volfied:
LGTM

shaoshing:
thanks :+1:

Issue #1758, opened 2015-07-17
Expanding groups expands before selecting resulting in UI selection jump


chadrolfs:
https://www.dropbox.com/s/zl4fv10fuyixpmj/expandB4select.mov?dl=0

iwehrman:
It bothers me that expanding results in changing the selection, but that's how PS does it. @designedbyscience @placegraphichere do you think we should adopt the same convention? If so, I guess we'll have to block rendering the new selection until the the expansion is complete.

chadrolfs:
Love the fix.

PR #1761, opened 2015-07-18
Expand/collapse all descendants

iwehrman:
0. Add a `modifier` store that holds the current state of the modifier keys (alt, command, control, shift).
1. Refactor `SuperselectOverlay.jsx` to use the modifier store instead of directly listening to `adapterFlagsChanged` events.
2. Expand/collapse all descendants when alt/option modifier key is held, using the `modifier` store to determine key state.
3. Suppress click on expand/collapse to prevent selection change. I could swear that I was matching standard PS behavior on this previously, but I guess I'm crazy. In any case, not changing layer selection at the same time as these expand/collapse seems pretty preferable.

Addresses #1758.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1402,30 +1402,34 @@ define(function (require, exports) {
      * @param {Document} document
      * @param {Layer|Immutable.Iterable.<Layer>} layers
      * @param {boolean} expand If true, expand the groups. Otherwise, collapse them.
+     * @param {boolean=} descendants Whether to expand all descendants of the given layers.
      * @return {Promise}
      */
-    var setGroupExpansion = function (document, layers, expand) {
+    var setGroupExpansion = function (document, layers, expand, descendants) {
         if (layers instanceof Layer) {
             layers = Immutable.List.of(layers);
         }
 
-        var layerRefs = layers
+        if (descendants) {
+            layers = layers.flatMap(document.layers.descendants, document.layers);
+        }
+
+        layers = layers
             .filter(function (layer) {
                 return layer.kind === layer.layerKinds.GROUP;
             });
 
-        if (layerRefs.isEmpty()) {
+        if (layers.isEmpty()) {
             return Promise.resolve();
         }
 
-        var documentRef = documentLib.referenceBy.id(document.id);
-
-        layerRefs = layerRefs
-            .map(function (layer) {
-                return layerLib.referenceBy.id(layer.id);
-            })
-            .unshift(documentRef)
-            .toArray();
+        var documentRef = documentLib.referenceBy.id(document.id),
+            layerRefs = layers
+                .map(function (layer) {
+                    return layerLib.referenceBy.id(layer.id);
+                })
+                .unshift(documentRef)
+                .toArray();
 
         var expandPlayObject = layerLib.setGroupExpansion(layerRefs, !!expand),
             expansionPromise = descriptor.playObject(expandPlayObject),
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -65,10 +65,11 @@ define(function (require, exports, module) {
 
         // Deeper selection changes
         var document = this.props.document,
+            nextDocument = nextProps.document,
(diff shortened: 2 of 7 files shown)


src/js/actions/layers.js
```diff
@@ -1402,25 +1402,33 @@ define(function (require, exports) {
...
-    var setGroupExpansion = function (document, layers, expand) {
+    var setGroupExpansion = function (document, layers, expand, descendants) {
         if (layers instanceof Layer) {
             layers = Immutable.List.of(layers);
         }
 
-        var layerRefs = layers
+        if (descendants) {
+            layers = layers
+                .flatMap(function (layer) {
```
volfied: can this be simplified with `.flatMap(document.layers.descendants)`

src/js/actions/layers.js
```diff
@@ -1402,25 +1402,33 @@ define(function (require, exports) {
...
 
-        if (layerRefs.isEmpty()) {
+        if (layers.isEmpty()) {
             return Promise.resolve();
         }
 
         var documentRef = documentLib.referenceBy.id(document.id);
 
-        layerRefs = layerRefs
+        var layerRefs = layers
```
volfied: It would be nice to collect these vars together.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -76,15 +76,11 @@ define(function (require, exports, module) {
...
         // component only needs to re-render when going from having a collapsed ancestor
         // (i.e., being hidden) to not having one (i.e., becoming newly visible).
-        var hadCollapsedAncestor = document.layers.hasCollapsedAncestor(this.props.layer);
-        if (!hadCollapsedAncestor) {
-            return false;
-        }
-
-        var nextDocument = nextProps.document,
+        var hadCollapsedAncestor = document.layers.hasCollapsedAncestor(this.props.layer),
+            nextDocument = nextProps.document,
```
volfied: We should move the nextDocument assignment to the `//Deeper selection changes` and use it in the childOfSelection if block as well.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -152,9 +148,14 @@ define(function (require, exports, module) {
             OS.addListener("externalMouseMove", this.mouseMoveHandler);
         },
 
-        componentDidUpdate: function () {
+        componentDidUpdate: function (prevProps, prevState) {
             if (!this.state.modalState) {
-                this._drawDebounced();
+                if (this.state.leafBounds !== prevState.leafBounds) {
+                    // Redraw immediately when the leaf modifier key changes
+                    this.drawOverlay();
```
volfied: :+1:

src/js/stores/modifier.js
```diff
@@ -0,0 +1,86 @@
...
+    var events = require("js/events");
+
+    /**
+     * 
+     * @constructor
+     */
+    var ModifierStore = Fluxxor.createStore({
+        /**
+         * @private
+         * @type {{alt: boolean, command: boolean, control: boolean, shift: boolean}}
```
volfied: Just out of curiosity, what made you go with alt instead of option?

volfied: I guess that's what coming to us from the adapter.

volfied:
Have couple nits, but I like this. Back to you @iwehrman

src/js/actions/layers.js
iwehrman: There are data-flow problems with that idea.

src/js/jsx/sections/layers/LayerFace.jsx
iwehrman: Good idea.

src/js/actions/layers.js
iwehrman: Good idea.

src/js/actions/layers.js
iwehrman: Oh, nm, there aren't. I'll fix this too.

iwehrman:
Suggestions implemented, branch rebased, ready for re-review.

volfied:
LGTM

PR #1762, opened 2015-07-18
Enable dropping at the bottom of the index directly below a group

iwehrman:
drop

1. Adds an invisible dummy layer face at the bottom of the layer index (when the last top-level layer is a group), which exists only as an index-0 drop target.
2. Removes the "batch" draganddrop registration infrastructure. It was no longer useful in a world in which not all document layers are mounted upfront. When the registration calls went through the action system it made sense to batch them, but now it doesn't make any real difference.

Addresses #1695.


volfied:
:+1:

Issue #1668, opened 2015-06-29
Tooltips should be hidden during drag


iwehrman:
They're an annoying distraction currently:
![tooltip](https://s3.amazonaws.com/f.cl.ly/items/2h1T2N0O3O3y0k3b3E0y/Screen%20Recording%202015-06-29%20at%2010.37%20AM.gif)

iwehrman:
Fixed by #1763.

PR #1763, opened 2015-07-18
Reset cursor after drag and disable tooltips during drag

iwehrman:
For some reason, the "grabbing" cursor does not automatically reset to the "default" cursor after a layer drag until the next mousemove event. This works correctly in modern web browsers, so perhaps this is a consequence of the details of the CEF integration provided by the Spaces plugin. The patch below hacks around the issue using the `os.resetCursor` API.

Ideally this would be part of the dragging infrastructure, but the comment below explains why that might not be sufficient. We can decide how to cross that bridge once @shaoshing and @volfied get to it with the drag-and-drop functionality of the CC Libraries panel.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -26,15 +26,46 @@ define(function (require, exports) {
 
     var Promise = require("bluebird");
 
-    var descriptor = require("adapter/ps/descriptor"),
+    var adapter = require("adapter"),
+        descriptor = require("adapter/ps/descriptor"),
         documentLib = require("adapter/lib/document"),
-        shortcuts = require("./shortcuts"),
-        system = require("js/util/system"),
-        adapterUI = require("adapter/ps/ui");
+        adapterUI = require("adapter/ps/ui"),
+        adapterOS = require("adapter/os");
 
     var events = require("js/events"),
         locks = require("js/locks"),
-        synchronization = require("js/util/synchronization");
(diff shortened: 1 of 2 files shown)


iwehrman:
I've also pushed c48db51, which addresses #1668 by disabling tooltips during layer the layer drag operation. It does so with new `ui.enableTooltips` and `disableTooltip` actions. The same comments above apply about moving this to

src/js/actions/ui.js
```diff
@@ -49,6 +80,28 @@ define(function (require, exports) {
...
+     * Globally disable tooltips and clear any current tooltip.
+     *
+     * @return {Promise}
+     */
+    var disableTooltips = function () {
+        return adapter.setPropertyValue(TOOLTIP_TIME_KEY, DISABLED_TOOLTIP_TIME).then(function () {
+            adapterOS.setTooltip("");
+        });
+    };
+    enableTooltips.writes = [locks.PS_APP];
```
volfied: C/P is bad for your health :)

volfied:
One bug, then I'll merge.

src/js/actions/ui.js
iwehrman: I wish there were a better way to do this so that it would be harder to make this mistake.

iwehrman: I'm only human, Barkin.

iwehrman:
Fixed and ready for re-review.

Issue #1760, opened 2015-07-18
Sibling hover overlay is incorrect before mousemove


iwehrman:
Here's a demonstration: http://cloud.wehrman.org/3K3G2K2v0w2s

To repro:

1. Create a group with two child shape layers, which mostly overlap. Select the group.
2. Double-click on top of the region of intersection of the two child layers. The top layer is correctly selected, but the overlay is for the bottom layer. This corrects itself on mousemove.

May be related to #1734.

volfied:
It's not related to #1734, but it's a bug nonetheless

iwehrman:
Fixed by #1765.

PR #1765, opened 2015-07-20
Get rid of topLayer highlighting during draw

volfied:
And instead let the actual function handle it.

This was a relic of D3 highlighting when we got the mouse events... but now that we have a complete control over highlighting function, we let it handle the highlight on draw.

Addresses #1760 @iwehrman since you opened this bug, can you review?


iwehrman:
It works and it removes a bunch of code, so let's not over-think this.

Issue #1764, opened 2015-07-20
Artboards can be dropped into groups, which causes play error


volfied:

volfied:
!!

PR #1770, opened 2015-07-20
Disallow dragging artboards onto groups

iwehrman:
Also disallow dragging an empty list of layers, which happens if you try to drag any selection that involves a background layer. Also delete some dead code.

Addresses #1764.

### DIFF
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -36,7 +36,6 @@ define(function (require, exports, module) {
     var TitleHeader = require("jsx!js/jsx/shared/TitleHeader"),
         LayerFace = require("jsx!./LayerFace"),
         DummyLayerFace = require("jsx!./DummyLayerFace"),
-        math = require("js/util/math"),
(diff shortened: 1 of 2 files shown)


iwehrman:
Please test! I changed something that seems dead simple, which means there is perhaps a high probability that I have missed something obvious.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -276,7 +258,7 @@ define(function (require, exports, module) {
                         layerTreeDepth = doc.layers.maxDescendantDepth(layer) - layerDepth,
                         extraDepth = dropPosition !== "above" ? 0 : 1,
                         nestDepth = layerTreeDepth + targetDepth + extraDepth,
-                        maxDepth = draggingArtboard ? 1 : PS_MAX_NEST_DEPTH;
+                        maxDepth = draggingArtboard ? 0 : PS_MAX_NEST_DEPTH;
```
iwehrman: This is the fix for dragging artboards onto groups.

volfied:
D'oh!

Issue #1627, opened 2015-06-20
Color picker drag jitter


iwehrman:
I strongly suspect that this started with @designedbyscience's recent drag-and-drop changes:
http://cloud.wehrman.org/3G1y3O442m1T

designedbyscience:
I think this is actually do to historyStates (as I didn't touch that code, and it seems to jump as history state things appear in the console...)

@mcilroyc Could you check if that seems accurate?

iwehrman:
Transferring @mcilroyc to this salt mine.

mcilroyc:
![joe](https://cloud.githubusercontent.com/assets/1168110/8298543/961dfdce-192f-11e5-8d41-f612c802edb0.jpg)

mcilroyc:
History *does* seem to be the culprit.  I turned history tracking off and the jitter goes back to the old mostly-bearable levels.  I'll look further to see what our options are...

iwehrman:
By way of #1773, good enough for me.

PR #1773, opened 2015-07-21
Reduce ColorPicker Drag Jitter

mcilroyc:
This addresses the color picker drag jitter noted in #1627.

This removes the ColorPicker's CWRP which was forcing state.color to be calculated on new props, and instead fixes the way the ColorInput component forces the picker to update by listening to a specific timetravel event from history store.

request review by señor @iwehrman

### DIFF
--- a/src/js/jsx/shared/ColorInput.jsx
+++ b/src/js/jsx/shared/ColorInput.jsx
@@ -209,7 +209,7 @@ define(function (require, exports, module) {
                 swatchClassProps = {
                     "color-input": true
                 };
-            
+
             // Normalize to arrays
             var defaultValue = this.props.defaultValue,
                 valueArray = !Immutable.Iterable.isIterable(defaultValue) ?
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/ColorInput.jsx
```diff
@@ -330,16 +330,17 @@ define(function (require, exports, module) {
...
-            // HACK: force update the color picker on undo/redo. We explicitly
-            // listen for changes here instead of with the StoreWatchMixin because
-            // there is no relevant history state and we don't want to re-render.
+            // Force color picker to update upon undo/redo.
+            // The picker otherwise does not necessarily re-render upon receiving new props.
+            // Furthermore, we explicitly listen for changes here instead using the StoreWatchMixin because
+            // we care only about this narrow type of history change (the 'current' pointer being moved)
             this.getFlux().store("history")
-                .on("change", this._handleHistoryStateChange);
+                .on("timetravel", this._handleHistoryStateChange);
```
mcilroyc: note that we also do this for the rotate component, but that was already correctly listening to `timetravel`

mcilroyc:
see also: https://github.com/adobe-photoshop/spaces-design/commit/10479064c798ede99b1daa2dc361e15fc45f26c6

iwehrman:
It's way better, though I still see some jitter, which I don't recall previously. Still, the change does not add complexity and results in a major improvement, so OK!

PR #1774, opened 2015-07-21
Load the font list lazily

iwehrman:
To improve startup performance, and to more evenly distribute load on the UI thread in some cases, this moves the loading of the font list from startup time until it's first necessary. Luckily, loading the font list doesn't seem to be quite as slow as it used to be, but in some cases this could shave at least a few hundred ms off of startup time.

Side note: it's kind of weird that we mount the `Type` component even when there are no type layers. I guess this is a consequence of having all the type-related logic in one component. That makes sense, but it also kind of seems like the `Type` component should only be mounted when there are selected `Type` layers. Maybe I'm over-thinking this.

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -366,13 +366,13 @@ define(function (require, exports) {
      * @private
      * @return {Promise}
      */
-    var afterStartup = function () {
+    var initFontList = function () {
         return descriptor.getProperty("application", "fontList")
             .bind(this)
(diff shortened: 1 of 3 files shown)


designedbyscience:
Would moving mounting of the Type component to only occur on the selection of Type layers slow down the selection of type layers?

iwehrman:
Probably.

designedbyscience:
On my machine, seems to give a single-digit percentage decrease in startup time.

A+++, would merge again

PR #1778, opened 2015-07-21
Use lodash sort for getMasterPolicyList

volfied:
I'm just gonna leave this here:

```
var scores = [1, 10, 2, 21]; 
scores.sort(); // [1, 10, 2, 21]
// Watch out that 10 comes before 2,
// because '10' comes before '2' in Unicode code point order.
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Helps solve #1650

### DIFF
--- a/src/js/models/eventpolicyset.js
+++ b/src/js/models/eventpolicyset.js
@@ -24,7 +24,8 @@
 define(function (require, exports, module) {
     "use strict";
 
-    var math = require("js/util/math");
+    var math = require("js/util/math"),
+        _ = require("lodash");
 
     /**
      * Manages a set of event policy lists.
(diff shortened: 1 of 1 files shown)


src/js/models/eventpolicyset.js
```diff
@@ -85,9 +86,10 @@ define(function (require, exports, module) {
      * @return {Array.<EventPolicy>}
      */
     EventPolicySet.prototype.getMasterPolicyList = function () {
-        return Object.keys(this._policyLists)
+        return _.chain(this._policyLists)
+            .keys()
             .map(math.parseNumber)
-            .sort()
+            .sortBy()
```
iwehrman: In theory I think the map could be included into the sortBy call.

iwehrman:
I also can't really find any promise in the lodash docs that numbers are sorted in ascending order, which is disturbing. But, it seems to work. If you feel like killing that extra map step, go for it. If not, I'll merge it.

volfied:
@iwehrman It is in the dev docs: https://lodash.com/docs#sortBy

Made the change, thank you for feedback.

Issue #1771, opened 2015-07-21
Dummy layer face should cover the rest of layer panel


volfied:
Sometimes my mouse moves fast enough that Dummy Layer Face never registers as a droppable during layer drags, so I have to move my drag back to the area to see the blue line. 

Feels like this could be fixed by making dummy face expand to the rest of the layer panel

iwehrman:
I'm kicking that one to @designedbyscience because it's all CSS. I'm not sure how to make the dummy face grow to fit the remaining panel size. If instead I make it a larger fixed size, it causes the panel to scroll vertically when it shouldn't.

pineapplespatula:
@chadrolfs  is this still happening?

chadrolfs:
Yep.
![layer-reorder-noop](https://cloud.githubusercontent.com/assets/8442631/12158273/67f21012-b48d-11e5-9ea8-8931383326ee.gif)

PR #1779, opened 2015-07-21
Minor layer drag tweaks

iwehrman:
Slightly expands the dummy layer area, and also makes the drop separator a little more responsive when dragging around invalid drop areas.

Addresses #1695, and to a lesser extent #1771.


volfied:
I'll allow it, but I don't think 1771 is still fixed by this.

Issue #1742, opened 2015-07-16
Layer Search dialog shouldn't dismiss by clicking in it's frame


chadrolfs:
open a file, cmd/ctrl + f, click in area around the input field

Result: layer search dialog is dismissed.

Issue #1720, opened 2015-07-10
Layer Search: selecting text in field can get jumbled


chadrolfs:
![screen shot 2015-07-10 at 2 14 34 pm](https://cloud.githubusercontent.com/assets/8442631/8629161/cb01eca8-270d-11e5-9443-ad82680fb63c.png)

chadrolfs:
check with @iwehrman or @volfied 
we had this problem before in other UI fields and I think one of them fixed it.

iwehrman:
Or @designedbyscience. I seem to recall that it had something to do with setting a background color on the text input versus it being transparent.

PR #1784, opened 2015-07-21
Search bugs: don't close dialog on click; fix for text jumbling when highlighted

hamburge:
for issues #1742 and #1720 

For #1742, I replaced dismissing the dialog with `event.stopPropagation()` because otherwise the item that was last selected (which doesn't necessarily match what you've inputted) will be set as the input value when you click on the dialog.


volfied:
:1234:

Issue #1711, opened 2015-07-09
Allow post condition validation to be turned off temporarily while in debug mode


mcilroyc:
Post condition validation is great.  It catches bugs.  But in some cases when tracking down performance issues, it would be nice to be able to selectively turn off PCV without ditching debug mode entirely.

volfied:
I thought this was a PR and I got so excited...

volfied:
Closing as this is now implemented, and a dev only feature.

PR #1785, opened 2015-07-21
Add debug menu items for postCond and PolicyVis toggling

volfied:
Under help menu, now we have better organization, and toggle items for postcondition verification and policy visualization, to help with debugging.

Addresses #1711, and couple other dev needs.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -293,7 +293,9 @@ define(function (require, exports, module) {
             post = action.post,
             parentActionName = actionReceiver.actionName,
             actionName = action.id,
-            actionTitle = "action " + parentActionName;
+            actionTitle = "action " + parentActionName,
+            preferences = this.flux.store("preferences").getState(),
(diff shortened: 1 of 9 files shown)


src/js/fluxcontroller.js
```diff
@@ -293,7 +293,9 @@ define(function (require, exports, module) {
             post = action.post,
             parentActionName = actionReceiver.actionName,
             actionName = action.id,
-            actionTitle = "action " + parentActionName;
+            actionTitle = "action " + parentActionName,
+            preferences = this.flux.store("preferences").getState(),
+            checkPost = preferences.get("postConditionsEnabled");
```
iwehrman: Is it too much to complain about having to do this lookup every for every action invocation even when not in debug mode?

iwehrman: On reflection, yes, that is too much to complain about.

iwehrman:
Looks great, thanks!

Issue #1683, opened 2015-07-07
After moving Artboard, old location for add UI shows before it's updated. 


chadrolfs:
https://www.dropbox.com/s/ym96xwttf8d0sop/redraw-oldloc.mov?dl=0

iwehrman:
FBNC to @chadrolfs.

PR #1788, opened 2015-07-21
Simplify overlay code, fix redraw issues with move/transform

volfied:
Couple changes:

 - Remove couple enable overlay calls that caused wrong re-draws
 - Change "editArtboardEvent" handler, so we reset bounds first, and then get layer order
 - During layer moves, if the hysteresis wasn't broken (0,0 move), we re-enable the overlay
 - Emit change events from tool store when tool modal state changes
 - Listen to tool store changes in SueprselectOverlay, so we can redraw overlay when modal state changes - this should be safe, as Toolbar, Scrim and SuperselectOverlay are the only listeners of tool store
 - If modal state is true, immediately clear the overlay, otherwise, debounce draw
 - Clean up the code to always clean things, but not draw anything in modal states

Addresses #1683, and probably also couple more transform/overlay bugs.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1080,28 +1080,27 @@ define(function (require, exports) {
 
     var beforeStartup = function () {
         _artboardTransformHandler = synchronization.debounce(function () {
-            this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
-            
             // After each action we call, document model changes
             // so we re-get it
             var appStore = this.flux.store("application"),
                 nextDoc = appStore.getCurrentDocument();
 
-            return this.flux.actions.layers.getLayerOrder(nextDoc, true)
+            return this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected)
                 .bind(this)
                 .then(function () {
                     nextDoc = appStore.getCurrentDocument();
                     this.flux.actions.layers.resetSelection(nextDoc);
                 }).then(function () {
                     nextDoc = appStore.getCurrentDocument();
-                    this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected);
+                    this.flux.actions.layers.getLayerOrder(nextDoc, true);
                 });
         }, this);
 
         _layerTransformHandler = synchronization.debounce(function (event) {
-            // If it was a simple click/didn't move anything, there is no need to update bounds
+            // If it was a simple click/didn't move anything, there is no need to update bounds,
+            // just redraw the overlay
             if (event.trackerEndedWithoutBreakingHysteresis) {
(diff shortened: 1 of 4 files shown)


src/js/actions/transform.js
```diff
@@ -1080,28 +1080,27 @@ define(function (require, exports) {
...
         _artboardTransformHandler = synchronization.debounce(function () {
-            this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
-            
             // After each action we call, document model changes
             // so we re-get it
             var appStore = this.flux.store("application"),
                 nextDoc = appStore.getCurrentDocument();
 
-            return this.flux.actions.layers.getLayerOrder(nextDoc, true)
+            return this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected)
```
iwehrman: I'm surprised we want to call `resetBounds` before `resetSelection`. If we're calling the latter, that must mean we may have lost track of the selection previously. But that previous set is the set whose bounds we're resetting. Shouldn't it be the other way around?

src/js/actions/transform.js
```diff
@@ -1080,28 +1080,27 @@ define(function (require, exports) {
...
                 });
         }, this);
 
         _layerTransformHandler = synchronization.debounce(function (event) {
-            // If it was a simple click/didn't move anything, there is no need to update bounds
+            // If it was a simple click/didn't move anything, there is no need to update bounds,
+            // just redraw the overlay
             if (event.trackerEndedWithoutBreakingHysteresis) {
-                return Promise.resolve();
+                return this.dispatchAsync(events.ui.TOGGLE_OVERLAYS, { enabled: true });
```
iwehrman: This is harmless, but handlers are allowed to dispatch synchronously because there is no chance that the calling function is on the call stack of a different dispatch handler.

Aside: I realize the distinction between `dispatch` and `dispatchAsync` is incredibly subtle. We should probably revisit offline the motivation for the latter helper function to see if it's still necessary or if there is a better solution.

iwehrman:
Clean-up looks good. Just one possibly tangential question. Back to you, @volfied.

src/js/actions/transform.js
volfied: Argh, good catch!

volfied: Actually, after testing a bit, I realized that this is correct. We want to resetBounds on the selected layer that was just transformed, THEN reset our selection to make sure any side effects like selection and order change is accounted for.

volfied:
@iwehrman Back to you, no changes needed.

src/js/actions/transform.js
iwehrman: Excellent. Thanks for confirming.

Issue #1690, opened 2015-07-08
Auto-nested artboards move on canvas and are not selectable or draggable


iwehrman:
Video: http://cloud.wehrman.org/0D1M0y3z3A1l

When a layer that is outside of an artboard is dragged into the artboard from the layers panel, the layer is automatically repositioned on a canvas. Unfortunately Design Space doesn't update its bounds as a result, and so the layer can't be selected, dragged, etc.

volfied:
Ideally: Reorder action descriptor response should indicate whether a layer's position is changed or not on reorder.

Realistically, we'll need to update bounds of the reordered layer IFF it's being moved into an artboard

joelrbrandt:
Possible dupe of #1614

iwehrman:
Fixed by #1790 and #1792.

PR #1790, opened 2015-07-21
 fix for 1614

pineapplespatula:
This is a bit of a heavy hammer, We only need to do this for arboard changes, but i'm not sure if there is a simple test to verify that we are making a layer drag between two different artboards 

fix for issue #1614 and #1690

### DIFF
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -429,6 +429,9 @@ define(function (require, exports, module) {
                 var dragSource = collection.pluck(this.state.dragTargets, "id");
 
                 flux.actions.layers.reorder(doc, dragSource, dropIndex)
+                    .then(function () {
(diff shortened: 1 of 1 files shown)


iwehrman:
Works great and the fix seems perfectly reasonable. Bravo!

iwehrman:
Actually, this doesn't completely fix #1690, but I know what's missing. I'll submit a follow-up PR.

pineapplespatula:
does it solve the issue chad is seeing ? https://www.dropbox.com/s/ur4nfhgoi4bqno1/1614repro.mov?dl=0

chadrolfs:
Yeah, doesn't totally fix it: ^^
select rect, drag out 2 shapes outside AB, select Select tool, drag shape layers into AB via layers panel

chadrolfs:
just checked #1792, looks like that did the trick!  nice teamwork!

iwehrman:
high five

Issue #1789, opened 2015-07-21
Layer Search: filtering is broken


chadrolfs:
open a doc, cmd/ctrl + f, click on input field for drop down, click or arrow+tab to a filter (layer or vector)

Result: no filter is set

PR #1791, opened 2015-07-21
Fix search filtering

hamburge:
issue #1789

### DIFF
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -410,8 +410,13 @@ define(function (require, exports, module) {
                     }) : null;
                 }
 
-                var suggestionID = suggestion ? suggestion.id : null,
+                var suggestionID = suggestion ? suggestion.id : this.state.id,
                     suggestionTitle = suggestion ? suggestion.title : "";
+
+                // If all the options are headers (no confirmable options, then set selected ID to null
+                if (!suggestion && collection.uniformValue(collection.pluck(options, "type"))) {
+                    suggestionID = null;
+                }
                
                 this.setState({
                     filter: value,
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/Datalist.jsx
```diff
@@ -410,8 +410,13 @@ define(function (require, exports, module) {
                     }) : null;
                 }
 
-                var suggestionID = suggestion ? suggestion.id : null,
+                var suggestionID = suggestion ? suggestion.id : this.state.id,
                     suggestionTitle = suggestion ? suggestion.title : "";
+
+                // If all the options are headers (no confirmable options, then set selected ID to null
+                if (!suggestion && collection.uniformValue(collection.pluck(options, "type"))) {
+                    suggestionID = null;
```
volfied: `this.state.id` instead of `null`?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -410,8 +410,13 @@ define(function (require, exports, module) {
                     }) : null;
                 }
 
-                var suggestionID = suggestion ? suggestion.id : null,
+                var suggestionID = suggestion ? suggestion.id : this.state.id,
```
volfied: Asking because you made the change here

src/js/jsx/shared/Datalist.jsx
hamburge: Yeah, in this case ID should be `null` instead of staying the same because it means that there are no options in the dropdown. If it doesn't get set to `null`, then when you hit enter when there are no options, it will select the last valid option, which is not good.

My earlier attempt to fix this bug caused the issue that this PR fixes (selecting filters sometimes not working).

volfied:
Alrighty then!

PR #1792, opened 2015-07-21
Reset bounds of selected layers after drag

iwehrman:
This is a follow up to #1790. That PR resets the bounds of the selected layers after being dragged and dropped in the layers panel to account for on-canvas movement that results from auto-nesting. But, if you drag an unselected layers, it will be selected after the index is re-ordered. This patch fetches the most recent document model to ensure that the list of selected layers is current after the reorder operation.

Addresses #1690.


iwehrman:
Credit where it's due: @volfied spotted the issue and suggested the fix.

PR #1793, opened 2015-07-22
Added Tabs for Document Header instead of past switching style

designedbyscience:
Review only as I need: 
a) @placegraphichere to look at it
b) @iwehrman (or someone) to let me know if we have anything for "all documents have loaded" as right now, I need to re-render the header every time a new document is parsed (and the application store changes)

### DIFF
--- a/src/js/jsx/DocumentHeader.jsx
+++ b/src/js/jsx/DocumentHeader.jsx
@@ -28,11 +28,10 @@ define(function (require, exports, module) {
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
+        Immutable = require("immutable"),
         classnames = require("classnames");
-
-    var Button = require("jsx!js/jsx/shared/Button"),
-        Gutter = require("jsx!js/jsx/shared/Gutter"),
-        strings = require("i18n!nls/strings");
+    
+    var DocumentHeaderTab = require("jsx!js/jsx/DocumentHeaderTab");
 
     var DocumentHeader = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("application", "document")],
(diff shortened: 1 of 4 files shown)


iwehrman:
For b) that doesn't necessarily seem like a huge problem. If it is, we could have components specifically for the tabs so that we can prevent them from re-rendering when other documents are added/removed/renamed/etc. How much of a performance impact do you think the current implementation will have?

iwehrman:
It _does_ seem worthwhile to write a `shouldComponentUpdate` method to prevent re-renders when something in the document other than the name/path changes, however. Dragging the slider around the color picker will cause continual virtual re-renders currently.

volfied:
Not to state the obvious, but I can't tell which document I'm editing.

iwehrman:
For the merely curious:
tabs

iwehrman:
I suspect this is as-yet essentially undesigned, but since we're piling on: I think the header is significantly too tall/thick. 

Also, I foresee the need very shortly to drag and drop these tabs.

And, generally, I think Chrome's tab interactions are generally awesome, and we should mimic those to the extent that it's possible.

placegraphichere:
First glance feedback:
- Active document hex e6e6e6, 
- inactive hex 8e8e8e
- background for the document bar is hex 2d2d2d
- Height should be 48px or I believe 6rem. 

I also think we need a better "unsaved" indicator, but that's probably outside of this PR.

designedbyscience:
Updated: 

Moved document tab into own component, added some shouldComponentUpdate goodness. 

Changed colors of things
**Did not** change height of tabs. The height remains the same as the previous document header. This height matches the height of 1 toolbar button, so that the unpinned state looks nice. @placegraphichere could change both of these sizes to match and allow the header to become shorter. Or something else entirely! But for the moment, they need to match.

Also, made my own scrollIntoViewIfNeeded because that function had some trouble with the showing the entirety of the last tab.

iwehrman:
@tiburbage: Would you please test this branch with a) lots of open documents; and b) VermillionArtboards.psd as the _active_ document and report on any performance implications?

Issue #1589, opened 2015-06-16
2nd swap returns incorrect location


chadrolfs:
create two shapes, select them, swap once, swap again.

Result: 2nd swap results in incorrect locations
https://www.dropbox.com/s/rgru01p7wrp39wd/swap-fail.mov?dl=0

chadrolfs:
happens under the following conditions: layers are inside an AB and one extends beyond AB bounds.

volfied:
Ah, this has nothing to do with artboards. Our swap algorithm does one thing for the first situation, and another thing for the second situationa fter they're swapped, due to size differences in the layers.

I'll look into it, but we never tried to make our swap algorithm symmetric.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Swapping ABs is kinda jacked:
https://www.dropbox.com/s/mjwaf9jh00ee1pc/AB-Swap-bug.mov?dl=0

chadrolfs:
Might be a dupe of #1532

volfied:
I got it! We need to turn off auto expand during swap

volfied:
@chadrolfs the core of this issue (back to back swap not being a no-op) was fixed by #1796. The AB swap issue is #1532 so I vote we close this one or hte other.

PR #1796, opened 2015-07-22
Make swap reflective

volfied:
Addresses #1589. When we first designed swap, we figured this might become a problem. Not swapping just top/left corners may be unexpected, but with this change, at least the features becomes reflective so user can swap back and forth to undo.


iwehrman:
I think this makes sense, though I'm not exactly sure why we came to a different conclusion originally...

Issue #1794, opened 2015-07-22
Type: unable to move via transform panel fields


chadrolfs:
create type layer and try to change X, Y
Result: you can't

Also, since we reverted to classic transform controls, should we also allow W and H to change?

volfied:
@iwehrman need your opinion on this, in regards to unlocking W/H for type layers

volfied:
You can assign it to me once a decision is made.

iwehrman:
Whoops, Github auto-closed this. FBNC to @chadrolfs.

iwehrman:
We'll need a card for it, but yes we should unlock the W/H controls. We'll also need to fix the action descriptors for use with the "implied" font sizes.

iwehrman:
https://trello.com/c/o5nH9ksM/527-allow-type-layer-transformations

PR #1798, opened 2015-07-22
Check for selection artboard count

volfied:
When I added the "artboards" property to layerstructure, I mistakenly started calculating the overall artboards count in position/size enablers. Oops.

Fixes https://github.com/adobe-photoshop/spaces-design/issues/1794


iwehrman:
:+1:

Issue #1797, opened 2015-07-22
First instance of creating a Type layer results in Typeface: missing


chadrolfs:
https://www.dropbox.com/s/qt9vlc9vbxewgip/Font-Missing.mov?dl=0

iwehrman:
I suspect that this is fixed by #1803.

volfied:
@chadrolfs since #1803 is merged, can you re-confirm?

chadrolfs:
Weeeeelllll, it still flashes for a split second.  https://www.dropbox.com/s/6a8ryol0uqszb98/Type-missing-flash.mov?dl=0

iwehrman:
That's a different issue, I think.

iwehrman:
It also happens exactly one time.

mcilroyc:
the new fix is in

Issue #1738, opened 2015-07-15
Errors accessing color picker after creating Type or shape


chadrolfs:
cmd/ctlr + n
select type tool
click on canvas and type, hit Enter (numpad) or Esc to commit
click on Type Color picker.
Result: `Action dialog.openDialog failed: TypeError: undefined is not a function`
`Action shortcuts.removeShortcut failed: Error: Unknown policy ID: 14`

2nd attempt succeeds

chadrolfs:
cmd/ctlr + n
drag out a rect shape within AB
click on Color picker for Stroke
BOOM

volfied:
This doesn't go boom anymore!

chadrolfs:
I made it go boom:
cmd/ctlr + n
select type tool
click on canvas and type, hit Enter (numpad) or Esc to commit
click on Type Color picker and set a color
hit enter or click on color picker to close
repeat steps above:
2nd attempt fails

chadrolfs:
![screen shot 2015-07-20 at 2 21 45 pm](https://cloud.githubusercontent.com/assets/8442631/8787883/79f20312-2eeb-11e5-8ccc-3e59cd0f0c37.png)

volfied:
This one is baffling me, but mostly because I don't know much about Dialogs.

@designedbyscience , I'll give this to you. Problem is two fold - if I create a new document, switch to type tool, type something, press esc to commit, and then click the Type color picker square (with the Aa), `setState({target: ...})` gets called, but also the dialog store emits a change event, and causes a re-render where _positionDialog gets an error. 

Other problem is, if target is defined, it still causes the problem above, because `this.state.target.offsetParent` is null in some cases.

designedbyscience:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent: offsetParent returns null when the element has style.display set to "none"

PR #1803, opened 2015-07-23
Dialog bugfix extravaganza

iwehrman:
1. Require at registration time that `Dialog`s all have unique IDs. If we don't require that then actions invocations like `openDialog(id)` and `closeDialog(id)` have unintended, negative consequences.
2. Elaborate all `Dialog` IDs in the app to be unique, mostly by jamming the document ID into the IDs of `ColorInput`s and `Datalist`s.
3. Fix a race in `Dialog` in which the `openDialog` action could execute before an asynchronous setState call was complete. (This could cause the target DOMElement to be null in some cases when computing the position of the dialog.)
4. Fix a bug in a `Dialog` error handler: `displayName` is a `ReactElement` property, not a method. (Thanks @volfied!)
5. Fix an unrelated `Datalist` bug that prevented re-rendering when options are updated (e.g., from an empty font list to a non-empty font list).

Addresses #1738 and I think #1797.

### DIFF
--- a/src/js/jsx/sections/libraries/LibraryList.jsx
+++ b/src/js/jsx/sections/libraries/LibraryList.jsx
@@ -83,9 +83,11 @@ define(function (require, exports, module) {
                 libraryName = this.props.selected.name;
             }
 
+            var listID = "libraries-" + this.props.document.id;
+
             return (
                 <Datalist
-                    list={"libraries"}
+                    list={listID}
                     disabled={false}
                     className="dialog-libraries"
                     options={libraryOptions}
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -49,6 +49,14 @@ define(function (require, exports, module) {
     var Dialog = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("dialog")],
 
+        /**
+         * The target of the event that opened the Dialog.
+         *
+         * @private
+         * @type {?DOMElement}
+         */
+        _target: null,
+
         propTypes: {
             id: React.PropTypes.string.isRequired,
             onOpen: React.PropTypes.func,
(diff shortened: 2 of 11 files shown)


iwehrman:
@mcilroyc @designedbyscience @volfied: please review. I have a low confidence in anything that changes `Dialog`.
@chadrolfs: please test. I have a low confidence in anything that involves dialogs.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -83,9 +83,11 @@ define(function (require, exports, module) {
                 libraryName = this.props.selected.name;
             }
 
+            var listID = "libraries-" + this.props.document.id;
+
             return (
                 <Datalist
-                    list={"libraries"}
+                    list={listID}
```
volfied: Pinging @shaoshing to make sure he's aware of these changes and he rebases when these are merged.

src/js/jsx/shared/Dialog.jsx
```diff
@@ -124,11 +132,10 @@ define(function (require, exports, module) {
                 id = this.props.id;
 
             if (this.state.open) {
+                this._target = null;
                 flux.actions.dialog.closeDialog(id);
             } else if (!this.props.disabled) {
-                this.setState({
-                    target: event.target
-                });
+                this._target = event.target;
```
volfied: Ah, there is the fix.

volfied:
Looks good to me

src/js/jsx/shared/Dialog.jsx
mcilroyc: would it also have been possible to call the `openDialog` action in the callback of setState?

iwehrman: I looked into that, but I don't think that's the intended use of that callback? In any case, I could not convince myself that that was safe. If you can point me to docs that say otherwise, I would be interested.

mcilroyc: This *seems* safe to me, but maybe I'm missing something.  From [the docs](http://facebook.github.io/react/docs/component-api.html):
"The second (optional) parameter is a callback function that will be executed once setState is completed and the component is re-rendered."

iwehrman: The question I have: does that happen before or after the render? Before or after componentDidUpdate? Before or after componentWillUpdate? What if the component doesn't update because of shouldComponentUpdate? So many questions!

mcilroyc: http://gph.is/1bqWRBT

PR #1817, opened 2015-07-23
dont use callback on Promisified function

pineapplespatula:
Since we aren't using a callback on startEditWithCurrentModalTool this needs to be adapted

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -41,8 +41,7 @@ define(function (require, exports, module) {
         TextInput = require("jsx!js/jsx/shared/TextInput"),
         system = require("js/util/system"),
         svgUtil = require("js/util/svg"),
-        strings = require("i18n!nls/strings"),
-        log = require("js/util/log");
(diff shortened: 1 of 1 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -180,9 +179,7 @@ define(function (require, exports, module) {
                         
                     if ((currentTool.id === "typeCreateOrEdit" || currentTool.id === "superselectType") &&
                         layer.kind === layer.layerKinds.TEXT && numSelectedLayers === 1) {
-                        UI.startEditWithCurrentModalTool(function (err) {
-                            log.error("startEditWithCurrentModalTool: " + err);
-                        });
+                        UI.startEditWithCurrentModalTool();
```
volfied: If you want to keep the error message around, you can always say:
```
UI.startEditWithCurrentModalTool()
    .catch(function (err) {
        log.error("startEditWithCurrentModalTool: " + err);
    });
```

pineapplespatula: I dont think it is important to do so

Issue #1804, opened 2015-07-23
Menu command results should use platform symbols for keyboard shortcuts


iwehrman:
![shortcuts](https://s3.amazonaws.com/f.cl.ly/items/2e440N4005063A0j0e2Z/Image%202015-07-23%20at%206.53.11%20AM.png)

This should use ⌘ instead of Cmd, ^ for Ctrl, etc. The correct symbols may be different on Mac and Windows.

iwehrman:
Also, there shouldn't be a + separating the terms because then you get things like this:
![++](https://s3.amazonaws.com/f.cl.ly/items/0P0k09472Q1G1i0L3B1c/Image%202015-07-23%20at%207.09.24%20AM.png)

Instead, just follow the platform menu conventions: 
![--](https://s3.amazonaws.com/f.cl.ly/items/2h1W232Q1R1L0Q221s3k/Image%202015-07-23%20at%207.10.35%20AM.png)

volfied:
@iwehrman please confirm the fix.

iwehrman:
Partially addressed, but a) the shortcuts should be right-justified somehow, and b) they shouldn't be wrapped in parentheses because that makes them harder to parse. Back to @hamburge!

iwehrman:
Actually, I guess I will relent on this and complain elsewhere about the styling. The shortcuts are indeed using the correct symbols. YOU WIN!

PR #1818, opened 2015-07-23
Use modifier symbols instead of strings for menu commands in search

hamburge:
Issue #1804. It seems like standard on Windows is to use the abbreviations "Ctrl," "Alt," etc. and the shift symbol.

This also fixes that the shortcut for flip vertical/horizontal wasn't showing up by adding an option to not shorten the paths of search type items.

### DIFF
--- a/src/js/stores/search.js
+++ b/src/js/stores/search.js
@@ -60,7 +60,7 @@ define(function (require, exports, module) {
      * @property {string} id ID used to perform action if item is selected
      * @property {string} name Displayable title
      * @property {Array.<string>} category Subset of filters that apply to item. All are keys of SEARCH.CATEGORIES
-     * @property {string} pathInfo A path separated by '/', or ""
+     * @property {string} pathInfo Optional path separated by '/'
      * @property {string} iconID Class for corresponding SVG
     */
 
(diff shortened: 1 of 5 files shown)


src/js/stores/search.js
```diff
@@ -227,7 +229,7 @@ define(function (require, exports, module) {
                     shortenedPaths = pathUtil.getShortestUniquePaths(ancestors).toJS();
 
                 var itemMap = items.map(function (item, index) {
-                    var newPathInfo = shortenedPaths[index] || item.pathInfo;
+                    var newPathInfo = searchTypeInfo.shortenPaths ? shortenedPaths[index] : item.pathInfo;
```
volfied: So I see a logic change here, and I have to ask. You were checking to see if shortenedPath existed, and then used `item.pathInfo`. What if you set shortenPaths to true, but shortenedPath does not exist for that item? Can you guarantee that won't happen?

volfied:
One Q, @hamburge

hamburge:
@volfied Made that check stronger-- and now you don't have to specify `pathInfo` for a search item if you don't want to

volfied:
:+1:

PR #1822, opened 2015-07-24
Don't show "mixed" typeface before font list is initialized

iwehrman:
Instead, just display nothing until the font list is ready. This is necessary because the font family is what should be shown, but that information isn't available anywhere except the font list.

Addresses #1797.


mcilroyc:
okie doke

PR #1824, opened 2015-07-24
Make sure to wait for acquireFocus to finish on _beginEdit

volfied:
Addresses #1682 
Problem was, textinput would call _beginEdit, which would call acquireFocus, and we'd have a race condition between acqurieFocus of Focusable and setState of React.

Now we make sure acquireFocus returns the promise, and in _beginEdit we wait for that promise to first set the state, and then use the callback of `React.setState`to actually do DOM editing, which is what React suggests we should do.

@iwehrman Can you review, since I bugged you about this.
@hamburge Please be an observer/reviewer as this is important for your work

### DIFF
--- a/src/js/jsx/shared/TextInput.jsx
+++ b/src/js/jsx/shared/TextInput.jsx
@@ -111,9 +111,16 @@ define(function (require, exports, module) {
         },
 
         componentDidUpdate: function () {
+            var node = React.findDOMNode(this.refs.input);
+
+            if (this.state.editing) {
+                node.removeAttribute("readOnly");
+                node.removeAttribute("disabled");
+                node.focus();
+            }
+
             if (!this.state.selectDisabled) {
                 // If the component updated and there is selection state, restore it
-                var node = React.findDOMNode(this.refs.input);
                 if (window.document.activeElement === node) {
                     node.setSelectionRange(0, node.value.length);
                 }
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/TextInput.jsx
```diff
@@ -272,16 +272,19 @@ define(function (require, exports, module) {
...
-            });
-            this.acquireFocus();
+            this.acquireFocus()
+                .bind(this)
+                .then(function () {
+                    this.setState({
+                        editing: true,
+                        selectDisabled: selectDisabled
+                    }, function () {
+                        var node = React.findDOMNode(this.refs.input);
```
iwehrman: I don't understand what's going on in this failure case. First, it seems like this should never fail, since in `acquireFocus` you're explicitly catching all errors. (I also don't think it makes sense in hindsight to catch errors from `acquireFocus`.) Second, I guess I just don't understand why you're doing what you're doing in the case of a failure.

iwehrman: Er, I meant to say: I also don't think it makes sense in hindsight to catch errors _within_ `acquireFocus`.

iwehrman: Oh, wait, that function is in the callback from `setState`? Are you sure that's how `setState` works??

volfied:
@iwehrman Updated the code based on our chat, this should be cleaner and easier to read.

iwehrman:
Live to fight another day.

PR #1826, opened 2015-07-24
Correctly set the first color of CharacterStyle

volfied:
Before this change, we were defaulting the color and showing the wrong value.

@pineapplespatula, this was your code, so please review.


pineapplespatula:
looks good

PR #1827, opened 2015-07-24
Styles Workflow 1 - Sampler Tool

volfied:
Styles Workflow Card 1 - Sampler tool to modify layer color

- Click on a source layer and grab it's color:
    - Text layer: text color, layer opacity
    - Shape layer: Fill color, fill opacity
    - Other: Pixel color at the value, layer opacity
- And apply it to selected layers:
    - Type layers: Type color and layer opacity
    - Shape layers: Fill color, fill opacity
    - Other: no op

Styles Workflow Card 2 - Shift modifier to sample layer FX

- Shift click on a source layer to copy all styles from that layer and apply the whole set to the selected layers. Existing effects will be deleted.

### DIFF
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -45,13 +45,13 @@ define(function (require, exports) {
      * @param {Document} document
      * @param {Immutable.List.<Layer>} layers
      * @param {boolean=} coalesce Whether to coalesce this operation's history state
-     * @param {string} type layer effect type, eg "dropShadow"
+     * @param {string|Array.<String>} type layer effect type, eg "dropShadow"
      * @return {Promise} returns the promised value from the photoshop adapter call
      */
     var _syncStoreToPs = function (document, layers, coalesce, type) {
         var documentStore = this.flux.store("document"),
             layerStruct = documentStore.getDocument(document.id).layers,
-            layerEffectPlayObjects,
+            layerEffectPlayObjects = [],
             options = {
                 paintOptions: {
                     immediateUpdate: true,
@@ -63,30 +63,38 @@ define(function (require, exports) {
                     coalesce: !!coalesce,
                     suppressHistoryStateNotification: !!coalesce
                 }
-            };
+            },
+            types = _.isArray(type) ? type : [type];
 
         // Map the layers to a list of playable objects
-        layerEffectPlayObjects = layers.map(function (curLayer) {
+        layers.forEach(function (curLayer) {
             // get this layer directly from the store and build a layer effect adapter object
             var layerFromStore = layerStruct.byID(curLayer.id),
-                layerEffectsFromStore = layerFromStore.getLayerEffectsByType(type),
                 referenceID = layerEffectLib.referenceBy.id(curLayer.id),
-                shadowAdapterObject = layerEffectsFromStore
-                    .map(function (shadow) {
-                        return shadow.toAdapterObject();
-                    }).toArray();
+                layerHasEffects = curLayer.usedToHaveLayerEffect;
 
-            if (curLayer.usedToHaveLayerEffect) {
-                return {
-                    layer: curLayer,
-                    playObject: layerEffectLib.setExtendedLayerEffect(type, referenceID, shadowAdapterObject)
-                };
-            } else {
-                return {
-                    layer: curLayer,
-                    playObject: layerEffectLib.setLayerEffect(type, referenceID, shadowAdapterObject)
-                };
-            }
+            types.forEach(function (type) {
+                var layerEffectsFromStore = layerFromStore.getLayerEffectsByType(type),
+                    shadowAdapterObject = layerEffectsFromStore
+                        .map(function (shadow) {
+                            return shadow.toAdapterObject();
+                        }).toArray();
+
+                if (layerHasEffects) {
+                    layerEffectPlayObjects.push({
+                        layer: curLayer,
+                        playObject: layerEffectLib.setExtendedLayerEffect(type, referenceID, shadowAdapterObject)
+                    });
+                } else {
+                    if (shadowAdapterObject.length > 0) {
+                        layerEffectPlayObjects.push({
+                            layer: curLayer,
+                            playObject: layerEffectLib.setLayerEffect(type, referenceID, shadowAdapterObject)
+                        });
+                        layerHasEffects = true;
+                    }
+                }
+            });
         });
 
         return layerActionsUtil.playLayerActions(document, Immutable.List(layerEffectPlayObjects), true, options);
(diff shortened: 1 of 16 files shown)


src/js/actions/layereffects.js
```diff
@@ -325,6 +333,89 @@ define(function (require, exports) {
...
+            documentID: document.id,
+            layerIDs: layerIDs,
+            layerEffectTypes: Immutable.List(masterEffectTypes),
+            layerEffectIndex: Immutable.Map(masterEffectIndexMap),
+            layerEffectProps: Immutable.Map(masterEffectPropsMap),
+            coalesce: false
+        };
+
+        // Synchronously update the stores
+        this.dispatch(events.document.history.optimistic.LAYER_EFFECTS_MASS_CHANGED, payload);
```
iwehrman: Why synchronous dispatch?

src/js/actions/layereffects.js
```diff
@@ -325,6 +333,89 @@ define(function (require, exports) {
     setShadowSpread.reads = [locks.PS_DOC, locks.JS_DOC];
     setShadowSpread.writes = [locks.PS_DOC, locks.JS_DOC];
 
+    /**
+     * Copies the layer effects of the source layer to all the layers
```
iwehrman: I don't think "copy" is the right word here. Maybe "transfer" would be clearer? Just kicking ideas around.

src/js/actions/sampler.js
```diff
@@ -0,0 +1,180 @@
...
+     * @return {Promise.<Color>} Either the primary color of the layer, or the color at the clicked pixel
+     */
+    var _getSourceColor = function (doc, sourceLayer, x, y) {
+        var uiStore = this.flux.store("ui"),
+            coords = uiStore.transformWindowToCanvas(x, y),
+            layerKinds = sourceLayer.layerKinds;
+
+        switch (sourceLayer.kind) {
+            case layerKinds.VECTOR:
+                return Promise.resolve(sourceLayer.fills.first().color);
```
iwehrman: The list of fills could be empty?

src/js/actions/sampler.js
```diff
@@ -0,0 +1,180 @@
...
+            coords = uiStore.transformWindowToCanvas(x, y),
+            layerTree = doc.layers;
+        
+        return uiUtil.hitTestLayers(doc.id, coords.x, coords.y)
+            .bind(this)
+            .then(function (hitLayerIDs) {
+                var hitLayerMap = hitLayerIDs.reduce(function (layerMap, id) {
+                        layerMap[id] = true;
+                        return layerMap;
+                    }, {}),
```
iwehrman: I think we generally prefer using `Map` or, in this case, `Set` for this sort of thing instead of objects.

src/js/actions/layereffects.js
volfied: This is the way we do effects. We update our own stores first, so later on when we call _syncStoreToPs, we get the overall effect data from the store and pass it to Photoshop.

src/js/actions/sampler.js
iwehrman: Maybe the return value should be `Promise.<?Color>`.

src/js/actions/sampler.js
volfied: This is identical to what I've been doing in Superselect. Reason I use a JS object is for fast lookups in the next line where I filter out non clicked leaves.

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+    // Used for debouncing the overlay drawing
+    var DEBOUNCE_DELAY = 200;
+
+    var SamplerOverlay = React.createClass({
+        mixins: [FluxMixin, StoreWatchMixin("tool", "document", "application", "ui", "modifier")],
+
+        /**
+         * Keeps track of current mouse position so we can rerender the overlaid layers correctly
+         * @type {Number}
```
iwehrman: Nit: primitive types are lower-cased in JSDoc annotations: `object`, `boolean`, `number`, etc.

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -0,0 +1,278 @@
...
+         */
+        _currentMouseY: null,
+
+        /**
+         * Owner group for all the overlay svg elements
+         *
+         * @type {<SVGElement>}
```
iwehrman: Nit: superfluous angle brackets.

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+        /**
+         * Attaches to mouse move events coming from Photoshop
+         * so we can set highlights manually. We have to resort to this
+         * because external mouse move events do not cause :hover states
+         * in DOM elements
+         *
+         * @param {CustomEvent} event EXTERNAL_MOUSE_MOVE event coming from _spaces.OS
+         */
+        mouseMoveHandler: function (event) {
```
iwehrman: Should be private?

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -0,0 +1,278 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER 
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING 
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
+ * DEALINGS IN THE SOFTWARE.
+ * 
+ */
+
+
+define(function (require, exports, module) {
```
iwehrman: There sure does seem to be a lot of duplication here w.r.t. `SuperSelectOverlay`. I guess it wasn't obvious how to share any code between them?

src/js/stores/document.js
```diff
@@ -85,6 +85,7 @@ define(function (require, exports, module) {
                 events.document.history.nonOptimistic.STROKE_ADDED, this._handleStrokeAdded,
                 events.document.history.optimistic.LAYER_EFFECT_CHANGED, this._handleLayerEffectPropertiesChanged,
                 events.document.history.optimistic.LAYER_EFFECT_DELETED, this._handleDeletedLayerEffect,
+                events.document.history.optimistic.LAYER_EFFECTS_MASS_CHANGED, this._handleLayerEffectsMassChanged,
```
iwehrman: "Batch" might be better than "mass".

src/js/util/ui.js
```diff
@@ -45,5 +53,53 @@ define(function (require, exports) {
...
+     * @param {number} x Horizontal coordinate
+     * @param {number} y Vertical coordinate
+     * @param {number?} opacity Opacity of the owner layer at the pixel
+     * @return {Promise.<Color>}
+     */
+    var colorAtPoint = function (id, x, y, opacity) {
+        var documentRef = documentLib.referenceBy.id(id),
+            hitPlayObj = hitTestLib.colorSampleAtPoint(documentRef, Math.round(x), Math.round(y));
+
+        opacity = opacity || 100;
```
iwehrman: This would convert `opacity === 0` into `opacity === 100`. You should declare `@param {number=}` and then test `opacity === undefined`.

iwehrman:
Done with a first pass. Looking good. Back to you, @volfied.

src/js/actions/layereffects.js
volfied: I think transfer is more wrong than copy, since they remain at the source layer as well.

src/js/jsx/tools/SamplerOverlay.jsx
volfied: Overlays are tough to share code with. On one hand, a lot of this can be moved to a mixin, but I am not a fan of mixins.  On the other hand, there is a lot of specific CSS classes we handle sprinkled through out the code, and even though function names are similar, behaviors and appearances differ enough that sharing code together would be more trouble than it's worth.

volfied:
@iwehrman Addressed your nits and questions. Thank you for the review.

src/js/actions/layereffects.js
iwehrman: The problem with "copy" is that it implies that it is not a destructive action. Copying something in a document does not change the document, but this does. Also, we'll soon have an actual "copy effects" command, which will do an actual non-destructive copy. How about "duplicate"? This would be analogous to Photoshop's layer duplicate command.

src/js/actions/sampler.js
iwehrman: I'm not talking about an Immutable Map/Set, I'm talking about a mutable `window.Map`, etc.

volfied:
:+1: Understood and made more changes @iwehrman

iwehrman:
image

PR #1829, opened 2015-07-24
removes Draggable's dependencies on its parent component

shaoshing:
This PR removes the Draggable component's dependencies on its parent component to pass in the **dragPosition** and **isDragTarget** by retrieving them from the DragAndDrop store directly. This will avoid duplicate in the LibrariesPanel and Tab components. More importantly, it makes the code easier to understand. 

##### Bonus 

Also resolved a position bug when you drag and move a layer fast.
**before** drag-before

**after** drag-after

##### Thoughts about future top-down refactoring

I against the way of wrapping the target component with higher level components (e.g. Draggable/Droppable, or  the similar [react-dnd](https://github.com/gaearon/react-dnd) package), as it tends to make the data flow and communication vague:

* higher level component (e.g. Draggable/Droppable) hides the props required by the target component (e.g. LayerFace). Basically we have to pass the props to Droppable/Draggable and then use {...this.state.props} to pass down to LayerFace. This tend to be buggy.

* the communications between the components are unclear. 

* configuration is difficult and confusing.

A better way might be wrapping the content of the component, instead of the component. Here is a [gist](https://gist.github.com/shaoshing/16a4fd0e2fb27a06d61c#file-example-jsx) I created as a demonstration of the Draggable component. The gist does not include every detail, but if you just look at the Item component, you have a much better understanding of the data flow and communication, and it is also very React style. The trick is to the `React.cloneElement(this.props.children[0], { ref: dragTarget })`, which allows us encapsulate the event listening inside the Draggable component, leaving the Item component clean and concise.

@iwehrman @designedbyscience  Please review and let me know you thoughts.

### DIFF
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -469,13 +469,6 @@ define(function (require, exports, module) {
 
                 return flux.actions.layers.reorder(doc, dragSource, dropIndex)
                     .bind(this)
-                    .then(function () {
-                        // The selected layers may have changed after the reorder.
-                        var documentStore = flux.store("document"),
-                            nextDocument = documentStore.getDocument(doc.id);
-
-                        flux.actions.layers.resetBounds(nextDocument, nextDocument.layers.allSelected);
-                    })
                     .finally(function () {
                         this.setState({
                             dropPosition: null,
@@ -551,7 +544,7 @@ define(function (require, exports, module) {
                         }
                     }, this)
                     .map(function (layer, visibleIndex) {
-                        var isDropTarget = dropTarget && dropTarget.key === layer.key,
+                        var isDropTarget = !!dropTarget && dropTarget.key === layer.key,
                             dropPosition = isDropTarget && this.state.dropPosition;
 
                         return (
@@ -585,6 +578,7 @@ define(function (require, exports, module) {
                         document={doc}
                         zone={doc.id}
                         isValid={this._validDropTarget}
+                        keyObject={{ key: "dummy" }}
                         onDrop={this._handleDrop}
                         isDropTarget={isBottomDropTarget} />
                 );
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -467,7 +467,7 @@ define(function (require, exports, module) {
 
                 var dragSource = collection.pluck(this.state.dragTargets, "id");
 
-                flux.actions.layers.reorder(doc, dragSource, dropIndex)
+                return flux.actions.layers.reorder(doc, dragSource, dropIndex)
                     .bind(this)
                     .then(function () {
                         // The selected layers may have changed after the reorder.
@@ -489,6 +489,8 @@ define(function (require, exports, module) {
                 this.setState({
                     dropPosition: null
                 });
+                
+                return Promise.resolve();
             }
         },
 
@@ -535,8 +537,7 @@ define(function (require, exports, module) {
                 dragTargets = this.state.pastDragTargets;
             }
 
-            var dragTargetSet = dragTargets && dragTargets.toSet(),
-                layerComponents = doc.layers.allVisibleReversed
+            var layerComponents = doc.layers.allVisibleReversed
                     .filter(function (layer) {
                         // Do not render descendants of collapsed layers unless
                         // they have been mounted previously
@@ -550,8 +551,7 @@ define(function (require, exports, module) {
                         }
                     }, this)
                     .map(function (layer, visibleIndex) {
-                        var isDragTarget = dragTargets && dragTargetSet.has(layer),
-                            isDropTarget = dropTarget && dropTarget.key === layer.key,
+                        var isDropTarget = dropTarget && dropTarget.key === layer.key,
                             dropPosition = isDropTarget && this.state.dropPosition;
 
                         return (
@@ -561,7 +561,7 @@ define(function (require, exports, module) {
                                 disabled={this.props.disabled}
                                 document={doc}
                                 layer={layer}
-                                axis="y"
+                                keyObject={layer}
                                 visibleLayerIndex={visibleIndex}
                                 dragPlaceholderClass="face__placeholder"
                                 zone={doc.id}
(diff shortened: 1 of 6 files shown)


designedbyscience:
@shaoshing You should use Vermillion to check this...It seems much slower than the current drag and drop (intolerably so)

designedbyscience:
For your proposed refactoring: I like the sound of that, I hadn't looked into the cloneElement trick

shaoshing:
Hey @designedbyscience , the performance issue is fixed by manually listening to the DragAndDrop store such that when dragging, there is only one layer (instead of all layer) that is responding to the drag updates. Now it is as fast as the original one :dancers:

designedbyscience:
Okay, now 1 bug with layers not dropping correctly and we can't drag multiple items at once

iwehrman:
I really like where this is going! But, before I dig further, I'd like to point out that there is still a very hot spot in the profile that happens when dragging around the area where the drag started:

drop

In that profile above, those areas of intense computation are happening when I drag the layers around where I originally picked them up. Performance otherwise seems fine, but the rendering delays there are huge. Can you look into that? Also, as @designedbyscience suggested, testing with VermilionArtboards.psd is essential. Also, you should expand _all_ the layers when doing this testing. We're still planning to limit the total number of layers present in the DOM at any time (updating the set on scroll, for example), but I want to keep the core rendering as efficient as possible in case we run into trouble with that plan.

shaoshing:
The reason of the peaks is  because the `isDropTarget` property changed from `false -> null ` (`null -> false`) and thus triggers all LayerFace re-rendered when you drag in to (out of) the original positions of the dragged layers. The fix is making sure to convert `null` to `false` before passing to LayerFace (https://github.com/adobe-photoshop/spaces-design/pull/1829/files#diff-599bf64b66bfd695d31114df0d8d0c63R554). (Note that the master branch has the same issue).

after

**Another note for future refactoring**

Since the LayerFace relies on the LayerPanel heavily, each re-render of the LayerPanel will trigger all LayerFace to re-render. Even though we are using `shouldComponentUpdate` to avoid re-rendering in the children, but the total time of all function calls (in React) could be a lot. We've removed the dependencies in Draggable, so I went a little bit further by removing the dependencies in the Droppable, and the result is 40% faster:

 droppabel-before Before 
 droppable-after After. Most of the `drag move` are below the 60FPS line, 

To make the Droppable independent, we need to use `onMouseEnter` and `onMouseLeave` to know which Droppable is the drop target, and also setting css `pointer-events: none;` to the dragged item to allow elements below it to receive mouse events.

@iwehrman  @designedbyscience  please review.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -467,7 +467,7 @@ define(function (require, exports, module) {
 
                 var dragSource = collection.pluck(this.state.dragTargets, "id");
 
-                flux.actions.layers.reorder(doc, dragSource, dropIndex)
+                return flux.actions.layers.reorder(doc, dragSource, dropIndex)
```
iwehrman: I know you didn't add this `then` handler, but it's a little odd that it's in the JSX instead of as part of some other action. I only mention it because there is not actually any reason why this method needs to return a promise.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -550,8 +551,7 @@ define(function (require, exports, module) {
                         }
                     }, this)
                     .map(function (layer, visibleIndex) {
-                        var isDragTarget = dragTargets && dragTargetSet.has(layer),
-                            isDropTarget = dropTarget && dropTarget.key === layer.key,
+                        var isDropTarget = !!dropTarget && dropTarget.key === layer.key,
```
iwehrman: This isn't really a nit, but I don't think it's actually necessary to convert the first parameter into a boolean here.

src/js/jsx/shared/Draggable.jsx
```diff
@@ -99,71 +117,77 @@ define(function (require, exports, module) {
...
                     dragging: false,
-
-                    // Start top/left of the DOM node
-                    startX: null, startY: null,
-                    dragStyle: null
+                    initialDragPosition: null
                 };
             },
+            
+            // Instead of responding to DragAndDrop's all change events, it listens to the "start-drag" event (fires
```
iwehrman: For consistency, methods and functions should be prefixed with block-style comments instead of line-style comments. Also, please mark it as `@private`.

src/js/jsx/sections/layers/LayersPanel.jsx
shaoshing: Adding the `!!` will make sure `isDropTarget` is always a boolean value, this is the cause of the performance issue when dragging over dragged items. So, before you drag over the dragged items, the `dropTarget` has value and `isDropTarget ` is false for all non-drop targets, but when you drag over the dragged items, `dropTarget` changed to null, so as the `isDropTarget`, and cause all non-drop targets to do a re-render because `this.props.isDropTarget !== nextProps.isDropTarget`.

iwehrman: Doh, yes, I get it.

src/js/jsx/sections/layers/LayersPanel.jsx
shaoshing: The reason behind this is that the "dragging" state should maintain until the drop process is finished. Without this, the dropped layer will snap back after dropping, or you have to add information into the draganddrop store to represents a `dropping` state. I think the promise way is a simpler and does not affect user experience.  I also added the reason here https://github.com/adobe-photoshop/spaces-design/pull/1829/files#diff-9adf1065d9513443f0903ee1061b1b0dR182

src/js/jsx/shared/Draggable.jsx
shaoshing: Got it.

src/js/jsx/sections/layers/LayersPanel.jsx
iwehrman: OK, I understand a little better. I agree that the `finally` handler makes sense here, but I still think the `then` handler should be part of the action.

shaoshing: the `then` part is from the master, but I agree with you on that. Let me see if I can move to the action safely.

iwehrman:
This is tremendous. Thank you for helping me understand your changes. I still have a few comment nits and possibly would like to push a little more of the drop handler back into reorder (if that's appropriate), but once that's cleaned up I'm ready to merge.

shaoshing:
Thanks for reviewing @iwehrman  , I updated the function doc and moved some drop handling code to where it belongs to (actions.layers). Now back to you.

iwehrman:
Looks good!

designedbyscience:
Awesome!

chadrolfs:
ahhhh, much better

Issue #1816, opened 2015-07-23
Always show up to ~10 results


iwehrman:
The total number of visible results should be ~10, and if all the valid results are from the same category then sobeit. 
![5](https://s3.amazonaws.com/f.cl.ly/items/1i030J3d2e1n3G0V1801/Image%202015-07-23%20at%202.42.43%20PM.png)

volfied:
FBNC to Ian

iwehrman:
Sure enough.

Issue #1807, opened 2015-07-23
Baseline of proposed search completion is shifted


iwehrman:
![baseline](https://s3.amazonaws.com/f.cl.ly/items/2I1q0o0t2u2e41153c45/Image%202015-07-23%20at%207.12.56%20AM.png)

Notice that "Others" sits below "Hide".

volfied:
FBNC to Ian

iwehrman:
Sure enough.

PR #1832, opened 2015-07-24
Search stylin'

hamburge:
Issues #1816, #1807, and also:
- Fix for if you had a filter set, the autocomplete suggestion would not always be lined up depending on your screen resolution.
- Fix for if there were no filter suggestions, there was an extra line on the top of the drop down.
- When you click on the dialog and the input loses focus, don't show the autofill suggestion
- Removes code that supported displaying multiple svgs for filters since it is not necessary.

### DIFF
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -443,7 +443,7 @@ define(function (require, exports, module) {
                 // Otherwise suggest based on last word typed
                 var valueLowerCase = value ? value.toLowerCase() : "",
                     lastWord = valueLowerCase.split(" ").pop(),
-                    options = this._filterOptions(valueLowerCase, true),
+                    options = this._filterOptions(valueLowerCase, false),
 
                     suggestion = (options && valueLowerCase !== "") ? options.find(function (opt) {
                             return (opt.type === "item" && opt.title.toLowerCase().indexOf(valueLowerCase) === 0);
@@ -475,6 +475,10 @@ define(function (require, exports, module) {
             }
         },
 
+        /**
+         * Sets autofill suggestion to empty string and re-renders
+         *
+         */
         removeAutofillSuggestion: function () {
             if (this.props.useAutofill) {
                 this.setState({
@@ -552,27 +556,35 @@ define(function (require, exports, module) {
                 filter = this.state.filter,
                 title = this.state.active && filter !== null ? filter : value,
                 searchableFilter = filter ? filter.toLowerCase() : "",
-                filteredOptions = this._filterOptions(searchableFilter, false),
+                filteredOptions = this._filterOptions(searchableFilter, true),
                 searchableOptions = filteredOptions;
 
             if (filteredOptions && collection.uniformValue(collection.pluck(filteredOptions, "type"))) {
                 searchableOptions = this.props.placeholderOption;
             }
             
-            var autocomplete,
-                hiddenTI,
-                hiddenSVG;
+            // Use hidden text input and hidden svg to find position of suggestion by updating their values and 
+            // getting width before calling render. Without using the hidden elements, in certain cases the 
+            // calculated position of suggestion would be one keystroke behind. 
+
+            // Hidden text input gets the value of the search input before rendering.
+            var hiddenTI,
+            // Hidden SVG is used to find width of a filter icon. It is just a div with the same width 
+            // (including margin) as the icons used for filters
+                hiddenSVG,
+                autocomplete;
+
  
             if (this.props.useAutofill) {
                 hiddenTI = (
                     <div ref="hiddenTextInput"
-                        className="hiddeninput">
+                        className="hidden-input">
                     </div>
                 );
 
                 hiddenSVG = (
                     <div ref="hiddenSVG"
-                        className="hiddenSVG">
+                        className="hidden-svg">
                     </div>
                 );
                 
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -84,15 +84,14 @@ define(function (require, exports, module) {
                 return true;
             }
 
-            // Update autofill here so that can check options based on the updated filter
-            if (this.props.options !== nextProps.options ||
+            return (this.props.options !== nextProps.options ||
                 this.state.filter !== nextState.filter ||
                 this.state.active !== nextState.active ||
-                this.state.suggestTitle !== nextState.suggestTitle) {
-                this._updateAutofill(nextState.filter, nextProps.filterIcon);
-                return true;
-            }
-            return false;
+                this.state.suggestTitle !== nextState.suggestTitle);
+        },
+
+        componentDidUpdate: function () {
+            this._updateAutofillPosition();
         },
 
         /**
@@ -347,6 +346,8 @@ define(function (require, exports, module) {
             this.setState({
                 filter: value
             });
+
+            this._updateAutofill(value);
         },
 
         /**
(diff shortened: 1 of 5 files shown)


src/js/jsx/Search.jsx
```diff
@@ -86,6 +86,7 @@ define(function (require, exports, module) {
                             searchID={SEARCH_BAR_DIALOG_ID}
                             dismissDialog={this._closeSearchBar}
                             executeOption={this._handleOption}
+                            maxOptions={14}
```
iwehrman: It will be easier to change this later if it's broken out into a `CONSTANT` with JSDoc.

iwehrman: Or, just set the default the way you want it for `Search` in `SearchBar` below and omit this entirely. (Future clients of `SearchBar` can change the default as they see fit, but we might as well make the default correct for the first client.)

src/js/jsx/shared/Datalist.jsx
```diff
@@ -451,6 +475,15 @@ define(function (require, exports, module) {
             }
         },
 
+        removeAutofillSuggestion: function () {
```
iwehrman: JSDoc needed.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -451,6 +475,15 @@ define(function (require, exports, module) {
...
+            if (this.props.useAutofill) {
+                this.setState({
+                    suggestTitle: ""
+                });
+                this.forceUpdate();
```
iwehrman: The `setState` call above will cause `render` to be run assuming the state passes through `shouldComponentUpdate`. Why is the extra `forceUpdate` call necessary?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -527,7 +560,8 @@ define(function (require, exports, module) {
             }
             
             var autocomplete,
-                hiddenTI;
+                hiddenTI,
+                hiddenSVG;
  
             if (this.props.useAutofill) {
                 hiddenTI = (
```
iwehrman: Can you leave a few `// comments` here that describe the point of these hidden elements?

src/style/search/search-bar.less
```diff
@@ -67,10 +71,17 @@
 .search-bar__dialog .drop-down .hiddeninput {
     display: inline-block;
     position: absolute;
-    left: 3.2rem;
+    margin-left: .1rem;
     visibility: hidden;
     white-space: pre;
 }
+.search-bar__dialog .drop-down .hiddenSVG {
```
iwehrman: The naming on some of these classes is not consistent with the rules in our style guide here: https://github.com/adobe-photoshop/spaces-design/wiki/Coding-Conventions#naming-and-syntax

iwehrman:
A few minor requests. Back to @hamburge for polish.

src/js/jsx/shared/Datalist.jsx
hamburge: Otherwise in `shouldComponentUpdate`, it sets `suggestTitle` to a value other than `""` again before rendering.

iwehrman: This is a bad code smell. `shouldComponentUpdate` really should not have side effects. Why is that necessary? What are other ways that we could get rid of this extra call? If I'm understanding this, I think the proposed control flow is too messy to commit.

volfied: Yeah this is bad... SCU should not ever update anything. The render method should see the conditions you're checking for in SCU and render correctly if the condition is being met.

hamburge: I moved the call I was making in SCU to `inputChange`, which I think addresses both of the issues.

hamburge:
@iwehrman I cleaned it up based on your and @volfied's comments so now there shouldn't be side effects from `shouldComponentUpdate`

src/js/util/svg.js
```diff
@@ -54,40 +54,33 @@ define(function (require, exports) {
...
             if (kind === "ARTBOARD") {
                 iconID += "artboard";
             } else if (kind === "BACKGROUND") {
                 iconID += layerLib.layerKinds.PIXEL;
             } else if (kind === "SMARTOBJECT" && isLinked) {
                 iconID += layerLib.layerKinds.SMARTOBJECT + "-linked";
-            } else if (kind !== "LAYER") {
+            } else if (kind.indexOf("LAYER") === -1) { // check if kind is "ALL_LAYER" or "CURRENT_LAYER"
                 iconID += layerLib.layerKinds[kind.toUpperCase().replace(" ", "")];
             }
```
volfied: What about the cases where none of these conditions match, is `"layer-"` an acceptable iconID?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -519,20 +552,35 @@ define(function (require, exports, module) {
...
             }
             
-            var autocomplete,
-                hiddenTI;
+            // Use hidden text input and hidden svg to find position of suggestion by updating their values and 
+            // getting width before calling render. Without using the hidden elements, in certain cases the 
+            // calculated position of suggestion would be one keystroke behind. 
+
+            // Hidden text input gets the value of the search input before rendering.
+            var hiddenTI,
```
volfied: I believe React prefers components to be defined, or `null`. Can you null initialize them here?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -580,22 +628,24 @@ define(function (require, exports, module) {
...
-
-            if (this.props.filterIcons) {
-                _.forEach(this.props.filterIcons, function (icon, index) {
-                    svg.push((<SVGIcon
-                            className="datalist__svg"
-                            CSSID={icon}
-                            key={index}
-                            viewbox="0 0 24 24"/>));
-                });
+                svg;
```
volfied: diitto

src/js/jsx/shared/Datalist.jsx
```diff
@@ -393,33 +389,58 @@ define(function (require, exports, module) {
...
+         * @private
+         * @param {string} value The current value of the input
+         * @param {string} icon The SVG icon for the input
+         */
+        _updateAutofillPosition: function (value, icon) {
+            // Base position off of hidden element widths because Datalist sub-components
+            // haven't rendered yet using the new value, and we don't want the suggestion to move
+            // after rendering
+
+            var hiddenInput = React.findDOMNode(this.refs.hiddenTextInput);
```
volfied: https://facebook.github.io/react/tips/dangerously-set-inner-html.html

Given this link, can you justify using innerHTML here?

volfied:
This code is diverging from React. I have couple nits, and one big problem.

src/js/jsx/shared/Datalist.jsx
hamburge: Wow so scary. For some reason I originally thought that moving the autofill suggestion in `componentDidUpdate` would be bad but it looks like that's possible, so I can avoid using `innerHTML` :grin:

src/js/util/svg.js
hamburge: That shouldn't ever happen. If the filter just contains "ALL_LAYER" or "CURRENT_LAYER" then it returns above. And otherwise the options are all in layerKinds, so they'll give the correct value.

I realize the logic to get the icons is very layer-specific right now, but once I get the other SVGs from Seth, I'll work on making it all more generic.

hamburge:
@volfied Ok reworked it so that it is more React-friendly, I think.

Issue #1776, opened 2015-07-21
Mixed state is inconsistent


chadrolfs:
![screen shot 2015-07-21 at 9 39 41 am](https://cloud.githubusercontent.com/assets/8442631/8806570/20f99c24-2f8c-11e5-9d81-bfccb8d046c6.png)

chadrolfs:
Fill mixed state is black, others are lined

volfied:
FBNC to Chad

PR #1833, opened 2015-07-24
Bug fix: Mixed state is inconsistent

shaoshing:
Bug fix for #1776, handle the following cases:

* select multiple layer with different fill color: style panel shows mixed color (used to show black color)
* select multiple layer with different stroke color: style panel shows mixed stoke color (used to show the stroke color of the first layer)

### DIFF
--- a/src/js/models/color.js
+++ b/src/js/models/color.js
@@ -25,7 +25,8 @@ define(function (require, exports, module) {
     "use strict";
 
     var Immutable = require("immutable"),
-        mathjs = require("mathjs");
+        mathjs = require("mathjs"),
+        tinycolor = require("tinycolor");
 
     var objUtil = require("js/util/object"),
         mathUtil = require("js/util/math");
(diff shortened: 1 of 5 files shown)


shaoshing:
Also added `Color#toTinyColor`, no more 

```
var tinycolor = require("tinycolor");
tinycolor(color.toJS());
```

src/js/models/color.js
```diff
@@ -180,11 +181,11 @@ define(function (require, exports, module) {
...
     /**
-     * Return string representation in CSS RGB format. 
-     * @return {string} Example: rgb(100, 100, 0)
+     * Return instance of TinyColor. Provide convenicen for accessing TinyColor functions
+     * @return {TinyColor} 
      */
-    Color.prototype.toCssRGB = function () {
-        return "rgb(" + [Math.round(this.r), Math.round(this.g), Math.round(this.b)].join(", ") + ")";
+    Color.prototype.toTinyColor = function () {
+        return tinycolor(this.toJS());
```
volfied: Can you confirm that tinycolor is doing the rounding for us now?

volfied: Nvm, tinycolor sources say that they first make a percentage out of the color value and then multiply by 255, so they would be rounding in there. :+1:

shaoshing:
Thanks @volfied

PR #1834, opened 2015-07-24
 first pass at adding new type layer. has warnings

pineapplespatula:
I am currently getting this Warning: 

`flattenChildren(...): Encountered two children with the same key, `.$1508=114`. Child keys must be unique; when two children share a key, only the first child will be used.`

I would like to chat with someone on monday about our layer implementation and how to add duplicate/ temporary layers to the model without it being so hacky.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1522,7 +1522,12 @@ define(function (require, exports) {
                 }
 
                 if (typeof event.layerID === "number") {
-                    this.flux.actions.layers.addLayers(currentDocument, event.layerID);
+                    var curLayer = currentDocument.layers.byID(event.layerID);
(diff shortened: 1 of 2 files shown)


iwehrman:
The simplest workaround is to ensure that they key of the new layer is unique. But, why do you want to add a duplicate layer temporarily?

pineapplespatula:
I would prefer that it just get added once when the layer is added during the modal event, but i dont think i can stop the notification when the tool is commited

iwehrman:
So we're getting the layer notification twice? If so, can't we just update the existing layer the second time?

pineapplespatula:
@iwehrman this is ready for a real review now

iwehrman:
:ship: :it:

Issue #1734, opened 2015-07-14
Smart guides and direct select overlays are inconsistent


iwehrman:
Direct (command) select seems to target different layers than are targeted by smart guides, which at least causes a lot of visual noise and at worst is really frustrating. To reproduce, open the VermillionArtboards.PSD file and try to directly target just about any object on any artboard, e.g., the seated woman in the top-left corner of the document.

It's not clear to me why we need two different overlays. If there's a good reason to have them both, it would be nice to make sure they're both implemented to the same specification so that they're consistent.

iwehrman:
CC @volfied.

volfied:
We need to discuss this, because I'm not sure what to do exactly. 

@placegraphichere: If we're always going to have smart guides on in Photoshop, what I think we should do is, disable cmd+ highlight for superselect (but still have it work), and match group highlighting to the same magenta color. 

This way, users will see the pop up ruler values (distance between selected layer and hovered layer) only when cmd key is down, and in other itmes, they'll see our same color highlight graphic.

volfied:
Keep in mind we can set the smart guide color to anything we want.

iwehrman:
I agree we should not have both at the same time, and that they should be styled consistently. (Conceivably though, if the user has disabled smart guides then perhaps they also don't want the hover highlighting?)

A separate issue though is the difference in algorithms used for highlighting. If we have one algorithm for highlighting with smart guides enabled and another one when they're disabled, it seems like the objects highlighted should still be the same in the same contexts.

volfied:
I do not see the inconsistency, can you provide a screenshot? Or better, a small file where I can easily repro?

iwehrman:
I can't provide a screenshot because smart guides. The repro steps in the original comment are the best I can do. It may be related to #1760.

designedbyscience:
![vermilionartboards_psd___66_7___desktop__rgb_8_____and_slack_and__1__playground_js___trello_and_issues_ _adobe-photoshop_spaces-design](https://cloud.githubusercontent.com/assets/22372/8780118/dfaa5a0e-2ed4-11e5-8330-0ffcfd888842.png)

designedbyscience:
Right now I think we'd have a problem if we simply turned off our overlay if smart guides is on as, like @iwehrman mentioned, I definitely see the case where an object appears active in smart guides but if you click you select a different object

volfied:
[...]

PR #1838, opened 2015-07-24
Do not target invisible layers

iwehrman:
Currently invisible layers are targetable on-canvas with selection and deep (command/alt) selection. This makes it impossible to target visible layers in VermilionArtboards.psd because there is a hidden grid system at the top of the layer index which covers the entire page. @placegraphichere correctly pointed out that invisible layers should not be targetable on-canvas, and that this was the source of frustration targeting layers in VermilionArtboards.psd. This change corrects that. Also, @volfied pointed out that groups with only visible children should also be treated as invisible, so he fixed that too.

Partially addresses #1734, by making smart guide overlays more consistent with the HTML overlays. (We still shouldn't draw both in different colors, but this this at least makes things less scatter-brained.)

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -170,6 +170,10 @@ define(function (require, exports) {
      */
     var _getContainingLayerBounds = function (layerTree, x, y) {
         return Immutable.Set(layerTree.all.reduce(function (layerSet, layer) {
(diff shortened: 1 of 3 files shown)


volfied:
They're gonna feel so alone and invisible.

PR #1839, opened 2015-07-25
Search for layers in any open document, not just the current document

hamburge:
Can now register a search module to either include layers in any open document or in just the current document.

If the layer is from the non-current document, the file name it comes from will appear as secondary information on the search option. Otherwise it shows the shortened layer path (just like before).

If a layer from the non-current document is confirmed, we switch to its document and select it.

### DIFF
--- a/src/js/actions/search/layers.js
+++ b/src/js/actions/search/layers.js
@@ -28,19 +28,20 @@ define(function (require, exports) {
         _ = require("lodash");
 
     var events = require("js/events"),
+        mathUtil = require("js/util/math"),
         svgUtil = require("js/util/svg");
     
     /**
      * Get the layer type and if it is linked or an artboard, as an array of strings.
-     * This is used for comparison against the inputted search terms and any search filters.
-     * Not directly user-visible, but needs to be able to match user input.
+     * This is used for comparison against any search filters.
      *
      * @private
      * @param {Layer} layer
+     * @param {string} type
      * @return {Array.<string>}
     */
-    var _getLayerCategory = function (layer) {
-        var layerType = ["LAYER"];
+    var _getLayerCategory = function (layer, type) {
+        var layerType = [type + "_LAYER"];
 
         if (layer.kind === layer.layerKinds.GROUP && layer.isArtboard) {
             layerType.push("ARTBOARD");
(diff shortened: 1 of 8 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -309,7 +309,7 @@ define(function (require, exports, module) {
                         }
                     });
 
-                    priority += (searchTerms.length - numTermsInTitle);
+                    priority += (searchTerms.length - numTermsInTitle + 3);
```
iwehrman: Mystery constants like this should definitely have an explanatory comment :)

src/js/actions/search/layers.js
```diff
@@ -87,55 +121,68 @@ define(function (require, exports) {
...
     };
 
     /**
-     * Register layer info for search
+     * Register layer info to search
      *
+     * @param {string} type "CURRENT" or "ALL". If not specified, defaults to current
      */
-    var registerLayerSearch = function () {
+    var _registerLayerSearch = function (type) {
```
iwehrman: It would be nicer to make this a boolean parameter so that the detail of which strings in particular are valid can be more limited.

src/js/util/svg.js
```diff
@@ -81,7 +81,7 @@ define(function (require, exports) {
                 iconID += layerLib.layerKinds[kind.toUpperCase().replace(" ", "")];
             }
 
-            if (kind !== "LAYER") {
+            if (kind.indexOf("LAYER") === -1) {
```
iwehrman: Can you add a comment here that clarifies what you're trying to match?

src/js/actions/search/layers.js
```diff
@@ -78,7 +79,40 @@ define(function (require, exports) {
...
+            layers = document.layers.allVisibleReversed;
+        
+        var layerMap = layers.map(function (layer) {
+            var ancestry = document.name,
+                layerType = _getLayerCategory(layer, type),
+                iconID = svgUtil.getSVGClassFromLayer(layer);
+
+            if (document.id === currentID) {
+                ancestry = _formatLayerAncestry(layer, appStore);
+            }
```
iwehrman: It would be easier to understand if `ancestry` was not initialized to the else value above, but instead was just declared above and initialize in one of two ways in a single if-else here.

src/js/actions/search/layers.js
```diff
@@ -78,7 +79,40 @@ define(function (require, exports) {
...
+     * @param {Document} document
+     * @param {string} type Type of layer search, for finding correct layer category
+     * @return {Immutable.List.<object>}
+    */
+    var _getLayerOptionsFromDoc = function (document, type) {
+        var appStore = this.flux.store("application"),
+            currentID = appStore.getCurrentDocumentID(),
+            layers = document.layers.allVisibleReversed;
+        
+        var layerMap = layers.map(function (layer) {
```
iwehrman: Also, it's fine to just directly `return layers.map(...`

src/nls/root/strings.js
```diff
@@ -601,7 +602,8 @@ define(function (require, exports, module) {
                 RECENT_DOC: "Recent Documents",
                 DOCUMENT: "Documents",
                 MENU_COMMAND: "Menu Commands",
-                LAYER: "Layers",
+                ALL_LAYER: "Layers",
+                CURRENT_LAYER: "Layers",
```
iwehrman: I wonder if there is some way we can distinguish between these two categories that is helpful but not overly verbose....

iwehrman:
Looks good! Just a few little nits and questions.

src/nls/root/strings.js
hamburge: @iwehrman Would `ALL_LAYER` for all open document layers and just `LAYER` for current document layers be better? The reason they need different category names is so that the search store can keep them separate.

iwehrman: Oh, I actually meant the user-visible strings; the constant names seem exactly right.

hamburge: I imagine they wouldn't both be used in the same search module, so I think we can get away with just saying that you are searching for layers.

hamburge:
@iwehrman fixed the things!

iwehrman:
Things: fixed.

PR #1841, opened 2015-07-27
More layer group drag-and-drop fixes

iwehrman:
1. Fix a bug that prevented artboards from being reordered, as reported in #1831.
2. Change the semantics of dropping "below" groups such that dropping below an open group places it inside the group at the top of that group, and dropping below a closed group places it after the group.

As with all layers panel drag-and-drop changes, I have only limited confidence that I haven't broken something. Please test if possible, @chadrolfs.

Addresses #1831.


volfied:
Changes look good to me!

Issue #1812, opened 2015-07-23
UI bounds don't account for both panels


iwehrman:
I can tell because when we `cloak` the non-UI bounds, the inner panel blinks away momentarily but not the outer panel.

iwehrman:
Easier repro:
1. Create a new artboard document.
2. Select the artboard.
3. View > Center selection.

Note that artboard is not centered in the non-UI bounds.

iwehrman:
Fixed by #1842.

PR #1842, opened 2015-07-27
Fixes cloaking issue

designedbyscience:
Update the calculation of the cloak bounds at the right time


iwehrman:
:+1:

iwehrman:
I believe this addresses #1812.

PR #1847, opened 2015-07-27
HOTFIX #1846: Ignore null layers during removeArtboardIDs

### LINKED ISSUE #1486: Allow text tool to work in modal state
with this PR modal text works with multiple runs. however there are some commented out chunks of code that I have some misgivings about. if someone could take a look and help me figure out either a workaround or if that code is safe to remove i would appreciate it.

-- @pineapplespatula at [2015-05-28T23:27:34Z](https://github.com/adobe-photoshop/spaces-design-archive/pull/1486)

### LINKED ISSUE #1846: Errors selecting layers in Vermilion
Open it, zoom into Desktop Large, start selecting layers (targeting via cmd/ctrl modifier), delete a couple, select some more.

Result: errors

volfied:
Fixes #1846 , for prerelease.

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -210,7 +210,9 @@ define(function (require, exports) {
      */
     var _removeArtboardIDs = function (layerTree, ids) {
         return ids.filterNot(function (id) {
-            return layerTree.layers.get(id).isArtboard;
+            var layer = layerTree.layers.get(id);
+            
+            return layer ? layer.isArtboard : true;
         });
     };
 


volfied:
This, by the way, is just a band-aid. The real problem here is, our models are not up to date with Photoshop, which means we're forgetting a lock in the click action. 

On master, the fix should be for the locks and making sure that `hitTest`does not return back a layer ID that's not in our model.

iwehrman:
Safe, but does not address the core issue. We might want to leave #1846 open.

Issue #1849, opened 2015-07-28
Select All layers fails due to missing locks


iwehrman:
```
Executing action layers.selectAll after waiting 0ms; 2/0
fluxcontroller.js:223 Missing read locks: jsApp, jsTool, psApp, jsPolicyfluxcontroller.js:223 FluxController._makeActionReceiver.Object.create.transfer.valuelayers.js:662 selectAllfluxcontroller.js:304 FluxController._applyActionfluxcontroller.js:399 (anonymous function)bluebird.js:4559 tryCatcherbluebird.js:2687 23.module.exports.Promise._settlePromiseFromHandlerbluebird.js:2761 23.module.exports.Promise._settlePromiseAtbluebird.js:150 (anonymous function)
fluxcontroller.js:416 Action layers.selectAll failed: Error: Next action requires additional read locks
    at Object.FluxController._makeActionReceiver.Object.create.transfer.value (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:224:31)
    at Object.selectAll (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/layers.js:662:21)
    at FluxController._applyAction (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:304:36)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:399:37)
From previous event:
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:395:22)
    at AsyncDependencyQueue._processNext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:292:27)
    at AsyncDependencyQueue.push (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:144:14)
    at Object.FluxController._synchronize (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:378:42)
[...]

PR #1850, opened 2015-07-28
Add resetPolicy locks to layers.selectAll

volfied:
Addresses #1849, lock I've forgotten during the transform work.


iwehrman:
image

Issue #1848, opened 2015-07-27
History state with modal text tool is broken


pineapplespatula:
Step backwards is not allowed during the modal tool. and currently we have update color modify our history state, but it does not match PS behavior during the modal type tool. it only matches when we are not in the modal state.

iwehrman:
We should disable undo/redo when in a modal tool state.

iwehrman:
And probably a whole bunch of other commands.

pineapplespatula:
there is a  watson bug related to this 4028040

mcilroyc:
I'm stealing this so I can try out a fix that @jsbache recommended, and which I've initially had success with.

iwehrman:
FBNC to @chadrolfs, but please read the caveats described in #2270. Short version: full-blown undo/redo in the modal type state is impossible in PS. We should now be doing what standard does, which is to support a single, combined, single-step undo/redo command.

PR #1854, opened 2015-07-28
Modify color change in text to only sometimes add history

pineapplespatula:
Our history model and modal text states do not currently coexist properly.
to work around this the color set routine which was a nonoptomistic history state
if set in the modal environment should instead append to history.

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -202,7 +202,7 @@ define(function (require, exports) {
      * @param {boolean=} coalesce Whether to coalesce this operation's history state
      * @return {Promise}
      */
-    var updateColor = function (document, layers, color, coalesce) {
+    var updateColor = function (document, layers, color, coalesce, fromSet) {
         var layerIDs = collection.pluck(layers, "id"),
             normalizedColor = null;
 
@@ -217,7 +217,11 @@ define(function (require, exports) {
                 calesce: coalesce
             };
 
-        return this.dispatchAsync(events.document.history.optimistic.TYPE_COLOR_CHANGED, payload);
+        if (fromSet) {
+            return this.dispatchAsync(events.document.history.optimistic.TYPE_COLOR_CHANGED, payload);
+        } else {
+            return this.dispatchAsync(events.document.history.amendment.TYPE_COLOR_CHANGED, payload);
+        }
     };
     updateColor.reads = [];
     updateColor.writes = [locks.PS_DOC, locks.JS_DOC];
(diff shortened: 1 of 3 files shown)


src/js/actions/type.js
```diff
@@ -202,7 +202,7 @@ define(function (require, exports) {
      * @param {boolean=} coalesce Whether to coalesce this operation's history state
      * @return {Promise}
      */
-    var updateColor = function (document, layers, color, coalesce) {
+    var updateColor = function (document, layers, color, coalesce, fromSet) {
```
mcilroyc: nit: this param name is not very meaningful.  Maybe `amendHistory` ?

also: needs to be added to JSDoc

pineapplespatula:
@mcilroyc good points. back to you

src/js/events.js
```diff
@@ -71,6 +71,7 @@ define(function (require, exports, module) {
                     DELETE_LAYERS: "deleteLayersNonOptimistic" // eg: ps deletes the entire layer after last path del
                 },
                 amendment: {
+                    TYPE_COLOR_CHANGED: "typeColorChanged",
```
mcilroyc: you can't use the same string for both events!  'typeColorChanged'

mcilroyc: this is causing the bug that ALL setColor calls to use the "amendment" flow

mcilroyc:
Aside from the mistake I mentioned above (which causes a bug in non-modal situations), this modal text flow doesn't really work.  It seems like when you exit the modal text editing state, photoshop "catches up" with several history state events.  This means that we get terribly out of sync.

Also, probably a separate issue: using the text input of the color picker (like, entering "red") does not work very well at all in the modal state.

pineapplespatula:
Photoshop seems to be doing two things oddly. I it is setting two events for each change we make. ( i'm not positivie about this one) and the other is ading a NoOp "Edit Type Layer" history state when we exit the modal state (if a change has been made)  I'm not seeing any issues with the color text input, can you be more specific.

mcilroyc:
Merging this because it prevents a new bogus history state from being captured with every slightest change while in the modal text state (such as arrow key navigation).  

But the problem still remains that making several (>1) color changes while in the modal text state will cause history store to *reset* upon text commit.  This seems to be due to the fact that historyState event notifications from Ps are "batched up" until the modal state is committed.  This causes our history store's internal state to get too far ahead of Ps, and it is not built to withstand the batch of historyState events that it receives some time in the future.

cc @iwehrman as an FYI

iwehrman:
Has somebody filed a watson bug for the problem with history states blocking while the modal tool state is active? This sounds like something that is important to fix for M3. CC @chadrolfs

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4028040

chadrolfs:
referenced in #1848

iwehrman:
Thx bro :facepunch:

PR #1855, opened 2015-07-29
Drag and drop graphic asset from libraries panel to canvas

shaoshing:
This PR implements the drag-and-drop of graphic asset from libraries panel to canvas. Works as below:

lib-dnd
( For now, you have to double click the dropped asset to quit the transform mode, before you can do anything else )

@volfied please review

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -317,35 +317,48 @@ define(function (require, exports) {
      * Places the selected asset in the document as a cloud linked smart object
      *  - Gets the path to the content from libraries
      *  - Sends the path to Photoshop with a place command
+     *  - Add the new layer to the document store to update the UI.
      *
      * Right now, this only works with image assets, for other types of assets we'll need
      * different actions and handlers.
      *
      * @param {AdobeLibraryElement} element
+     * @param {object} location of canvas
      *
      * @return {Promise}
      */
-    var createLayerFromElement = function (element) {
+    var createLayerFromElement = function (element, location) {
         var appStore = this.flux.store("application"),
             libStore = this.flux.store("library"),
+            uiStore = this.flux.store("ui"),
             currentDocument = appStore.getCurrentDocument(),
-            currentLibrary = libStore.getCurrentLibrary();
+            currentLibrary = libStore.getCurrentLibrary(),
+            pixelRatio = window.devicePixelRatio;
 
         if (!currentDocument || !currentLibrary) {
             return Promise.resolve();
         }
+        
+        location.x = uiStore.zoomWindowToCanvas(location.x) / pixelRatio;
+        location.y = uiStore.zoomWindowToCanvas(location.y) / pixelRatio;
 
-        var docRef = docAdapter.referenceBy.id(currentDocument.id),
-            location = { x: 100, y: 100 },
-            representation = element.getPrimaryRepresentation();
+        return Promise
+                .fromNode(function (cb) {
+                    element.getPrimaryRepresentation().getContentPath(cb);
+                })
+                .bind(this)
+                .then(function (path) {
+                    var docRef = docAdapter.referenceBy.id(currentDocument.id),
+                        placeObj = libraryAdapter.placeElement(docRef, element, path, location);
 
-        return Promise.fromNode(function (cb) {
-            representation.getContentPath(cb);
-        }).then(function (path) {
-            var placeObj = libraryAdapter.placeElement(docRef, element, path, location);
-
-            return descriptor.playObject(placeObj);
-        });
+                    return descriptor.playObject(placeObj);
+                })
+                .then(function () {
+                    return descriptor.getProperty("layer", "layerID");
+                })
+                .then(function (newLayerID) {
+                    this.flux.actions.layers.addLayers(currentDocument, newLayerID);
+                });
     };
     createLayerFromElement.reads = [locks.JS_LIBRARIES, locks.JS_DOC];
     createLayerFromElement.writes = [locks.JS_LIBRARIES, locks.JS_DOC, locks.PS_DOC];
@@ -536,7 +549,6 @@ define(function (require, exports) {
     };
     afterStartup.reads = [locks.JS_LIBRARIES];
     afterStartup.writes = [locks.JS_LIBRARIES];
-
 
     exports.beforeStartup = beforeStartup;
     exports.afterStartup = afterStartup;
--- a/src/js/jsx/Scrim.jsx
+++ b/src/js/jsx/Scrim.jsx
@@ -28,14 +28,17 @@ define(function (require, exports, module) {
     var React = require("react"),
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
-        StoreWatchMixin = Fluxxor.StoreWatchMixin;
+        StoreWatchMixin = Fluxxor.StoreWatchMixin,
+        classnames = require("classnames"),
+        Promise = require("bluebird");
 
     var adapterOS = require("adapter/os");
 
-    var PolicyOverlay = require("jsx!js/jsx/tools/PolicyOverlay");
-
+    var PolicyOverlay = require("jsx!js/jsx/tools/PolicyOverlay"),
+        Droppable = require("jsx!js/jsx/shared/Droppable");
+    
     var Scrim = React.createClass({
-        mixins: [FluxMixin, StoreWatchMixin("tool", "ui", "application", "preferences")],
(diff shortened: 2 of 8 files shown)


src/js/jsx/Scrim.jsx
```diff
@@ -28,14 +28,25 @@ define(function (require, exports, module) {
...
+        Droppable = require("jsx!js/jsx/shared/Droppable");
+
+    /**
+     * Indicates whether the mouse is currently within the view port of the Scrim component. It is shared between the 
+     * Scrim component and the "canDropGraphic" callback. 
+     * @private
+     * 
+     * FIXME need a better way to share information between component and Droppable settings.
+     */
+    var _isMouseInsideScrim = false;
```
shaoshing: For the current drag-and-drop implementation, this seems to be the only way to share information between component and droppable [callbacks](https://github.com/adobe-photoshop/spaces-design/pull/1855/files#diff-768f12e2b3c6382be0ac86fe532542feR323), if you don't want to get store involved.

src/js/jsx/Scrim.jsx
```diff
@@ -28,14 +28,25 @@ define(function (require, exports, module) {
...
+    
     var Scrim = React.createClass({
-        mixins: [FluxMixin, StoreWatchMixin("tool", "ui", "application", "preferences")],
+        mixins: [FluxMixin, StoreWatchMixin("tool", "ui", "application", "preferences", "library", "draganddrop")],
```
volfied: Why does the scrim need to listen to Library store?

src/js/jsx/Scrim.jsx
```diff
@@ -249,21 +287,28 @@ define(function (require, exports, module) {
             var document = this.state.document,
                 disabled = document && document.unsupported,
                 transform = this.state.transform,
+                isDragging = this.state.isDragging,
                 overlays = !disabled && this.state.overlaysEnabled,
                 transformString = this._getTransformString(transform),
-                toolOverlay = (overlays && transform) ? this._renderToolOverlay(transformString) : null,
+                toolOverlay = (overlays && transform && !isDragging) ? this._renderToolOverlay(transformString) : null,
```
volfied: This check to whether render tool overlay is broken across two places. Can you consolidate them in renderToolOverlay function instead?

src/js/jsx/Scrim.jsx
```diff
@@ -274,6 +319,40 @@ define(function (require, exports, module) {
             );
         }
     });
-
-    module.exports = Scrim;
+    
+    var canDropGraphic = function () {
```
volfied: JSDoc for this please, also, let's keep this function definition near top of the file, prepend `_` and mark it private in JSDoc.

src/js/jsx/Scrim.jsx
```diff
@@ -274,6 +319,40 @@ define(function (require, exports, module) {
...
+            return {
+                // Don't allow dropping if the position of the draged targets is not inside the Scrim component.
+                valid: _isMouseInsideScrim,
+                compatible: true
+            };
+        },
+        handleDropGraphic = function () {
```
volfied: JSDoc for all functions here, you can break them apart with `var`s

src/js/jsx/sections/libraries/Library.jsx
```diff
@@ -28,10 +28,11 @@ define(function (require, exports, module) {
 
     var os = require("adapter/os");
 
-    var Image = require("jsx!./assets/Image"),
+    var Graphic = require("jsx!./assets/Graphic"),
```
volfied: good call.

src/js/jsx/Scrim.jsx
```diff
@@ -198,6 +230,11 @@ define(function (require, exports, module) {
             if (!nextState.activeDocumentInitialized || !nextState.recentFilesInitialized) {
                 return false;
             }
+            
+            if (this.state.isDragging === nextState.isDragging &&
```
volfied: There is a bug here. If I'm not dragging at all... but transform changes, this is going to prevent us from re-rendering the scrim with correct transform.

volfied: Overall, for shouldCOmponentUpdate functions, it's better to calculate when you *should* render, instead of when you shouldn't.

src/js/jsx/sections/libraries/assets/Graphic.jsx
```diff
@@ -54,17 +57,51 @@ define(function (require, exports, module) {
...
+            this.getFlux().actions.libraries.createLayerFromElement(this.props.element, { x: 100, y: 100 });
+        },
+        
+        /**
+         * Create preview of the dragged graphic asset under the cursor.
+         * @private
+         * @return {ReactComponent?}
+         */
+        _renderDragPreview: function () {
+            if (!this.props.dragPosition) {
```
volfied: Right now, when I let go of the mouse, the drag preview stays on screen. Can you fix that?

src/style/sections/libraries/libraries-section.less
```diff
@@ -77,3 +77,25 @@
         }
     }
 }
+
+.assets__graphic_dragging {
+    opacity: 0.7;
+}
+
+@preview-size: 64px;
```
volfied: This constant should be at the top of the file.

src/js/jsx/Scrim.jsx
shaoshing: that's the leftover from the previous dnd code, removing it now.

src/js/jsx/Scrim.jsx
shaoshing: Good idea and excellent catch! I will change it.

src/js/actions/libraries.js
```diff
@@ -335,17 +337,20 @@ define(function (require, exports) {
...
+                })
+                .bind(this)
+                .then(function (path) {
+                    var docRef = docAdapter.referenceBy.id(currentDocument.id),
+                        placeObj = libraryAdapter.placeElement(docRef, element, path, location);
+
+                    return descriptor.playObject(placeObj);
+                })
+                .then(function () {
+                    this.flux.actions.documents.updateDocument();
```
volfied: Is there a way we can do this without an updateDocument call? Can we get the layer and pass this to addLayer somehow?

iwehrman: Yes, this is very important. `updateDocument` is extremely slow in large documents. If we need Photoshop to emit an event with the layer IDs and positions of the placed document, we should file a high-priority Watson bug for that.

volfied:
@shaoshing I still see the drag preview hanging on the screen after I drop a graphic. Can you fix that before I merge?

volfied:
Also, as an added challenge, let's do this: After placing, instead of calling updateDocument, you should call to see the ID of the selected layer, and transfer to  `addLayer` action on that layer. We really don't want to add another updateDocument call as it is incredibly slow and expensive.

src/js/jsx/sections/libraries/assets/Graphic.jsx
shaoshing: I found out that PS will halt the JavaScript before the end of the transform mode, so I set a 50ms delay to allow the draganddrop store finish its updating before placing the new layer. Please let me know if there is a better way to deal with this. https://github.com/adobe-photoshop/spaces-design/pull/1855/files#diff-768f12e2b3c6382be0ac86fe532542feR334

shaoshing:
Thanks for reviewing @volfied . I've replace the updateDocument with addLayer as suggested. now it is read for another review, back to you now.

src/js/actions/libraries.js
```diff
@@ -317,35 +317,48 @@ define(function (require, exports) {
...
+                })
+                .bind(this)
+                .then(function (path) {
+                    var docRef = docAdapter.referenceBy.id(currentDocument.id),
+                        placeObj = libraryAdapter.placeElement(docRef, element, path, location);
+
+                    return descriptor.playObject(placeObj);
+                })
+                .then(function () {
+                    return descriptor.getProperty("layer", "layerID");
```
volfied: Can you write a short comment describing what this getProperty does? It's obvious once you know it's pointing at the target layer, but explain that PS logic. Say, in the future, PS changes something and placed layer is no longer the target layer, there will be a bug here and it'll be easier to debug if the next dev sees that comment.

volfied:
The 50 ms delay is not working for me, and I'd like to see a comment about the getProperty call... but I'm going to make those into issues you can fix in your next PR as this has been open for long enough and updateDocument call is gone which was the major blocker for me. (Also I'll be out the next 2-3 days so I don't want this PR to atrophy)

Issue #1843, opened 2015-07-27
Undo failure: drag and drop nested group


chadrolfs:
`Action edit.undo failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Select” is not currently available.`

https://www.dropbox.com/s/znpizclvk9z7wdk/Undo-Fail.mov?dl=0

chadrolfs:
occurs in prerelease build m.177 as well as current pg-dev.412 and master @ 6a2880c

iwehrman:
Actually, this might be for @mcilroyc instead.

chadrolfs:
this may or may not be a prerelease blocker.  doesn't seem like anything bad happens afterwards.  thoughts?

mcilroyc:
@chadrolfs : good news and bad news.  The good news is that I think this is easily addressed by the above noted PR (still open as of right now).  The *bad* news is that this issue is much worse than indicated by the title of the issue.  I think ALL layer reordering is busted (vis a vis history) in that undo doesn't actually work on the first try (it just changes the selection state).

PR #1856, opened 2015-07-29
Post-reorder bounds reset should be non-history-changing

mcilroyc:
This addresses #1843 
@iwehrman: you break it, you bought it: 9d285585eb80f49b49703de96013441742e359e1
cc @shaoshing just as an FYI since you recently ported this particular code from the LayersPanel to the action.

The resetBounds action is an accident waiting to happen.  I will attempt to make this less error-prone during the next round of history changes.  In the meantime, the *default* behavior is that resetBounds is a *history changing* event.  It is generally used as a generic "final step" in several other history-changing workflows.  But, in cases like this, it must explicitly be told *not* to change history.  I agree this is confusing.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1137,7 +1137,7 @@ define(function (require, exports) {
             .then(function () {
                 // The selected layers may have changed after the reorder.
                 var nextDocument = this.flux.store("document").getDocument(document.id);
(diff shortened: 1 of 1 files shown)


iwehrman:
image

shaoshing:
:+1:

PR #1861, opened 2015-07-29
Post-reorder bounds reset should be non-history-changing - hotfix on release branch

mcilroyc:
This is analogous to PR #1856
fixing #1843 on release branch


iwehrman:
OK!

Issue #1551, opened 2015-06-12
Cannot paste into search field


DesignSpaceBot:
-- @chadrolfs at [2015-06-09T20:59:51Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1551)

volfied:
With #1824, I'm able to c/p into text field, will wait til we merge to confirm

PR #1864, opened 2015-07-29
Allow copy, cut, paste, and select all in search bar

hamburge:
Addresses issue #1551 and @placegraphichere's request to be able to select all (cmd-A) in the search bar.

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -131,88 +131,91 @@ define(function (require, exports, module) {
      * @param {boolean} hasPreviousHistoryState is there a previous history state in the list
      * @param {boolean} hasNextHistoryState is there a next history state in the list
      * @param {boolean} appIsModal is there a global modal that should disable most menu items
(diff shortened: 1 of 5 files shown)


src/js/models/menubar.js
```diff
@@ -131,88 +131,90 @@ define(function (require, exports, module) {
...
-                    document.layers.hasLinkedSmartObjects,
-                "multiple-documents":
-                    Object.keys(openDocuments).length > 1,
-                "earlier-history":
-                    (document !== null) && hasPreviousHistoryState,
-                "later-history":
-                    (document !== null) && hasNextHistoryState
-            };
-        }
+        } return {
```
volfied: woah, what?

hamburge: Sorry missed a new line. I just removed the else and untabbed everything.

volfied: Ok, I get it... but this looks really bad when it's on the same line with the ending of if

PR #1865, opened 2015-07-30
Restyling of search bar

hamburge:
Search bar styling to match the new designs. There are some missing svgs--rectangles currently replace all of the missing item type svgs, the circle on the right will be an X to clear the inputted value, and the magnifying glass will be a different version that is flipped the other direction.

Other changes:
- Show all possible filters, not just if they have corresponding options available. For example, if there are no Smart Objects in the document, still show the Smart Objects filter option.
- Fix broken datalist updating- the filter icon would not always show until the render after selecting the filter.

The designs:
design-space-search-001
design-space-search-002
design-space-search-003
application-search-002
application-search-005

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -35,13 +35,7 @@ define(function (require, exports, module) {
     var svgUtil = require("js/util/svg"),
         strings = require("i18n!nls/strings");
 
-    /**
-     * Strings for labeling search options
-     *  
-     * @const
-     * @type {object} 
-    */
-    var CATEGORIES = strings.SEARCH.CATEGORIES;
+    var Button = require("jsx!js/jsx/shared/Button");
 
     var SearchBar = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("search")],
@@ -75,7 +69,7 @@ define(function (require, exports, module) {
             return {
                 dismissDialog: _.identity,
                 searchID: "",
-                maxOptions: 14
+                maxOptions: 30
             };
         },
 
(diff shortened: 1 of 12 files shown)


iwehrman:
The search bar does not seem to be centered in the non-UI bounds:
center

iwehrman:
The inner hairline border is too light/bright compared to the comp:
border

iwehrman:
All the filter icons are missing:
filter

hamburge:
I fixed the colors and icons, but I'm having trouble centering it based on the new layout. Any tips @designedbyscience?

Also for now I changed the maximum number of options back to a smaller number, because there's an issue with scrolling, so I'll put that in another PR.

iwehrman:
This probably doesn't help, but it looks like the inner div is bleeding off the left edge of the correctly centered dialog element:
dialog

iwehrman:
Also, the clear-search icon is currently a circle. Are you in need of an X icon? How about `&times;`?

hamburge:
-Fixed scrolling so now we can have more options
-Replaced the circle with an X to clear (shoutout to unicode and @iwehrman)
-Placeholder text now tells you what you are searching for if you have a filter selected
-If your item doesn't have a name it now places "Untitled" in italics as its name
-Adjusted styling for filter options after chatting with Seth

Changes abound!

iwehrman:
But the centering!?!

src/js/stores/search.js
```diff
@@ -231,18 +239,26 @@ define(function (require, exports, module) {
...
+                        style;
                     // Don't show the path info if it is just the same as the item name 
-                    if (item.name === newPathInfo) {
+                    if (name === newPathInfo) {
                         newPathInfo = "";
                     }
 
+                    if (name === "") {
+                        name = "Untitled";
+                        style = { "font-style": "italic" };
```
iwehrman: This should be `fontStyle`: https://facebook.github.io/react/tips/inline-styles.html

src/style/search/search-bar.less
```diff
@@ -85,38 +98,68 @@
...
+.search-bar__dialog .select__option__selected {
+    color: @search-highlight;
+    padding-left: .5rem;
+    border-left: .3rem solid @search-highlight;
+    background-color: @search-dropdown-bg;
+}
+
+.search-bar__dialog .select__header {
+    color: @search-border;
+    border-top: 1px solid @search-border;
```
iwehrman: Also, you can't use pixel constants in the CSS. Everything has to be in rems.

src/js/stores/search.js
```diff
@@ -200,16 +209,15 @@ define(function (require, exports, module) {
...
-                    
-                    search.searchFilters.push(filters);
-                    return items.push(options);
-                }.bind(this), Immutable.List());
+                var filterItems = this._getFilterOptions(types),
+                    searchItems = _.reduce(types, function (items, type) {
+                        var options = this._getOptions(type);
+
+                        return items.push(options);
+                    }.bind(this), Immutable.List());
```
iwehrman: It would be better to build this with an array and then convert the complete array to an immutable list at the end.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -379,17 +330,22 @@ define(function (require, exports, module) {
...
                         useAutofill={true}
                         onChange={this._handleChange}
-                        onKeyDown={this._handleKeyDown}
-                    />
+                        onKeyDown={this._handleKeyDown} />
+                    <Button
+                        title="Clear Search Input"
+                        className="button-clear-search"
+                        onClick={this._clearInput}
+                       >
```
iwehrman: This delimiter should go back up at the end of the previous line.

iwehrman:
I give this PR :hamburger: :hamburger: :hamburger: :hamburger: :hamburger:

PR #1866, opened 2015-07-30
Render the visible subset of layers in the panel based on scroll position

iwehrman:
This changes how the layers panel renders the list of layers so that only a window of layers outside the visible area are rendered at once. When the selection changes, or when the scroll position changes, the set of rendered layers changes accordingly. For documents with many (visible) layers, this improves startup time significantly (~30%). It also seems to improve layer selection time from the panel somewhat.

But, there are also two known performance downsides. First, direct on-canvas selection can be slower because it can cause the layers panel to jump far away from it's current scroll position, which can require all the layers in the panel to be re-rendered. Second, scroll is jittery in some instances, especially when scrolling fast, because the necessary DOM changes require a layout, which is synchronous.

The good news is that **both of these slow-downs are significantly less bad in compiled mode** than in debug mode. I don't necessarily know if that makes this PR a net improvement overall. I'm looking for some very honest feedback here. Of course, I would also be very happy to hear suggestions about how to make this faster. For the moment, I'm out of ideas about that. I'm especially looking for feedback from @designedbyscience and @shaoshing, but I would appreciate testing and feedback from everyone! I suggest testing either with the VermilionArtboards document (with all descendants expanded) or with [this helpful document that contains ~1000 layers](http://cloud.wehrman.org/3f2N0T3n1L1h).

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -30,8 +30,7 @@ define(function (require, exports, module) {
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         classnames = require("classnames"),
-        Immutable = require("immutable"),
-        _ = require("lodash");
+        Immutable = require("immutable");
 
     var Draggable = require("jsx!js/jsx/shared/Draggable"),
         Droppable = require("jsx!js/jsx/shared/Droppable"),
@@ -57,6 +56,12 @@ define(function (require, exports, module) {
             this.props.dragPosition !== nextProps.dragPosition ||
             this.props.dragStyle !== nextProps.dragStyle ||
             this.props.isDropTarget !== nextProps.isDropTarget) {
+            return true;
+        }
+
+        // Margin change
+        if (this.props.marginTop !== nextProps.marginTop ||
+            this.props.marginBottom !== nextProps.marginBottom) {
             return true;
         }
         
@@ -327,9 +332,9 @@ define(function (require, exports, module) {
 
             // Super Hack: If two tooltip regions are flush and have the same title,
             // the plugin does not invalidate the tooltip when moving the mouse from
-            // one region to the other. This is used to make the titles to be different,
+            // one region to the other. This is used to make adjancent titles different,
             // and hence to force the tooltip to be invalidated.
-            var tooltipPadding = _.repeat("\u200b", layerIndex),
+            var tooltipPadding = (this.props.visibleLayerIndex % 2) === 1 ? "\u200b" : "",
                 tooltipTitle = layer.isArtboard ? strings.LAYER_KIND.ARTBOARD : strings.LAYER_KIND[layer.kind],
                 iconID = svgUtil.getSVGClassFromLayer(layer),
                 showHideButton = layer.isBackground ? null : (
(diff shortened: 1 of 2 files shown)


iwehrman:
Another experiment that I could perform: instead lazily mount the layer faces as needed for scrolling and group expansion, but always keep them in the DOM afterwards but with a `visible` prop that immediately cuts off `shouldComponentUpdate` in `LayerFace` if the layer is hidden. Pros and cons: 

1. The first scroll through the layers would be equally as jittery, but presumably after the faces are mounted the resulting DOM updates will be much faster, resulting in less (maybe?) jitter. On the other hand, it might still be just as jittery because there could be far more layers to (quickly?) update. 
2. The startup speedup would be just as good.
3. Somewhat more complicated logic. (And it's already pretty complicated.)

shaoshing:
Hey @iwehrman , :+1:  terrific work! I tested with the 1000-layers file and the before and after difference is significant. 

Could you explain a bit more about how you control whether to render a LayerFace? I added some `console.info` in its shouldComponentUpdate and render functions, and found out that the call times are not equal, so I guess you must added something else to control this. 

Like you mentioned, the afterward scrolling is slower than the previous implementation, and I spotted a lot of re-rendering while scrolling, even if they were already rendered. I guess it is related to the newly added `marginTop` and `marginBottom` props. Maybe it is worth of investigation as well.

screen shot 2015-07-31 at 9 33 44 am

I also looked up the initial rendering of the previous implementation, and the several `_setStateFromFlux` calls caught my attention. I haven't dig any deeper on these, but it might be a potential problem of the slow initialization. Maybe you already check them out? Do you have any interesting results?

screen shot 2015-07-31 at 9 14 22 am

iwehrman:
This is sort of a long story. Virtual rendering of mounted layer faces, like for all React components, is controlled by `shouldComponentUpdate`. We've done our best to make that method fast and cheap, but it isn't free. Our hypothesis going in was that re-rendering the layers panel was slow because of the _virtual_ rendering step, which requires virtual rendering of ~1000 child layer faces, which requires at least calling `shouldComponentUpdate` ~1000 times. Even if most of those don't result in a DOM rendering step, it seemed that with enough child faces the virtual rendering was still slowing down. So, the experiment here is to see if we can do better by not keeping all the face components mounted so that each virtual rendering step only needs to render at most ~100 faces. The problem is that this requires a lot of DOM manipulation during scroll, because we have to completely unmount or remount from scratch layer faces to cover the visible window plus padding. So, once a face has been unmounted, re-mounting it is fairly expensive, requires a DOM render, and can't be helped by `shouldComponentUpdate`. And unmounting is probably fairly expensive itself.

You're seeing a lot of `setStateFromFlux` calls because the layers panel re-renders during scroll as a result of `setState` calls that result from the scroll event handler, and the `StoreWatchMixin` always calls `getStateFromFlux` before each update. I don't think that in particular is the expensive thing (it's just a few synchronous, DOM-free store queries), but I could be wrong. Instead, I think scrolling is jittery because of all the DOM-manipulation that happens in the scroll handler.

iwehrman:
Re: the experiment [mentioned here](https://github.com/adobe-photoshop/spaces-design/pull/1866#issuecomment-126563626), I tried this out and it doesn't seem worth pursuing. Hiding and unhiding the existing elements in the DOM seems to cause just as much scroll jank as unmounting and remounting them. I'm kind of at a loss about what to do here. @designedbyscience: it seemed like when you did this experiment, you felt like scroll performance wasn't an issue. Do you have any ideas about what I'm doing differently/wrong here?

designedbyscience:
Instead of margin, I had used height/padding in my version. Does that change anything: 

``             var styleObj = {
                    height: (doc.layers.allVisible.size*height) + "px",
                    "paddingTop": (offsetLayers * height) + "px"
                };``

iwehrman:
Using `height/paddingTop` does not seem to change performance compared to `marginTop/marginBottom`, at least not that I can detect. 

I still don't think the jerkiness is all that bad in compiled mode. Load time is improved somewhat (maybe by a couple hundred ms for Vermilion in compiled mode), but I can't really convince myself that this actually improves the performance of `layers.select` compared to master. Frankly both branches seem to perform pretty well these days in compiled mode. Maybe we should instead consider just a hack to improve the performance of the very first render?

iwehrman:
I officially give up on this one.

Issue #1859, opened 2015-07-29
CC Lib: importing asset does not update layers panel


chadrolfs:
new doc, load a library and add an asset to canvas by clicking on (+), hit Enter to place

Result: layers panel is not updated 
Expect: layer panel to update reflecting newly added asset (and it should be focused/selected)

If you attempt to move the artboard, you get 
`Action layers.getLayerOrder failed: Error: Too many layers in layer index`

shaoshing:
The + button is working but it is placing at the top left corner. I can make it to place at the center of the view port, but I am wondering with the drag-n-drop, do we really need the click-n-add function? It is not as natural as drag-n-drop, and is not more efficient either (you can not batch add given that you have to quit the transform mode first). @volfied @chadrolfs  what do you think?

volfied:
Let's definitely get @davealonzo's input, but as far as the designs show, there is no click to apply for Graphic assets, it should be drag&drop.

chadrolfs:
Agreed with @volfied that we can probably do away with the (+) and just do drag&drop but the final call is @davealonzo's

shaoshing:
With the implementation of drag and drop graphic asset, I believe we can safely close this one. Any further discussion on whether to support mouse click event will be in a separate issue.

PR #1891, opened 2015-07-31
Start on adding libraries to search bar

hamburge:
If the library item is a graphic, it will be placed on the document. If it is a character/layer style it will be applied if text/a layer is selected. Doesn't show colors because colors can't be placed in the document yet.

When placing a graphic you have to hit "enter" after selecting it to confirm it and close the dialog.

Can filter by library name or by type of library asset (but not both at the moment).

@shaoshing, could you review? Is there anything you are working on that will directly affect this in the near future?


src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,129 @@
...
+define(function (require, exports) {
+    "use strict";
+
+    var Immutable = require("immutable");
+
+    var events = require("js/events"),
+        collection = require("js/util/collection");
+
+    // Map from element IDs to corresponding element
+    var _idMap = {};
```
shaoshing: Can you match the jsdoc style comment and add `@private` here?

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,129 @@
...
+        var libActions = this.flux.actions.libraries,
+            elementInfo = _idMap[id];
+
+        if (elementInfo) {
+            var asset = elementInfo.asset;
+            libActions.selectLibrary(asset.library.id);
+
+            switch (elementInfo.type) {
+                case "IMAGE":
+                    var location = { x: 100, y: 100 };
```
shaoshing: Instead of using a fixed position, can you use the center of the viewport as the new layer position? You can use uiStore#transformWindowToCanvas to do the transformation.

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,129 @@
...
+    };
+
+    /**
+     * Register library info to search
+     *
+     * @param {Array.<AdobeLibraryComposite>} libraries
+     */
+    var registerLibrarySearch = function (libraries) {
+        var libraryNames = libraries.length > 0 ? collection.pluck(libraries, "name") : Immutable.List(),
+            assetTypes = ["LIBRARY", "IMAGE", "COLOR", "LAYERSTYLE", "CHARACTERSTYLE"];
```
shaoshing: The "COLOR" type should be removed if it is not supported.

src/js/util/svg.js
```diff
@@ -61,7 +61,9 @@ define(function (require, exports) {
             return null;
         }
         // currently only have icons for layers
-        if (filter.length < 2 || filter.join(" ").indexOf("LAYER") === -1) {
+        // check for libraries because of "layer styles"
+        if (filter.length < 2 || filter.join(" ").indexOf("LAYER") === -1 ||
+                filter.join(" ").indexOf("LIBRARY") !== -1) {
```
shaoshing: You can simplify here with `String#search(/LAYER|LIBRARY/)`

src/js/actions/search/libraries.js
shaoshing: Can you also use "GRAPHIC", instead of "IMAGE", to keep consistency?

shaoshing:
Great job @hamburge ! I tried the libraries search and it works great! There are few small problems need more tweaking. In the meanwhile, I am wondering if there is any discussion on showing more context info or different icons for layerstyle and graphic asset? I am asking this because when you add a layerstyle or graphic asset, Photoshop will use its layer name as default name, which means the result list might contain different type of assets with identical names.  

screen shot 2015-07-31 at 5 36 39 pm

iwehrman:
Ideally I think the thumbnails themselves would be in the results...

src/js/util/svg.js
hamburge: Not sure that would work, since it's looking for if it does contain "LIBRARY" or does not contain "LAYER"

hamburge:
@shaoshing Thanks for looking at this. I'll work on making the context clearer with extra info text and/or thumbnails.

hamburge:
@shaoshing I made some more changes to this to make it a little smoother. Could you or @volfied review it again when you get a chance?

In regards to adding extra context about what type of asset each item is, with the new svgs I think it is much clearer, and I also put it alongside the library name (mostly so you can use that as an additional search term, since you can't apply multiple filters). @iwehrman I could give the thumbnails a shot, but I don't think they would be big enough if they replaced the svg alongside the item name. Maybe they'd appear on hover?

At this point this branch *mostly* doesn't break. The outstanding issues I can find are related to improving the library actions themselves, like the new layers having `kind` undefined and checking for what type of layers are selected when using an asset type. Should I write up issues for these or are they too low priority at the moment?

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,173 @@
...
+                }
+            });
+            return allMaps.concat(libMap);
+        }, Immutable.List());
+    };
+
+    /*
+     * Find SVG class for library item based on what type of asset it is
+     *
+     * @param {Array.<string>} categories
```
shaoshing: Missing the @private statement.

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,173 @@
...
+        var type = categories[categories.length - 1];
+        switch (type) {
+        case "GRAPHIC":
+            return "libraries-addGraphic";
+        case "CHARACTERSTYLE":
+            return "libraries-addCharStyle";
+        case "LAYERSTYLE":
+            return "libraries-addLayerStyle";
+        default:
+            return "libraries-CC";
```
shaoshing: How about using a map data here, to eliminate the use of `switch`? For example:

```js
var CATEGORY_MAP_ICON = {
   "GRAPHIC": "libraries-addGraphic"
};

return CATEGORY_MAP_ICON[type] || "libraries-CC"
```

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,173 @@
...
+            libActions.selectLibrary(asset.library.id);
+
+            switch (elementInfo.type) {
+                case "GRAPHIC":
+                    var uiStore = this.flux.stores.ui,
+                        centerOffsets = uiStore.getState().centerOffsets,
+                        midX = (window.document.body.clientWidth + centerOffsets.left - centerOffsets.right) / 2,
+                        midY = (window.document.body.clientHeight + centerOffsets.top - centerOffsets.bottom) / 2,
+                        location = uiStore.transformWindowToCanvas(midX, midY);
+                    libActions.createLayerFromElement(asset, location);
```
shaoshing: A line break above might improve readability.

src/js/actions/search/libraries.js
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Register library info to search
+     *
+     * @param {Array.<AdobeLibraryComposite>} libraries
+     */
+    var registerLibrarySearch = function (libraries) {
+        var libraryNames = libraries.length > 0 ? collection.pluck(libraries, "name") : Immutable.List(),
+            assetTypes = ["LIBRARY", "GRAPHIC", "LAYERSTYLE", "CHARACTERSTYLE"];
+        
+        var filters = Immutable.List(libraryNames.concat(assetTypes));
```
shaoshing: `concat` will return Immutable result, so the constructor here is not required.

src/js/jsx/Search.jsx
```diff
@@ -64,10 +64,9 @@ define(function (require, exports, module) {
          * @param {string} itemID ID of selected option
```
shaoshing: This is a @private function too.

shaoshing:
Hi, @hamburge the changes look great. I love the asset icons and the brief context; now the search result is much more clear. I added some more comments, and you might want to rebase from master as well, since we have updated all the actions regarding the lock mechanism. So, back to you.

hamburge:
@shaoshing Thanks, I fixed up the things you mentioned. I'm thinking that this branch is actually in good enough shape that I can remove the "Review Only" tag-- do you agree or do you think it should wait until more of the general libraries functionality is in master?

Another thing is that I think issue #1859 (updating the layers panel when you place a graphic) is relevant to this branch

shaoshing:
Hi @hamburge , thanks for the updates, they look great! I agree that the branch has made its goal. There might be user experience issues need discussion (for example, we should disable or filter out the layer/text style in the list if nothing selected), but they can be in separate tickets. Before merging, you might want to do more checking on asset types because I can see brushes and color themes are also within the search result. I can share my testing libraries with you if that would be helpful.

screen shot 2015-08-11 at 8 51 41 am
screen shot 2015-08-11 at 8 51 50 am

hamburge:
@shaoshing Great! I added a check if it is a valid asset type before adding it to search.
Removing "Review Only"!

hamburge:
Oh just realized that with the rebase, one of the icons is missing

hamburge:
@shaoshing Icon is in!

shaoshing:
great work!

Issue #1868, opened 2015-07-31
Search: menu commands: Application should be Photoshop CC


chadrolfs:
open a doc, cmd/ctrl + f, select Menu Commands

Result: Show All   Application > Show All

iwehrman:
FBNC to @chadrolfs.

PR #1893, opened 2015-08-01
In search for menu commands, replace 'Application' with 'Photoshop CC…

hamburge:
…' on Mac

For issue #1868

### DIFF
--- a/src/nls/root/strings.js
+++ b/src/nls/root/strings.js
@@ -65,7 +65,7 @@ define(function (require, exports, module) {
         APP_NAME: "Photoshop",
         MENU: {
             APPLICATION: {
-                $MENU: "Application",
+                $MENU: "Photoshop CC",
                 ABOUT_MAC: "About Photoshop…",
                 PREFERENCES: "Preferences…",
                 HIDE_APPLICATION: "Hide Photoshop",


src/js/actions/search/menucommands.js
```diff
@@ -53,7 +53,16 @@ define(function (require, exports) {
             }
 
             if (labels[part].$MENU) {
-                path += labels[part].$MENU + ">";
+                // Replace "Application" with "Photoshop CC" on Macs
+                // Don't show "Application" options on Windows
+                if (part !== "APPLICATION") {
+                    path += labels[part].$MENU + ">";
+                } else if (system.isMac) {
+                    path += "Photoshop CC >";
```
iwehrman: At the very least this needs to be localized. But, what happens if you just change this line? https://github.com/adobe-photoshop/spaces-design/blob/master/src/nls/root/strings.js#L68

hamburge:
@iwehrman Yep, that did it. I can't test on Windows though...

src/js/actions/search/menucommands.js
```diff
@@ -53,7 +53,13 @@ define(function (require, exports) {
             }
 
             if (labels[part].$MENU) {
-                path += labels[part].$MENU + ">";
+                // Don't show "Application" options on Windows
```
iwehrman: I don't think any of this is necessary? There is no "application" menu on Windows...

iwehrman: Which is to say that it seems like the one-line diff to the strings file is all that should be needed?

hamburge:
@iwehrman Ok reverted it back to what we had before except for the strings file

iwehrman:
:+1:

PR #1894, opened 2015-08-01
Libraries - drag and drop layer

shaoshing:
Now you can drag and drop layers into the libraries panel to create graphic asset.

Here are few things worth of mention:

* Adding a single linked layer is prohibited, but multiple or mixed multiple linked layers are accepted (same as standard PS)
* Asset + buttons are removed for now. More discussion is at #1859 
* Dragging a layer down to the libraries panel might cause the libraries panel's scroll event unexpectedly. 

lib-layer-dnd

### DIFF
--- a/src/js/jsx/sections/libraries/LibrariesPanel.jsx
+++ b/src/js/jsx/sections/libraries/LibrariesPanel.jsx
@@ -29,7 +29,9 @@ define(function (require, exports, module) {
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
-        classnames = require("classnames");
+        classnames = require("classnames"),
+        Promise = require("bluebird"),
+        Immutable = require("immutable");
 
     var TitleHeader = require("jsx!js/jsx/shared/TitleHeader"),
         LibraryList = require("jsx!./LibraryList"),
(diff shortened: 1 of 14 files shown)


src/js/jsx/sections/libraries/LibrariesPanel.jsx
```diff
@@ -167,26 +180,92 @@ define(function (require, exports, module) {
             var sectionClasses = classnames({
                 "libraries": true,
                 "section": true,
-                "section__collapsed": !this.props.visible
+                "section__collapsed": !this.props.visible,
+                "libraries_no-drop": this.state.isDropTarget && !this.state.selectedLibrary
```
volfied: Should probably also check for connection status here to disallow dropping.

shaoshing: `selectedLibrary` will always be null when no connection, so it should be enough of checking here.

src/js/jsx/sections/libraries/LibrariesPanel.jsx
```diff
@@ -106,47 +117,49 @@ define(function (require, exports, module) {
...
 
             if (connected) {
-                var libraries = this.state.libraries,
-                    currentLibrary = libraries.get(this.state.selectedLibrary);
-
-                containerContents = this.props.visible && !this.props.disabled && (
-                    <div>
-                        <div className="formline">
-                            <LibraryList
+                if (this.props.visible && !this.props.disabled) {
```
volfied: Can these two if lines be merged?

volfied:
Just one nit, and I will merge this, opening a more general problem as an issue. Linked layers should not all be banned because we want users to be able to change their local linked smart objects to cloud linked smart objects, but you can fix that in an issue.

Back to you @shaoshing

src/js/jsx/sections/libraries/LibrariesPanel.jsx
shaoshing: I gave it a look and turns out you have to keep both, because the panel might be visible while it is disabled for unsupported document.

volfied: What do we render in those cases?

shaoshing: Currently we show the Libraries panel with only its title

volfied:
Alright then.

Issue #1892, opened 2015-07-31
Search bar should be centered in the viewport


hamburge:
Currently it is only centered when both toolbar panels are open.

iwehrman:
Help, @designedbyscience!

designedbyscience:
Fixed by #1898

PR #1898, opened 2015-08-03
Center the search bar, no matter how many panels are open

hamburge:
Changes the class of main based on how many panels are open. Left margin of search bar changes based on that class.

@designedbyscience Is this along the lines of what you were suggesting?

issue #1892

### DIFF
--- a/src/style/search/search-bar.less
+++ b/src/style/search/search-bar.less
@@ -30,9 +30,8 @@
     background-size: 1.2rem;
     background-position: 1.9rem 1.8rem;
     top: 25%;
-    left: 25%;
+    left: 50%;
     transform: none;
-    margin-left: -(@properties-width - @toolbar-width)/2;
 
     width: @search-bar-width;
     min-height: 5rem;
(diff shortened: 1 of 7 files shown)


src/style/search/search-bar.less
```diff
@@ -44,6 +43,19 @@
     overflow: visible;
 }
 
+.main__no-panel .search-bar__dialog {
+    margin-left: -@search-bar-width / 2;
+}
+
+
+.main__one-panel .search-bar__dialog {
+    margin-left: -(@properties-width / 2)  - (@search-bar-width / 2);
```
iwehrman: Isn't `@properties-width` going to vary at runtime?

designedbyscience: No, it is a CSS variable (it is the width of one panel...maybe we should rename that)

designedbyscience: Also, extra line

designedbyscience:
Yep, this is what I was thinking. Could you go rename `@properties-width` to `@panel-column-width`? One other nit about lines

hamburge:
@designedbyscience Great! Made those changes.

designedbyscience:
@hamburge In the 2 column case, I think you are 0.5rem off center (it calculates -50rem but it should be -50.5rem;

hamburge:
@designedbyscience Got it, subtracted .5rem-- doesn't seem to be stored anywhere as a CSS variable.

hamburge:
Not ready yet-- just realized that a rogue commit got in here...

hamburge:
Ok removed the unrelated commit

designedbyscience:
:guitar: on

Issue #1888, opened 2015-07-31
Can't search unless a document is open


iwehrman:
I want to use the search bar to open recent files!

placegraphichere:
When there are no open docs, I might suggest the default view once you click into the app search would be recent files, instead of filters. Auto fill for filters would still work of course

iwehrman:
Fixed by #1900.

chadrolfs:
can't search in zero doc state

volfied:
@chadrolfs please confirm the fix.

PR #1900, opened 2015-08-03
Search for recent docs and menu commands when there are no documents

hamburge:
You can now open the search bar when there is no document and search recent documents and menu commands to your heart's content. (issue #1888)

I added an optional property to the registered searches in the search store specifying if a category will have any options when there is no document. Then we can ignore those filters when rendering the drop down, so we don't have a bunch of filter options for layers and current documents when there is no open document.

### DIFF
--- a/src/js/actions/search/layers.js
+++ b/src/js/actions/search/layers.js
@@ -89,9 +89,10 @@ define(function (require, exports) {
     var _getLayerOptionsFromDoc = function (document, type) {
         var appStore = this.flux.store("application"),
             currentID = appStore.getCurrentDocumentID(),
-            layers = document.layers.allVisibleReversed;
+            layers = document.layers.allVisibleReversed,
+            layerItems = [];
         
-        return layers.map(function (layer) {
+        layers.forEach(function (layer) {
             var ancestry,
                 layerType = _getLayerCategory(layer, type),
                 iconID = svgUtil.getSVGClassFromLayer(layer);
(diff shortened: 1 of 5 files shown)


src/js/actions/search/layers.js
```diff
@@ -136,17 +136,20 @@ define(function (require, exports) {
         // Get list of layers
         var appStore = this.flux.store("application"),
             currDoc = appStore.getCurrentDocument(),
-            documents = appStore.getOpenDocuments(),
+            allLayerMaps = Immutable.List();
```
iwehrman: This seems to create a bunch of unnecessary garbage Immutable objects in the process of building one final Immutable list. Can you find a way to do it that creates as little unnecessary garbage as possible? Like, for example by using a single array as an intermediate data structure and then converting it to an immutable list at the end?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -180,11 +180,18 @@ define(function (require, exports, module) {
 
                     var title = option.title.toLowerCase(),
                         category = option.category || [];
+ 
+                    if (option.id.indexOf("FILTER") === 0) {
```
iwehrman: I know this isn't new, but we should figure out a better way to distinguish between filter and regular options than with a string comparison. What about if filter options just have a true `filter` property?

hamburge: I changed filters to have type of "filter" instead of "item," which gets rid of this string comparison, but I don't think there's a simple workaround for `_handleChange` in `SearchBar.jsx`. We could look up the object by ID in the search options, but the string comparison seems more efficient.
https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/jsx/search/SearchBar.jsx#L110

hamburge:
@iwehrman I cleaned up a few Immutable loops that I found throughout the search code. :ballot_box_with_check: :it: :outbox_tray:

iwehrman:
OK!

Issue #1830, opened 2015-07-24
The first doc opened after switching to DS not centered


ktaki:
Switch to DS and open a new doc. The doc is not centered in the open canvas area and often partially behind the property panel. The second doc shifts a bit, but  still not centered. The third doc on, they are centered correctly.

iwehrman:
Possibly related to #1812 and/or #1709.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #1901, opened 2015-08-04
Center document when the panel is resized

shaoshing:
The previous implementation only center the selected document once after startup, leaves the other documents shifted when opening multiple document. This PR will make sure document is centered when the panel is resize (including initial startup), which is the same behavior in standard PS.

Fix issue: #1830

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -482,25 +482,13 @@ define(function (require, exports) {
     beforeStartup.modal = true;
 
     /**
-     * Center the document after startup.
+     * Handle afterStartup event. 
      *
      * @private
-     * @param {boolean} reset Indicates whether this is being called as part of a reset
      * @return {Promise}
      */
-    var afterStartup = function (reset) {
-        var document = this.flux.store("application").getCurrentDocument();
-
-        if (document && !reset) {
-            // Flag sets whether to zoom to fit app window or not
-            return this.transfer(updateTransform)
-                .bind(this)
-                .then(function () {
-                    return this.transfer(centerBounds, document.bounds, false);
-                });
-        } else {
-            return Promise.resolve();
-        }
+    var afterStartup = function () {
+        return Promise.resolve();
     };
     afterStartup.reads = [locks.PS_APP, locks.JS_DOC, locks.JS_APP, locks.JS_TOOL];
     afterStartup.writes = [locks.JS_UI, locks.PS_APP, locks.JS_POLICY];
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -482,19 +482,6 @@ define(function (require, exports) {
     beforeStartup.modal = true;
 
     /**
-     * Handle afterStartup event. 
-     *
-     * @private
-     * @return {Promise}
-     */
-    var afterStartup = function () {
-        return Promise.resolve();
-    };
-    afterStartup.reads = [locks.PS_APP, locks.JS_DOC, locks.JS_APP, locks.JS_TOOL];
-    afterStartup.writes = [locks.JS_UI, locks.PS_APP, locks.JS_POLICY];
-    afterStartup.modal = true;
-
-    /**
      * Remove event handlers.
      *
      * @private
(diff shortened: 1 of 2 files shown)


shaoshing:
doc-centering

designedbyscience:
@shaoshing Does this respect zoom levels? Namely, if the document is zoomed all the way out and centered, great, recenter with panel expand/collapse. But if I'm zoomed in, it should not recenter the document

iwehrman:
I don't think we want to reposition the canvas when the window is resized. Imagine you're zoomed into some particular part of the document and want to expand the window just a little...

shaoshing:
Hi, @designedbyscience @iwehrman, I agree that re-center zoomed in documents are unexpected to the users, so I made some tweak and now it will only center the docs once at the beginning.

designedbyscience:
@shaoshing Behavior is slightly more complicated than that...not sure how critical is it to recenter when zoomed out all the time: (https://www.dropbox.com/s/mksmka3wy87jqr0/zoomed_recentering.mov?dl=0)

Not sure if this is an MVP thing, but it seems useful to me. I am pestering @placegraphichere for his feedback

shaoshing:
@designedbyscience thanks for the video; it does seem more complicated than I thought. However, this PR now only center the document once for its initial display, so it should be safe in terms of breaking user experience.

src/js/actions/ui.js
```diff
@@ -482,25 +482,13 @@ define(function (require, exports) {
...
-            // Flag sets whether to zoom to fit app window or not
-            return this.transfer(updateTransform)
-                .bind(this)
-                .then(function () {
-                    return this.transfer(centerBounds, document.bounds, false);
-                });
-        } else {
-            return Promise.resolve();
-        }
+    var afterStartup = function () {
```
iwehrman: If this doesn't do anything it can just be removed.

shaoshing: Removed in a separate commit, for recovery in the future.

placegraphichere:
So it looks like there are some "rules" being followed in standard, shown in @designedbyscience's video.

1) When full document width is within visible bounds, and maybe >75% of height, recenter the document when expanding columns. 
2) When full document width is within visible bounds, and maybe <75% of height, do NOT recenter the document when expanding columns.
3) When full document width is not visible, do NOT recenter the document when expanding columns.

I bet there is a better way to describe this, but it seems like that's whats going on. I think the basic PR as described by @shaoshing is a good start.

designedbyscience:
Added those rules to a trello card

src/js/stores/ui.js
```diff
@@ -392,8 +415,37 @@ define(function (require, exports, module) {
...
+            });
+        },
+        
+        /**
+         * Center the selected document once.
+         *
+         * @private
+         */
+        _centerCurrentDocumentOnce: function () {
+            var flux = require("js/main").getController().flux,
```
iwehrman: The better way to do this is just `this.flux`. http://fluxxor.com/documentation/stores.html#-store-flux-

iwehrman:
I supplied @shaoshing with fresh basket of nits, so back to him for now.

shaoshing:
Hey @iwehrman , thanks for reviewing and all the feedback. I've made some new changes, please review again.

src/js/stores/ui.js
```diff
@@ -117,6 +119,15 @@ define(function (require, exports, module) {
...
         _toolbarWidth: null,
+        
+        /**
+         * A list of document IDs that are centered after panel resized. If a document's id is not within the list,
+         * it will be centered when it is selected.
+         *
+         * @private
+         * @type {Array.<number>}
+         */
+        _centeredDocumentIDs: null,
```
iwehrman: Why not use a `Set` here instead of an Array? That way you can do constant-time membership checks, additions and removals. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

iwehrman:
Changes look good. Back to @shaoshing for one final question about the data structure in use.

shaoshing:
I agree that it is a perfect scenario for using Set; I've made those changes. Back to you @iwehrman , and thank you for your brilliant inputs.

iwehrman:
Looks great!

Issue #1853, opened 2015-07-28
To reduce mouse travel, always show panel well or add option to pin


chadrolfs:
![screen shot 2015-07-28 at 2 44 19 pm](https://cloud.githubusercontent.com/assets/8442631/8944684/d8966d12-3536-11e5-8d4c-e33e7c609bb0.png)

It's easier to toggle/expand/collapse panel visibility if the well is always there and also more discoverable than the triangles

designedbyscience:
I think we're gonna have it here all the time

placegraphichere:
I think we're gonna do this for other reasons, but yes, agreed

iwehrman:
FBNC to @chadrolfs.

Issue #1368, opened 2015-06-12
Reordering layers has two UI drop zone dividers


DesignSpaceBot:
http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png

-- @chadrolfs at [2015-04-30T18:19:33Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1368)

iwehrman:
Fixed by #1904! Woohoo! I hated this one. @designedbyscience is like :santa: FBNC to @chadrolfs

chadrolfs:
Thanks :santa: !

PR #1904, opened 2015-08-04
Style updates for the layers panel

designedbyscience:
Followed @placegraphichere design. He will be doing additional icon work.

Panel Columns are a little wider than before

Groups and Artboard layers are supposed to have a lighter/easier to find icon.

This does not include warn/error states for smart objects (need to talk to @volfied about those)

I made an adjustment to how the layers/libraries panels interact (a single one open looks better, and both open split the space evenly).

There are also changes to address #1368 and #1853

@placegraphichere Please review this


iwehrman:
I'm not sure what my constructive suggestion here is, but the vertical spacing feels weird at the intersection of the header and the layers panel: 
vert

iwehrman:
I think the names of layers hidden on canvas are way too hard to read due to lack of contrast:
contrast

iwehrman:
Again, I'm not sure what to suggest, but the transition from one vertical panel with a scroll gutter to one without looks weird:
gutter

iwehrman:
All the datalist menus are misaligned:
menu

iwehrman:
Maybe this is on the roadmap, but I think that we should get rid of the little panel-hover arrows to collapse the columns since we've got the buttons. I might still be in favor of a similar panel-hover arrow thing to _swap_ the columns though.

iwehrman:
Color picker is misaligned too:
color

iwehrman:
In the special case of two-panel columns like Layers/Libraries, it seems like collapsing one row panel should always expand the other one: 
expand

iwehrman:
The layers/libraries column bounces a bit when opening and closing the styles/export column: 
bounce

iwehrman:
This isn't your fault, but I think it's time for the ALL-CAPS tooltips to go:
caps

placegraphichere:
For the look of the scrollbar and panel vertical alignment:  Unless we go to a JS library for scrolling, I think we should minimize the width to 0.4rem, and remove the background color of the scrollbar. This looks a bit better. @designedbyscience had concerns the width will not work on WIN, which may be a problem for another day when I can get my VM running.

screen shot 2015-08-06 at 3 17 08 pm

placegraphichere:
Cases where the user tries to close the last possible panel, and could arrive in a state where no panels are open in a column. We'd like to avoid this. One option is to simply close the entire column, however that seems a little unexpected to me. 

Instead, I think we should just not allow the user to close the last open panel. This will help standardize the rules for both the Properties column (3+ panels) and the Layers column (2 panels). 

The down side to just "opening something else instead" would be the Properties column, we'd need to have a preferred "open" panel in a case where the user closes the last panel—and what would we open instead? Instead, in this proposal we're keeping the same panel functionality for both columns, under one simple rule.

Logic might look like:
`Panel Double-Click = Am I the last panel open ? Do nothing : Close this panel;`

placegraphichere:
Icon and hidden layer contrast: we should up the value of the text to @label-secondary-color: #545454;. For the icons, I believe that has been improved. We should also get a better contrast with the icons once the new icons are in place (aka defer).

designedbyscience:
Lots o' updates, addressed above concerns and bugs.

I have left the little black "Hide this column" arrows for the moment

The behavior for the panels collapsing is as Seth described above. We can test the different versions when we have more panels in there

@iwehrman Back to you

iwehrman:
Needs a rebase.

designedbyscience:
@iwehrman rebased

iwehrman:
Is it intentional that the hairlines at the top and the bottom of the libraries panel are different colors?

iwehrman:
Without the underline, the floating down arrow is a little mysterious:
image

iwehrman:
Eh, I just realized I'm complaining about the wrong panel. I'll move those complaints elsewhere. SORRY.

iwehrman:
Actually about layers panel: the layer drag style is too short horizontally. As soon as you start the drag, the layers are reduced in width by ~25%.

iwehrman:
Bad text color in layer rename mode:
image

iwehrman:
Also, is this still the desired text selection color? 
image

iwehrman:
Error when attempting to collapse the libraries panel:
```
Uncaught TypeError: Cannot read property 'parentElement' of null PanelSet.jsx:140
 React.createClass._handlePanelVisibilityToggle react-with-addons.js:3311
 executeDispatch react-with-addons.js:17428
 SimpleEventPlugin.executeDispatch react-with-addons.js:3299
...
```

iwehrman:
Look who I found!
image

designedbyscience:
You win the prize!

iwehrman:
Also, something is screwed up with `ui.togglePinnedToolbar`. It never finishes executing:
```
Enqueuing action ui.togglePinnedToolbar; 0/0
Executing action ui.togglePinnedToolbar after waiting 0ms; 1/0
Enqueuing action preferences.setPreference; 1/0
Enqueuing action documents.createNew; 1/1
Enqueuing action documents.createNew; 1/2
Enqueuing action documents.createNew; 1/3
Enqueuing action documents.createNew; 1/4
```

iwehrman:
He's over here too:
image

iwehrman:
Gotta say I still think the contrast is still too low here:
image

iwehrman:
`ui.updatePanelSizes` is really slow, and seems to get called twice in a row quite often. Anything we can do about that?

iwehrman:
Expanding or collapsing layers does not cause the tab bar to resize the document titles. Instead, it seems like they only resize when the window is resized.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -595,14 +595,6 @@ define(function (require, exports, module) {
                 </ul>
             );
 
-            var layerCount = (
-                <div
-                    title={strings.TOOLTIPS.LAYER_COUNT}
-                    className="layer-count">
-                    {doc.layers.selected.size}<span className="text-fancy"> oƒ </span>{doc.layers.count}
```
iwehrman: I like this thing, but oh well. :cocktail:

iwehrman: We could probably rip out the associated `count` derived property, since I'm pretty sure it was only being used there.

src/js/util/svg.js
```diff
@@ -46,6 +46,10 @@ define(function (require, exports) {
             iconID += layer.kind;
         }
 
+        if ((layer.isArtboard || layer.layerKinds.GROUP) && !layer.expanded) {
```
iwehrman: Artboard layers are groups, so I think the first disjunct is redundant.

iwehrman:
OK, done with a second pass. Beyond the usual nits, the main concern is the `ui` action weirdness and performance issues. Happy to chat more offline. Other than that, it's obviously getting close. Everyone should be prepared to hold onto their butts.

designedbyscience:
1. Drag Style is not wide enough - Now it is wide enough
2. Bad text color - new text color
3. Selection color - keeping same color for now
4. Errors collapsing Library Panel - Fixed that
5. Borders on document header - now only shown in withdoc state
6. Pinned Toolbar errors - Fixed now
7. Contrast - adjusted a litte more. However, we may need to reconsider the normal text color in order to get more contrast
8. updatePanelSizes - now called just once
9. Document Title - this is a known bug (1890)
10. Count property - removed
11. Fixed expanded icon bug

Okay, back to @iwehrman

iwehrman:
Looks great. I filed a few follow-on bugs which I'm sure you'll be excited to address in your next pass. Bombs away!

Issue #1869, opened 2015-07-31
Search: clicking on icon or search window gives app focus


chadrolfs:
open a file and select a layer,
cmd/ctrl + f
click to expand the list
click on the magnifier icon or click on window somewhere (now no blinking cursor in field)
hit delete key

Result: layer gets deleted

iwehrman:
FBNC to @chadrolfs.

PR #1908, opened 2015-08-04
Search won't lose focus when you click on the dialog itself.

hamburge:
When you click inside the search dialog or on the "no options match" option, the dialog will no longer lose focus or close. Instead it will remove the autosuggest text (if it exists) and then highlight all the text in the input.

issue #1869


iwehrman:
OK!

PR #1910, opened 2015-08-04
String localization: French, German and Japanese

iwehrman:
This PR makes it possible to additionally load French, German and Japanese language dictionaries based on the locale in use by Photoshop in debug and compiled builds. For now, the dictionaries are copied as-is except for the `About Photoshop...` menu item, which has been edited in the other language dictionaries to show the current language. Later, the localization team will work with translators to update these dictionaries with translations as appropriate.

In debug builds, we just need to configure require to use the locale specified by `navigator.language` before loading the main module. In compiled builds, it's a little more complicated. We concatenate/compile the JavaScript once for each locale, and then at runtime we configure the locale and also alias `js/main` to `js/main-<lang>`, again using `navigator.language`, so that require can figure out that the symbolic path name `js/main` corresponds to the file `build/js/main-de.js`.

### DIFF
--- a/src/nls/de/strings.js
+++ b/src/nls/de/strings.js
@@ -22,9 +22,15 @@
  */
 
 /*global define */
+/*jscs:disable maximumLineLength*/
+/*jshint -W101*/
 
 define(function (require, exports, module) {
     "use strict";
 
-    module.exports = require("./locales");
+    module.exports = {
+        SEARCH: {
+            PLACEHOLDER: "Search, open & select IN GERMAN"
+        }
+    };
 });
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,6 +26,30 @@
 module.exports = function (grunt) {
     "use strict";
 
+    var getRequireOptions = function (locale) {
+        return {
+            options: {
+                baseUrl: "src/",
+                mainConfigFile: "src/js/config.js",
+                name: "js/main",
+                out: "build/js/main-" + locale + ".js",
+                // optimize: "none",
+                paths: {
+                    "react": "../bower_components/react/react-with-addons.min",
+                    "JSXTransformer": "../bower_components/jsx-requirejs-plugin/js/JSXTransformer"
+                },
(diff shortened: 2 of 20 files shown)


src/nls/de/strings.js
```diff
@@ -0,0 +1,627 @@
...
+/*global define */
+/*jscs:disable maximumLineLength*/
+/*jshint -W101*/
+
+define(function (require, exports, module) {
+    "use strict";
+
+    module.exports = {
+        ACTIONS: {
+            ADD_STROKE: "Add Stroke",
```
pineapplespatula: these dont seem very german. Is the plan to have this file be edited by the localization team.

iwehrman: Yes, the localization team is going to update those files with translations. They will submit pull requests. You can test that the different string dictionaries are being used in different locals by looking at the `ABOUT_PHOTOSHOP` menu string.

iwehrman: Er, I guess I meant `ABOUT_MAC`:
https://github.com/adobe-photoshop/spaces-design/pull/1910/files#diff-11e574773c7aed3a15d3183fdd0b0f31R69

volfied:
This is a good start.. however keeping these locale files in sync is going to be tough. 

Does nls support missing strings with warnings/fallbacks to root locale? 

I'm good to merge this, but I want to make sure as we add more strings we have as little potential pitfalls as possible.

iwehrman:
Yes, each non-root locale has both its dictionary and the root dictionary, and requirejs-i18n uses the root dictionary as a fallback. This PR could omit all keys in the other dictionaries other than the one key that I changed. Maybe that would be better.

iwehrman:
I rebased and cleaned up the dictionaries so that the untranslated languages just have a few entries to help the loc team figure out how to fill in the blanks. I also made a single `js/locales.js` file that holds all the supported locales, and that's imported by all the dictionary definition files. Back to @volfied for re-review.

volfied:
ook

Issue #1835, opened 2015-07-24
Templates "Android 1080p" and "Microsoft Surface Pro" are invalid


ktaki:
"Microsoft Surface Pro" is not available in Default New Doc Size.txt. "Android 1080p" exists in the file, but failing for some reason.

Action documents.createNew failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25923 Photoshop message: error: -25923 message: The parameters for command “Make” are not currently valid.

iwehrman:
FBNC to @chadrolfs

chadrolfs:
Verified the fix but assigning to @ktaki since he originally reported it.

ktaki:
Verified the fix.

PR #1911, opened 2015-08-04
fix for ussue 1835, underlying template changes

pineapplespatula:
fix for issue #1835


iwehrman:
Microsoft Surface Pro, we hardly knew ye.

src/static/templates.json
```diff
@@ -31,13 +31,7 @@
...
         "id": "ANDROID_1080P",
-        "preset": "Android 1080p (1920, 1080)",
-        "width": 1920,
-        "height": 1080
-    },
-    {
-        "id": "MS_SURFACE_PRO",
-        "preset": "Microsoft Surface Pro (1920, 1080)",
+        "preset": "Android 1080p (1080, 1920)",
         "width": 1920,
```
iwehrman: I think the Android 1080p width and height values should be swapped.

pineapplespatula: I just swapped them in order to match the PS template. #That'sTheFix

pineapplespatula:
@iwehrman back to you

src/static/templates.json
iwehrman: No, the width and height _properties_ need to be switched, not the values in the strong. When you create an `Android 1080p (1080, 1920)` document you get a document of width 1080 and height 1920, but this file has `width: 1920` and `height: 1080`. Those values are used here:
android

pineapplespatula:
ahh i see .

iwehrman:
:grapes:

PR #1913, opened 2015-08-05
Add read lock for JS_DOC to click and drag so they wait for delete

volfied:
Should address #1846, at least during my testing I got up to 30 enqueued clicks/deletes and nothing crashed.


iwehrman:
See `iwehrman/transfer` work-in-progress branch...

Issue #1836, opened 2015-07-24
Invalid new doc template will append the template names to New From Template menu


ktaki:
With no doc opened, click on either "Android 1080p" or "Microsoft Surface Pro" in the TEMPLATES panel. They are currently invalid templates as reported in https://github.com/adobe-photoshop/spaces-design/issues/1835. This should fail.

Check File -> New From Template. 
RESULT: The menu contents were doubled.

pineapplespatula:
since i fixed the two broken templates i cannot reproduce this anymore. I'm not sure if it is worth breaking in order to fix this other bug.

ktaki:
Since current template implementation in Ps is fragile, we should fix this bug. Modifying one of the template names in Default New Doc.txt will break the template. Let me know if you cannot repro.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #1918, opened 2015-08-05
clear template menu before appending new templates

pineapplespatula:
fix for issue #1836

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -235,21 +235,13 @@ define(function (require, exports, module) {
             templateIndex = _.findIndex(fileMenu.submenu, function (menu) {
                 return menu.id === "NEW_FROM_TEMPLATE";
             }),
-            templateMenu = fileMenu.submenu[templateIndex],
-            templateSubmenu = templateMenu.submenu;
-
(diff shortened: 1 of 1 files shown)


src/js/models/menubar.js
```diff
@@ -238,6 +238,7 @@ define(function (require, exports, module) {
             templateMenu = fileMenu.submenu[templateIndex],
             templateSubmenu = templateMenu.submenu;
 
+        templateSubmenu.length = 0;
```
volfied: Why not just start templateSubmenu as an empty array?

volfied:
Got a nit against redundancy of the redundancy, @pineapplespatula

src/js/models/menubar.js
pineapplespatula: because we need to make sure we clear the current submenu in this case to avoid duplicate entries.

iwehrman: It would be less scary to do something like `templateMenu.submenu = templates.map(...)`. Can you try that instead?

pineapplespatula:
@iwehrman it is prettier now, so back to you

iwehrman:
Brilliant.

Issue #1912, opened 2015-08-04
Sort menu commands by menu order by default


iwehrman:
It seems that the menu entries are in reverse order compared to how they're drawn in the menu:
![menu](https://s3.amazonaws.com/f.cl.ly/items/0P1M1G3P3U421z330x04/Image%202015-08-04%20at%204.53.00%20PM.png)

iwehrman:
Fixed by #1922.

Issue #1825, opened 2015-07-24
Localize keyCode strings


hamburge:
Currently using a lookup table in `util/key.js` to find strings for keys, but they should be localized.

iwehrman:
Fixed by #1922.

PR #1922, opened 2015-08-05
Search bugs

hamburge:
1. Issue #1912: puts menu commands in their default order
2. Issue #1825: we now have a lovely table for looking up localized keyCode names by their integer values
3. Fixes bug where input text was highlighted when you click on the search bar itself or when applying a filter where you had more text (for example if I typed "mylayer tex" and then applied the text filter, the remaining "mylayer" would be highlighted instead of leaving the cursor at the end)
4. Prevent error from not having registered search when finding search items and from closing dialog with escape


iwehrman:
Well, darned if I can find anything to complain about. Bombs away!

Issue #1923, opened 2015-08-05
Help > Reset DS defaults to Type tool and then errors selecting another tool


chadrolfs:
```Action tools.select failed: TypeError: Cannot read property 'layers' of null
    at Object.deselectHandler (file:///Users/crolfs/playground-repos/spaces-design/src/js/tools/type.js:147:41)
    at Object.<anonymous> (file:///Users/crolfs/playground-repos/spaces-design/src/js/actions/tools.js:184:52)```

volfied:
Also, can you please explain here why type tool deselect handler is calling reset layers? Tool switch is a common thing and we shouldn't be blocking with expensive operations.

pineapplespatula:
On exiting the type tool, if the current character style has been changed to match a selection made in the modal state, we need to reset it to the baseline character style from PS.

volfied:
Aren't all the other listeners you have handling that case?

pineapplespatula:
no

PR #1925, opened 2015-08-05
fix for undefined case in deselect handler

pineapplespatula:
fix for #1923


volfied:
Better to move the code discussion here.

I'm still not convinced by the need for calling resetLayers on tool deselect. Are we not getting all character style change events from Photoshop? What's the difference between the scenarios where _typeChangedHandler vs this event are required?

pineapplespatula:
we are not getting a type changed notification when we exit the tool or modal state. We could log that as a core bug, if you think the performance gain would be worth it, but deselecting the tool also removes the handler for those notifications so there are some timing concerns

Issue #1873, opened 2015-07-31
Sample tool missing tool tip


chadrolfs:

iwehrman:
I GOT THIS.

PR #1926, opened 2015-08-05
Localize keyboard shortcuts

iwehrman:
1. Menu and global keyboard shortcuts are now localized per platform at `i18n!nls/shortcuts-mac` and `i18n!nls/shortcuts-win`. Previously the shortcuts were coded directly in the menu definitions with `keyChar` and `keyCode` properties. Those properties now have a boolean value, and when true the `menubar` will attempt to resolve the shortcut key from the localized shortcut files instead. 
2. Localized menu labels now live in their own file in the i18n menu structure at `i18n!nls/menu` to make them a little easier to manage.
3. The sampler tooltip is fixed, addressing #1873.


volfied:
Drop the rebase @iwehrman

src/js/models/menubar.js
```diff
@@ -260,19 +260,63 @@ define(function (require, exports, module) {
...
+        entries.forEach(function (entry) {
+            var id = entry.id;
+
+            if (entry.hasOwnProperty("shortcut")) {
+                var shortcut = entry.shortcut;
+                if (!shortcut.keyCode && !shortcut.keyChar) {
+                    throw new Error("Menu entry " + id + " has a shortcut without a key.");
+                }
+
+                if (!shortcuts) {
```
volfied: This check doesn't look like it belongs in the forEach, `shortcuts` is the fn parameter.

volfied:
I like it, got one nit and a rebase.

src/js/models/menubar.js
iwehrman: This a little convoluted but intentional. The point is that it's OK if there is no localization dictionary for a particular submenu ID _only if_ there are no entries that have keyboard shortcuts. So, it's an error as soon as you have a shortcut key that needs to be localized. Yes, I could do a first pass over the array and check for the existence of shortcuts first, but this seemed just as well to me...

iwehrman:
Rebased and ready 2 rock.

src/js/models/menubar.js
iwehrman: Though I did push a fix to correct the JSDoc on that param to indicate that it is optional.

volfied:
Alrighty then

Issue #1871, opened 2015-07-31
Search: filters missing their icons


chadrolfs:

iwehrman:
FBNC to @chadrolfs.

PR #1928, opened 2015-08-05
Put new icons into search

hamburge:
Also moved the getting of SVG class names to the search store so it can more easily be determined based on the option's category.
issue #1871


src/js/jsx/search/SearchBar.jsx
```diff
@@ -126,7 +125,7 @@ define(function (require, exports, module) {
             var idArray = id ? id.split("-") : [],
                 filterValues = _.drop(idArray),
                 updatedFilter = id ? _.uniq(this.state.filter.concat(filterValues)) : [],
-                filterIcon = svgUtil.getSVGClassesFromFilter(updatedFilter);
+                filterIcon = id ? this.getFlux().store("search").getSVGClass(updatedFilter) : null;
```
iwehrman: Hot tip: if you've got some expression `foo` which may be falsey, then you can just do `foo && bar` instead of `foo ? bar : null`.

src/js/stores/search.js
```diff
@@ -294,26 +315,26 @@ define(function (require, exports, module) {
...
                             title: title,
                             svgType: icon,
-                            category: categories,
+                            category: itemCategories,
                             style: { "fontStyle": "italic" },
                             haveDocument: searchInfo.haveDocument,
                             type: "filter"
                         });
-                    });
+                    }.bind(this));
```
iwehrman: Hot tip: Both `Array.prototype.forEach` and `Immutable.List.prototype.forEach` (and all their collection-traversing friends) allow you to pass a final "receiver" parameter so you don't have to explicitly bind a `this` to the function.

iwehrman: And, in case you were wondering, we do prefer receiver arguments over explicit binding :grimacing:

iwehrman:
This is for @placegraphichere as much as @hamburge, but I think the magnifying glass icon here is too tiny compared to the other inline icons: 
search

hamburge:
@iwehrman Thanks for the hot tips. I made the search icon a little bigger too because I agree, it was too little.

iwehrman:
:fire: :point_left:

Issue #1806, opened 2015-07-23
Selected search options shift horizontally when selected


iwehrman:
![shift](https://s3.amazonaws.com/f.cl.ly/items/113I2U3t0U0h2P0C211c/Screen%20Recording%202015-07-23%20at%2007.07%20AM.gif)

iwehrman:
This is still happening in your most recent branch, @hamburge.

iwehrman:
Fixed by brute force.

PR #1939, opened 2015-08-06
Search options should not shift horizontally when selected

hamburge:
Chrome on Retina displays apparently doesn't know how to add rem values. @designedbyscience said this is currently the best workaround option to properly calculate the rem value.

issue #1806


designedbyscience:
She's not wrong

iwehrman:
If @designedbyscience approves then I approve.

iwehrman:
Man if only there were some WebKit reviewers around here that could fix that bug.

Issue #1932, opened 2015-08-06
Menu shortcut to open search bar should also close it if open


iwehrman:
Just like Spotlight and Alfred.

iwehrman:
Hmm, this still doesn't work for me. (I guess I should have tested it before merging #1943.) Back to @hamburge!

Issue #1931, opened 2015-08-06
Search should refer to "results" not "options"


iwehrman:
![options](https://s3.amazonaws.com/f.cl.ly/items/3T1n1B1J3M1y2y2G1C2f/Image%202015-08-06%20at%206.57.26%20AM.png)

iwehrman:
Fixed by #1943.

Issue #1930, opened 2015-08-06
Filter results not selected by default


iwehrman:
Maybe this is intentional, but it just tripped me up. I wanted to just filter by menu commands, so I hit cmd-F, "me", enter. I expected this to filter by menu commands, but the first result wasn't highlighted. It seems like it only highlights the first non-filter result? Also, some unexpected command was executed when I hit enter with no (visible) results highlighted. Maybe an off-screen result was highlighted? That shouldn't happen either!

iwehrman:
![me](https://s3.amazonaws.com/f.cl.ly/items/1L3u0e2K0i0H400F262N/Image%202015-08-06%20at%206.53.33%20AM.png)

iwehrman:
Fixed by #1943.

Issue #1929, opened 2015-08-06
Add Super Search to keyboard shortcut list


joelrbrandt:
the "Help > Keyboard Shortcuts" menu lists keyboard shortcuts. But, Super Search isn't in the list! :scream:

joelrbrandt:
Actually, I guess it's under the "Layers" menu, so maybe it doesn't belong in the keyboard shortcuts? Maybe @placegraphichere or @iwehrman has an opinion? This feature has gotten so much bigger than just layers...

iwehrman:
I agree that the `Layers` menu probably is no longer the right place for it. Actually we should still have a `Find Layers` entry under `Layers`, which will trigger @designedbyscience's in-panel layers-only search feature. We should have an `App Search` menu entry elsewhere; possibly under `Edit`?

So far we've only put non-menu keyboard shortcuts in that dialog. But I agree that it is important enough to jam it in there anyway.

hamburge:
Putting it under `Window` makes more sense to me than `Edit`

iwehrman:
This is now in the keyboard shortcuts dialog, and the search menu command has been moved under `Edit`. The placement in the keyboard shortcuts dialog is a little weird, but that doesn't matter because we'll have to spruce that whole dialog up before M3 anyway. Closing this out since @joelrbrandt is busy with other things.

PR #1943, opened 2015-08-06
More search bugs

hamburge:
:bug: :bee: :beetle: :snail: :ant: 

-#1929 Put search shortcut in the Keyboard Shortcuts dialog and moved the command to `Window` from `Layer`, but if the executive decision is to put it in `Edit` (or somewhere else), I can move it again.

-I was also dissatisfied with how things looked after 507b561bf74b2acfe2b800472121b2013010b704, since the text would sometimes highlight momentarily before going to the end of the word, so I undid most of those changes and added a prop to TextInput for when we will never be selecting all text. I think it now looks cleaner and means there are fewer changes in `TextInput.jsx`

-Also #1930, #1931, #1932

### DIFF
--- a/src/nls/root/strings.js
+++ b/src/nls/root/strings.js
@@ -125,7 +125,10 @@ define(function (require, exports, module) {
                 ELLIPSE: "Ellipse",
                 PEN: "Pen",
                 TYPE: "Type",
-                SAMPLER: "Sampler"
+                SAMPLER: "Sampler",
+                SEARCH: "Search",
+                SEARCH_INSTRUCTION_MAC: "Cmd + F",
+                SEARCH_INSTRUCTION_WIN: "Ctrl + F"
             },
             SELECT_TOOL: {
                 TARGET: "Target Nested Group / Layers",
@@ -382,7 +385,7 @@ define(function (require, exports, module) {
         SEARCH: {
             PLACEHOLDER: "Search, open & select",
             PLACEHOLDER_FILTER: "Search ",
-            NO_OPTIONS: "No options match your search",
+            NO_OPTIONS: "No results match your search",
             HEADERS: {
                 ALL_LAYER: "Layers",
                 CURRENT_LAYER: "Layers",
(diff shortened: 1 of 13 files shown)


iwehrman:
@placegraphichere: do you approve of the placement of the `Search` menu command in the `Window` menu? Another plausible alternative might be the `Edit` menu.

iwehrman:
Also, generally speaking, it is preferable to have smaller pull requests compared to larger ones. Next time, if it's possible, it would be nice to have one PR per issue. Sometimes that's not possible, or it's really inconvenient, and that's fine. But smaller PRs tend to be much easier to review, which means there tends to be a much shorter turnaround time before merging!

iwehrman:
And, finally, needs a rebase!

hamburge:
@iwehrman Just rebased-- not sure how some of your commits got in here?

hamburge:
^jk, fixed that

src/nls/root/strings.js
```diff
@@ -63,6 +63,216 @@ define(function (require, exports, module) {
             UNGROUP_LAYERS: "Ungroup layers"
         },
         APP_NAME: "Photoshop",
+        MENU: {
```
iwehrman: This is a rebase error. There is no longer a `MENU` property in strings.js. Instead, this has all moved to `i18n!nls/menu`.

iwehrman:
Besides the rebase error, this looks good. I'm going to file a separate issue about the use of string encoding to identify filters, etc. Back to @hamburge to fix the menu stuff.

iwehrman:
Opened #1966 for the aforementioned string encoding cleanup request.

placegraphichere:
We’re using CMD+f which is the same as "find" for most apps. That is generally housed in the Edit menu. Usually that “help” or search finder located in the window menu is meant for it’s own OS specific application search capability (at least on mac). That kind of search is definitely different than ours. 

So for this case, I might lean towards Edit as a home. That way we can also break it out into more categorical searches in the future, say for instance *Edit > Search > Stock* or *Edit > Search > Libraries*. Housing them under Window would imply (to me anyway) it will take me out of the app to the web.

hamburge:
@iwehrman Great, thanks. I moved the menu command over to Edit as well.

iwehrman:
OK!

PR #1957, opened 2015-08-09
Local lock declarations

iwehrman:
The controller currently assumes that actions will specify the entire set of read and write locks necessary to safely execute the given action _as well as all other sub-actions to which the action may transfer_. Suppose, for example, we have:

```javascript
var foo = function () {
  // write A
};
foo.writes = [A];

var bar = function () {
  // write B, and then
  return this.transfer(foo);
}
bar.writes = [A, B]; // A is needed because of the transfer to foo
```
Above, the `foo` action writes to state protected by the `A` lock, and so includes the appropriate write lock declaration. The `bar` action itself only writes to state protected by the `B` lock, but also transfers to `foo`, and so must also include all of the locks specified by `foo` in its lock declaration. 

Suppose now we update `foo` so that it additionally writes to state protected by the `C` lock:
```javascript
var foo = function () {
  // write A
  // and, now, also write C
};
foo.writes = [A, C];
```
Unfortunately now when `bar` is executed, there will be a runtime error because it transfers to `bar` which requires both `A` and `C` locks, but `bar` only declares that it requires locks `A` and `B`. To solve this problem, we have to also amend the lock declaration of `bar` to include `C`.
```javascript
var bar = function () {
  // write B, and then
  return this.transfer(foo);
}
bar.writes = [A, B, C]; // now A AND C are needed because of the transfer to foo
```
This is hugely problematic for maintainability because small, local changes almost always require perfunctory global changes. Furthermore, because there is so much "noise" in the lock declarations related to sub-actions, mistakes in these declarations have propagated so thoroughly that **nearly 100% of our lock declarations are currently incorrect.** 

This PR makes a major change to the controller to make it easier to correctly declare the correct lock sets. Instead of requiring that an action declare all the locks needed for its execution as well as of its transferrable sub-actions, with this change it is only necessary to specify the locks needed immediately for the given action. To avoid deadlock though, the controller still acquires all the required locks up-front. In order to do this, it needs to know additionally to which sub-actions the given action may transfer. With this information, it can then calculate the complete, transitive set of locks required for each action. 

With this PR, the original example above is now correctly rewritten as follows: 
```javascript
var foo = function () {
  // write A
};
foo.writes = [A];

var bar = function () {
  // write B, and then
  return this.transfer(foo);
}
bar.writes = [B];
bar.transfers = [foo]; // If this isn't declared, the transfer will fail at runtime
```
Note that `bar` _only_ explicitly declares the `B` lock and not the `A` lock because the latter is not immediately required by `bar`, only by the sub-action `foo` to which it transfers. Furthermore, this transfer too `foo` is declared up-front in the set of `transfers` declared for `bar`. Now, if `foo` is updated as before to additional write to stated required by a `C` lock, we update `foo` as before but _no update is required to `bar`_. 

Currently, the controller validates the correctness of the lock sets when transferring from one action to another. In this new regime, it instead validates that _transfers_ have been declared up-front when performing the transfer. Because the controller calculates the lock sets statically, there is no need to validate the lock sets at runtime. That is to say, when `bar` is executed before the change it still waits until locks `A,B` are available, and after the change it waits until locks `A,B,C` are available. This doesn't change what locks are acquired as a precondition for execution; only what locks need to be explicitly declared and where.

In summary, this PR makes two changes to the controller and to action declarations:

1. We now declare only the immediately required read and write locks instead of the locks needed by all potentially transferrable sub-actions.
2. We now also declare transferrable sub-actions, and the controller validates these transfers at runtime.

**Aside 1:** Both before and after this PR, the controller is unable to validate whether the set of locks specified for a particular actions is sufficient to maintain safety in the presence of concurrently executing actions. It was and is up to the developer to choose an appropriate set of locks. This change only affects how the overall set of locks is calculated relative to sub-action transfers.

**Aside 2:** this change relies upon an assumption that there are **no circular transfer declarations**. That is, it is not currently possible for `foo` to transfer to `bar` and for `bar` to transfer to `foo`. I don't think this is a fundamental requirement, but a) I haven't quite worked out how to solve that case; and b) it doesn't happen in our application at all. 

**Aside 3:** In this PR, it is possible to specify action transfers either by direct reference to the unsynchronized action function (e.g., `toolActions.select`) OR with a string pathname (e.g., `"tools.select"`). I had to at least allow the latter to solve the problem of circular module dependencies. For example, the `layers` and `documents` modules are circularly dependent, which means one is loaded first, which means it can't refer at module-initialization time to the other's exports. I think in a future PR I will probably make the additional change of ONLY allowing string pathnames to be used, both in the transfer declarations and also in the first parameter of the `transfer` method itself. Why? Because a) it's necessary at least sometimes; b) it probably improves readability by enforcing uniformity of the declarations; and c) it will eliminate some otherwise unnecessary module dependencies and imports.

### DIFF
--- a/src/js/actions/edit.js
+++ b/src/js/actions/edit.js
@@ -32,6 +32,7 @@ define(function (require, exports) {
     var events = require("../events"),
         locks = require("../locks"),
         layers = require("js/actions/layers"),
+        menu = require("./menu"),
         collection = require("js/util/collection"),
         headlights = require("js/util/headlights"),
         history = require("js/actions/history");
@@ -88,13 +89,14 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var nativeCut = function () {
-        return this.flux.actions.menu.nativeModal({
+        return this.transfer(menu.nativeModal, {
             commandID: CUT_NATIVE_MENU_COMMMAND_ID
         });
     };
     nativeCut.modal = true;
     nativeCut.reads = [];
     nativeCut.writes = [];
+    nativeCut.transfers = [menu.nativeModal];
 
     /**
      * Execute a native copy command.
@@ -103,13 +105,14 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var nativeCopy = function () {
-        return this.flux.actions.menu.nativeModal({
+        return this.transfer(menu.nativeModal, {
             commandID: COPY_NATIVE_MENU_COMMMAND_ID
         });
     };
     nativeCopy.modal = true;
     nativeCopy.reads = [];
     nativeCopy.writes = [];
+    nativeCopy.transfers = [menu.nativeModal];
 
     /**
      * Execute a native paste command.
@@ -118,13 +121,14 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var nativePaste = function () {
-        return this.flux.actions.menu.nativeModal({
+        return this.transfer(menu.nativeModal, {
             commandID: PASTE_NATIVE_MENU_COMMMAND_ID
         });
     };
     nativePaste.modal = true;
     nativePaste.reads = [];
     nativePaste.writes = [];
+    nativePaste.transfers = [menu.nativeModal];
 
     /**
      * Execute a native selectAll command.
@@ -136,7 +140,7 @@ define(function (require, exports) {
     var nativeSelectAll = function (waitForCompletion) {
         waitForCompletion = waitForCompletion || false;
 
-        return this.flux.actions.menu.nativeModal({
+        return this.transfer(menu.nativeModal, {
             commandID: SELECT_ALL_NATIVE_MENU_COMMMAND_ID,
             waitForCompletion: waitForCompletion
         });
@@ -144,6 +148,7 @@ define(function (require, exports) {
     nativeSelectAll.modal = true;
     nativeSelectAll.reads = [];
     nativeSelectAll.writes = [];
+    nativeSelectAll.transfers = [menu.nativeModal];
 
     /**
      * Execute either a cut or copy operation, depending on the value of the parameter.
@@ -186,9 +191,9 @@ define(function (require, exports) {
                 // If we're on modal state (type edit), we should go with native copy/cut
                 if (this.flux.store("tool").getModalToolState()) {
                     if (cut) {
-                        this.flux.actions.edit.nativeCut();
+                        return this.transfer(nativeCut);
                     } else {
-                        this.flux.actions.edit.nativeCopy();
+                        return this.transfer(nativeCopy);
                     }
                 } else if (!cut) {
                     var applicationStore = this.flux.store("application"),
@@ -221,8 +226,9 @@ define(function (require, exports) {
         return _cutOrCopy.call(this, true);
     };
     cut.modal = true;
-    cut.reads = [];
-    cut.writes = [locks.JS_DOC, locks.PS_DOC];
+    cut.reads = [locks.JS_TOOL, locks.PS_APP];
+    cut.writes = [locks.JS_DOC, locks.PS_DOC, locks.OS_CLIPBOARD];
+    cut.transfers = [nativeCut];
 
     /**
      * Execute a copy operation on the currently active HTML element.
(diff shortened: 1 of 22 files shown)


src/js/actions/edit.js
```diff
@@ -88,13 +89,14 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var nativeCut = function () {
-        return this.flux.actions.menu.nativeModal({
+        return this.transfer(menu.nativeModal, {
```
volfied: Note to self to make sure this enqueue -> transfer change is safe.

src/js/actions/history.js
```diff
@@ -186,8 +186,9 @@ define(function (require, exports) {
     var incrementHistory = function (document) {
         return _navigateHistory.call(this, document, 1);
     };
-    incrementHistory.reads = [locks.PS_DOC, locks.JS_APP, locks.JS_TOOL];
-    incrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC, locks.PS_APP, locks.JS_POLICY];
+    incrementHistory.reads = [locks.JS_DOC, locks.JS_APP];
+    incrementHistory.writes = [locks.JS_HISTORY, locks.JS_DOC, locks.PS_DOC];
+    incrementHistory.transfers = [toolActions.resetBorderPolicies, "documents.updateDocument"];
```
volfied: Any particular reason `documents.updateDocument` is stringified?

src/js/fluxcontroller.js
```diff
@@ -148,6 +171,7 @@ define(function (require, exports, module) {
      */
     FluxController.prototype._actionReceivers = null;
 
+
```
volfied: :space_invader:

src/js/fluxcontroller.js
```diff
@@ -106,7 +94,10 @@ define(function (require, exports, module) {
         EventEmitter.call(this);
 
         var cores = window.navigator.hardwareConcurrency || 8;
+
         this._actionQueue = new AsyncDependencyQueue(cores);
+        this._initActionNames();
```
volfied: Wouldn't it be better to merge `initActionNames` and `initActionLocks in `_createActionReceivers` so we loop through the actions once?

src/js/fluxcontroller.js
```diff
@@ -218,16 +363,9 @@ define(function (require, exports, module) {
...
-                        throw new Error("Next action requires additional read locks");
-                    }
-
-                    var nextWrites = nextAction.writes || locks.ALL_LOCKS;
-                    if (!_subseteq(nextWrites, currentWrites)) {
-                        log.error("Missing write locks:", _.difference(nextWrites, currentWrites).join(", "));
-                        throw new Error("Next action requires additional write locks");
+                    if (!currentTransfers.has(nextAction)) {
+                        var nextActionName = self._actionNames.get(nextAction);
+                        throw new Error("Invalid transfer from " + actionName + " to " + nextActionName);
```
volfied: You should probably have a more descriptive error here telling the dev they forgot to declare a transfer.

volfied:
Just two nits, but otherwise looks good. This clean up is great, and I wish we had this before I made resetBorderPolicies :) Back to you @iwehrman

iwehrman:
Nits fixed. In particular: https://github.com/adobe-photoshop/spaces-design/pull/1957/files#diff-66f2530292e4c08200d2c82ef4860d35R367

volfied:
Well, this is a great architectural change, and looking through the actions everything seems safe to me. Thank you for working on this.

shaoshing:
The lock declaration now look much more clear with the changes! Awesome job @iwehrman

src/js/actions/shapes.js
```diff
@@ -716,8 +724,9 @@ define(function (require, exports) {
 
         return this.transfer(combineSubtract, currentDocument, currentDocument.layers.selected);
     };
-    combineSubtractSelectedInCurrentDocument.reads = [locks.PS_DOC, locks.JS_DOC, locks.JS_APP];
-    combineSubtractSelectedInCurrentDocument.writes = [locks.PS_DOC, locks.JS_DOC];
+    combineSubtractSelectedInCurrentDocument.reads = [locks.JS_APP];
+    combineSubtractSelectedInCurrentDocument.writes = [];
+    combineSubtractSelectedInCurrentDocument.transfer = [combineSubtract];
```
shaoshing: A tiny typo.

Issue #1955, opened 2015-08-07
Paste (cmd/ctrl + v) on new doc errors


chadrolfs:
new doc (Layer 1 focused), cmd/ctrl + v
Result: `Action edit.nativePaste failed: Error: Menu command not available: 105`

iwehrman:
Downgrading this to low-priority since it only happens in debug builds. But, on the upside, FIP.

PR #1958, opened 2015-08-10
Ignore clipboard-related errors due to menu.nativeModal failure

iwehrman:
`menu.nativeModal` and `menu.native` fail in debug mode when the underlying menu command isn't available. In the case of the native clipboard edit operations, these errors are harmless and can safely be ignored.

Addresses #1955.


volfied:
Looks good.

Issue #882, opened 2015-06-12
Layer drag-copy is slow


DesignSpaceBot:
The culprit is [an `updateDocument` call](https://github.com/adobe-photoshop/playground-design/blob/026cd42c02f1850ae87abf224b3417f578db122c/src/js/actions/superselect.js#L559) which leads to an unnecessarily expensive state reset.

-- @iwehrman at [2015-03-22T00:25:56Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/882)

volfied:
https://trello.com/c/ibANMsqv/293-layer-option-drag-to-copy-should-trigger-an-event-that-provides-the-newly-created-layer-ids

iwehrman:
@isonno has answered all of our prayers. We're no longer blocked by core on this. LET'S ROLL :us: :us:

chadrolfs:
@iwehrman can we close this now?

iwehrman:
Fixed by #1959.

PR #1959, opened 2015-08-10
Faster layer drag-copy

iwehrman:
This changes the implementation of layer drag-copy to use `layers.addLayers` instead of `documents.updateDocument`. The former is much faster than the latter. It relies on a new event added to Photoshop a few weeks back by @isonno. 

Addresses #882.


volfied:
A sub 1000 issue being addressed is always awesome :)

Issue #1940, opened 2015-08-06
cmd/ctrl + click selects AB instead of leaf layers


chadrolfs:

iwehrman:
I think @volfied said he was going to look into this.

iwehrman:
2 @chadrolfs 2 furious.

PR #1962, opened 2015-08-10
Superselect Fixes

volfied:
Addresses #1940, and likely another superselect bug we may not have noticed.

Problem with #1940 was, I was using PS hit layer IDs for deep select, without artboards taken out.

Other bug is more nuanced, we start with `topBelowArtboards` and then systematically replace selected layers with relatives. However, if a relative is in the `topBelowArtboards`, it would remain at the beginning of the list, breaking z-order.


iwehrman:
What I meant to say was: looks great!

Issue #1895, opened 2015-08-02
transform.flip, align and distribute should have an underscore prefix


iwehrman:
... because they're private. They should also be moved to the top of the file for consistency and readability.

iwehrman:
Also, `transform.flip`. CC @mcilroyc.

volfied:
#1921 will fix this.

iwehrman:
@DivyaPrabhakar will finish painting the back of this fence.

iwehrman:
Also `transform.nudgeLayers`.

pineapplespatula:
@iwehrman  you wrote it, so you get to close it. 
congrats

iwehrman:
Challenge accepted.

PR #1963, opened 2015-08-10
Added an underscore prefix to transform.flip, align, and distribute 

DivyaPrabhakar:
Code cleanup: Added an underscore prefix to transform.flip, align, and distribute to indicate that they are private. This addresses #1895.


pineapplespatula:
looks great.

PR #1964, opened 2015-08-10
Fix Japanese localization

iwehrman:
I thought I tested this, but I guess I didn't! The Japanese locale prefix is "ja" not "jp", but my initial PR used the latter instead of the former.

Addresses #1953.


iwehrman:
CC @chadrolfs for testing.

volfied:
:bow:

PR #1970, opened 2015-08-10
allow for adding vector masks to layers

pineapplespatula:
Allows users to edit existing vector masks, requires https://github.com/adobe-photoshop/spaces-adapter/pull/143 

there are a few blocking bugs, but i thought it would be helpful to get this PR back up with my more recent changes. 

the blocking bugs are :
entering the modal state in the mask tool blocks CEF from responding to commands, 
vector masks spoil our bounds in a way that makes resizing layers impossible, 
and the pen tool does not edit the vector mask, instead it edits the work path, ( which is useless to us)

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -30,6 +30,7 @@ define(function (require, exports) {
         
     var photoshopEvent = require("adapter/lib/photoshopEvent"),
         artboardLib = require("adapter/lib/artboard"),
+        boundsLib = require("adapter/lib/bounds"),
         descriptor = require("adapter/ps/descriptor"),
         documentLib = require("adapter/lib/document"),
         layerLib = require("adapter/lib/layer"),
@@ -72,7 +73,7 @@ define(function (require, exports) {
         "layerLocking",
         "itemIndex",
         "background",
-        "boundsNoEffects",
+        "boundsNoMask",
         "opacity",
         "layerFXVisible",
         "mode"
@@ -456,7 +457,7 @@ define(function (require, exports) {
             } else if (layer.kind === layer.layerKinds.TEXT) {
                 property = "boundingBox";
             } else {
-                property = "boundsNoEffects";
+                property = "boundsNoMask";
             }
             return [
                 documentLib.referenceBy.id(document.id),
@@ -1731,6 +1732,19 @@ define(function (require, exports) {
     duplicate.transfers = ["documents.updateDocument", addLayers, select];
 
     /**
+     * Dispatches a command to  select the VEctor mask for the currently selected layer
+     *
+     * @return {Promise}
+     */
+    var selectVectorMask = function () {
+        return descriptor.playObject(vectorMaskLib.selectVectorMask());
+    };
+
+    selectVectorMask.reads = [locks.JS_TOOL];
+    selectVectorMask.writes = [locks.PS_DOC];
+    selectVectorMask.modal = true;
+
+    /**
      * Dispatches a layer translate for all layers in the document
      *
      * @param {object} event Action Descriptor of autoCanvasResizeShift event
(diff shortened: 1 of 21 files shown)


src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
+            return this.transfer(tools.select, superselectVector)
+                .bind(this)
+                .then(function () {
+                    return descriptor.batchPlayObjects([vectorMaskLib.makeBoundsWorkPath(bounds),
+                        vectorMaskLib.makeVectorMaskFromWorkPath(),
+                        vectorMaskLib.deleteWorkPath()]);
+                });
+        }
+    };
+
```
volfied: :space_invader:

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
     editVectorMask.writes = [locks.PS_DOC];
     editVectorMask.transfers = [tools.select];
 
+    var addVectorMask = function () {
```
volfied: JSDoc

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
+        var applicationStore = this.flux.store("application"),
+            currentDocument = applicationStore.getCurrentDocument(),
+            toolStore = this.flux.store("tool"),
+            superselectVector = toolStore.getToolByID("superselectVector");
+
+        if (currentDocument !== null) {
```
volfied: If the current Document is null, this action will cause a throw, because it doesn't return a promise.

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
+            var layer = currentDocument.layers.selected.first(),
+            bounds = boundsLib.bounds(layer.bounds);
+
+            var payload = {
+                documentID: applicationStore.getCurrentDocumentID(),
```
volfied: You can use `currentDocument.id`

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
 
+    var addVectorMask = function () {
+        var applicationStore = this.flux.store("application"),
+            currentDocument = applicationStore.getCurrentDocument(),
+            toolStore = this.flux.store("tool"),
+            superselectVector = toolStore.getToolByID("superselectVector");
+
+        if (currentDocument !== null) {
+            var layer = currentDocument.layers.selected.first(),
+            bounds = boundsLib.bounds(layer.bounds);
```
volfied: indent

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
     editVectorMask.transfers = [tools.select];
 
+    var addVectorMask = function () {
+        var applicationStore = this.flux.store("application"),
+            currentDocument = applicationStore.getCurrentDocument(),
+            toolStore = this.flux.store("tool"),
+            superselectVector = toolStore.getToolByID("superselectVector");
+
+        if (currentDocument !== null) {
+            var layer = currentDocument.layers.selected.first(),
```
volfied: What if layer is null?

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
+            var payload = {
+                documentID: applicationStore.getCurrentDocumentID(),
+                layerIDs: Immutable.List.of(layer.id),
+                vectorMaskEnabled: true
+            };
+
+            this.dispatch(events.document.history.optimistic.ADD_VECTOR_MASK_TO_LAYER, payload);
+
+            return this.transfer(tools.select, superselectVector)
+                .bind(this)
```
volfied: You don't need to bind this, you don't use it at all below.

src/js/actions/layers.js
```diff
@@ -1528,6 +1529,37 @@ define(function (require, exports) {
...
+                .then(function () {
+                    return descriptor.batchPlayObjects([vectorMaskLib.makeBoundsWorkPath(bounds),
+                        vectorMaskLib.makeVectorMaskFromWorkPath(),
+                        vectorMaskLib.deleteWorkPath()]);
+                });
```
volfied: How does the user get back to their original tool? Do they get stuck in superselectVector tool?

src/js/models/menubar.js
```diff
@@ -227,6 +227,20 @@ define(function (require, exports, module) {
                 (document.layers.selectedNormalized.size === 1) &&
                 (document.layers.selected.every(function (layer) {
                     return layer.vectorMaskEnabled === true;
+                })),
```
volfied: These two rules should be just testing for hasVectorMask/canAddVectorMask, and you should add `layers-selected-1` as the other rule.

src/static/menu-actions.json
```diff
@@ -233,6 +233,10 @@
         "EDIT_VECTOR_MASK": {
             "$action": "layers.editVectorMask",
             "$enable-rule": "one-layer-selected-has-vector-mask"
+        },
+        "ADD_VECTOR_MASK": {
+            "$action": "layers.addVectorMask",
+            "$enable-rule": "one-layer-selected-can-add-vector-mask"
```
volfied: Per comment above, these two should be a combination of `layers-selected-1` and some other simpler rule

src/static/menu-mac.json
```diff
@@ -331,6 +331,9 @@
                             "command": true
                         }
                     }
+                },
+                {
+                    "id": "ADD_VECTOR_MASK"
```
volfied: No shortcut for this one?

On that note, can we have "Edit Vector Mask" be one menu item, and it adds it if there isn't one?

volfied:
Got some nits and comments. Back to you @pineapplespatula

src/js/actions/layers.js
pineapplespatula: They just added a vector mask, I assume here that they want to do something with it.

src/js/models/menubar.js
pineapplespatula: are comma separated rules ANDed or ORed?

pineapplespatula:
I got some new direction from seth. i vote we table this PR until i incorporate more of the new design

pineapplespatula:
addresses https://github.com/adobe-photoshop/spaces-design/issues/2048

PR #1972, opened 2015-08-10
Search bar would not reopen after hitting escape on a search with a filter applied

hamburge:
After applying a filter, the selected element was not updated properly, so the filter element was still selected, and the dialog did not close properly.

Now, after applying a filter, nothing is selected and the modal will fully close when you hit escape.


hamburge:
This is actually fixed as a side effect of PR #1975

PR #1975, opened 2015-08-10
Better ordering of search results

hamburge:
Order search results so that the group of options that contains the suggestion is on top. Also if there is no suggestion, select the first item. With these changes, you will always be able to see what is selected (with the one exception of the first render, when you open the dialog and click on the text input).

As a side effect, also fixes a bug where the search bar would not reopen after hitting escape on a search with a filter applied. The dialog would not close properly in that case, because the filter element was still selected. Now, after applying a filter, the first element of the list will be selected and the modal will fully close when you hit escape.

### DIFF
--- a/src/js/stores/search.js
+++ b/src/js/stores/search.js
@@ -227,7 +227,8 @@ define(function (require, exports, module) {
                         return items.concat(options);
                     }.bind(this), []);
 
-                search.searchItems = Immutable.List(searchItems).unshift(filterItems);
+                searchItems.push(Immutable.List(filterItems));
+                search.searchItems = Immutable.List(searchItems);
             }
             this.emit("change");
         },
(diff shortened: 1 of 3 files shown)


iwehrman:
> also fixes a bug where the search bar would not reopen after hitting escape on a search with a filter applied. 

Woohoo! I think I hit this a few times but couldn't figure out how to repro it.

src/js/stores/search.js
```diff
@@ -227,7 +227,8 @@ define(function (require, exports, module) {
                         return items.concat(options);
                     }.bind(this), []);
 
-                search.searchItems = Immutable.List(searchItems).unshift(filterItems);
+                searchItems.push(Immutable.List(filterItems));
```
volfied: You're pushing a list onto a list... did you mean concat?

volfied:
One potential bug, @hamburge

src/js/stores/search.js
hamburge: Nope-- searchItems is an list of lists, and `filterItems` is one of the inner lists.

hamburge:
@volfied Responded to your comment. Back to you!

volfied:
Great, thanks for clarifying.

PR #1979, opened 2015-08-11
Use correct descriptor name for libraries

volfied:
I've removed the IMS connection from Photoshop as we don't use it. And changed the descriptor names... because there is no IMS information anymore.

This will require pg-dev-mac 440 to work. So @shaoshing, please merge this once you grab the latest Jenkins build as to not disturb your library work progress.


shaoshing:
Looks good. thanks

Issue #1862, opened 2015-07-29
CC Lib: unable to add Adobe Shapes (svg) files from library to canvas


chadrolfs:
Using Adobe Shape mobile app, add assets to a library
Load library in DS and click on (+) to add to canvas

Result: nothing is added

shaoshing:
Graphic asset created from the mobil app is SVG file instead of PS file. I checked the log from the script listener and it added additional params for SVG asset. @volfied , do you think I can fix this in the adapter by adding those params, or I need supports from the core? 

#### Regular asset 

``` js
_playground.ps.descriptor.play("placeEvent", {
 "ID": 7,
 "null": {
  "path": "/Users/shalee/Library/Application Support/Adobe/Creative Cloud Libraries/LIBS/EED3CFB4557B1FE67F000101_adobe.com/creative_cloud/dcx/30629ecc-a840-4946-a6c4-e4c891214491/components/a4656939-d652-481e-ac36-de24294e3829.psd"
 },
 "layerName": "Artboard 1",
 "linked": true,
 "freeTransformCenterState": {
  "_enum": "quadCenterState",
  "_value": "QCSAverage"
 },
 "offset": {
  "_obj": "offset",
  "_value": {
   "horizontal": {
    "_unit": "pixelsUnit",
    "_value": 0
   },
   "vertical": {
    "_unit": "pixelsUnit",
    "_value": 0
   }
  }
 }
}, {}, function(){})
``` 

#### SVG asset

```js
_playground.ps.descriptor.play("placeEvent", {
 "as": {
  "_obj": "PDFGenericFormat",
  "_value": {
   "selection": {
    "_enum": "pdfSelection",
    "_value": "page"
   },
   "pageNumber": 1,
   "crop": {
    "_enum": "cropTo",
    "_value": "boundingBox"
   }
  }
 },
 "ID": 5,
 "null": {
  "path": "/Users/shalee/Library/Application Support/Adobe/Creative Cloud Libraries/LIBS/EED3CFB4557B1FE67F000101_adobe.com/creative_cloud/dcx/30629ecc-a840-4946-a6c4-e4c891214491/components/e1bd7c99-7e6b-488b-b8c7-4368242c32eb.pdf"
 },
 "layerName": "Hello Keyboard",
 "linked": true,
 "freeTransformCenterState": {
  "_enum": "quadCenterState",
  "_value": "QCSAverage"
 },
 "offset": {
  "_obj": "offset",
  "_value": {
[...]

PR #1982, opened 2015-08-11
Find placable representation for image elements

volfied:
Will address #1862. When we're trying to place a graphic element into the PS document, we go through it's representations to find one PS will accept. In .shape files, this is the PDF format.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -64,6 +64,43 @@ define(function (require, exports) {
     };
 
     /**
+     * List of acceptable image representations that PS can place as
+     *
+     * @type {Array}
+     */
+    var _EDITABLE_IMAGE_REPRESENTATIONS = [
+        "image/png",
+        "image/jpeg",
+        "image/jpg",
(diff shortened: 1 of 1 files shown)


src/js/actions/libraries.js
```diff
@@ -64,6 +64,43 @@ define(function (require, exports) {
...
+    var _findPlacableImageRepresentation = function (element) {
+        var representations = element.representations;
+
+        for (var i = 0; i < representations.length; i++) {
+            if (_EDITABLE_IMAGE_REPRESENTATIONS.indexOf(representations[i].type) !== -1) {
+                return representations[i];
+            }
+        }
+            
+        throw new Error("Can't find a usable representation for image element", element.name);
```
shaoshing: you might want to concatenate the strings because the second param of Error is for `filename`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

shaoshing:
Looks great. Works great as well

Issue #1966, opened 2015-08-10
Search result kinds should not be encoded in the result ID


iwehrman:
Special search results like "filters" are currently denoted by string encoding; i.e., by including the string "FILTER" as part of the ID. This is inelegant at best and dangerous at worst: what happens if a provider returns a result that happens to have the string "FILTER" as part of the ID? :bomb: is what! Instead the result objects should have additional properties that can be used to distinguish them from other kinds of results, e.g., a boolean `filter` property.

iwehrman:
Fixed by #1985.

PR #1985, opened 2015-08-11
Don't encode search item type in IDs

hamburge:
Instead, keeps a list of IDs that correspond to items with type "filter" to check if an item is a filter, or look directly at the item's type if that is easily available.

Issue #1966

### DIFF
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -48,10 +48,14 @@ define(function (require, exports, module) {
             live: React.PropTypes.bool,
             startFocused: React.PropTypes.bool,
             placeholderText: React.PropTypes.string,
-            // option to render when there are no other valid options to display
-            placeholderOption: React.PropTypes.instanceOf(Immutable.Iterable),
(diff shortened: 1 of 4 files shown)


src/js/jsx/shared/Datalist.jsx
```diff
@@ -48,10 +48,14 @@ define(function (require, exports, module) {
             live: React.PropTypes.bool,
             startFocused: React.PropTypes.bool,
             placeholderText: React.PropTypes.string,
-            // option to render when there are no other valid options to display
-            placeholderOption: React.PropTypes.instanceOf(Immutable.Iterable),
-            useAutofill: React.PropTypes.bool, // displays a suggested option next to the inputted text
-            neverSelectAllInput: React.PropTypes.bool // if true, will not highlight input text on commit
+            // Option to render when there are no other valid options to display
```
iwehrman: These are really useful comments. At some point, since you've spent so much time inside `Datalist`, would you mind trying to provide the same level of documentation for all the state properties as well?

iwehrman:
Brilliant.

Issue #1993, opened 2015-08-11
Can't select empty artboards on canvas


iwehrman:
It doesn't seem to be possible currently to select empty artboards on canvas either by clicking their title or using marquee select.

iwehrman:
Fixed by #1997.

PR #1997, opened 2015-08-11
Consider empty artboards to have visible descendants so they're sSele…

volfied:
…ctable

Addresses #1993


iwehrman:
What could go wrong?

Issue #1956, opened 2015-08-09
UI not cloaking on document selection


iwehrman:
![cloak](https://s3.amazonaws.com/f.cl.ly/items/0l1z412B0d0F2m332u0F/Screen%20Recording%202015-08-09%20at%2011.25%20AM.gif)

iwehrman:
Fixed by #1998.

Issue #1960, opened 2015-08-10
Clean up Toggle Overlay dispatches


volfied:
These calls were mainly to help with on canvas transform guides... However, with cloaking and on canvas transform being driven by Photoshop, a lot of these dispatches can be cleaned up.

iwehrman:
For example, these calls in `selectNext` and `selectPrevious` should be pushed into `selectDocument`: https://github.com/adobe-photoshop/spaces-design/blob/915d5540fab46e6635400baf5e470a5a83b9705e/src/js/actions/documents.js#L580-L582

iwehrman:
Maybe what we want is a utility which takes an action, or an array of play objects, and executes them after disabling the overlay, and then reenables the overlay afterward?

PR #1998, opened 2015-08-11
Disable overlays during document switch

volfied:
So #1956 is addressed.


iwehrman:
Splendid. Also partially addresses #1960.

PR #2010, opened 2015-08-12
Clean up of a few older components that weren't using classnames() to its full potential

mcilroyc:
The new version of `classnames()`, which replaced the old react add-on, is more powerful and smart and stuff.


volfied:
:+1:

Issue #1992, opened 2015-08-11
Secondary search info is hard to read


iwehrman:
The contrast is low, there isn't a lot of space between the primary and secondary info, and it's relatively tiny. 

![image](https://cloud.githubusercontent.com/assets/1075154/9211903/d73b9f4c-4039-11e5-8288-ad7609f8e66a.png)

placegraphichere:
Primary 
* color #e6e6e6 
* font-size:1.4rem (14px)

Secondary 
* color #7b7b7b
* font-size: 1.1rem (11px)

Highlight 
* color #2d8ceb

chadrolfs:
@DivyaPrabhakar coordinate with @iwehrman & @placegraphichere

DivyaPrabhakar:
@iwehrman @placegraphichere I got Seth's branch https://github.com/adobe-photoshop/spaces-design/tree/sewalker/search-tweaks and looked at the changes. I agree it an improvement and definitely fixes the contrast issue between varying levels of information in the search tool. I think the color differences Seth made are subtle, noticeable, and clear. My three remaining ideas about search are: 

1. Make the height of each list item in the search feature taller, giving more space to each one. We can change the font size too if we deem appropriate but I think the more vertical space itself allows it to breathe more. We already have scroll enabled so it makes little difference. In addition, the user types queries so quickly, the extra scrolling is negligible. 

![alt text](https://www.dropbox.com/s/8gqlmg3psuu6s6e/Screen%20Shot%202015-08-21%20at%202.23.30%20PM.png?raw=1 "Current state")

![alt text](https://www.dropbox.com/s/qs5ul9ebhenmmgq/Screen%20Shot%202015-08-21%20at%202.23.20%20PM.png?raw=1 "Proposed state with heightened list items")

2. Increase the space a bit between Primary and Secondary info still, since the difference in font size is not too drastic and the color is muted (rightfully so) so adding some space would further indicate they are two different but related bits of information. 

3. I think the text color, the autocomplete color, and the cursor are crowded all together. Maybe we can lessen the opacity on the autocomplete text or darken the color? 

![alt text](https://www.dropbox.com/s/86yroet2ouhdpo0/Screenshot%202015-08-21%2014.57.38.png?raw=1 "Current state")

![alt text](https://www.dropbox.com/s/sscxvo3esqupvip/Screenshot%202015-08-21%2014.58.34.png?raw=1 "Proposed state (lowered opacity")

[...]

PR #2012, opened 2015-08-12
Increase contrast between primary and secondary search info

hamburge:
Adjust colors and sizes of search text to @placegraphichere's specifications (issue #1992)

### DIFF
--- a/src/style/search/search-bar.less
+++ b/src/style/search/search-bar.less
@@ -39,7 +39,7 @@
     line-height: normal;
     box-shadow: none;
     border-radius: @dialog-radius;    
-    font-size: 1.6rem;
+    font-size: 1.4rem;
     overflow: visible;
 }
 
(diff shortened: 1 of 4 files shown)


iwehrman:
Hmmm, I guess this is more of a complaint for @placegraphichere, but I think this is going the wrong direction. Here are before and after screenshots:

image

image

This mostly seems to make all of the text slightly smaller, which I think makes it harder to read. Personally I'd be going to opposite direction and making things bigger and with more contrast. What say you, @placegraphichere?

iwehrman:
Re-assigning this to @DivyaPrabhakar. Figure out what to do!

placegraphichere:
@DivyaPrabhakar try out https://github.com/adobe-photoshop/spaces-design/tree/sewalker/search-tweaks

iwehrman:
I think that is an improvement.

iwehrman:
Obviated by recent work by @DivyaPrabhakar.

Issue #2003, opened 2015-08-12
Search dialog is not centered in the no-doc UI


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9230208/cd959b2c-40d5-11e5-9202-9099068f1b87.png)

iwehrman:
Fixed by #2013!

PR #2013, opened 2015-08-12
Center search dialog when there is no document

hamburge:
Before it would center the dialog based on your preferences, even though there is always one panel when there is no document. Now, we always assume there's one panel when there is no document open. #2003


iwehrman:
I suspect none of this takes the possibility of a minimized toolbar into account, but that's a problem for another day. Looks great, thanks!

PR #2014, opened 2015-08-12
CC Libraries - improvements

shaoshing:
this PR includes the following improvements over the CC Libraries:

* Memorize the last selected library, and restore it for new documents or reload - 1909
* Do not center library content vertically
* Can add graphic asset with long layer name - 2001
* Correctly render text style with no color - 1994
* Allow drag-n-drop on the initial library - 2008

@volfied please review

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -42,9 +42,17 @@ define(function (require, exports) {
         layerActions = require("./layers"),
         searchActions = require("./search/libraries"),
         documentActions = require("./documents"),
-        pathUtil = require("js/util/path");
+        pathUtil = require("js/util/path"),
+        preferencesActions = require("./preferences");
 
-    /** 
+    /**
+     * Preference name for storing the ID of the last selected library.
+     *
+     * @private
+     */
+    var _LAST_SELECTED_LIBRARY_ID_PREF = "lastSelectedLibraryID";
+
+    /**
      * For image elements, their extensions signify their representation type
      *
      * @type {Object}
@@ -97,7 +105,7 @@ define(function (require, exports) {
                 return representations[i];
             }
         }
-            
+
         throw new Error("Can't find a usable representation for image element: " + element.name);
     };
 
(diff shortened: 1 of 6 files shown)


src/js/actions/libraries.js
```diff
@@ -141,14 +149,14 @@ define(function (require, exports) {
                 representationType = _EXTENSION_TO_REPRESENTATION_MAP[fileExtension];
             }
         }
+        
+        var tempLayerName = (new Date().getTime()).toString();
```
shaoshing: this is a walk-around for #2015

src/js/actions/libraries.js
```diff
@@ -544,10 +552,18 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var selectLibrary = function (id) {
-        return this.dispatchAsync(events.libraries.LIBRARY_SELECTED, { id: id });
+        var updatePreferencePromise = this.transfer(
+            preferencesActions.setPreference, _LAST_SELECTED_LIBRARY_ID_PREF, id);
+        
+        return Promise.resolve(updatePreferencePromise)
```
volfied: I'd say dispatching the library selected event is higher priority than setting the preference here.

src/js/jsx/sections/libraries/assets/CharacterStyle.jsx
```diff
@@ -88,10 +92,14 @@ define(function (require, exports, module) {
                 <div className="libraries__asset__desc">
                     <div>{font.name} {font.style}</div>
                     <div className="libraries__asset__desc-details">
-                        {fontSize.value}{fontSize.type}, {fontColorHex}
+                        {fontSize.value}{fontSize.type}
```
volfied: what if `fontSize` is null?

volfied: This happened to me with one of my older libraries, and I couldn't get things to work again.

volfied:
Looks good, but while you're in there, if you can make CharacterStyle.jsx a bit more robust with some null checks, it would help.

Also needs rebase.

shaoshing:
Added more checking when rendering Charstyle. Back to you @volfied

volfied:
:+1:

iwehrman:
Hey @shaoshing: if you stick a # in front of your bug numbers (e.g., in your PR summary), it will both link to that bug automatically and also create back-links from the issue to this PR. The former is useful in the short term, and the latter is often useful in the long term. #protip.

shaoshing:
@iwehrman thanks for the tip :100:

PR #2016, opened 2015-08-13
Allow setting the font size on transformed layers

iwehrman:
Longer term, we want to scale the font sizes in text layers that are transformed on canvas for both typographic and asset-export reasons. For the short term, this pull request changes how font sizes are interpreted in affine-transformed text layers such that a font size that is implied by that transform is now used instead of the literal font size used by the text rendering engine.

implied font size

More concretely, this pull request: 1) loads `impliedFontSize` into the `textSize` property of the `CharacterStyle` models, 2) removes the `hasTransform` property in text models, and 3) removes the checks that disable the `Size` component in the transform panel and the size input of the `Text` component. It's also now necessary to 4) update the `textSize` property whenever a layer is text layer resized, so when that happens we call `resetLayers` instead of just `resetBounds`. (I figured that since the `textStyle` property is apparently by far the most expensive property to fetch from layer descriptors, it didn't make sense to add a special `resetFontSize` action, though we certainly could for **_extreme performance_**.)

Aside: surprisingly, this does not seem to require an adapter update because setting the `size` property of the `textStyle` property apparently does the same thing as setting the `impliedFontSize` property, which sure seems like a bug to me.

This is a strategic retreat, but, fear not @timriot, we are still fighting this good fight!

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1124,9 +1124,18 @@ define(function (require, exports) {
             this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
 
             var appStore = this.flux.store("application"),
-                currentDoc = appStore.getCurrentDocument();
+                currentDoc = appStore.getCurrentDocument(),
+                textLayers = currentDoc.layers.allSelected.filter(function (layer) {
+                    // Reset these layers completely because their impliedFontSize may have changed
+                    return layer.kind === layer.layerKinds.TEXT;
+                }),
+                otherLayers = currentDoc.layers.allSelected.filterNot(function (layer) {
+                    return layer.kind === layer.layerKinds.TEXT;
+                }),
+                textLayersPromise = this.flux.actions.layers.resetLayers(currentDoc, textLayers),
+                otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers);
 
-            return this.flux.actions.layers.resetBounds(currentDoc, currentDoc.layers.allSelected);
+            return Promise.join(textLayersPromise, otherLayersPromise);
         }, this);
 
         descriptor.addListener("transform", _layerTransformHandler);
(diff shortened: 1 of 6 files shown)


iwehrman:
PS: I hit [4035824](https://watsonexp.corp.adobe.com/#bug=4035824) frequently when working on this.

src/js/actions/transform.js
```diff
@@ -1124,9 +1124,18 @@ define(function (require, exports) {
...
+                currentDoc = appStore.getCurrentDocument(),
+                textLayers = currentDoc.layers.allSelected.filter(function (layer) {
+                    // Reset these layers completely because their impliedFontSize may have changed
+                    return layer.kind === layer.layerKinds.TEXT;
+                }),
+                otherLayers = currentDoc.layers.allSelected.filterNot(function (layer) {
+                    return layer.kind === layer.layerKinds.TEXT;
+                }),
+                textLayersPromise = this.flux.actions.layers.resetLayers(currentDoc, textLayers),
+                otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers);
```
mcilroyc: I guess the right thing will happen downstream if there are zero "non-text selected layers" but it still catches my attention that we may be calling this action needlessly in many cases.

mcilroyc:
LGTM.  @iwehrman tell me you're not concerned about my other comment, and I'll merge it.

src/js/actions/transform.js
iwehrman: These sorts of checks ought to happen exactly once, as deep as possible, so that clients do not constantly have to check for this condition. (I think this is the right way to do it because 1 check has many fewer lines of potentially buggy code than N different checks.) 

Both `resetLayers` and `resetBounds` correctly handle the case when they are passed empty lists, but they don't drop out early. I've pushed commit 7bedadc which additionally causes those actions to resolve immediately when they're applied to empty lists. So I'm now satisfied that it is correct to omit the check here. Good point of discussion.

iwehrman:
Thanks for the review, @mcilroyc. Back to you.

PR #2017, opened 2015-08-13
Prevent incorrect search option from being selected occasionally on click

hamburge:
This would happen because the updating of the search suggestion would occur when Datalist's filter value went from `null` to `""`. This meant that if you had no text in the input and clicked on an item, the first item would always be selected.


iwehrman:
Lovin' it :hamburger: :fries:

PR #2027, opened 2015-08-13
Sewalker/icons more sharper

placegraphichere:
Almost all current icons have been refreshed as well as their LESS size and position information if that was necessary.


designedbyscience:
Noticed couple things, probably from my merge:
untitled-9___48_5___group_1__rgb_8____

Is the Not Visible icon supposed to be bigger?
And, the colors got messed up there for the selected state

placegraphichere:
Addressed both of these comments
- Is the Not Visible icon supposed to be bigger?
- And, the colors got messed up there for the selected state

iwehrman:
image

placegraphichere:
tumblr_inline_ns4ch2aofc1r1fj7r_250

shaoshing:
shiny

designedbyscience:
These look great

Issue #2024, opened 2015-08-13
Duplicate keys possible with Libraries Search


iwehrman:
I think this happens when you have two assets in a Library with the same name?
```
Warning: flattenChildren(...): Encountered two children with the same key, `.$FILTER-LIBRARY-New Library`. Child keys must be unique; when two children share a key, only the first child will be used.
```
![image](https://cloud.githubusercontent.com/assets/1075154/9259533/aa3b0c84-41b6-11e5-86b8-43ab41c6afa7.png)

iwehrman:
Fixed #2030.

PR #2030, opened 2015-08-13
Handle searching for libraries that have identical names

hamburge:
Issue #2024. 
There's currently no way to tell the difference between filtering for one library versus another when they have the same name, but it will at least properly filter when you select either of the two options.

I also changed back how Datalist updates when you apply a filter to a forceUpdate instead of checking if the icon changes, because when you have the "all libraries" filter selected and then pick a specific library, the icon won't change and the Datalist has no way of updating.

### DIFF
--- a/src/js/actions/search/libraries.js
+++ b/src/js/actions/search/libraries.js
@@ -91,7 +91,7 @@ define(function (require, exports) {
                         name: title,
                         pathInfo: library.name + ": " + strings.SEARCH.CATEGORIES[categoryKey],
                         iconID: _getSVGClass([categoryKey]),
-                        category: ["LIBRARY", library.name, categoryKey]
+                        category: ["LIBRARY", library.id.replace(/-/g, "."), categoryKey]
                     });
                 }
             });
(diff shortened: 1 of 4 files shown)


src/js/actions/search/libraries.js
```diff
@@ -91,7 +91,7 @@ define(function (require, exports) {
                         name: title,
                         pathInfo: library.name + ": " + strings.SEARCH.CATEGORIES[categoryKey],
                         iconID: _getSVGClass([categoryKey]),
-                        category: ["LIBRARY", library.name, categoryKey]
+                        category: ["LIBRARY", library.id.replace(/-/g, "."), categoryKey]
```
iwehrman: Why is this necessary? Is there still some instance of string encoding in which hyphens are treated specially?

iwehrman:
I don't yet have any major complaints, but I'd like to learn about the string encoding thing before pushing the big green button.

src/js/actions/search/libraries.js
hamburge: @iwehrman Yes, the search store makes every item ID its broad category and some unique ID, separated by a hyphen. Then later the search store splits the ID on the hyphen. This should only really be an issue when a category is a user inputted string, because otherwise we are using the keys to localized strings.

The reason there's been a tendency to store information in the ID is because the whole Datalist/Select structure only passes around IDs when something is selected or confirmed, so I've been trying not to overhaul that.

iwehrman: OK, this is a longer term architectural problem which we should figure out how to address in the future. This seems good for now though. Thanks!

Issue #2025, opened 2015-08-13
Long result titles should be clipped


iwehrman:
:scream: 
![image](https://cloud.githubusercontent.com/assets/1075154/9259779/800944f6-41b8-11e5-8955-714258db0afa.png)

iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9259805/a8666172-41b8-11e5-972b-8a9740fad5e8.png)

iwehrman:
Fixed by #2032.

PR #2032, opened 2015-08-13
Clip search bar suggestion if it is longer than the text box

hamburge:
Issue #2025. Adjust the suggestion width to be (width of the whole text input) - (width of entered text)

@iwehrman can now open up an image of his favorite tweets without screaming :scream:


iwehrman:
Now it trims too much! :scream_cat: 
image

hamburge:
@iwehrman True-- try now :smile_cat:

iwehrman:
Perfectamundo.

Issue #1983, opened 2015-08-11
CC Lib: unable to apply Fill/Stroke color or Layer Styles


chadrolfs:
new doc, drag out a rect
load a library that has fill and stroke colors and layer styles

mouse over cc lib assets (Click to apply): click 
Result: nothing is applied

Drag and drop doesn't work, double-click doesn't work

shaoshing:
Hi @chadrolfs , this is implemented and merged into master. note that you have to click on the preview to apply an asset.

PR #2033, opened 2015-08-13
CC Libraries - apply char style / layer style / color / color theme

shaoshing:
Now you can apply char style / layer style / color / color theme to the selected layers.

This also addresses issue #1983


src/js/actions/libraries.js
```diff
@@ -514,31 +520,34 @@ define(function (require, exports) {
...
 
         var representation = element.getPrimaryRepresentation(),
             styleData = representation.getValue("characterstyle", "data"),
-            layerRef = textLayerAdapter.referenceBy.current,
+            textLayerIDs = collection.pluck(textLayers, "id"),
+            layerRef = textLayerIDs.map(textLayerAdapter.referenceBy.id).toArray(),
             applyObj = textLayerAdapter.applyTextStyle(layerRef, styleData);
 
-        return descriptor.playObject(applyObj)
+        return layerActionsUtil.playSimpleLayerActions(currentDocument, textLayers, applyObj, true)
```
shaoshing: I have to make the changes here to only apply char style to text layers to address an issue (Set command is not available) when you select all layers (including the art board).

src/js/actions/type.js
```diff
@@ -264,7 +265,7 @@ define(function (require, exports) {
             playObject = [playObject].concat(setOpacityPlayObjects);
         }
         var updatePromise = this.transfer(updateColor, document, layers, color, coalesce, true),
-            setColorPromise = locking.playWithLockOverride(document, layers, playObject, typeOptions);
+            setColorPromise = layerActionsUtil.playSimpleLayerActions(document, layers, playObject, true, typeOptions);
```
shaoshing: This also address the `Set command is not available` issue when apply font color to all layers (including the art board)

shaoshing:
@volfied please review.

src/js/actions/libraries.js
```diff
@@ -514,31 +520,34 @@ define(function (require, exports) {
...
-     * @param {AdobeLibraryElement} element [description]
+     * @param {AdobeLibraryElement} element
      *
      * @return {Promise}
      */
     var applyCharacterStyle = function (element) {
         var appStore = this.flux.store("application"),
-            currentDocument = appStore.getCurrentDocument();
+            currentDocument = appStore.getCurrentDocument(),
+            selectedLayers = currentDocument ? currentDocument.layers.selected : new Immutable.List(),
```
volfied: you don't need `new` here.

src/js/actions/libraries.js
```diff
@@ -546,6 +555,50 @@ define(function (require, exports) {
...
+                return textLayers.isEmpty() ? null :
+                    this.transfer(typeActions.setColor, currentDocument, textLayers, color, true, true);
+            })
+            .then(function () {
+                return vectorLayers.isEmpty() ? null :
+                    this.transfer(shapeActions.setFillColor, currentDocument, vectorLayers, 0, color,
+                        true, true, true);
+            })
+            .then(function () {
+                return this.transfer(layerActions.resetLayers, currentDocument, textLayers.concat(vectorLayers));
```
volfied: Do you need to resetLayers for color changes? Make sure it's absolutely necessary

src/js/actions/libraries.js
```diff
@@ -546,6 +555,50 @@ define(function (require, exports) {
...
+            vectorLayers = selectedLayers.filter(function (l) { return l.isVector(); });
+
+        if (textLayers.isEmpty() && vectorLayers.isEmpty()) {
+            return Promise.resolve();
+        }
+
+        // FIXME: Setting font color and fill color at the same time will result in two histroy states.
+        //        We should merge the two states when transaction is supported.
+        return Promise.bind(this)
+            .then(function () {
```
volfied: You should use Promise.join instead, look at actions/sampler.js/pasteStyle for an example.

src/js/actions/type.js
volfied: Given that you've fixed applyColor in libraries, and you only pass text layers to this function, this change should be unnecessary.

volfied: Also, the only reason playSimpleLayerActions worked but playWithLockOverride didn't is latter skips errors.

src/js/jsx/sections/libraries/assets/AssetSection.jsx
```diff
@@ -0,0 +1,127 @@
...
+         *
+         * @private
+         */
+        _listenTitleEvent: function () {
+            if (!this.refs.assetTitle) {
+                return;
+            }
+            
+            var assetTitleNode = React.findDOMNode(this.refs.assetTitle);
+            assetTitleNode.addEventListener("dblclick", this._handleRename);
```
volfied: Is there a reason for not using React's own `onDoubleClick` handler?

src/js/jsx/sections/libraries/assets/AssetSection.jsx
```diff
@@ -0,0 +1,127 @@
...
+        },
+
+        /**
+         * Handle select asset event.
+         * @param  {SynthenticEvent} event
+         *
+         * @private
+         */
+        _handleSelect: function (event) {
+            if (event.target !== React.findDOMNode(this)) {
```
volfied: This seems weird as well.. clickEvent should only come here if this component is indeed the target. If you're getting click events from one of it's children, you probably want to stopPropagation in the children's click handlers.

volfied:
Looks good, but I have a lot of reservations about the implementation of AssetSection. Back to you @shaoshing

src/js/actions/type.js
shaoshing: The previous use of `playSimpleLayerActions` is for `applyCharacterStyle`, so it is necessary here to make `applyColor` works. I checked how `playSimpleLayerActions` works and looks like it will first select the desired layers for "Set" command, and then restore the previous select; this fix the issue when all layers are selected while apply fond color.

src/js/actions/libraries.js
shaoshing: Confirmed that both fills#setColor and shapes#setFillColor will dispatch the color change event, so we don't need the resetLayers action here.

src/js/jsx/sections/libraries/assets/AssetSection.jsx
shaoshing: assetTitle is a reference of its children `this.props.children[0]`, and the regular React way of binding event is not available here. However, I figured out I can set the `onDoubleClick` option in the `React.cloneElement`. A optimized version will arrive soon

src/js/jsx/sections/libraries/assets/AssetSection.jsx
shaoshing: The reason is that we are trying to support both click (select) and double click (rename / edit), and a double click event will also trigger a click event. I agree with you that this is not the optimal way, and I will find a better solution when I am working on the rename and edit feature.

shaoshing:
hey @volfied , I made some improvements (thanks for the feedback), along with two additional fixes:

* missing collaborator icon
* missing add-text-color button in the contextual bar.

Please review again.

shaoshing:
Resolved a bug of setting color for unexpanded group(s). https://github.com/adobe-photoshop/spaces-design/pull/2033/files#diff-d036559733279e3c642c1fffe3d0fbc8R568

src/js/actions/libraries.js
```diff
@@ -157,6 +162,7 @@ define(function (require, exports) {
             .bind(this)
             .then(function (tempFilename) {
                 // Export the selected layers
+
```
volfied: :space_invader: (unnecessary newline)

src/js/actions/libraries.js
```diff
@@ -41,9 +42,13 @@ define(function (require, exports) {
         locks = require("../locks"),
         layerActions = require("./layers"),
         searchActions = require("./search/libraries"),
+        shapeActions = require("./shapes"),
```
volfied: While you're in here, can you also sort this require list to make it a bit more readable.

src/js/actions/libraries.js
```diff
@@ -546,6 +555,39 @@ define(function (require, exports) {
...
+     *  - Text layer: will set layer's font color
+     *  - Vector layer: will set layer's fill color
+     *
+     * @param {Color} color
+     *
+     * @return {Promise}
+     */
+    var applyColor = function (color) {
+        var currentDocument = this.flux.store("application").getCurrentDocument(),
+            selectedLayers = currentDocument ? currentDocument.layers.allSelected : null;
```
volfied: You probably want to set it to Immutable.List() as well like above.

src/js/jsx/sections/libraries/LibraryBar.jsx
```diff
@@ -165,7 +170,22 @@ define(function (require, exports, module) {
 
         render: function () {
             var adobeStockURL = "https://stock.adobe.com";
-            
+
+            var addFillButton = this._createColorButton(function (layer) {
+                return layer.fills.isEmpty() ? null :
```
volfied: Do we want these to be null, or disabled? Is that going to be in a separate PR?

src/js/jsx/sections/libraries/assets/AssetSection.jsx
```diff
@@ -0,0 +1,105 @@
...
+        },
+
+        /**
+         * Handle rename asset event.
+         * @param  {SynthenticEvent} event
+         *
+         * @private
+         */
+        _handleRename: function (event) {
+            event.stopPropagation();
```
volfied: Can you add a `// FIXME: Implement` here?

src/js/jsx/sections/libraries/assets/CharacterStyle.jsx
```diff
@@ -80,40 +74,41 @@ define(function (require, exports, module) {
                 fontSize = charStyle.fontSize,
```
volfied: Probably a safe bet to just return `null` if charStyle.adbeFont is null.

src/js/jsx/sections/libraries/assets/ColorTheme.jsx
```diff
@@ -29,56 +29,57 @@ define(function (require, exports, module) {
...
-                
-            var colorSwatchComponents = colorSwatches.map(function (color) {
-                var colorHex = tinycolor(color[0].value).toHexString().toUpperCase();
-                return (<div key={colorHex} style={{ background: colorHex }} title={colorHex}/>);
+
+            var colorSwatchComponents = colorSwatches.map(function (colors) {
+                var colorData = colors[0],
+                    colorHex = tinycolor(colorData.value).toHexString().toUpperCase();
+
+                return (<div key={colorHex}
```
volfied: If the color theme has the same colors, this is going to have identical keys and React will be mad at us. Maybe use the `index` coming from the map function here instead?

volfied:
One more round, looks great! @shaoshing

src/js/jsx/sections/libraries/LibraryBar.jsx
shaoshing: It will be null here, same as standard PS. But we can definitely have a separate PR if we decide to show disabled button instead.

src/js/jsx/sections/libraries/assets/ColorTheme.jsx
shaoshing: Good catch!

shaoshing:
Thanks for the review @volfied . I added more changes to address your comments.

volfied:
You forgot two nits... but they're minor enough to not stop progress

PR #2034, opened 2015-08-13
JSON localization dictionaries

iwehrman:
The Loc teams prefers that our localization dictionaries be in JSON files instead of JS files. This pull request accommodates that preference. Unfortunately there doesn't seem to be a good way to do this with the RequireJS i18n plugin (see requirejs/i18n#12), so I'm doing it a bad way! This is also going to create rebase difficulties for anyone who is working on a branch with changed strings, but ¯\\_(ツ)_/¯


volfied:
Needs more rebase

volfied:
Oh boy, rebases galore.

PR #2035, opened 2015-08-13
Close search when window loses focus

hamburge:
Otherwise the search will lose focus but remain open, and bad things will happen when you come back to the window.

This is better than the potential for those bad things, but is also maybe not the most elegant solution-- the dialog does not actually close until you come back to the window, which means you see it disappear after a short delay.

### DIFF
--- a/src/js/jsx/Search.jsx
+++ b/src/js/jsx/Search.jsx
@@ -82,6 +82,7 @@ define(function (require, exports, module) {
                         dismissOnCanvasClick={true}
                         dismissOnWindowClick={true}
                         dismissOnWindowResize={false}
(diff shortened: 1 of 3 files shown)


iwehrman:
Hmm, I don't know if I like this. What if I start typing something but then need to peek at another window to figure out what to keep typing? Also, when I hit cmd-F again, my partial text is gone, so there could be frustration. What if instead when the window gets focus, the search bar re-focuses the input if it's open?

hamburge:
Made issue #2057 for this

iwehrman:
Closing this out. I'll tackle #2057 separately.

Issue #2029, opened 2015-08-13
Panels blink when scrolling or zooming a document


chadrolfs:
https://www.dropbox.com/s/mqm4n09hm108ymj/PanelBlink.mov?dl=0

iwehrman:
I smell an overlay offsets bug!

designedbyscience:
@chadrolfs Anything particular to replicate this (not seeing it with a few different variations on opening/making new doc/refreshing DS)?

chadrolfs:
grab the latest jenkins build (pgdev 454), launch into DS (optional: connect to master), cmd/ctrl + n, use mousewheel to scroll doc.

chadrolfs:
once you toggle a panel, it stops.

iwehrman:
#2045 attempted to fix this, but did not do so completely.

designedbyscience:
Addressing again by #2046, ready for checking

Issue #2022, opened 2015-08-13
Panel toolbar icons need tooltips


iwehrman:
These guys:
![image](https://cloud.githubusercontent.com/assets/1075154/9258814/767d1d1e-41b2-11e5-992b-a644f09e240f.png)

iwehrman:
Fixed by #2045.

Issue #2021, opened 2015-08-13
Libraries panel horizontal rule should not be visible when collapsed


iwehrman:
The little grey/beige hairline at the bottom of the screen:
![image](https://cloud.githubusercontent.com/assets/1075154/9258773/38a4e3f0-41b2-11e5-868a-e3671c66b838.png)

iwehrman:
Not sure if @designedbyscience or @shaoshing should get this one. Assigning to @designedbyscience for triage.

iwehrman:
Fixed by #2045.

Issue #1814, opened 2015-07-23
Unpinned toolbar is lonely in no-doc view


iwehrman:
![lonely](https://s3.amazonaws.com/f.cl.ly/items/2L3c3P3W0v2j3T1C310T/Image%202015-07-23%20at%202.31.58%20PM.png)

CC @placegraphichere

iwehrman:
Modest proposal: if the toolbar is unpinned/minimized, just hide it in the no-doc view.

iwehrman:
Fixed by #2045.

PR #2045, opened 2015-08-14
Several UI Fixes

designedbyscience:
Fixes for:

#1814 Unpinned toolbar is lonely in no-doc view
#2022 Panel toolbar icons need tooltips
#2021 Libraries panel horizontal rule should not be visible when collapsed
#2029 Panels blink when scrolling or zooming a document


iwehrman:
This doesn't completely fix #2029. I still see the panel being partially cloaked when going from no documents to one document, especially right after a new doc is created. The first pan of the canvas triggers it. But there's nothing wrong with the proposed fix (the rest of the fix probably lies elsewhere), and everything else looks good, so I'm merging this.

PR #2046, opened 2015-08-14
Proper check for no doc to some doc state

designedbyscience:
No more flashing from no-doc state to new or open document


mcilroyc:
LGTM.  My minor concerns were allayed via slack.

PR #2052, opened 2015-08-14
Allow copy/paste/select all in search bar open when there is no document

hamburge:
These actions only need a document if there is no input modal open.

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -141,13 +141,13 @@ define(function (require, exports, module) {
             return {
                 "always": true,
                 "always-except-modal": false,
-                "modal-have-document": appIsInputModal || (document !== null)
+                "super-search-or-have-document": appIsInputModal
             };
         }
         return {
             "always": true,
             "always-except-modal": true,
-            "modal-have-document":
+            "super-search-or-have-document":
                 (document !== null),
             "have-document":
                 (document !== null),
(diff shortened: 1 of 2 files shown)


iwehrman:
I'm not familiar enough with this code to understand the ramifications of this change. Over to @mcilroyc?

src/js/models/menubar.js
```diff
@@ -141,7 +141,7 @@ define(function (require, exports, module) {
             return {
                 "always": true,
                 "always-except-modal": false,
-                "modal-have-document": appIsInputModal && (document !== null)
+                "modal-have-document": appIsInputModal || (document !== null)
```
mcilroyc: I think this that has one unintended consequence, and that "modal-have-document' is no longer a good name for this prop.

I think what we are now *intending* is "*super search OR have-document*" which currently implies to "*cut/copy/paste/selectAll is enabled*". 

But the logic as of this commit is "super search OR have-document OR (help screens && have-document)".  So, for example, we now allow select-all while looking at the "keyboards shortcut" modal dialog, which has the effect of selecting all layers on the underlying document.  I don't *think* we want that.

I suggested changing this (and associated references to):
`super-search-or-have-document: appIsInputModal`

mcilroyc:
Fixed this PR per my comment above.  back to @iwehrman for a second opinion & merge.

iwehrman:
LGTM!

PR #2056, opened 2015-08-14
Export style fixes; adding first asset enables layer for export; moved 'add' button

mcilroyc:
This PR contains a few minor style tweaks, and a semi-major refactoring of LayerExports & ExportPanel.

Style...

1. Remove Asset (trashcan) was sized wrong
1. Export headers alignment (esp suffix) was wrong
1. Collapse Export Panel didn't work at all (only partially works now)
1. The "add asset" button needed to be moved up (adjacent to panel title)

New feature: When adding the first asset, the layer is automatically enabled for export (ie, it is "checked" in the export all dialog)

Refactor: as part of moving the "add asset" plus button, I needed to move logic up in to the ExportPanel, which is now the component that watches the export store.  I think this improved the organization a bit.

### DIFF
--- a/src/js/jsx/sections/export/ExportPanel.jsx
+++ b/src/js/jsx/sections/export/ExportPanel.jsx
@@ -26,16 +26,33 @@ define(function (require, exports, module) {
     "use strict";
 
     var React = require("react"),
-        classnames = require("classnames");
+        Fluxxor = require("fluxxor"),
+        FluxMixin = Fluxxor.FluxMixin(React),
+        StoreWatchMixin = Fluxxor.StoreWatchMixin,
+        classnames = require("classnames"),
+        _ = require("lodash");
 
     var os = require("adapter/os");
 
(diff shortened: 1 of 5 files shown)


src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -26,16 +26,33 @@ define(function (require, exports, module) {
...
-        synchronization = require("js/util/synchronization");
+        synchronization = require("js/util/synchronization"),
+        collection = require("js/util/collection");
+
+    /**
+     * A simple array of scale values, used to determine "next" scale when adding a new asset
+     * @private
+     * @type {Array.<number>}
+     */
+    var _allScales = [0.5, 1, 1.5, 2];
```
iwehrman: Maybe in the next iteration you could put the supported scales and formats in the export store so that this, e.g., wouldn't need to be duplicated in two different UI components?

iwehrman:
One suggestion for the next pass. Looks good otherwise.

Issue #2053, opened 2015-08-14
Search bar overflow ellipsis uses blue color


placegraphichere:
![screen shot 2015-08-14 at 3 03 50 pm](https://cloud.githubusercontent.com/assets/3229167/9283176/4e421b48-4297-11e5-8654-ac8b43b8da94.png)

iwehrman:
Fixed by something or other.

PR #2058, opened 2015-08-14
Make search item overflow ellipsis color match secondary info

hamburge:
Uses flex display to cut off additional search info. If the primary info is very long it will have a primary-info-sized ellipsis as well. Issue #2053


designedbyscience:
Looks good

Issue #1969, opened 2015-08-10
Can't create new point text layers


iwehrman:
As of the 8/8 PG-Dev build, when you click the canvas to create a new text layer you get both a "make" event and a "createTextLayer" event, both of which call the `layers.addLayers` action to add the same layer ID, which causes problems.

At one point we were not getting "make" events at the _start_ of text editing, which is why @pineapplespatula added the new "createTextLayer" event. But it seems like we're getting both.

iwehrman:
```
Enqueuing action tools.changeModalState; 0/0
fluxcontroller.js:538 Executing action tools.changeModalState after waiting 0ms; 1/0
fluxcontroller.js:517 Enqueuing action tools.changeModalState; 1/0
fluxcontroller.js:517 Enqueuing action layers.addLayers; 1/1
fluxcontroller.js:526 Killing modal state for action layers.addLayers
fluxcontroller.js:517 Enqueuing action type.updateSize; 2/1
fluxcontroller.js:517 Enqueuing action type.updatePostScript; 2/2
fluxcontroller.js:517 Enqueuing action type.updateColor; 2/3
fluxcontroller.js:517 Enqueuing action type.updateTracking; 2/4
fluxcontroller.js:517 Enqueuing action type.updateAlignment; 2/5
fluxcontroller.js:517 Enqueuing action type.updateLeading; 2/6
fluxcontroller.js:548 Finished action tools.changeModalState in 33ms with RTT 33ms; 2/7
fluxcontroller.js:517 Enqueuing action tools.changeModalState; 2/7
fluxcontroller.js:517 Enqueuing action tools.changeModalState; 2/8
history.js:271 History state event from photoshop (raw): currentState (index) 5, total states: 5
fluxcontroller.js:517 Enqueuing action history.handleHistoryState; 2/8
fluxcontroller.js:538 Executing action tools.changeModalState after waiting 42ms; 2/9
fluxcontroller.js:517 Enqueuing action layers.addLayers; 2/9
fluxcontroller.js:538 Executing action layers.addLayers after waiting 0ms; 2/10
fluxcontroller.js:548 Finished action tools.changeModalState in 48ms with RTT 90ms; 2/10
history.js:525 Pushing history for an action for which we expect to have gotten a rogue history event
history.js:624 History state added or merged: 5
fluxcontroller.js:466 Verifying 2 postconditions for action layers.addLayers
[...]

PR #2059, opened 2015-08-14
Work around occasional duplicate PS make event

iwehrman:
This works around an apparent PS race condition in which we sometimes get both a "make" and a "createTextLayer" event at the beginning of the type modal tool state. If it's the latter, and it happens after the make event, then bad things happen. With this PR, if that happens we now log a warning and reset the selected layers. 

Addresses #1969


pineapplespatula:
Looks good to me

PR #2060, opened 2015-08-14
Singularize stroke and fill for layer models

volfied:
This one is a hairy PR, so @chadrolfs and @ktaki, please slam on the fill/stroke related workflows a little on this PR before it's merged.

Clears out the layer model to have a singular stroke and fill information, since we essentially kept a list of them when there could only be one. :sword: Gets rid of all the pesky fill/strokeIndex variables being thrown around, and makes sure style sampling/pasting and library style application work right.

@iwehrman can you review this bad boy?

### DIFF
--- a/src/js/jsx/sections/style/Fill.jsx
+++ b/src/js/jsx/sections/style/Fill.jsx
@@ -58,7 +58,7 @@ define(function (require, exports, module) {
         getInitialState: function () {
             return {
                 layers: Immutable.List(),
-                fill: {}
+                fill: null
             };
         },
 
@@ -177,7 +177,7 @@ define(function (require, exports, module) {
 
         render: function () {
             // If there are no vector layers, hide the component
-            if (this.state.layers.isEmpty()) {
+            if (!this.state.fill || this.state.layers.isEmpty()) {
                 return null;
             }
 
--- a/src/js/jsx/sections/style/Fill.jsx
+++ b/src/js/jsx/sections/style/Fill.jsx
@@ -22,7 +22,7 @@
  */
 
 
-define(function (require, exports) {
+define(function (require, exports, module) {
     "use strict";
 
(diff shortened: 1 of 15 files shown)


src/js/jsx/sections/style/Fill.jsx
```diff
@@ -52,10 +50,31 @@ define(function (require, exports) {
...
-                this.props.readOnly !== nextProps.readOnly;
+        shouldComponentUpdate: function (nextProps, nextState) {
+            return !Immutable.is(this.state.fill, nextState.fill) ||
+                this.props.disabled !== nextProps.disabled;
+        },
+
+        getInitialState: function () {
+            return {
+                layers: Immutable.List(),
+                fill: {}
```
iwehrman: That's weird, since even when it's set it's not a bare object. Seems like this should be null?

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -53,10 +51,31 @@ define(function (require, exports) {
...
-                this.props.readOnly !== nextProps.readOnly;
+        shouldComponentUpdate: function (nextProps, nextState) {
+            return !Immutable.is(this.state.stroke, nextState.stroke) ||
+                this.props.disabled !== nextProps.disabled;
+        },
+
+        getInitialState: function () {
+            return {
+                layers: Immutable.List(),
+                stroke: {}
```
iwehrman: Ditto.

iwehrman:
Just one set of nits. Entirely possible that I'm missing something since this is, as you indicated, rather nasty. Probably best to just merge it ASAP to get wider testing.

volfied:
@iwehrman Thank you for the review, I've set them to be null in initial state, and not render if they're.

iwehrman:
LGTM!

Issue #1905, opened 2015-08-04
Artboards: Add new artboard using file menu


placegraphichere:
Allow addition of a adding new artboard templates to canvas using the file menu.

**Layer > New Artboard > [options]**

We can grab the [options] from current Ps template options for artboards (I believe there are 10 or so).

![filemenu-layers](https://cloud.githubusercontent.com/assets/3229167/9048852/ffebe07a-3a01-11e5-914a-598138b609b8.png)

chadrolfs:
Basically duplicating File > New From Template, right?

chadrolfs:
Should this be a trello card?  This isn't really a bug/issue.

DivyaPrabhakar:
https://github.com/adobe-photoshop/spaces-design/pull/2063

- Added the menu specified above
- If no preset is defined and an artboard is selected, when new artboard is created, it takes the size of the selected artboard. Else, if no preset is defined and an artboard is not selected, the new one takes the default size.

iwehrman:
![ab](https://s3.amazonaws.com/f.cl.ly/items/2S2F293S0U1P3n3a3M45/Screen%20Recording%202015-08-19%20at%2003.25%20PM.gif)

placegraphichere:
this is radical and exciting

DivyaPrabhakar:
Wooooooooo

iwehrman:
@placegraphichere you have the power to close this bug once you are satisfied with the product.

placegraphichere:
I am coming across an issue where I add an artboard and it will often times freeze the UI completely. I am able to return to classic, but am not able to return to Ds. I'm on dev 507.

placegraphichere:
I've bower updated, npm updated, updated both my repos and pulled down dev 511 and still hit this issue on creation of a second artboard with the file menu

DivyaPrabhakar:
I can look into this in about an hour or so and see if I can find the problem/repro it

DivyaPrabhakar:
@placegraphichere I assigned this to myself because I have talked to Jesper about it and I think the problem is on our side, therefore I am going to try and isolate it. In the meantime, can you try to repro this problem but DO NOT deselect the new artboard after you have created it, just keep creating artboards. On my end, it is able to do this without problem.

iwehrman:
In the future, it would have been better to close this when the feature landed. The PR #2227 is really addressing a different issue entirely.

iwehrman:
[...]

PR #2063, opened 2015-08-15
Added new functionality to the Menu bar - Create Art From Template

DivyaPrabhakar:
createArtboard can now handle inputs of type Bounds, payload, and undefined; if artboard selected, uses that as default size if no preset or Bounds have been passed in. Addresses issue #1905

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -49,7 +49,11 @@ define(function (require, exports) {
         locks = require("js/locks"),
         locking = require("js/util/locking"),
         headlights = require("js/util/headlights"),
-        strings = require("i18n!nls/strings");
+        strings = require("i18n!nls/strings"),
+        Bounds = require("js/models/bounds");
+
+    var templatesJSON = require("text!static/templates.json"),
+        templates = JSON.parse(templatesJSON);
 
     var PS_MAX_NEST_DEPTH = 9;
 
@@ -1452,45 +1456,151 @@ define(function (require, exports) {
     };
 
     /**
+     * Helper function to find the right most artboard. It assumes a non-empty list of artboards. 
+     * 
+     * @private 
+     * @param {Immutable.List.<Layer>} artboards
+     * @return {Layer}
+     */
+    var _findRightMostArtboard = function (artboards) {
+        var layer = artboards.reduce(function (selectedLayer, currentLayer) {
+            if (currentLayer.bounds.right > selectedLayer.bounds.right) {
+                return currentLayer;
+            } else {
+                return selectedLayer;
+            }
+        }, artboards.first());
+        return layer;
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     *
+     * @private
+     * @param {object} specs 
+     * @param {Immutable.List.<Layer>} artboards
+     * @return {{finalBounds: Bounds, changeLayerRef: boolean}}
+     */
+    var _getBoundsFromTemplate = function (specs, artboards) {
+        var preset = specs.preset,
+            index = _.findIndex(templates, function (obj) {
+                return obj.preset === preset;
+            });
+
+        var height = templates[index].height,
+            width = templates[index].width,
+            changeLayerRef = false,
+            finalBounds;
+
+        // if artboards are empty, insert template at first position 
+        if (artboards.isEmpty()) {
+            changeLayerRef = true;
+            finalBounds = new Bounds({
+                top: DEFAULT_ARTBOARD_BOUNDS.top,
+                bottom: height,
+                left: DEFAULT_ARTBOARD_BOUNDS.left,
+                right: width
+            });
+        } else {
+            // find the rightmost artboard and insert the template after this 
+            var rightMostLayer = _findRightMostArtboard(artboards),
+                offset = 100;
+            finalBounds = rightMostLayer.bounds.merge({
+                bottom: rightMostLayer.bounds.top + height,
+                left: rightMostLayer.bounds.left + offset + rightMostLayer.bounds.width,
+                right: rightMostLayer.bounds.right + offset + width
+            });
+        }
+
+        return {
+            finalBounds: finalBounds,
+            changeLayerRef: changeLayerRef
+        };
+    };
+
+    /**
+     * Helper function to get the bounds of an artboard with no prior input
+     * 
+     * @private
+     * @param {Immutable.List.<Layer>} artboards
+     * @param {Immutable.List.<Layer>} selectedArtboards 
+     * @return {{finalBounds: Bounds, changeLayerRef: boolean}}
+     */
+    var _getBoundsFromNoInput = function (artboards, selectedArtboards) {
+        var changeLayerRef = false,
+            finalBounds;
+
+        // if there are no artboards in the document, the new artboard should have dimensions of default
+        if (artboards.isEmpty()) {
+            changeLayerRef = true;
+            finalBounds = new Bounds({
+                top: DEFAULT_ARTBOARD_BOUNDS.top,
+                bottom: DEFAULT_ARTBOARD_BOUNDS.bottom,
+                left: DEFAULT_ARTBOARD_BOUNDS.left,
+                right: DEFAULT_ARTBOARD_BOUNDS.right
+            });
+        } else {
+            // else we want to get the right most artboard 
+            var rightMostLayer = _findRightMostArtboard(artboards),
+                offset = 100;
+            // if there is a selected artboard, we want the new one to inherit the size of the selected
+            if (selectedArtboards.size === 1) {
+                var selectedbounds = selectedArtboards.first().bounds;
+                finalBounds = rightMostLayer.bounds.merge({
+                    bottom: rightMostLayer.bounds.top + selectedbounds.height,
+                    left: rightMostLayer.bounds.left + rightMostLayer.bounds.width + offset,
+                    right: rightMostLayer.bounds.right + offset + selectedbounds.width
+                });
+            } else {
+                // else we want it to insert after the right most artboard with default size 
+                finalBounds = rightMostLayer.bounds.merge({
+                    bottom: rightMostLayer.bounds.top + DEFAULT_ARTBOARD_BOUNDS.bottom,
+                    left: rightMostLayer.bounds.left + rightMostLayer.bounds.width + offset,
+                    right: rightMostLayer.bounds.right + offset + DEFAULT_ARTBOARD_BOUNDS.right
+                });
+            }
+        }
+
+        return {
+            finalBounds: finalBounds,
+            changeLayerRef: changeLayerRef
+        };
+    };
+
+    /**
      * Create a new Artboard on the PS doc
      * if no bounds are provided we place this 100 px to the right of selected artboard 
-     * or we add a default sized "iphone" artboard 
-     * otherwise passed in bounds are used
+     * or we add a default sized "iphone" artboard otherwise passed in bounds are used
      *
-     * @param {Bounds=} artboardBounds where to place the new artboard
+     * @param {Bounds=|object=} boundsOrSpecs
      * @return {Promise}
      */
-    var createArtboard = function (artboardBounds) {
+    var createArtboard = function (boundsOrSpecs) {
         var document = this.flux.store("application").getCurrentDocument(),
             artboards = document.layers.all.filter(function (layer) {
                 return layer.isArtboard;
             }),
+            selectedArtboards = document.layers.selected.filter(function (layer) {
+                return layer.isArtboard;
+            }),
             layerRef = layerLib.referenceBy.none,
+            boundsAndLayerRef,
             finalBounds;
 
-        if (artboardBounds !== undefined) {
-            finalBounds = artboardBounds.toJS();
-        } else if (artboards.isEmpty()) {
-            // If there are no artboards selected, use current selection
-            layerRef = layerLib.referenceBy.current;
-            finalBounds = DEFAULT_ARTBOARD_BOUNDS;
+        if (boundsOrSpecs instanceof Bounds) {
+            finalBounds = boundsOrSpecs.toJS();
+        } else if (boundsOrSpecs === undefined) {
+            boundsAndLayerRef = _getBoundsFromNoInput(artboards, selectedArtboards);
+            finalBounds = boundsAndLayerRef.finalBounds.toJS();
+            if (boundsAndLayerRef.changeLayerRef) {
+                layerRef = layerLib.referenceBy.current;
+            }
         } else {
-            var layer = artboards.reduce(function (selectedLayer, currentLayer) {
-                if (currentLayer.bounds.right > selectedLayer.bounds.right) {
-                    return currentLayer;
-                } else {
-                    return selectedLayer;
-                }
-            }, artboards.first());
-
-            var offset = layer.bounds.width + 100;
-            
-            finalBounds = {
-                    top: layer.bounds.top,
-                    bottom: layer.bounds.bottom,
-                    left: layer.bounds.left + offset,
-                    right: layer.bounds.right + offset
-                };
+            boundsAndLayerRef = _getBoundsFromTemplate(boundsOrSpecs, artboards);
+            finalBounds = boundsAndLayerRef.finalBounds.toJS();
+            if (boundsAndLayerRef.changeLayerRef) {
+                layerRef = layerLib.referenceBy.current;
+            }
         }
 
         var createObj = artboardLib.make(layerRef, finalBounds);
@@ -1513,6 +1623,7 @@ define(function (require, exports) {
                 return this.transfer(resetIndex, document, true, true);
             });
     };
+
     createArtboard.reads = [locks.JS_APP];
     createArtboard.writes = [locks.PS_DOC, locks.JS_DOC];
     createArtboard.transfers = [resetIndex];
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -272,6 +272,43 @@ define(function (require, exports, module) {
     };
 
     /**
+     * Incorporates artboard templates into the menu and menu actions
+     * 
+     * @privates
+     * @param {object} menus
+     * @param {object} actions
+     * @param {Array.<object>} templates
+     */
+    var _artboardFromTemplates = function (menus, actions, templates) {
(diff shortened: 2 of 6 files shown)


src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
     };
 
     /**
+     * Helper function to find the right most artboard 
+     * @param {Immutable.List.of<Layer>} artboards
+     * @return {Layer}
+     */
+
```
iwehrman: Extra newline.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
     };
 
     /**
+     * Helper function to find the right most artboard 
+     * @param {Immutable.List.of<Layer>} artboards
```
iwehrman: Just the constructor should be used in the type expression, so this should be `Immutable.List.<Layer>`.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+     * @return {Layer}
+     */
+
+    var _getRightMost = function (artboards) {
```
iwehrman: I suggest naming this instead `_findRightMostArtboard`. Putting `Artboard` in the name makes it obvious what it's finding (even without reading the docs), and `find` hints that this is an O(N) search instead of a constant-time lookup.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
     };
 
     /**
+     * Helper function to find the right most artboard 
+     * @param {Immutable.List.of<Layer>} artboards
+     * @return {Layer}
+     */
```
iwehrman: Also, `@private`.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+            }, artboards.first());
+        return layer;
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     * @param {string} specs, {Immutable.List.of<Layer>} artboards, {Layer} layerRef
+     * @return {Bounds} finalBounds
+     */
+
```
iwehrman: Extra newline.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+                } else {
+                    return selectedLayer;
+                }
+            }, artboards.first());
+        return layer;
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     * @param {string} specs, {Immutable.List.of<Layer>} artboards, {Layer} layerRef
```
iwehrman: Also, we generally put an empty line between the summary and the tags in function JSDoc headers for readability.

iwehrman: Separate `@param` declarations needed for each parameter. Also, the type expression on the second parameter is incorrect as above.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+
+    var _getBoundsFromSpecs = function (specs, artboards, layerRef) {
+        var preset,
+            width, 
+            height,
+            rightmostlayer,
+            finalBounds;
+
+        //get the preset template from the specs
+        preset = specs.preset;
```
iwehrman: It's better to declare and initialize in one step when it's not awkward to do so. This is a good candidate for combined declaration and initialization.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+        //find the measurements of that preset 
+        _.each(templates, function (value, index) {
```
iwehrman: It would be better to use something like `_.findIndex` for this so that a) you don't have to iterate over the entire array unnecessarily, and b) you can cleanly separate the search from what you're going to do with the search result.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+    var _getBoundsFromSpecs = function (specs, artboards, layerRef) {
+        var preset,
+            width, 
+            height,
+            rightmostlayer,
+            finalBounds;
+
+        //get the preset template from the specs
+        preset = specs.preset;
+        //find the measurements of that preset
```
iwehrman: Put a space after the `//` in line comments for readability.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+            layerRef = layerLib.referenceBy.current;
+            finalBounds = new Bounds(
+                {top: DEFAULT_ARTBOARD_BOUNDS.top,
+                 bottom: height,
+                 left: DEFAULT_ARTBOARD_BOUNDS.left,
+                 right: width
+                });
+        } else {
+            //find the rightmost artboard and insert the template after this 
+            rightmostlayer = _getRightMost(artboards);
```
iwehrman: If you're _only_ using `rightmostLayer` right here in this else branch, it's preferable to declare it here so that it doesn't muddle overall understanding of the function.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+                height = templates[index].height;
+                width = templates[index].width;
+            } 
+        });
+
+        //if artboards are empty, insert template at first position 
+        if (artboards.isEmpty()) {
+            layerRef = layerLib.referenceBy.current;
+            finalBounds = new Bounds(
+                {top: DEFAULT_ARTBOARD_BOUNDS.top,
```
iwehrman: Bad object literal formatting. The example below is better.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+                 bottom: height,
+                 left: DEFAULT_ARTBOARD_BOUNDS.left,
+                 right: width
+                });
+        } else {
+            //find the rightmost artboard and insert the template after this 
+            rightmostlayer = _getRightMost(artboards);
+            var offset = 100; 
+            finalBounds = rightmostlayer.bounds.merge({
+                        top: rightmostlayer.bounds.top,
```
iwehrman: Too many tab stops here.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+                {top: DEFAULT_ARTBOARD_BOUNDS.top,
+                 bottom: height,
+                 left: DEFAULT_ARTBOARD_BOUNDS.left,
+                 right: width
+                });
+        } else {
+            //find the rightmost artboard and insert the template after this 
+            rightmostlayer = _getRightMost(artboards);
+            var offset = 100; 
+            finalBounds = rightmostlayer.bounds.merge({
```
iwehrman: I think you just want to create a new `Bounds` object here instead of merging into `rightmostlayer`? You don't seem to be inheriting anything from the previous bounds object.

src/js/actions/layers.js
iwehrman: `specs` does not seem to be a string?

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,147 @@ define(function (require, exports) {
...
+                    return selectedLayer;
+                }
+            }, artboards.first());
+        return layer;
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     * @param {string} specs, {Immutable.List.of<Layer>} artboards, {Layer} layerRef
+     * @return {Bounds} finalBounds
```
iwehrman: You seem to be returning an array?

src/js/models/menubar.js
```diff
@@ -268,6 +268,35 @@ define(function (require, exports, module) {
         });
     };
 
+    var _artboardFromTemplates = function (menus, actions, templates) {
```
iwehrman: Missing JSDoc header.

src/nls/root/menu.js
```diff
@@ -121,6 +121,21 @@ define(function (require, exports, module) {
             },
             CREATE_CLIPPING_MASK: "Create Clipping Mask",
             CREATE_ARTBOARD: "Create New Artboard",
+            NEW_ARTBOARD_FROM_TEMPLATE: {
```
iwehrman: I think we need to find a way to _not_ duplicate all this information into this strings file, since it already exists elsewhere.

iwehrman:
Pausing the review for now because you've probably got plenty to think about :) Many of these nits will be caught for you with our default `grunt` task. 

Once you've made some changes, feel free to just push an additional commit to this branch, which will automatically update the PR. When you're ready for me to look at it again, re-assign the pull request back to me. We can deal with rebasing and squashing the commits again later before merging.

This is a great start! Ask @volfied how many complaints I left on his first pull requests :grimacing:

iwehrman:
Also, please amend the PR summary to note which issue this addresses, and link to it with the special #XYZ syntax.

src/js/actions/layers.js
DivyaPrabhakar: Would like to clarify this one with you in person.

src/js/actions/layers.js
volfied: Drive by comment, please use camelCasing for these variables `rightmostLayer` :car:

src/js/actions/layers.js
DivyaPrabhakar: Do I need to specify the types of the objects within the array?

src/js/actions/layers.js
DivyaPrabhakar: Need to come back to this

src/js/actions/layers.js
iwehrman: If it is a uniform array, yes. If it's not a uniform array, then you shouldn't use an array; you should use an object instead, and declare it _exactly_ like this:
```
@param {{foo: number, bar: string, baz: object}}`
```
(Yes, there are two sets of curly braces. See [here](https://developers.google.com/closure/compiler/docs/js-for-compiler?hl=en#types) for the explanation.)

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
     };
 
     /**
+     * Helper function to find the right most artboard \
```
iwehrman: Trailing slash. Might also want to say "assumes a non-empty list of arboards".

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     *
+     * @private
+     * @param {object} specs 
+     * @param {Immutable.List.<Layer>} artboards
+     * @param {Layer} layerRef
+     * @return {{finalBounds: Bounds, layerRef: Layer}}
+     */
+    var _getBoundsFromSpecs = function (specs, artboards, layerRef) {
```
iwehrman: The name of this is sort of confusing. Maybe it should just be "getNextArtboardBounds"?

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+        return layer;
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     *
+     * @private
+     * @param {object} specs 
+     * @param {Immutable.List.<Layer>} artboards
+     * @param {Layer} layerRef
```
iwehrman: This parameter, and the fact that it's also in the return object, is weird. My intuition says that this is not the right way to do this. I see that in one path through this function you're changing the ref, but maybe you can just return a flag that would be used by the caller to do something with the ref instead?

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+     * @return {{finalBounds: Bounds, layerRef: Layer}}
+     */
+    var _getBoundsFromSpecs = function (specs, artboards, layerRef) {
+        var preset = specs.preset,
```
iwehrman: I think it would make more sense to group the decls and initializations as follows:
```javascript
        var preset = specs.preset,
            index = _.findIndex(templates, function (obj) {
                return obj.preset === preset;
            });

        var height = templates[index].height;
            width = templates[index].width,
            finalBounds;
```

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+            height,
+            finalBounds;
+
+        var index = _.findIndex(templates, function (obj) {
+                return obj.preset === preset;
+            });
+
+        height = templates[index].height;
+        width = templates[index].width;
+
```
iwehrman: Extra newline. AFAIK, we never condone more than a single empty line in a row.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+            layerRef = layerLib.referenceBy.current;
+            finalBounds = new Bounds({
+                top: DEFAULT_ARTBOARD_BOUNDS.top,
+                bottom: height,
+                left: DEFAULT_ARTBOARD_BOUNDS.left,
+                right: width
+            });
+        } else {
+            // find the rightmost artboard and insert the template after this 
+            var rightMostLayer = _findRightMostArtboard(artboards);
```
iwehrman: Group these decls.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+    /**
+     * Helper function to get Bounds from specs (template preset)
+     * 
+     * @private
+     * @param {Immutable.List.<Layer>} artboards
+     * @param {Immutable.List.<Layer>} selected 
+     * @param {Layer} layerRef
+     * @return {{finalBounds: Bounds, layerRef: Layer}}
+     */
+
```
iwehrman: There should not be a newline separating the JSDoc header from its associated syntactic object.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+            finalBounds = new Bounds({
+                top: DEFAULT_ARTBOARD_BOUNDS.top,
+                bottom: DEFAULT_ARTBOARD_BOUNDS.bottom,
+                left: DEFAULT_ARTBOARD_BOUNDS.left,
+                right: DEFAULT_ARTBOARD_BOUNDS.right
+            });
+        } else {
+            // else we want to get the right most artboard 
+            var rightMostLayer = _findRightMostArtboard(artboards);
+            var offset = 100;
```
iwehrman: Group these decls too.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
+        }
+
+        return {
+            finalBounds: finalBounds,
+            layerRef: layerRef
+        };
+    };
+
+    /**
+     * Helper function to get Bounds from specs (template preset)
```
iwehrman: I think this is an incorrect description?

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
      * or we add a default sized "iphone" artboard 
      * otherwise passed in bounds are used
      *
-     * @param {Bounds?} artboardBounds where to place the new artboard
+     * @param {Bounds=|object=} boundsorspecs
      * @return {Promise}
      */
-    var createArtboard = function (artboardBounds) {
+
+    var createArtboard = function (boundsorspecs) {
```
iwehrman: camelCase is preferred for variable names.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,154 @@ define(function (require, exports) {
...
         var document = this.flux.store("application").getCurrentDocument(),
             artboards = document.layers.all.filter(function (layer) {
                 return layer.isArtboard;
             }),
+            selectedArtboards = document.layers.selected.filter(function (layer) {
+                return layer.isArtboard;
+            }),
             layerRef = layerLib.referenceBy.none,
+            boundsandlayerref,
```
iwehrman: Ditto for this variable name.

iwehrman:
OK, done with a second pass. It's looking better! We may need to chat more about this `layerRef` param though.

chadrolfs:
addresses #2090 too.  :tada:

src/js/actions/layers.js
DivyaPrabhakar: Do you think _getBoundsFromTemplate makes it clearer? getNextArtboardBounds is technically what both the current _getBoundsFromSpecs and _getBoundsFromNoInput both do.

src/js/actions/layers.js
DivyaPrabhakar: Do you mean rightMostLayer, offset with the declaration of var finalBounds at the top? I included them down here because I thought in the last review, you said that I should only declare them in the part of the if statement that uses them? Maybe I misunderstood...

iwehrman:
I do not believe this will address #2090 once it's merged into master.

src/js/actions/layers.js
iwehrman: Still need to group these decls.

src/js/actions/layers.js
```diff
@@ -1377,45 +1382,153 @@ define(function (require, exports) {
...
+        };
+    };
+
+    /**
+     * Helper function to get the bounds of an artboard with no prior input
+     * 
+     * @private
+     * @param {Immutable.List.<Layer>} artboards
+     * @param {Immutable.List.<Layer>} selectedArtboards 
+     * @return {{finalBounds: Bounds, changeLayerRef: Boolean}}
```
iwehrman: Basic types are expressed in lower case: `object`, `string`, `number`, `boolean`, `undefined`, etc.

src/js/actions/layers.js
iwehrman: Sorry, I just meant:
```javascript
var rightMostLayer = _findRightMostArtboard(artboards),
    offset = 100;
```

iwehrman:
Looking good! Just a few more instances of declarations to group into block declarations, and then we can 1) squash the commits, and 2) rebase against master. After that, I'll merge! :white_check_mark:

iwehrman:
LGTM!

chadrolfs:
@iwehrman @DivyaPrabhakar what trello card is this attached to?

iwehrman:
None.

Issue #1414, opened 2015-06-12
Adding an artboard is slow.


DesignSpaceBot:
The culprit is a call to `updateCurrentDocument` [here](https://github.com/adobe-photoshop/spaces-design/blob/6620eb210c2e834a8e820aa83f449a5d93412f7e/src/js/actions/layers.js#L1299). This can and should be implemented optimistically. 

-- @iwehrman at [2015-05-09T14:34:07Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1414)

iwehrman:
@volfied: did we ever make this core request? If not, can you do so? It would be great to solve this for M3.

iwehrman:
Watson: 
https://watsonexp.corp.adobe.com/#bug=4034977

iwehrman:
Fixed by #2065.

PR #2065, opened 2015-08-15
Faster artboard creation

iwehrman:
Speed up artboard creation by using new output from the `make` event that includes the resulting name and layer IDs for the new artboard to update the model without resetting the entire document. This reuses the existing group creation infrastructure, but requires updating the layer index from PS afterward due to auto-nesting.

Side note: I renamed the `layers.getLayerOrder` action to `layers.resetIndex` because that's more consistent with all the other reset-model-from-PS actions.

Requires CL 1036241 (pgdev ~465 from Friday afternoon). Addresses #1414.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1459,6 +1459,7 @@ define(function (require, exports) {
                     groupEndID: result.layerSectionEndID,
                     groupname: result.layerName,
                     isArtboard: true,
+                    bounds: result.artboardRect,
                     // don't redraw UI until after resetting the index
                     suppressChange: true
                 };
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1226,14 +1226,15 @@ define(function (require, exports) {
     unlockSelectedInCurrentDocument.transfers = [setLocking];
 
     /**
-     * Updates our layer information based on the current document 
+     * Reset the layer z-index. Assumes that all layers are already in the model,
+     * though possibly out of order w.r.t. Photoshop's model.
      *
      * @param {Document} document Document for which layers should be reordered
      * @param {boolean=} suppressHistory if truthy, dispatch a non-history-changing event.
      * @param {boolean=} amendHistory if truthy, update the current state (requires suppressHistory)
      * @return {Promise} Resolves to the new ordered IDs of layers as well as what layers should be selected
      **/
-    var getLayerOrder = function (document, suppressHistory, amendHistory) {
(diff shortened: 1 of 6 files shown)


src/js/actions/layers.js
```diff
@@ -1451,14 +1452,24 @@ define(function (require, exports) {
         
         return descriptor.playObject(createObj)
             .bind(this)
-            .then(function () {
-                log.debug("Warning: calling updateDocument to add a single artboard is very slow!");
-                return this.transfer(documentActions.updateDocument);
+            .then(function (result) {
+                var payload = {
+                    documentID: document.id,
```
mcilroyc: indent further?  I mean, you're the decider, but this seems to me a reasonable request.

src/js/models/layerstructure.js
```diff
@@ -1140,19 +1140,24 @@ define(function (require, exports, module) {
...
-     * Given IDs of group start and end, and group name, will create a new group and
-     * put all selected layers in those groups
-     * Emulates PS behavior on group - group gets created at the top most selected layer index
+     * Given IDs of group start and end, and group name, creates a new group.
+     * If the group is NOT an artboard, puts all selected layers in the new
+     * group. NOTE: If the group IS an artboard then the caller is responsible
+     * for correctly reordering the layers after this operation!
+     * 
+     * Emulates PS behavior - group gets created at the top most selected layer
+     * index.
```
mcilroyc: pretty sure this would fit on the previous line :us:

src/js/actions/layers.js
iwehrman: I think it matches the way we do this:
```javascript
var foo = function () {
    // blah
};
```
instead of this:

```javascript
var foo = function () {
        // blah
    };
```
We make an exception when there is more than one decl though.
```javascript
var foo = 123,
    bar = function () {
        // blah
    };
```

I can be convinced otherwise, but this is the pattern I'm currently following.

iwehrman:
Comment cleanup pushed in 335196a.

src/js/actions/layers.js
mcilroyc: withdrawn.  it looks/looked weird to my eye, but a quick grep proved that this is an established pattern (for simple objects too, not just functions)

iwehrman:
Now also include the artboard bounds from the result when creating a new artboard model. (I forgot that artboard groups specifically do have bounds, unlike other groups.) Nice catch, @mcilroyc.

Issue #1063, opened 2015-06-12
Placing smart objects is slow


DesignSpaceBot:
The culprit is a call to [`documents.updateDocument`](https://github.com/adobe-photoshop/playground-design/blob/c6d3ee313935fb8cdf4580dc9b2613627eebb1b2/src/js/actions/documents.js#L659-L663). This should be replaced with a call to `layers.addLayer` once the `placeEvent` includes the affected layer IDs. 

Core request here: http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png

-- @iwehrman at [2015-04-08T22:16:31Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1063)

volfied:
Closing as this is a dev bug.

PR #2066, opened 2015-08-15
Faster smart object placement

iwehrman:
Thanks to @TomRuark, there is now a layer ID included with the `placeEvent` event, and we can use this to call `layers.addLayers` to update the document model instead of calling `documents.updateDocument`, which is much slower.

Addresses #1063.


volfied:
Nothing could go wrong!

Issue #1978, opened 2015-08-11
Consider adding slight translucency to UI


iwehrman:
I still kind of miss the slight translucency that we used to have in the non-dialog parts of the UI. I think this might improve visibility into a bigger part of the canvas:
![opaque](https://s3.amazonaws.com/f.cl.ly/items/2d0i1P1O143z1i1a0g2O/Image%202015-08-10%20at%205.46.00%20PM.png)

Incidentally, adding some slight translucency might also help with #1976.

DivyaPrabhakar:
A couple opinions about this: 

- I asked Charles Pearson about whether taking away the translucency was part of user feedback or branding necessity -- the answer is no for both questions, taking away the translucency was just something that happened over time by accident and no one commented either way. 

- In focus groups held last summer, some said it gave the perception of depth and space, traits we are presumably aiming for in Design Space, especially in contrast to classic Photoshop UI. It also may be worth a try now because the properties panel is thicker than the initial designs (drafted last summer), so the idea of more space could be useful.

iwehrman:
Welp, I guess I got out-voted on this one. I shall return!

PR #2068, opened 2015-08-16
[Armchair designer] UI transparency

iwehrman:
This reintroduces a slight (3%) transparency to the UI, including the toolbar, header, and properties panel. It looks like this:
transparency

It correctly maintains full opacity for all the dialogs, like the color picker and search bar.

Addresses #1978. See in particular @DivyaPrabhakar's [recent note](https://github.com/adobe-photoshop/spaces-design/issues/1978#issuecomment-129993462).

Over to @designedbyscience and @placegraphichere for discussion.

### DIFF
--- a/src/style/shared/colors.less
+++ b/src/style/shared/colors.less
@@ -24,8 +24,8 @@
 
 /* Basic Colors */
 @bg-app: #1d1d1d;
-@bg-border: #262626;
-@bg-panel: #2d2d2d;
+@bg-border: rgba(38, 38, 38, 0.98);
+@bg-panel: rgba(45, 45, 45, 0.98);
 @bg-alt: #34332F;
 
 @item: #FFFFFF;


iwehrman:
yes

designedbyscience:
I'll bother @placegraphichere about this today. I think maybe 2% just to reduce the noise a little further (I know, obnoxious)

iwehrman:
Commit amended with 2% transparency. Now it looks like this:
image

iwehrman:
yes

iwehrman:
Rebased after #2115, but the at least the search bar is no longer fully opaque like it should be. Can you help, @placegraphichere?

placegraphichere:
Both @davealonzo and I don't think this what we should do. It can look ok at times, but we just don't think it provides the clarity the UI needs at all times. Sorry @iwehrman

Issue #1991, opened 2015-08-11
Kill the disabled cursor softly


iwehrman:
Let's get rid of the "disabled" cursor styling for disabled buttons in the UI. I'm not aware of other apps that do this, and it's annoying me more than it's delighting me lately. :knife:

placegraphichere:
yes, I agree

iwehrman:
One shot one kill.

PR #2070, opened 2015-08-16
Eliminate disabled cursor styling on buttons

iwehrman:
What it says. The disabled cursor styling is not disabled elsewhere, e.g., in Libraries drag and drop.

Addresses: concerns that this was a bad idea.

### DIFF
--- a/src/style/sections/style/style-section.less
+++ b/src/style/sections/style/style-section.less
@@ -48,7 +48,6 @@
 
 .style-button__disabled {
     opacity: 0.20 !important;
-    cursor: not-allowed;
 }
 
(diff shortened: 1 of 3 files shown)


iwehrman:
Addresses #1991.

designedbyscience:
Are we leaving that libraries overlay one?: libraries-section.less, line 181

iwehrman:
I left it because it seemed like it had to do with drag-and-drop, which seemed like a plausible use case for these cursors. Of course, they should be used consistently, but that's a different issue.

designedbyscience:
Alrighty then

Issue #2077, opened 2015-08-17
If fill is not visible, alpha of stroke appears to be 100% on canvas


ktaki:
Could be the same root cause as #2076, but reporting this anyway.
Draw a rectangle. Add a stroke and set the alpha to be 20%.
Click on Toggle Fill icon to make the fill invisible.
RESULT: Stroke will be drawn as the alpha value 100% on canvas.

volfied:
Dupe of https://github.com/adobe-photoshop/spaces-design/issues/526

Issue #2076, opened 2015-08-17
Sets fill color to black when Toggle Fill is clicked


ktaki:
Draw a rectangle. Set the color to red. Click Toggle Fill icon. Now fill is black.

iwehrman:
Possibly related to #2060. Over to @volfied for investigation.

ktaki:
Verified the fix.

PR #2078, opened 2015-08-17
Preserve fill color on toggle enabled

volfied:
Addresses #2076 and #2077, casualties of a refactor.

### DIFF
--- a/src/js/jsx/sections/style/Fill.jsx
+++ b/src/js/jsx/sections/style/Fill.jsx
@@ -85,10 +85,12 @@ define(function (require, exports, module) {
          * @param {boolean} isChecked
          */
         _toggleFillEnabled: function (event, isChecked) {
+            var color = this.state.fill && collection.uniformValue(this.state.fill.colors) || Color.DEFAULT;
+
             this.getFlux().actions.shapes.setFillEnabled(
                 this.props.document,
                 this.state.layers,
-                this.state.fill && this.state.fill.color || Color.DEFAULT,
+                color,
                 isChecked
             );
         },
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/style/Fill.jsx
```diff
@@ -88,7 +88,7 @@ define(function (require, exports, module) {
             this.getFlux().actions.shapes.setFillEnabled(
                 this.props.document,
                 this.state.layers,
-                this.state.fill && this.state.fill.color || Color.DEFAULT,
+                this.state.fill && this.state.fill.colors.first() || Color.DEFAULT,
```
iwehrman: Should this be the result of `collection.uniformValue`?

ktaki:
#2076 is fixed, but I still see #2077.

ktaki:
OK. I spoke too fast. I think #2080 contributes #2076.
Create a rect. Set fill color. Toggle the visibility. This works. Now, set the alpha to 90%. Toggle the visibility. It resets the fill color to black.

volfied:
Now it also fixes #2080

ktaki:
Confirmed the fixes for #2076 and #2080.

src/js/models/layerstructure.js
```diff
@@ -1213,13 +1212,19 @@ define(function (require, exports, module) {
...
-                return map.set(layerID, Immutable.Map({
-                    fill: nextFill
-                }));
-            }, new Map()));
+        var nextLayers = Immutable.Map(layerIDs.reduce(function (map, layerID) {
+            var layer = this.byID(layerID),
+                fill = layer.fill;
+
+            if (!fill) {
+                throw new Error("Unable to set fill properties: no fill for layer " + layer.name);
```
iwehrman: Seems like it would be better to just make a new Fill here, but I guess that can wait for a later day.

iwehrman:
Let's do this.

Issue #2071, opened 2015-08-17
UI isn't updated in modal tool state with super-select type subtool


iwehrman:
The necessary handlers are only installed using the primitive type tool, and not the super-select type subtool.

iwehrman:
Fixed by #2081.

Issue #2009, opened 2015-08-12
Stuck in multiple text editing modal states


ktaki:
With type tool selected, selecting a text layer in the layers panel immediately goes into modal text editing mode. With this mode, clicking on another text layer in the layers panel seems to go into another level of modal state. Eventually, the UI freezes. Repeating Esc keys can escape from the editing mode, but bad things may happen in the process.

Steps.  
Create two text layers
Select type tool
1. Select one of them in the layers panel.
2. Select another text layer in the layers panel.
Repeat 1 and 2.
Eventually UI freezes. Use Esc key a few times to recover the state. This might cause some errors.

chadrolfs:
related to or dupe of #1969?

iwehrman:
I agree that this is a high priority bug. It's distinct from #1969. One of the problems is that we get a "createdTextLayer" event when we click with the text tool, but then if the modal tool state is ended without committing the text we never get a corresponding "delete" event. (And I'm not aware of any other way to detect this.) The other problem is that, when the type tool is selected, we seem to be entering the type modal tool state automatically when selecting the layer, which is unexpected and broken.

@pineapplespatula: Any idea about either one of these issues? They both seem like core bugs to me. Let me know if you think I should file issues about either one.

ktaki:
Verified the fix.

Issue #1927, opened 2015-08-05
Type tool does not work in mixed selection states


pineapplespatula:
Steps: 
Make two layers 
* one type layer
* one shape layer
Attempt to change type properties

Making Changes in the property panel have no effect

iwehrman:
Fixed by #2081.

Issue #1883, opened 2015-07-31
Text Edit: Align icons should be highlighted to indicate current setting


ktaki:

DivyaPrabhakar:
@pineapplespatula said that this might be a good bug to get lost in for a while

ktaki:
Do we have a label "get lost"? It surely will become very popular. :-)

placegraphichere:
#2d86de is the color you're looking for a highlight state

ktaki:
Verified the fix.

PR #2081, opened 2015-08-17
Modal type bug fix grab bag

iwehrman:
This is modal-type bug-fix PR. I believe it addresses issues #2071, #2009, #1927, #1883, #1881, #1857, #1691. There are still a number of type issues it does not address, so please take a look at the [Type issue label](https://github.com/adobe-photoshop/spaces-design/labels/Type) if you see a problem during testing to see if an issue has already been filed.

Highlights:

* Removes the `*Assigned` properties in `ParagraphStyle` and `CharacterStyle` in favor of a simpler `mergeWith` operation for resolving inconsistencies.
* Simplifies and throttles the `updatedTextProperties` event handler, which PS emits whenever the style properties for the current cursor position/text selection change in the modal state.
* Applies this handler in the select type sub-tool, as well as in the full type tool.
* Makes use of a new `deleteTextLayer` event to catch the case of a new text layer being canceled without committing.
* Factors a `layers.removeLayers` action out of `deleteSelected`, which can be used to remove layers from the JS-side layers model.
* Adds a `type.updateProperties` action to set type properties en masse instead of with separate actions that redraw the UI multiple times redundantly. (TODO: eliminate the other `type.update*` actions in favor of this one.)

@mcilroyc: I'm not sure if I've used the history events correctly. Can you double-check that part of this PR for me?

Requires at least Mac Jenkins pgdev.469 from the afternoon of Monday the 17th.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -790,9 +790,40 @@ define(function (require, exports) {
     selectAll.post = [_verifyLayerSelection];
 
     /**
+     * Remove the given layers from the JavaScript model. This is useful from
+     * event handlers that tell us about layers which have already been removed
+     * from the model. To delete layers from a Photoshop document, use the
+     * deleteSelected action.
+     *
+     * @param {Document} document
+     * @param {Layer|Immutable.Iterable.<Layer>} layers
+     * @return {Promise}
+     */
+    var removeLayers = function (document, layers) {
+        if (layers instanceof Layer) {
+            layers = Immutable.List.of(layers);
+        }
+
+        return _getSelectedLayerIndices(document)
+            .bind(this)
+            .then(function (selectedLayerIndices) {
+                var payload = {
+                    documentID: document.id,
+                    layerIDs: collection.pluck(layers, "id"),
+                    selectedIndices: selectedLayerIndices
+                };
+
+                this.dispatch(events.document.history.nonOptimistic.DELETE_LAYERS, payload);
+            });
+    };
+    removeLayers.reads = [locks.PS_DOC];
+    removeLayers.writes = [locks.JS_DOC];
+    removeLayers.post = [_verifyLayerIndex, _verifyLayerSelection];
+
+    /**
      * Deletes the selected layers in the given document, or in the current document if none is provided
      *
-     * @param {?document} document
+     * @param {?Document} document
      * @return {Promise}
      */
     var deleteSelected = function (document) {
@@ -808,12 +839,7 @@ define(function (require, exports) {
 
         var documentID = document.id,
             layers = document.layers.allSelected,
-            layerIDs = collection.pluck(layers, "id"),
             deletePlayObject = layerLib.delete(layerLib.referenceBy.current),
-            payload = {
-                documentID: documentID,
-                layerIDs: layerIDs
-            },
             options = {
                 historyStateInfo: {
                     name: strings.ACTIONS.DELETE_LAYERS,
@@ -823,14 +849,13 @@ define(function (require, exports) {
 
         return locking.playWithLockOverride(document, layers, deletePlayObject, options, true)
             .bind(this)
-            .then(_.wrap(document, _getSelectedLayerIndices))
-            .then(function (selectedLayerIndices) {
-                payload.selectedIndices = selectedLayerIndices;
-                return this.dispatchAsync(events.document.history.nonOptimistic.DELETE_LAYERS, payload);
+            .then(function () {
+                return this.transfer(removeLayers, document, layers);
             });
     };
-    deleteSelected.reads = [locks.JS_APP];
-    deleteSelected.writes = [locks.PS_DOC, locks.JS_DOC];
+    deleteSelected.reads = [locks.JS_APP, locks.JS_DOC];
+    deleteSelected.writes = [locks.PS_DOC];
+    deleteSelected.transfers = [removeLayers];
     deleteSelected.post = [_verifyLayerIndex, _verifyLayerSelection];
 
 
@@ -1822,6 +1847,7 @@ define(function (require, exports) {
     exports.rename = rename;
     exports.selectAll = selectAll;
     exports.deselectAll = deselectAll;
+    exports.removeLayers = removeLayers;
     exports.deleteSelected = deleteSelected;
     exports.groupSelected = groupSelected;
     exports.groupSelectedInCurrentDocument = groupSelectedInCurrentDocument;
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -39,6 +39,7 @@ define(function (require, exports) {
     var Layer = require("js/models/layer"),
         collection = require("js/util/collection"),
         documentActions = require("./documents"),
+        historyActions = require("./history"),
         searchActions = require("./search/layers"),
         log = require("js/util/log"),
         events = require("../events"),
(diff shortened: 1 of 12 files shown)


src/js/models/characterstyle.js
```diff
@@ -119,23 +96,17 @@ define(function (require, exports, module) {
             textStyle = characterStyleDescriptor.textStyle;
 
         var rawColor = textStyle.hasOwnProperty("color") ?
-            textStyle.color :
-            baseParentStyle.color,
-            color = Color.DEFAULT;
+            textStyle.color : baseParentStyle.color;
```
volfied: can you indent this, it's hard to follow.

src/js/models/characterstyle.js
```diff
@@ -187,64 +140,67 @@ define(function (require, exports, module) {
...
-            }, new CharacterStyle());
+                return CharacterStyle.fromCharacterStyleDescriptor(documentDescriptor,
+                    layerDescriptor, characterStyleDescriptor, baseParentStyle);
+            });
+
+        var reducedModel = Object.keys(_characterStylePrototype)
+            .reduce(function (reducedModel, property) {
+                var values = collection.pluck(characterStyles, property);
+
+                reducedModel[property] = collection.uniformValue(values);
```
volfied: No protection against null?

src/js/tools/superselect/type.js
```diff
@@ -43,6 +53,14 @@ define(function (require, exports, module) {
...
             }
         });
+
+        if (_typeChangedHandler) {
+            descriptor.removeListener("updateTextProperties", _typeChangedHandler);
+        }
+
+        _typeChangedHandler = TypeTool.updateTextPropertiesHandler.bind(this);
+
+        descriptor.addListener("updateTextProperties", _typeChangedHandler);
```
volfied: should deselectHandler be removing this listener?

src/js/models/characterstyle.js
iwehrman: In the case of these reduce models, null will be interpreted in the UI to mean "mixed".

src/js/tools/superselect/type.js
iwehrman: Mmm, yeah. Good call.

src/js/models/characterstyle.js
iwehrman: OK, sure.

iwehrman:
Fixes in commit 912c291.

volfied:
Not merging this yet as I'd like to go through the bugs and code one more time.

src/js/actions/layers.js
```diff
@@ -790,9 +790,40 @@ define(function (require, exports) {
...
+        return _getSelectedLayerIndices(document)
+            .bind(this)
+            .then(function (selectedLayerIndices) {
+                var payload = {
+                    documentID: document.id,
+                    layerIDs: collection.pluck(layers, "id"),
+                    selectedIndices: selectedLayerIndices
+                };
+
+                this.dispatch(events.document.history.nonOptimistic.DELETE_LAYERS, payload);
```
mcilroyc: It seems like this *might* need to call a non-history flavor of DELETE_LAYERS when called by the "deleteTextLayer" handler.  I need to get the newest build to test this - just wanted to register the suspicion.

iwehrman: Do you suggest a flag?

mcilroyc: I believe this scenario (clicking canvas to create a text layer and then ESCing before typing) causes two extraneous history states in our store that do not have corresponding PS history.  These are created by the layer actions `addLayers()` and `removeLayers()`, both of which dispatch history-changing events.

You can test this by creating a rectangle, changing its color twice, doing the point-text-empty-ESC dance, and then undoing twice.  You should see the rectangle change color on canvas, but not updated in the panel.

It seems like a simple fix for the removeLayers case (conditionally dispatching a non-history version of the event).  If it is not that simple for the addLayers case, and we therefore might need to do some retrospective history manipulation, then LMK.

volfied:
#2081 is still not fixed in this PR.

volfied:
#1882 is still not fixed.

volfied:
Other issues listed at the top are good. At this point I'm passing this review to @mcilroyc so he can wrap up the history code review.

iwehrman:
@volfied: #1927 seems fixed to me. Can you explain how to repro a problem? 
Re #1882: I may have misunderstood the bug originally, but I'm not sure PS even supports paragraph alignments at the text-run level?

volfied:
@iwehrman #1927: Select a shape layer and a text layer (I selected shape layer last) and try changing the text color. Now I realize I may have to rebuild (I thought i rebuilt yesterday after 2:30 PM when Jeffrey checked his changes in).

#1882 - If paragraph alignment is not usable during modal type state, we should remove/disable those buttons.

iwehrman:
@volfied: that's what I'm doing for #1927.
@pineapplespatula: should the alignment buttons be disabled during modal text editing?

iwehrman:
It looks like the alignment buttons _do_ work in standard PS, so I'll see if either a) the action descriptors aren't working, in which case I think we should proceed and leave #1882 open, or b) if they are working, I'll fix it.

pineapplespatula:
@iwehrman the alignment buttons should be enabled. I'm pretty sure this one is a core bug

iwehrman:
The next check-in will also address #1883.

iwehrman:
@pineapplespatula yes, that's consistent with what I'm seeing. I officially retract that this addresses #1882.

iwehrman:
The next check-in will also address #1881.

volfied:
#1927 confirmed, I had an earlier build :X

volfied:
#1881 fix confirmed. @mcilroyc, please merge when ready.

iwehrman:
I've pushed the _narrow_ fix for the history state problem discussed after the standup this morning. I notice that once I clear the document history, though, undo/redo are no longer enabled despite the fact that the document is still dirty. Is this a known issue? It seems independent of what I'm doing here. In any case, back to @mcilroyc for review.

iwehrman:
Also note again that 7c6bfa4 claims to address #1881 and #1883, but I think @volfied only verified the former.

src/js/actions/layers.js
```diff
@@ -790,9 +791,46 @@ define(function (require, exports) {
...
+
+                if (resetHistory) {
+                    return this.transfer(historyActions.resetHistory, document);
```
mcilroyc: Per the above-noted issue with undo/redo not being available in the menu, try...

changing this line to `return this.transfer(historyActions.queryCurrentHistory, document.id);`
(and the associated `.transfers` change)

I'm not saying this is the ideal solution, but it is more correct because it asks photoshop what the current pointer should be at, and then re-inits the  history store with its known-unknowns

This throws a warning ...which could probably be forcibly quelled with a `iKnowWhatImDoing` flag

src/js/tools/type.js
```diff
@@ -70,80 +84,66 @@ define(function (require, exports, module) {
...
+                    return;
+                }
+
+                var layerID = event.layerID,
+                    layer = document.layers.byID(layerID);
+
+                if (layer) {
+                    if (_layersReplaced) {
+                        // See comment above at the _layersReplaced declaration
+                        this.flux.actions.documents.updateDocument();
```
mcilroyc: should _layersReplaced be set to false here, so the subsequent deleteTextLayer events aren't misinterpreted, assuming there was *not* a createTextLayer in the interim?  Maybe that's not possible?

iwehrman: Shouldn't be possible? But, better safe than sorry?

iwehrman: Actually I think I'm against adding that extra line, since it will make the supposed state machine more obscure from our end, and there's no way to defend against all possible PS bugs.

mcilroyc:
Alright, droppin this bomb.

chadrolfs:
set appropriate labels and re-assigned bugs addressed in this PR.

PR #2088, opened 2015-08-18
Styles Workflow - Sampler HUD

volfied:
With sampler tool selected, pressing Spacebar will pop up a HUD with sampling possibilities shown.

Current possibilities are:
Sample fill color / stroke color / type style / layer effects / graphics (between smart objects)


src/js/actions/layereffects.js
```diff
@@ -337,11 +337,13 @@ define(function (require, exports) {
      * Duplicates the layer effects of the source layer on all the target layers
      *
      * @param {Document} document
-     * @param {Immutable.Iterable.<Layer>} targetLayers
+     * @param {Immutable.Iterable.<Layer>?} targetLayers Default is selected in the document
```
mcilroyc: ? goes at the beginning

src/js/actions/layereffects.js
```diff
@@ -408,6 +410,8 @@ define(function (require, exports) {
             coalesce: false
         };
 
+        this.dispatch(events.style.HIDE_HUD);
```
mcilroyc: I think the rule now is that dispatches must use `dispatchAsync` *unless* it is already inside a promise chain.  @iwehrman correct me if I'm misremembering our recent conversation.

src/js/actions/sampler.js
```diff
@@ -176,7 +289,64 @@ define(function (require, exports) {
     };
     click.reads = [locks.PS_DOC, locks.JS_UI];
     click.writes = [];
-    click.transfers = [layerFXActions.duplicateLayerEffects, shapeActions.setFillColor, typeActions.setColor];
+    click.transfers = [
+        layerFXActions.duplicateLayerEffects,
+        shapeActions.setFillColor,
+        typeActions.setColor, applyColor
```
mcilroyc: nit: this should be split in to two lines

src/js/actions/type.js
```diff
@@ -524,15 +524,19 @@ define(function (require, exports) {
      * Applies the given text style to target layers
      *
      * @param {Document} document
-     * @param {Immutable.Iterable.<Layer>} targetLayers
+     * @param {Immutable.Iterable.<Layer>?} targetLayers Default is selected layers
```
mcilroyc: same nit as above, ? at beginning, = at the end

src/js/actions/type.js
```diff
@@ -524,15 +524,19 @@ define(function (require, exports) {
...
      * @return {Promise}
      */
     var applyTextStyle = function (document, targetLayers, style) {
+        targetLayers = targetLayers || document.layers.selected;
+
         var layerIDs = collection.pluck(targetLayers, "id"),
             layerRefs = layerIDs.map(textLayerLib.referenceBy.id).toArray(),
             applyObj = textLayerLib.applyTextStyle(layerRefs, style);
 
+        this.dispatch(events.style.HIDE_HUD);
```
mcilroyc: same possible gripe as above, wrt dispatchAsync

src/js/actions/layereffects.js
iwehrman: That is indeed the rule.

src/js/models/layer.js
```diff
@@ -529,5 +558,24 @@ define(function (require, exports, module) {
         return this.kind === this.layerKinds.VECTOR;
     };
 
+    /**
+     * Returns the smart object type for smart object layers, null otherwise
+     *
+     * @return {SmartObjectType}
```
iwehrman: If you're returning null, should be `?SmartObjectType`

src/js/models/layer.js
```diff
@@ -37,6 +37,29 @@ define(function (require, exports, module) {
         Text = require("./text");
 
     /**
+     * Possible smart object types for a layer
```
iwehrman: Should this be in adapter/lib?

src/js/tools/sampler.js
```diff
@@ -70,13 +85,33 @@ define(function (require, exports, module) {
...
+            flux.actions.sampler.click(currentDocument, event.pageX, event.pageY, event.shiftKey);
+        }
+    };
+
+    /**
+     * Handler for keydown events, installed when the tool is active.
+     *
+     * @param {CustomEvent} event
+     */
+    SamplerTool.prototype.onKeyUp = function (event) {
```
iwehrman: Can this happen on keydown instead? It feels really sluggish due to the fact that it's on key up.

iwehrman:
A few times I accidentally clicked in the hole in the middle of the stroke icon... it's easier to do than you you might think. Maybe we can make the hole slightly opaque so that JS side gets that event?

iwehrman:
This could be due to the different UI size on a big 5k iMac, but I find the HUD to be a little too small compared to the rest of the UI:
image
In particular, it seems like the icons should roughly the same size as the toolbar icons? CC @placegraphichere

iwehrman:
If the HUD is up and I switch tools then switch back to the sampler, the HUD is still up with the wrong colors.

iwehrman:
Currently sampling the stroke with the HUD sets the fill of the target layer. @placegraphichere thinks that it should set the stroke of the target layer. Fight.

iwehrman:
Also, re: hole in the icon/icons are too small: the Fx icon is also hard to click. Maybe I'm just bad at clicking.

iwehrman:
If I hold down the command key and move the mouse, the cursor turns into a move arrow. That ain't right.

src/js/actions/sampler.js
```diff
@@ -176,7 +289,64 @@ define(function (require, exports) {
...
+                var hitLayerMap = new Set(hitLayerIDs.toJS()),
+                    clickedLeafLayers = layerTree.leaves.filter(function (layer) {
+                        return hitLayerMap.has(layer.id);
+                    }),
+                    topLayer = clickedLeafLayers.last();
+
+                return topLayer;
+            })
+            .then(function (source) {
+                return _calculateSampleTypes.call(this, doc, source, x, y);
```
iwehrman: This can be moved into the previous then handler.

src/js/actions/sampler.js
```diff
@@ -197,9 +367,37 @@ define(function (require, exports) {
...
             return Promise.resolve();
         }
+
+        var fillColor = null,
+            strokeColor = null,
+            typeStyle = null;
+
+        switch (source.kind) {
+        case source.layerKinds.VECTOR:
+            fillColor = source.fill ? source.fill.color : null;
```
iwehrman: Could be just `fillColor = source.fill && source.fill.color`.

iwehrman:
OK, done with this pass. Looks cool! Big questions are:

1. correct behavior of the stroke button?
2. keydown instead of keyup?
3. bigger/easier click targets?

src/js/models/layer.js
volfied: I thought about that... but these are values that we defined, unlike layer kinds.

volfied:
@iwehrman Made the review changes. 

Re: stroke behavior: this is annoying to me as well.. it feels like shape->shape stroke sampling should sample... well, stroke! But it still applies to the main color of text layers, so it might be unexpected. I'd like to leave that one on the back burner, and see how it feels to QE and testers.

iwehrman:
I'm not qualified to judge, but unless I'm mistaken @placegraphichere said he thought stroke should go to stroke, and that he and you discussed that. It's fine if it changes later, but I'd like to at least get everyone on the same page about what the intended design is.

volfied:
I must have remembered wrong. It seems to make sense to everyone, so I'll make the change!

iwehrman:
Filed #2135 about the change in behavior for the stroke button. Looks good otherwise, thanks!

PR #2089, opened 2015-08-18
Export enhancements: quick selection, and improved style in export all panel

mcilroyc:
Export All panel:

- Improved UI including grey tri-state checkboxes!
- Added two new checkboxes to quickly select (mark for export) All artboards, and All non-artboards
- Layers are listed in proper layer order, including artboards in reverse (per dave, I think)

Tech Debt Payments:

- Moved the definition of supported scales and formats into the model
- Improved error handling of asset export, including adding a timeout since generator seems to sometimes just not respond

http://cl.ly/2w3z3U350J2i

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -67,7 +67,8 @@ define(function (require, exports) {
      */
     var _syncLayerExportMetadata = function (documentID, layerID) {
         var documentExports = this.flux.stores.export.getDocumentExports(documentID),
-            layerExportsArray = documentExports && documentExports.layerExportsArray(layerID),
+            layerExports = documentExports && documentExports.getLayerExports(layerID),
+            layerExportsArray = layerExports && layerExports.toJS(),
             document = this.flux.stores.document.getDocument(documentID),
             layer = document && document.layers.byID(layerID),
             exportEnabled = document && layer && layer.exportEnabled;
@@ -301,7 +302,7 @@ define(function (require, exports) {
      * @param {Document} document Owner document
      * @param {Layer|Immutable.List.<Layer>} layers Either a Layer reference or array of Layers
      * @param {boolean=} exportEnabled
-     * @returns {Promise}
+     * @return {Promise}
      */
     var setLayerExportEnabled = function (document, layers, exportEnabled) {
         var layerIDs = Immutable.List.isList(layers) ?
@@ -330,7 +331,7 @@ define(function (require, exports) {
      * @private
      * @param {Document} document Owner document
      * @param {boolean=} exportEnabled
-     * @returns {Promise}
+     * @return {Promise}
      */
     var setAllArtboardsExportEnabled = function (document, exportEnabled) {
         var documentExports = this.flux.stores.export.getDocumentExports(document.id);
@@ -352,7 +353,7 @@ define(function (require, exports) {
      * @private
      * @param {Document} document Owner document
      * @param {boolean=} exportEnabled
-     * @returns {Promise}
+     * @return {Promise}
      */
     var setAllNonABLayersExportEnabled = function (document, exportEnabled) {
         var documentExports = this.flux.stores.export.getDocumentExports(document.id);
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -103,7 +103,16 @@ define(function (require, exports) {
                     status: ExportAsset.STATUS.STABLE
                 };
                 return this.transfer(updateLayerExportAsset, document, layer, assetIndex, assetProps);
-            });
+            })
+            .catch(function (e) {
+                log.error("Export Failed for asset %d of layerID %d, documentID %d, with error %s",
+                    assetIndex, layer.id, document.id, e);
+                var assetProps = {
+                    filePath: "",
+                    status: ExportAsset.STATUS.ERROR
+                };
+                return this.transfer(updateLayerExportAsset, document, layer, assetIndex, assetProps);
+            }) ;
     };
 
     /**
@@ -317,6 +326,50 @@ define(function (require, exports) {
     setLayerExportEnabled.writes = [locks.JS_DOC, locks.PS_DOC];
 
     /**
+     * Sets the exportEnabled flag for all artboards
+     * @private
+     * @param {Document} document Owner document
+     * @param {boolean=} exportEnabled
+     * @returns {Promise}
+     */
+    var setAllArtboardsExportEnabled = function (document, exportEnabled) {
+        var documentExports = this.flux.stores.export.getDocumentExports(document.id);
+
+        if (!documentExports) {
+            return Promise.resolve();
+        }
+
+        var layersWithExports = documentExports.getLayersWithExports(document, true);
+
+        return this.transfer(setLayerExportEnabled, document, layersWithExports, exportEnabled);
+    };
+    setAllArtboardsExportEnabled.reads = [locks.JS_DOC];
+    setAllArtboardsExportEnabled.writes = [];
+    setAllArtboardsExportEnabled.transfers = [setLayerExportEnabled];
+
+    /**
+     * Sets the exportEnabled flag for all layers that have at least one configured asset
+     * @private
+     * @param {Document} document Owner document
+     * @param {boolean=} exportEnabled
+     * @returns {Promise}
+     */
+    var setAllNonABLayersExportEnabled = function (document, exportEnabled) {
+        var documentExports = this.flux.stores.export.getDocumentExports(document.id);
+
+        if (!documentExports) {
+            return Promise.resolve();
+        }
+
+        var layersWithExports = documentExports.getLayersWithExports(document, false);
+
+        return this.transfer(setLayerExportEnabled, document, layersWithExports, exportEnabled);
+    };
+    setAllNonABLayersExportEnabled.reads = [locks.JS_DOC];
+    setAllNonABLayersExportEnabled.writes = [];
+    setAllNonABLayersExportEnabled.transfers = [setLayerExportEnabled];
+
+    /**
      * Export all assets for the given document for which export has been enabled (layer.exportEnabled)
      *
      * @param {Document} document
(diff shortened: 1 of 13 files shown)


src/js/actions/export.js
```diff
@@ -317,6 +326,50 @@ define(function (require, exports) {
     setLayerExportEnabled.writes = [locks.JS_DOC, locks.PS_DOC];
 
     /**
+     * Sets the exportEnabled flag for all artboards
+     * @private
+     * @param {Document} document Owner document
+     * @param {boolean=} exportEnabled
+     * @returns {Promise}
```
iwehrman: `@return` throughout

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -193,26 +227,41 @@ define(function (require, exports, module) {
...
-                            <TitleHeader
-                                title={strings.EXPORT.EXPORT_LIST_ARTBOARDS} />
+                            <div className="exports-panel__asset-list__quick-selection">
+                                <div className="column-2" >
+                                    <CheckBox
+                                        checked={allArtboardsExportEnabled}
+                                        onChange={this._setAllArtboardsExportEnabled} />
+                                </div>
+                                <Gutter />
+                                All&nbsp;{strings.EXPORT.EXPORT_LIST_ARTBOARDS}
```
iwehrman: "All" should be localized.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -193,26 +227,41 @@ define(function (require, exports, module) {
...
                             </div>
                         </div>
                     </div>
-                    <hr />
                     <div className="exports-panel__button-group">
                         <Button
                             onClick={this.props.dismissDialog}>
                             {strings.EXPORT.BUTTON_CANCEL}
                         </Button>
+                        <div className="exports-panel__button-group__separator" ></div>
```
iwehrman: Extra space!

mcilroyc:
I pushed up fixes to the nits you found so far.

src/style/shared/button.less
```diff
@@ -199,3 +199,55 @@ button {
...
+        &:checked + label:after {
+          opacity: 1;
+        } 
+    }
+}
+
+.button-checkbox__mixed {
+    label {
+        &:after {
+            top: 2px;
```
iwehrman: I am reluctant to merge pixel-based measurements into master because it's hard to remember to get them back out, and they always cause trouble down the road. I realize this isn't the final styling pass though. Would you be interested in trying to do an eyeball-quality conversion from px to rem here? Or, I guess get @designedbyscience to give you a free hall pass.

src/js/models/exportasset.js
```diff
@@ -35,10 +35,25 @@ define(function (require, exports, module) {
...
+     * @type {Imutable.List.<string>}
+     */
+    var FORMATS = Immutable.List(["png", "jpg", "svg", "pdf"]);
+
+    /**
+     * Supported scales of export assets
+     *
+     * @type {Imutable.List.<number>}
+     */
+    var SCALES = Immutable.List([0.5, 1, 1.5, 2]);
```
iwehrman: Seems like we might as well be a little more generous here... At least 3x!

src/js/models/documentexports.js
```diff
@@ -55,7 +85,7 @@ define(function (require, exports, module) {
      * @param {number} layerID
      * @return {Array.<object>} An array of ExportAsset-like JS objects
      */
-    DocumentExports.prototype.layerExportsArray = function (layerID) {
+    Object.defineProperty(DocumentExports.prototype, "layerExportsArray", objUtil.cachedLookupSpec(function (layerID) {
```
iwehrman: It is dangerous to cache mutable objects. You Have Been Warned.

src/js/models/documentexports.js
```diff
@@ -48,6 +49,35 @@ define(function (require, exports, module) {
...
+                return [];
+            }
+        },
+
+        /**
+         * Get a list of layer IDs that have a non-empty set of configured export assets
+         *
+         * @return {Immutable.IndexedSeq.<number>}
+         */
+        layerIDsWithExports: function () {
```
iwehrman: I'm curious about why you chose a Seq here. I usually think of Seqs as useful when incrementally constructing complicated data structures because they obviate construction of the intermediate data structures. But it doesn't seem like that's how you're using this?

src/js/models/documentexports.js
```diff
@@ -48,6 +49,35 @@ define(function (require, exports, module) {
...
+    Object.defineProperties(DocumentExports.prototype, objUtil.cachedGetSpecs({
+        /**
+         * Produce an array of ExportAssets for the root of this document, converted to a JS object
+         * This is used for serializing the data to Ps metadata store
+         *
+         * @return {Array.<object>} An array of ExportAsset-like JS objects
+         */
+        rootExportsArray: function () {
+            if (this.rootExports && this.rootExports.size > 0) {
+                return this.rootExports.toJS();
```
iwehrman: Again, I'm generally curious about the mixture of mutable and immutable data structures. I'm not saying that it's necessarily bad, but I wonder what the constraints were that lead to a mix of both. It's almost always _simpler_ if you only use mutable or immutable data structures, though of course there are valid reasons to use a mixture.

src/js/jsx/shared/CheckBox.jsx
```diff
@@ -89,9 +97,11 @@ define(function (require, exports, module) {
                     title={this.props.title}
                     className={className} >
                     <input
+                        id={this.state.uniqueId}
                         type="checkbox"
                         checked={checked}
                         onChange={!this.props.disabled && this._handleClick} />
+                    <label htmlFor={this.state.uniqueId}></label>
```
iwehrman: Huh, I didn't know about `htmlFor`.

mcilroyc: I tried to use `for` and react was all Uh I think you meant `htmlFor`

src/js/jsx/sections/export/LayerExports.jsx
```diff
@@ -45,48 +46,30 @@ define(function (require, exports, module) {
...
-        "svg": {
-            id: "svg",
-            title: "SVG"
-        },
-        "pdf": {
-            id: "pdf",
-            title: "PDF"
-        }
-    });
+    var _formatOptions = ExportAsset.FORMATS
```
iwehrman: More Seq questions here. Keep in mind that if you use this data structure as a lookup table then every time you perform a lookup you have to re-do the `reduce` part!

src/js/actions/export.js
```diff
@@ -317,6 +326,50 @@ define(function (require, exports) {
...
+
+        if (!documentExports) {
+            return Promise.resolve();
+        }
+
+        var layersWithExports = documentExports.getLayersWithExports(document, false);
+
+        return this.transfer(setLayerExportEnabled, document, layersWithExports, exportEnabled);
+    };
+    setAllNonABLayersExportEnabled.reads = [locks.JS_DOC];
```
iwehrman: Seems like maybe there should be a JS_EXPORT lock which is read here, and probably written elsewhere? Or were you thinking that the JS_DOC lock would cover all export stuff? That's plausible, but sort of violates the vague association I have with locks to models/stores.

iwehrman:
OK, actually done with a first pass. (Sorry I got distracted half-way through yesterday.) Biggest questions: 1) Seqs; 2) pixels; and 3) locks.

src/style/shared/button.less
mcilroyc: This was easy to fix

src/js/models/documentexports.js
mcilroyc: ya... I'll get rid of this monkey business.

src/js/models/documentexports.js
mcilroyc: All I want is set of the layer IDs, which in this case come from the *key* of the map I'm operating on. So, I used `.keySeq()` to get just the keys (after filtering).  If that seems wrong to you, another obvious solution would be for this function to just return the filtered `map`, and have the consumer test with `map.has()` instead of `seq.includes()` like it is now.

src/js/models/documentexports.js
mcilroyc: I removed `rootExportsArray` entirely as it is both unused, and silly in hindsight.

src/js/actions/export.js
mcilroyc: My vague feeling is that this won't buy us much, since there is a high coupling between the two stores.  BUT, I'll give it a shot because it might increase clarity... and I should probably let the machine decide if the more granular lock helps, instead of my intuition.

src/js/models/documentexports.js
iwehrman: You could either do `.keySeq().toList()` or use `.map` to get the keys.

src/js/actions/export.js
iwehrman: Also, a lot of actions request JS_DOC read locks, so it would be nice _in theory_ to be able to write to _just_ JS_EXPORTS concurrently.

mcilroyc:
Comments addressed.  Back to @iwehrman

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -151,41 +167,59 @@ define(function (require, exports, module) {
...
+            var layerExportComponent = function (layer) {
+                var layerExports = layerExportsMap.get(layer.id);
+
+                return (
+                    <LayerExportsItem
+                        document={document}
+                        layer={layer}
+                        layerExports={layerExports}
+                        freshState={freshState}
+                        key={"layer-" + layer.id} />
```
iwehrman: Layers have a `.key` property which could be useful here?

mcilroyc: ok!  (squashed into second commit)

iwehrman:
:+1: :heavy_multiplication_x: :100:

PR #2095, opened 2015-08-19
During start up, ensure that generator is enabled.  Forcibly start it if necessary.

mcilroyc:
depends on https://github.com/adobe-photoshop/spaces-adapter/pull/133

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -34,6 +34,7 @@ define(function (require, exports) {
     var dialog = require("./dialog"),
         events = require("js/events"),
         locks = require("js/locks"),
+        objUtil = require("js/util/object"),
         collection = require("js/util/collection"),
         log = require("js/util/log"),
         ExportAsset = require("js/models/exportasset"),
@@ -414,14 +415,14 @@ define(function (require, exports) {
     exportAllAssets.transfers = [updateLayerExportAsset];
 
     /**
-     * Before start up, Ensure that generator is enabled, and then initialize the export service
+     * After start up, ensure that generator is enabled, and then initialize the export service
      * @return {Promise}
      */
-    var beforeStartup = function () {
+    var afterStartup = function () {
         return descriptor.playObject(generatorLib.getGeneratorStatus())
             .bind(this)
             .then(function (status) {
-                var enabled = status.generatorStatus.generatorStatus === 1;
+                var enabled = objUtil.getPath(status, "generatorStatus.generatorStatus") === 1;
                 if (!enabled) {
                     log.info("Enabling Generator...");
                     return descriptor.playObject(generatorLib.setGeneratorStatus(true))
@@ -451,8 +452,8 @@ define(function (require, exports) {
                 return Promise.resolve("Export Service not enabled, but giving up");
             });
     };
-    beforeStartup.reads = [];
-    beforeStartup.writes = [];
+    afterStartup.reads = [];
+    afterStartup.writes = [locks.JS_EXPORT, locks.GENERATOR];
 
     /**
      * Handle the standard onReset action
@@ -460,8 +461,10 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var onReset = function () {
-        _exportService = null;
-        return Promise.resolve();
+        return _exportService.close()
+            .finally(function () {
+                _exportService = null;
+            });
     };
     onReset.reads = [];
     onReset.writes = [];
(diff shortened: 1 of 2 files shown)


src/js/actions/export.js
```diff
@@ -358,23 +359,41 @@ define(function (require, exports) {
...
      */
     var beforeStartup = function () {
-        _exportService = new ExportService();
-        return _exportService.init()
+        return descriptor.playObject(generatorLib.getGeneratorStatus())
             .bind(this)
-            .then(function () {
-                return this.dispatchAsync(events.export.SERVICE_STATUS_CHANGED, { serviceAvailable: true });
+            .then(function (status) {
+                var enabled = status.generatorStatus.generatorStatus === 1;
```
mcilroyc: I should use objUtil.getPath() to protect against response structure variations

src/js/actions/export.js
```diff
@@ -358,23 +359,41 @@ define(function (require, exports) {
...
 
     /**
-     * before start up, initialize the export service
+     * Before start up, Ensure that generator is enabled, and then initialize the export service
      * @return {Promise}
      */
     var beforeStartup = function () {
-        _exportService = new ExportService();
-        return _exportService.init()
+        return descriptor.playObject(generatorLib.getGeneratorStatus())
```
iwehrman: I feel rather strongly that this should go in `afterStartup`, not `beforeStartup`. The UI and store should be robust to Generator not being enabled for some period of time, which means we shouldn't block rendering the UI while waiting for Generator to start. As a rule, if you don't want to block showing the UI, do it `afterStartup`.

src/js/actions/export.js
```diff
@@ -358,23 +359,41 @@ define(function (require, exports) {
...
+                    return descriptor.playObject(generatorLib.setGeneratorStatus(true))
+                        .catch(function (e) {
+                            throw new Error("Could not enable generator", e);
+                        });
+                }
             })
-            .catch(Promise.TimeoutError, function (e) {
-                log.error("Could not connect to generator plugin!", e);
-                return Promise.resolve("Export Service not enabled, but giving up because of timeout");
+            .delay(500)
```
iwehrman: I'm confused about this delay. Why is it necessary?

iwehrman: Maybe because you're waiting for the plugin to initialize? Seems like it would be better to find a way to have the export service poll the websocket server?

src/js/actions/export.js
```diff
@@ -358,23 +359,41 @@ define(function (require, exports) {
...
+            .then(function () {
+                _exportService = new ExportService();
```
iwehrman: Also, I may have wrongly complained about this before, but I just want to be sure that your intention is _not_ to clean this up on reset. Generally, state like this should be reset on reset in `onReset`. But, maybe this is intentional.

iwehrman: The minimum bar is: it would be OK for this to be garbage collected without explicitly calling "disconnect" or something like that.

src/js/actions/export.js
mcilroyc: After generator startup (well, after the response to the photoshop command), it takes some time before the plugin is initialized and the websocket will accept requests.  Even after this 500ms delay, you'll notice that it takes another few seconds before the websocket connections don't fail.  Maybe there is a more positive way to do this (maybe get an event from Ps when generator has started?)

mcilroyc: There is already polling in place inside NodeConnection, so I don't think it is necessary to have a second layer of polling.  We could get rid of the delay, but that just means that the polling will start earlier and will fail more times.  Side note: I still think we can squelch some of the errors that hit the console when the polling fails (prior to max retries reached)... but obviously not a high priority.

mcilroyc:
back to @iwehrman for a second look.

mcilroyc:
also I rebased.

src/js/actions/export.js
iwehrman: I think the follow-up bug should be: squelch errors and make the NodeConnection polling work so reliably that we can safely remove this delay. E.g., don't have polling give up after N attempts, but instead after M seconds.

iwehrman: RT if you agree, fav if you don't.

iwehrman:
LGTM. @mcilroyc to file a follow-up bug for removing the delay and sprucing up NodeConnection polling.

Issue #2092, opened 2015-08-18
Text Layer name not restored after undo+redo


mcilroyc:
1. Create a point-text layer by clicking on the canvas with the text tool
1. Type JELLO
1. Hit ESC
1. Note in the Layer Panel that the layer name becomes JELLO
1. Undo
1. Redo
1 Note the Layer Panel shows a default layer name, not JELLO
1. Help > Update Current Document ; see that it now reflects JELLO

mcilroyc:
Heads up @pineapplespatula and @iwehrman in case there is overlap with your bugs ... but I will otherwise assign this to myself.  My hunch is that we are updating the document out of band (after the history state has been captured)

volfied:
I think @DivyaPrabhakar should look at this bug and learn about the wonderful world of `shouldComponentUpdate`, my 2Çents

volfied:
I retract my statement, this might be something far more sinister

iwehrman:
@DivyaPrabhakar should not look directly into this bug :sunglasses:

iwehrman:
Even though @mcilroyc filed this, I'd like to assign it back to @chadrolfs to close. Seems likely that he'll hit another few bugs in this area when verifying :grimacing:

PR #2096, opened 2015-08-19
After type layer commit, 'make' event handler resets Layers and then amends current history with the result

mcilroyc:
After committing a new text layer, some time in the future get a 'make' event.  If we determine that this event is related to the type layer we've just added to the model (and for which we've already captured a history state), then we need to *amend* the history state with the late-breaking information.

In practice this means the layer name has changed.  It is now informed by the type layer's text content.

This PR also spruces up history state 'amendment' handling;  it is now generally handled more efficiently, and less chatty.

addresses #2092

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -386,9 +386,10 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer|Immutable.Iterable.<Layer>} layers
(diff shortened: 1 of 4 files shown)


iwehrman:
This looks fair to me. I was just going to ask if this applied to #2097 as well, but I see you have already done so! I think it all makes sense, but I guess the real test will be whether I can figure out the next instance on my own :) Would you mind summarizing this issue during the standup today?

Issue #2093, opened 2015-08-18
Text transformation by changing H does not update font size


ktaki:
With a text layer selected, change H value in the property panel. Font size is not updated till the layer goes to editing mode next time. The font size should be updated immediately.

iwehrman:
Note to self: when fixing this, remove `transform.setBounds`, which is now dead code.

iwehrman:
Indeed, I missed this case in my recent PR. But I made a branch with the additional change, and then discovered that Photoshop is _also_ not sending us the right "impliedFontSize" after we play the "transform" action descriptor. So this is now blocked on a Photoshop core bug, and requires a Watson issue. And I'm probably the best person to do that, so leaving this assigned to me.

mcilroyc:
FBNC, back to @ktaki

ktaki:
Verified the fix.

PR #2097, opened 2015-08-19
Reset type size after resizing text layers in the transform panel

iwehrman:
Calls `resetLayers` on type layers at the end of `transform.setSize`. Also, removes `transform.setBounds`, which was dead code.

Note for testing: Photoshop doesn't change the implied font size if you just change the layer's width; it seems to only actually care about the height. (I'm not sure if that's guaranteed though, so I'm not factoring that into the logic here.)

Addresses #2093.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -57,7 +57,6 @@ define(function (require, exports) {
         quality: "draft"
     };
 
-
     /**
      * Helper function that will break down a given layer to all it's children
      * and return layerActionsUtil compatible layer actions to move all the kids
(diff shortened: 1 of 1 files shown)


src/js/actions/transform.js
```diff
@@ -524,6 +476,17 @@ define(function (require, exports) {
...
 
+        // Reset type layers to pick up their implicit font size changes
+        sizePromise = sizePromise
+            .bind(this)
+            .then(function () {
+                var typeLayers = layerSpec.filter(function (layer) {
+                    return layer.kind === layer.layerKinds.TEXT;
+                });
+
+                return this.transfer(layerActions.resetLayers, document, typeLayers);
```
mcilroyc: `turns out` you may want to come back here after #2096 has merged, and tack on a third `true` param here.  Otherwise, this also suffers from the undo+redo combo ... such that the implied font size is not correctly restored.

I know this seems like it is getting *more complicated* instead of less, but this is helping at least to codify these edge cases where it makes sense to amend the current history. That was not a concept built in to the history store originally, so hopefully it can tightened up in round 2.

iwehrman: I understand the issue now. Your suggestion makes perfect sense. Thank you! I've amended the PR and force-pushed a replacement commit.

PR #2100, opened 2015-08-19
Layer panel search

designedbyscience:
You can search the layers panel, using the work done by @hamburge 

Hides the search when the panel is collapsed 

@placegraphichere Assigning to you for some spacing/sizing review

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -113,6 +113,11 @@ define(function (require, exports, module) {
                 this.refs.datalist.forceUpdate();
             }
         },
+        
+        _updateItems: function () {
+            var searchStore = this.getFlux().store("search");
+            searchStore._updateSearchItems(this.props.searchID);
+        },
 
         /**
          * Dismiss the parent dialog
@@ -398,7 +403,7 @@ define(function (require, exports, module) {
                     type: "placeholder"
                 };
 
-            var placeholderText = searchStrings.PLACEHOLDER,
+            var placeholderText = this.props.placeholder,
                 filter = this.state.filter;
 
             // If we have applied a filter, change the placeholder text
@@ -412,13 +417,15 @@ define(function (require, exports, module) {
 
             return (
                 <div
+                    className="dialog-search-bar__container"
                     onClick={this._handleDialogClick}>
                    <Datalist
                         ref="datalist"
+                        list={this.props.searchID + "_search_results"}
                         live={false}
                         className="dialog-search-bar"
                         options={this.state.options}
-                        startFocused={true}
+                        startFocused={this.props.startFocused}
                         placeholderText={placeholderText}
                         placeholderOption={noMatchesOption}
                         filterIcon={this.state.icon}
@@ -426,6 +433,7 @@ define(function (require, exports, module) {
                         dontCloseDialogIDs={this.state.filterIDs}
                         useAutofill={true}
                         neverSelectAllInput={true}
+                        onFocus={this._updateItems}
                         onChange={this._handleChange}
                         onClick={this._handleDialogClick}
                         onKeyDown={this._handleKeyDown} />
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -38,9 +38,11 @@ define(function (require, exports, module) {
         DummyLayerFace = require("jsx!./DummyLayerFace"),
         strings = require("i18n!nls/strings"),
         collection = require("js/util/collection"),
-        synchronization = require("js/util/synchronization");
+        synchronization = require("js/util/synchronization"),
+            SearchBar = require("jsx!js/jsx/search/SearchBar");
 
-    var PS_MAX_NEST_DEPTH = 10;
+    var PS_MAX_NEST_DEPTH = 10,
+        LAYER_SEARCH_ID = "layer_search";
 
     /**
      * Get the layer faces that correspond to the current document. Used for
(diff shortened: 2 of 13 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -113,6 +113,11 @@ define(function (require, exports, module) {
                 this.refs.datalist.forceUpdate();
             }
         },
+
```
iwehrman: Missing JSDoc.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -107,13 +109,18 @@ define(function (require, exports, module) {
...
+                pastDragTargets: dragAndDropState.pastDragTargets,
+                searchID: LAYER_SEARCH_ID + "_" + this.props.document.id
             };
         },
 
         componentWillMount: function () {
             this._setTooltipThrottled = synchronization.throttle(os.setTooltip, os, 500);
             this._mountedLayerIDs = new Set();
+            
+
```
iwehrman: Extra newline.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -38,9 +38,11 @@ define(function (require, exports, module) {
         DummyLayerFace = require("jsx!./DummyLayerFace"),
         strings = require("i18n!nls/strings"),
         collection = require("js/util/collection"),
-        synchronization = require("js/util/synchronization");
+        synchronization = require("js/util/synchronization"),
+            SearchBar = require("jsx!js/jsx/search/SearchBar");
 
-    var PS_MAX_NEST_DEPTH = 10;
+    var PS_MAX_NEST_DEPTH = 10,
```
iwehrman: It would be good to JSDoc `@type` annotate both of these.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -520,6 +527,29 @@ define(function (require, exports, module) {
             this._setTooltipThrottled("");
             this._boundingClientRectCache = this.state.dragTargets ? new Map() : null;
         },
+
```
iwehrman: Two extra newlines. Newlines don't grow on trees, son!

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -615,6 +646,14 @@ define(function (require, exports, module) {
                         visible={this.props.visible}
                         disabled={this.props.disabled}
                         onDoubleClick={this.props.onVisibilityToggle}>
+                        <SearchBar
+                            ref="searchBar"
+                            searchID={this.state.searchID}
+                            startFocused={false}
+                            dismissDialog={this._closeSearchBar}
+                            executeOption={this._handleOption}
+                            placeholder={""}
```
iwehrman: This should just be omitted and the `SearchBar` should make "" the default prop.

src/style/search/search-bar.less
```diff
@@ -21,14 +21,30 @@
...
-@drop-down-width: 46rem;
-@dialog-radius: .4rem;
+
+
+.button-simple.button-clear-search {
+    position: absolute;
+    top: 1.2rem;
+    right: 1.4rem;
+    color: @search-text;
+    font-size:1.5rem;
```
iwehrman: :space_invader:

src/style/search/search-bar.less
```diff
@@ -171,76 +214,171 @@
...
+        background-size: 1.2rem;
+        background-position: 0.5rem 0.5rem;
+
+        margin-top: 0.5rem;
+        transform: none;
+
+        text-transform: none;
+        width: @search-bar-width;
+        min-height: 3rem;
+        max-height: 60rem; /*this is important for use with dialog.showModa()*/
```
iwehrman: `showModal`

src/style/search/search-bar.less
```diff
@@ -171,76 +214,171 @@
...
-    color: @search-border;
-    line-height:2.1rem;
-}
+.layer-search {
+    @search-bar-width: 23rem;
+    @drop-down-width: 19rem;
+    @dialog-radius: .4rem;    
+    
+    .dialog-search-bar__container {
+        background: url(../img/ico-search-white.svg) no-repeat left @search-bg;
```
iwehrman: Shouldn't this URL string be wrapped in quotes?

iwehrman:
I have filed my usual round of JSDoc/whitespace nits. The JS/JSX changes look fine to me. Over to @placegraphichere for CSS nits.

designedbyscience:
Oh, also of note: removed little white collapse panel arrows

iwehrman:
Unhandled exception when "deleting" a filter:
```
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null Select.jsx:540
 React.createClass._scrollToIfNeeded Select.jsx:607
 React.createClass.componentDidUpdate react-with-addons.js:1063 
```

iwehrman:
Also, @placegraphichere will correct me if I'm wrong, but I thought the point here was to filter the list of layers _in the layers panel_? So, for example, the user should be able to pick the filter and have the layers panel show just the artboard layers in the document.

iwehrman:
The reason I say that is because @placegraphichere originally felt like this was the primary purpose of the in-panel layer search: to filter the layers panel by some criteria, and then work in the panel with just those layers visible. That was the reason why we moved away the idea of _only_ using the modal search, and instead wish for this two-search solution.

placegraphichere:
@iwehrman is correct, @designedbyscience and I talked about this too. 

We may differ the filtering for now, as we have lots of UI to get to. 

Some Issues I'm seeing: 
- Clicking the search icon specifically does not focus the search field, you must click to the left of it. 
- Double clicking the search icon closes the panel.
- Icons and text seem slightly mis-aligned vertically. The text is 3px or so too low.
- Default open search field height seems to cut off the last item by just a few px. Can we either have it cut it off in the middle (half the 15th item), or account for the 15 full list-items? 
- type "ar" so autocomplete will show "Artboards". *Select* any layer in returned layers result area. The search term "Artboards" will instead be entered into the field. Seems like an error. The same will occur with any layer type autocomplete, and then selection of a layer instead of the layer type.
- layer selection from the search results seems to flash the UI, especially the toolbar.
- layer selection often takes a second or more to update the UI (dev build)
- close "x" should only appear when there are search terms present
- Using mouse to select a layer-type does not re-focus the search bar for additional typing

volfied:
Closing as we're not going ahead with this this release.

PR #2102, opened 2015-08-19
Guides in Design Space!

volfied:
Allows: 
- Dragging of existing guides
- Dragging a guide off screen to delete
- Overlay at the edges of canvas, allowing new guide creation

Technicals:
- Policies are created around each guide to allow mouse events to go to Photoshop during move tool
- Because guides of inactive documents can't be read, we re-read guides every time document changes
- Because guides are 1 indexed, rangeOpts fails if there are 0 guides, so we immediately return empty array in no guide cases

Requires latest spaces-adapter#134

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -33,7 +33,8 @@ define(function (require, exports) {
         selectionLib = require("adapter/lib/selection"),
         PS = require("adapter/ps");
 
-    var historyActions = require("./history"),
+    var guideActions = require("./guides"),
+        historyActions = require("./history"),
         layerActions = require("./layers"),
         toolActions = require("./tools"),
         searchActions = require("./search/documents"),
@@ -80,7 +81,8 @@ define(function (require, exports) {
         "targetLayers",
         "guidesVisibility",
         "smartGuidesVisibility",
-        "format"
+        "format",
+        "numberOfGuides"
     ];
 
     /**
@@ -149,6 +151,20 @@ define(function (require, exports) {
                     layers: layers
                 };
             });
+    };
+
+    /**
+     * Get an array of guide descriptors for the given document descriptor.
+     *
+     * @private
+     * @param {object} doc Document descriptor
+     * @return {Promise.<Array.<object>>}
+     */
+    var _getGuidesForDocument = function (doc) {
+        var docRef = documentLib.referenceBy.id(doc.documentID),
+            numberOfGuides = doc.numberOfGuides;
+
+        return guideActions._getGuidesForDocumentRef(docRef, numberOfGuides);
     };
 
     /**
@@ -232,9 +248,12 @@ define(function (require, exports) {
                 var indexRef = documentLib.referenceBy.index(index);
                 return _getDocumentByRef(indexRef)
                     .bind(this)
-                    .then(_getLayersForDocument)
-                    .then(function (payload) {
-                        this.dispatch(events.document.DOCUMENT_UPDATED, payload);
+                    .then(function (doc) {
+                        return Promise.join(_getLayersForDocument(doc), _getGuidesForDocument(doc),
+                            function (payload, guides) {
+                                payload.guides = guides;
+                                this.dispatch(events.document.DOCUMENT_UPDATED, payload);
+                            }.bind(this));
                     });
             }, this);
 
@@ -266,14 +285,17 @@ define(function (require, exports) {
                         var currentDocLayersPromise = _getLayersForDocument(currentDoc),
                             historyPromise = this.transfer(historyActions.queryCurrentHistory,
                                 currentDoc.documentID, true),
+                            guidesPromise = _getGuidesForDocument(currentDoc),
                             deselectPromise = descriptor.playObject(selectionLib.deselectAll());
 
                         return Promise.join(currentDocLayersPromise,
                             historyPromise,
+                            guidesPromise,
                             deselectPromise,
-                            function (payload, historyPayload) {
+                            function (payload, historyPayload, guidesPayload) {
                                 payload.current = true;
                                 payload.history = historyPayload;
+                                payload.guides = guidesPayload;
                                 this.dispatch(events.document.DOCUMENT_UPDATED, payload);
                                 this.dispatch(events.application.INITIALIZED, { item: "activeDocument" });
                             }.bind(this))
@@ -316,19 +338,25 @@ define(function (require, exports) {
                 var layersPromise = _getLayersForDocument(doc),
                     historyPromise = current ?
                         this.transfer(historyActions.queryCurrentHistory, doc.documentID, true) :
-                        Promise.resolve(null);
+                        Promise.resolve(null),
+                    guidesPromise = _getGuidesForDocument(doc);
 
-                return Promise.join(layersPromise, historyPromise,
-                    function (payload, historyPayload) {
+                return Promise.join(layersPromise, historyPromise, guidesPromise,
+                    function (payload, historyPayload, guidesPayload) {
                         payload.current = current;
                         payload.history = historyPayload;
+                        payload.guides = guidesPayload;
                         return this.dispatchAsync(events.document.DOCUMENT_UPDATED, payload);
-                    }.bind(this));
+                    }.bind(this))
+                    .bind(this)
+                    .then(function () {
+                        this.transfer(toolActions.resetBorderPolicies);
+                    });
             });
     };
     updateDocument.reads = [locks.PS_DOC];
     updateDocument.writes = [locks.JS_DOC];
-    updateDocument.transfers = [historyActions.queryCurrentHistory];
+    updateDocument.transfers = [historyActions.queryCurrentHistory, toolActions.resetBorderPolicies];
     updateDocument.lockUI = true;
 
     /**
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -164,7 +164,11 @@ define(function (require, exports) {
         var docRef = documentLib.referenceBy.id(doc.documentID),
             numberOfGuides = doc.numberOfGuides;
 
-        return guideActions._getGuidesForDocumentRef(docRef, numberOfGuides);
+        if (numberOfGuides === 0) {
+            return Promise.resolve([]);
+        }
+
+        return guideActions._getGuidesForDocumentRef(docRef);
     };
 
     /**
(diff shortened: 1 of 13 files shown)


src/js/stores/document.js
```diff
@@ -1011,8 +1014,51 @@ define(function (require, exports, module) {
...
+         */
+        _handleGuideMoved: function (payload) {
+            var documentID = payload.documentID,
+                index = payload.index,
+                document = this._openDocuments[documentID],
+                orientation = payload.orientation,
+                position = payload.position;
+
+            var nextGuide = document.guides.get(index)
+                .merge({
```
mcilroyc: I think .merge should be indented further IF you are going to include other vars below in the same stmt.

also this might throw if there is no guide at `index`

src/js/stores/document.js
```diff
@@ -1011,8 +1014,51 @@ define(function (require, exports, module) {
...
+         * @private
+         * @param {{documentID: number, index: number}} payload
+         */
+        _handleGuideDeleted: function (payload) {
+            var documentID = payload.documentID,
+                index = payload.index,
+                document = this._openDocuments[documentID];
+
+            var nextGuides = document.guides.delete(index),
+                nextDocument = document.set("guides", nextGuides);
```
mcilroyc: just for grins, this could be combined into a `deleteIn()` call

src/js/actions/guides.js
```diff
@@ -0,0 +1,286 @@
...
+        adapterUI = require("adapter/ps/ui"),
+        adapterOS = require("adapter/os");
+
+    var locks = require("js/locks"),
+        events = require("js/events"),
+        policy = require("./policy"),
+        EventPolicy = require("js/models/eventpolicy"),
+        PointerEventPolicy = EventPolicy.PointerEventPolicy;
+
+
```
mcilroyc: extra space

src/js/actions/guides.js
```diff
@@ -0,0 +1,286 @@
...
+            },
+            getOpts = {
+                failOnMissingProperty: true
+            };
+
+        // FIXME: Should we reverse these?
+        return descriptor.getPropertiesRange(docRef, rangeOpts, _guideProperties, getOpts);
+    };
+
+    var _currentGuidePolicyID = null;
```
mcilroyc: jsdoc

src/js/stores/document.js
```diff
@@ -1011,8 +1014,51 @@ define(function (require, exports, module) {
...
+
+            this.setDocument(nextDocument, true);
         }
     });
 
+
```
mcilroyc: :space_invader:

src/js/models/guide.js
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Constructs the guide descriptors of a document into a list of Guide objects
+     *
+     * @param {object|Immutable.Record} documentDescriptor
+     * @param {Array.<object>} guideDescriptors
+     * @return {Immutable.List.<number, Guide>}
+     */
+    Guide.fromDescriptors = function (documentDescriptor, guideDescriptors) {
+        return guideDescriptors.reduce(function (guides, descriptor) {
```
mcilroyc: I would *personally* prefer the construct: Immutable.List(descs.map(...))

volfied:
@mcilroyc Updated this PR with the later changes as well... Can you review again (I addressed your old comments too)

src/js/actions/documents.js
```diff
@@ -235,7 +255,7 @@ define(function (require, exports) {
                     .then(_getLayersForDocument)
                     .then(function (payload) {
                         this.dispatch(events.document.DOCUMENT_UPDATED, payload);
-                    });
+                    }.bind(this));
```
mcilroyc: this shouldn't necessary?  see the bluebird .bind(this) above

volfied: Leftover from another change, oops.

iwehrman:
hey you guides

src/js/actions/guides.js
```diff
@@ -0,0 +1,329 @@
...
+            } else {
+                return Promise.resolve();
+            }
+        }
+
+        var guides = currentDocument.guides;
+
+        // If selection is empty, remove existing policy
+        if (!guides || guides.isEmpty()) {
+            if (currentPolicy) {
```
mcilroyc: DRY :dromedary_camel:  this seems like a repeat of the block a few lines up

src/js/actions/guides.js
```diff
@@ -0,0 +1,329 @@
...
+            }
+            
+            return new PointerEventPolicy(adapterUI.policyAction.ALWAYS_PROPAGATE,
+                adapterOS.eventKind.LEFT_MOUSE_DOWN,
+                {}, // no modifiers
+                guideArea);
+        }).toArray();
+
+        var removePromise;
+        if (currentPolicy) {
```
mcilroyc: :dromedary_camel:

src/js/actions/guides.js
```diff
@@ -0,0 +1,329 @@
...
+            _currentGuidePolicyID = null;
+            removePromise = this.transfer(policy.removePointerPolicies,
+                currentPolicy, false);
+        } else {
+            removePromise = Promise.resolve();
+        }
+
+        return removePromise.bind(this).then(function () {
```
mcilroyc: wrap these to two new lines, I think, for consistency

src/js/actions/guides.js
```diff
@@ -0,0 +1,329 @@
...
+
+    /**
+     * Register event listeners for guide edit/delete events
+     * 
+     * @return {Promise}
+     */
+    var beforeStartup = function () {
+        // Listen for guide set events
+        _guideSetHandler = function (event) {
+            var target = event.null._ref;
```
mcilroyc: are you pretty sure `.null` will always be there?  If not, you could use objUtil.getPath(event, "null._ref")

src/js/actions/guides.js
```diff
@@ -0,0 +1,329 @@
...
+
+            if (_.isArray(target) && target[0]._ref === "document" && target[1]._ref === "good") {
```
mcilroyc: also test of length>0?  I'm probably just being paranoid

src/js/actions/tools.js
```diff
@@ -204,15 +207,23 @@ define(function (require, exports) {
             removePromise = Promise.resolve();
         }
 
-        return removePromise.bind(this).then(function () {
-            return this.transfer(policy.addPointerPolicies, pointerPolicyList);
-        }).then(function (policyID) {
-            _currentTransformPolicyID = policyID;
-        });
+        return guidePromise
```
mcilroyc: would it also work to just `return Promise.join(guidePromise, removePromise).bind(this).then(...` ? 
My gut says that would be easier to read.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -0,0 +1,251 @@
...
+    var DEBOUNCE_DELAY = 200;
+
+    var GuidesOverlay = React.createClass({
+        mixins: [FluxMixin, StoreWatchMixin("document", "tool", "application", "ui")],
+
+        /**
+         * Keeps track of current mouse position so we can rerender the overlaid layers correctly
+         * @type {number}
+         */
+        _currentMouseX: null,
```
mcilroyc: This may just be a re-hashing of old conversations, but my immediate reaction when I see a bunch of private props as the opening salvo of a React component class is: why aren't these just tracked in the component state?

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -0,0 +1,251 @@
...
+                this.updateMouseOverHighlights();
+            }
+        },
+
+        /**
+         * Calls all helper functions to draw sampler overlay
+         * Cleans it first
+         * @private
+         */
+        drawOverlay: function () {
```
mcilroyc: I think all these private methods should be prefixed with `_`

src/style/tools/guides.less
```diff
@@ -0,0 +1,32 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+.guide-edges {
+    fill-opacity: 0.0;
+    stroke-opacity: 0.0;
+}
+
+.guide-edges-hover {
```
mcilroyc: *technically* our pattern would suggest this should be `guide-edges__hover` but.... feel free to argue that

src/js/stores/document.js
```diff
@@ -1011,6 +1016,73 @@ define(function (require, exports, module) {
...
+         * 
+         * @private
+         * @param {{documentID: number, index: number}} payload
+         */
+        _handleGuideDeleted: function (payload) {
+            var documentID = payload.documentID,
+                index = payload.index,
+                document = this._openDocuments[documentID];
+
+            var nextDocument = document.deleteIn(["guides", index]);
```
mcilroyc: :8ball:

mcilroyc:
done with first pass.  just a few nits or questions.

src/js/jsx/tools/GuidesOverlay.jsx
volfied: These aren't pure React components. They're empty `<g>`s that allow direct manipulation of their own DOM by D3.js. Interweaving the React component state and D3 drawing state would be much more trouble than it's worth so I skipped doing that.

src/js/jsx/tools/GuidesOverlay.jsx
volfied: Given these are D3/React components, my thinking was: main drawing methods should not be prepended with a _, just to give it a bit more structure in my mind as I was writing them.

volfied:
Back to you @mcilroyc, thanks for reviewing.

src/js/actions/guides.js
```diff
@@ -0,0 +1,313 @@
...
+    var onReset = function () {
+        descriptor.removeListener("set", _guideSetHandler);
+        descriptor.removeListener("delete", _guideDeleteHandler);
+
+        return Promise.resolve();
+    };
+    onReset.reads = [];
+    onReset.writes = [];
+
+    exports._getGuidesForDocumentRef = _getGuidesForDocumentRef;
```
mcilroyc: this is just a "maybe for a future revision" comment: Something rubs me the wrong way about exporting this "private" method.  Gut says there's a better way to factor this.

src/js/actions/tools.js
```diff
@@ -108,17 +109,16 @@ define(function (require, exports) {
             uiStore = this.flux.store("ui"),
             currentDocument = appStore.getCurrentDocument(),
             currentPolicy = _currentTransformPolicyID,
-            currentTool = toolStore.getCurrentTool();
+            currentTool = toolStore.getCurrentTool(),
+            removePromise = currentPolicy ?
```
mcilroyc: this really helped a lot!  thanks

src/js/jsx/tools/GuidesOverlay.jsx
mcilroyc: Fair.  This might suggest that it would breathe a little better if it were factored out into a separate module.  food for thought.  not for now.

Issue #2087, opened 2015-08-18
Effects: Outer (drop) shadow set to white does not show


placegraphichere:
- create shape layer overlaid on a dark background layer
- add an outer shadow to the shape layer
- change the color of the outer shadow to white
- nothing shows

expected: outer shadow should be white. 

Is there a blend mode on the outer shadow?

volfied:
Yeah, blend mode is multiply by default, which multiplies to nothing in this situation.

volfied:
What's the expected blend mode for drop shadows?

chadrolfs:
standard PS default for drop shadow blend mode is multiply

chadrolfs:
I'm a bit confused by this bug.  There is Outer Glow and Drop Shadow, but not Outer Shadow.

placegraphichere:
outer glow has no directional component, but has added "noise". Outer glow and shadow are very similar. 

The point of this issue is simply that we are using multiply as the default blend mode that makes some colors impossible to use. If we do not offer the ability to change blend modes, I'd rather we use a Normal blend mode and rely on opacity to soften the amount of shadow. This way more colors are available to users, and confusion I came across would not occur.

volfied:
Let's bring this up tomorrow at stand up and go from there. @placegraphichere do you see a good way to stick a blend mode popup in the drop shadow detail UI?

placegraphichere:
fixed with #2104. Closing

PR #2104, opened 2015-08-19
Set the default blend mode for shadows to be "Normal"

placegraphichere:
Ability to achieve lighter colors is impossible with current default
set to "multiply". This is a stop gap until we can get blend modes into
effects panel options.

This will take forever to review.

addresses #2087

### DIFF
--- a/src/js/models/shadow.js
+++ b/src/js/models/shadow.js
@@ -124,7 +124,7 @@ define(function (require, exports, module) {
          * Blend mode of the shadow
          * @type {BlendMode}
          */
-        blendMode: "multiply"
+        blendMode: "normal"
 
     });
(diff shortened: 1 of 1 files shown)


placegraphichere:
For reference (top is multiply, bottom is normal)

screen shot 2015-08-19 at 2 41 53 pm

screen shot 2015-08-19 at 2 43 28 pm
screen shot 2015-08-19 at 2 46 09 pm

chadrolfs:
The other possible alternative is to add Outer Glow (default: screen), which will provide for what you want to happen with Drop Shadow (default: multiply)

volfied:
Lighting the fuse.

Issue #1938, opened 2015-08-06
Show/Hide Guides has wrong shortcut, conflicts with Group (cmd/ctrl + g)


chadrolfs:

iwehrman:
What should the correct shortcuts be? CC @placegraphichere @designedbyscience

chadrolfs:
it was cmd/ctrl + ;
why did it change?

iwehrman:
It's possible that I screwed it up in #1926.

chadrolfs:
6a20b769818918de6976525279c024088434a204 is the first bad commit
commit 6a20b769818918de6976525279c024088434a204
Author: Ian Wehrman <wehrman@adobe.com>
Date:   Wed Aug 5 14:40:08 2015 -0700

    Localize menu and global keyboard shortcuts. Also factor out menu strings into their own file.

iwehrman:
I changed my mind. @DivyaPrabhakar should fix this.

DivyaPrabhakar:
https://github.com/adobe-photoshop/spaces-design/pull/2109

PR #2109, opened 2015-08-19
Fixed the keyboard shortcut for Toggle Guides

DivyaPrabhakar:
Previously it was CMD + g. Now it is corrected to CTRL/CMD ; 
Regarding issue #1938


volfied:
can't argue with that!

Issue #1881, opened 2015-07-31
Text Edit: color picker does not update to reflect color of selected text


chadrolfs:
select some characters of a text layer, open color picker and select a red, select some other characters that are black

Result: color picker remains unchanged (red)

pineapplespatula:
this might be a good one for @DivyaPrabhakar

iwehrman:
https://watsonexp.corp.adobe.com/#bug=4038142

chadrolfs:
As soon as I try to initially set the text color, it blows up on me:
`Action type.setColor failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Select” is not currently available.`

new doc
hit T for type tool
click on canvas and start typing creating mulitple lines (type, enter, type, enter)
hit Esc to commit the text
hit V for select tool
double-click on type and select one line
click on the Type Color picker and set it to a red
BOOM

iwehrman:
This is the recent change which is causing this bug:
https://github.com/adobe-photoshop/spaces-design/commit/4ae20a279e2416926bc9cd9cdafc39b767a05795#diff-66fb006bb46f8f716c009f317aded016R268

iwehrman:
@volfied @shaoshing @mcilroyc: Let us discuss this in ye olde parking lot tomorrow morning.

mcilroyc:
FBNC @chadrolfs

PR #2110, opened 2015-08-19
Skip the selection dance if possible

iwehrman:
This adds a short-circuit return to `playSimpleLayerActions` when the one target layer is equal to the one selected layer. This helps avoid problems with selecting layers in modal tool states. Presumably this could be pushed into `playLayerActions` more made more wide-ranging, but I couldn't immediately convince myself that that was safe. (E.g., I think if there is more than one layer, the existing function will play the play object once for each layer, in each case with only the one layer selected, and so we can't necessarily skip the dance in that case?)

Addresses #1881.


mcilroyc:
:tangerine:

PR #2115, opened 2015-08-19
Sewalker/color me bad

placegraphichere:
- consolidated duplicate colors
- remove incorrect colors
- removed color variables that are used only once, consolidated into existing variables
- icons have proper hover and active states across toolbar and panels
- transparency grid is white with light grey background for libraries, color, swatch and now library sample swatch
- library item selected and hover states match layers panel
- small color additions and changes for design accuracy
- refined layer rename input look
- refined disabled states for icons
- icon for dropdown hover state
- consistent hover and focus state cleanup

### DIFF
--- a/src/style/shared/color-picker.less
+++ b/src/style/shared/color-picker.less
@@ -60,7 +60,7 @@
     flex-direction: column;
     justify-content: center;
     margin: 0 @dialog-margin @dialog-margin @dialog-margin;
-    color: @warm-black;
+    color: @bg-alt;
     overflow-y: auto;
 }
 
@@ -74,7 +74,6 @@
     display: flex;
     flex-direction: column;
     justify-content: flex-end;
-
 }
 
 .color-picker .drop-down {
@@ -83,8 +82,8 @@
   width: 7rem;
 }
 
-.color-picker .drop-down input{
-  color: @warm-black;
+.color-picker .drop-down input {
+  color: @bg-alt;
   font-size: @text-small;
 }
 
@@ -121,7 +120,7 @@
     }
 
     &.color-picker-map__dark .color-picker-map__pointer {
-        border-color: @warm-white;
+        border-color: @item-active;
     }
 
     &.color-picker-map__light .color-picker-map__pointer {
@@ -204,9 +203,9 @@
     }
 
     .color-picker-slider__track,
-    .color-picker-slider__track-overlay {    
+    .color-picker-slider__track-overlay {
         border-radius: 0.3rem;
-        border: 1px solid @warm-black;
+        border: 1px solid @bg-alt;
     }
 
     .color-picker-slider__pointer {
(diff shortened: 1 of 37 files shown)


src/style/shared/colors-light-theme.less
```diff
@@ -0,0 +1,100 @@
+/*
```
iwehrman: This is awesome, but I don't think the world is quite ready. I suggest removing this file from the PR and maintaining it in a separate branch. (But _definitely_ keep it in a separate branch. We might well want to merge it before we ship as a sort of easter egg...)

src/style/shared/color-picker.less
```diff
@@ -121,7 +121,7 @@
     }
 
     &.color-picker-map__dark .color-picker-map__pointer {
-        border-color: @warm-white;
+        border-color: @white-active;
```
iwehrman: Some day it would be great to rename all these variables so that they are color agnostic. For example, instead of `@white-active` it might be `@foreground-active` or something like that.

iwehrman:
I'm happy to merge this because it seems like a nice cleanup. Preconditions:

1. Remove the light-theme file from this branch, and put it in a different branch. Ideally do this in a way that, when I check out said branch, I'm running the light theme. (Later we'll figure out a better way to switch themes, etc.) Please note the name of this new branch in this PR. 
2. Squash your commits and force-push a single, new commit to this branch. @DivyaPrabhakar can help with that if you've never done it before.
3. (optional) Have @designedbyscience take a look too. But, he's busy, and the best way to get him to look at these changes will be my merging them right in!

iwehrman:
hover

iwehrman:
This is just my opinion, but I don't think the active text inputs feel active enough any more:
text

iwehrman:
Armchair designer: I like the new colors, but maybe I think the default grey text color is sliiiiightly too dark grey? 
image
The rectangle above is the grey color I'm talking about.

iwehrman:
I've been saving this one for just this PR:
just a touch

placegraphichere:
What is the issue on the library name hover gif @iwehrman ?

iwehrman:
The arrow seems to change size/shape?

placegraphichere:
armchair-of-thrones

chadrolfs:
my new wallpaper!

iwehrman:
Uneasy lies the head that wears a crown.

placegraphichere:
back to @iwehrman for review

iwehrman:
Good news: you get to practice your rebase skills again!

iwehrman:
There seems to be a LESS parse error in button.less:223:
image

iwehrman:
Looks like `@warm-black` is used but never defined.

placegraphichere:
maybe now @iwehrman

iwehrman:
butts

placegraphichere:
http://stream1.gifsoup.com/view7/4503231/samuel-l-jackson-shark-attack-o.gif

Issue #1462, opened 2015-06-12
Disabled input hover styling is confusing


DesignSpaceBot:
The hover styling for disabled inputs is the same as for enabled inputs, which is confusing:
![hover](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)
The hover styling for disabled inputs should be muted somehow to indicate that click is not an enabled gesture. (But ideally it would still hint that text inside of disabled inputs is selectable.)

-- @iwehrman at [2015-05-22T15:41:27Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1462)

iwehrman:
Hey @placegraphichere, what should this look like? @DivyaPrabhakar will fix it!

iwehrman:
@DivyaPrabhakar If @placegraphichere doesn't respond here, HUNT HIM DOWN.

placegraphichere:
Disabled state from spec for a quiet text input (no box around text) looks like: 

Text color: #535353 
Text input underline color: #3f3f3f (no change)

![screen shot 2015-08-12 at 11 13 19 am](https://cloud.githubusercontent.com/assets/3229167/9229521/2d4cbe62-40e3-11e5-95a1-10d2885d7440.png)

DivyaPrabhakar:
https://github.com/adobe-photoshop/spaces-design/pull/2148

iwehrman:
Fixed by #2148.

PR #2116, opened 2015-08-19
Disabled the blue outline on disabled text inputs to signify it is inactive

DivyaPrabhakar:
Regarding #1462

### DIFF
--- a/src/style/shared/number-input.less
+++ b/src/style/shared/number-input.less
@@ -30,6 +30,10 @@
     border: @hairline solid @highlight;
 }
 
+.number-input:disabled:hover {
+    border: none;
+}
+
 .number-input__dirty:focus,
 .number-input__clean:focus {
     .background-transparency(@warm-white, 1);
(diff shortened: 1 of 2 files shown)


iwehrman:
Did you figure out how to get a disabled text input??

DivyaPrabhakar:
@iwehrman Yup!

Create document
Places some shapes on it 
Deselect everything -- The W and H inputs are disabled. 

I think it also works if you:
Create new document
Create Artboard 
Have nothing selected -- all the inputs are inactive

src/style/shared/number-input.less
```diff
@@ -30,6 +30,10 @@
     border: @hairline solid @highlight;
 }
 
+.number-input:disabled:hover {
+	border: none;
```
iwehrman: This is a tab character. Thou shalt not use tab characters! Instead, you must configure your editor to _always_ create four-space tab stops.

iwehrman: Looks good otherwise :)

Issue #2106, opened 2015-08-19
CC Lib: Style naming is inconsistent, rename Text Styles > Character Styles


chadrolfs:
Tooltip at bottom of panel: Add Character Style
Label within library: Text Styles

Standard PS: Character Styles

Issue #2042, opened 2015-08-13
CC Lib: hitting Esc from libraries drop down leaves app unresponsive


chadrolfs:
click on libraries drop down (field selects all text of loaded library)
hit Esc

Result:  `Warning: Failed propType: Required prop `library` was not specified in `Library`. Check the render method of `LibrariesPanel`.`
`Action libraries.selectLibrary failed: TypeError: Cannot read property 'elements' of undefined`
app unresponsive now

Issue #2037, opened 2015-08-13
CC Lib: library sorting doesn't match standard PS


chadrolfs:
Compare library sorting on mac/win/standard/DS

Issue #2036, opened 2015-08-13
CC Lib: you can get into a state where library and assets are out of sync


chadrolfs:
https://www.dropbox.com/s/h9zk3m1fxnnocuh/CCLib-mismatch.mov?dl=0

PR #2125, opened 2015-08-20
CC Libraries: more complete support for editing library and asset

shaoshing:
This PR adds support to the following:

* create / rename / delete library
* rename / delete asset
* library / asset buttons are all functional

It also addresses the following issues:

* #1984 
* #1989
* #2042
* #2036
* #1987
* #2037
* #2106

@volfied please review.

lib-editing


src/js/stores/library.js
```diff
@@ -109,36 +111,109 @@ define(function (require, exports, module) {
...
+         */
         _handleLibraryRemoved: function (payload) {
-            var id = payload.id;
+            var modified = this.getCurrentLibrary().modified,
+                nextLibraryID = null;
+
+            this._libraries = this._libraries.delete(payload.id);
+
+            // Pick another library as selected library.
+            if (!this._libraries.isEmpty()) {
```
shaoshing: This is the switching rule in CEP.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -39,28 +77,47 @@ define(function (require, exports, module) {
...
+            var selectedCommand = libraryID,
+                newLibraryName = selectedCommand === _RENAME_LIBRARY ? this.props.selected.name : "";
+
+            this.setState({
+                command: selectedCommand,
+                newLibraryName: newLibraryName
+            });
+
+            // Cancel Datalist selection
+            return false;
```
volfied: Is this function supposed to return bool, can you reflect this in the JSDoc, and also return something in the first if.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -39,28 +77,47 @@ define(function (require, exports, module) {
...
 
         /**
          * Given the libraries, creates the datalist friendly
          * options for the library picker
          *
-         * @param {Array.<AdobeLibraryComposite>} libraries
-         *
          * @private
+         * @param {Array.<AdobeLibraryComposite>} libraries
          * @return {{title: String, id: string}}
```
volfied: Please reflect the newly added returned keys in this JSDOC line as well (or just use object)

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -96,24 +160,203 @@ define(function (require, exports, module) {
...
+         *
+         * @private
+         * @return {?ReactComponent}
+         */
+        _renderDeleteConfirmationDialog: function () {
+            if (this.state.command !== _DELETE_LIBRARY) {
+                return null;
+            }
+
+            var collaborationMapBody = {};
```
volfied: Why not just an object with keys pre-defined?

volfied: And this object looks like something you can define in top level

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -96,24 +160,203 @@ define(function (require, exports, module) {
...
                 listID = "libraries-" + this.props.document.id;
 
-            return (
+            var sharedLibrary = false,
+                libraryLink,
+                shareLink,
+                collaborateLink;
+
+            if (selectedLibrary) {
+                sharedLibrary = selectedLibrary.collaboration !== _REGULAR_LIBRARY;
```
volfied: maybe call this `isSharedLibrary` so it's obvious it's not a library itself

src/js/jsx/shared/Datalist.jsx
```diff
@@ -56,6 +59,8 @@ define(function (require, exports, module) {
             useAutofill: React.PropTypes.bool,
             // If true, will not highlight input text on commit
             neverSelectAllInput: React.PropTypes.bool,
+            // If true, mouse over an item will automatically select it and trigger the onChange callback.
+            autoSelect: React.PropTypes.bool,
```
volfied: Why do you need this when we already have `live` prop?

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -96,24 +160,203 @@ define(function (require, exports, module) {
...
+         * @return {?ReactComponent}
+         */
+        _renderLibraryNameInput: function () {
+            var command = this.state.command;
+
+            if (!command || command === _DELETE_LIBRARY) {
+                return null;
+            }
+
+            var rename = command === _RENAME_LIBRARY;
```
volfied: Another variable where it's not clear whether it's a flag or an action.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -96,24 +160,203 @@ define(function (require, exports, module) {
...
+
+            return (<div className="libraries__bar__top__content libraries__bar__top__content-input">
+                <TextInput
+                    type="text"
+                    live={true}
+                    continuous={true}
+                    className="libraires__bar__input"
```
volfied: typo here

volfied: Also, as far as these CSS classes are concerned, `__` should be only used to separate the "states" of a class from each other. So most of these should be `libraries-bar_input`, `libraries-bar_top-content` etc.

src/js/jsx/shared/TextInput.jsx
```diff
@@ -1,26 +1,26 @@
 /*
  * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
- *
```
iwehrman: You gotta turn off your remove-all-trailing whitespace plugin because it makes these diffs really messy. But if somebody wants to go to the trouble to a) in one commit, remove all trailing whitespace from the repo; b) augment one of our grunt tasks to verify the absence of trailing whitespace, optionally c) suggest an editor plugin that will help everyone enforce this automatically, I would merge it!

src/js/jsx/sections/libraries/LibraryList.jsx
shaoshing: Unfortunately Chrome does not support using variable as property name in {} syntax and I want to avoid repeating the category strings here. Note that this is one of the features in ECMAScript 6, which will allow us to write something like these:

```js
var collaborationMapBody = {
  [_OUTGOING_LIBRARY]: ...,
  [_INCOMING_LIBRARY]: ...,
}
```

Given that `collaborationMapBody` does not look like a constant, and putting it in top level will cache the translation strings (although we cannot change language in DS), this might be a better place for it.

src/js/jsx/shared/TextInput.jsx
shaoshing: I totally agree with you, and the setting is turned off.

src/js/jsx/shared/Datalist.jsx
shaoshing: By default, Datalist will 1) automatically select any item that is hovered or selected via up/down keys. With the `live` option turned off, Datalist will not fire change event for every item select event, but 2) it will still fire a change event when the dialog is closed (by click on an item, or press ESC). The `autoSelect` option allows you to turn off behavior 1 and 2, and It will fire a change event only when a menu item is clicked.

src/js/jsx/sections/libraries/LibraryList.jsx
shaoshing: The bottom three items in the libraries menu are commands and should not be selected by the user. Returning false will indicate Datalist to not set these items as selected. I added the callback comment here https://github.com/adobe-photoshop/spaces-design/pull/2125/files#diff-2679935a0e56611617e48e9185145f97R46

iwehrman:
Drive-by UX suggestion: after choosing "create" or "rename", select all the default text in the input to help people understand that they should edit the text before taking action. 

Drive-by comment: this looks so great, @shaoshing. 
billy dee

src/js/jsx/sections/libraries/LibraryList.jsx
iwehrman: Hopefully we'll get this and many other ES6 improvements right after M3!

volfied:
Ok then!

shaoshing:
thank you @iwehrman . I actually added the auto-focus to the input, but it doesn't always work somehow. I will include a fix in a separate PR.

src/js/jsx/sections/libraries/LibraryList.jsx
shaoshing: Ohhhh yeah.

Issue #2050, opened 2015-08-14
Unable to collapse export panel


iwehrman:
```
Uncaught TypeError: Cannot read property 'parentElement' of null PanelSet.jsx:156 
```

iwehrman:
CC @designedbyscience @mcilroyc. I suspect @designedbyscience knows how to fix this since he fixed a similar problem previously with the Libraries panel.

designedbyscience:
It needs `ref={UI.EXPORT_PANEL}` in PanelSet.jsx

mcilroyc:
I implemented eric's suggestion in PR #2056 and it *helps*, but there is still some work to be done.  Lucky for me, he volunteered, so assigning this over...

Issue #2094, opened 2015-08-18
Incorrect CC Libraries panel height


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9344720/87c0f88c-45bf-11e5-87fa-be45dd7fe177.png)

PR #2133, opened 2015-08-20
Updates to expand/collapse behavior

designedbyscience:
Makes panel expansion more sane, also addresses #2050

### DIFF
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -181,7 +181,7 @@
 
 .layers .section-container {
     padding: 0;
-    flex-shrink: 1;    
+    flex-shrink: 1;
 }
 
 .section-container, .section-container__no-collapse {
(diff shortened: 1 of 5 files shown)


designedbyscience:
Oh, and #2094

src/style/panel.less
```diff
@@ -181,7 +181,7 @@
 
 .layers .section-container {
     padding: 0;
-    flex-shrink: 1;
+    flex-shrink: 1;
```
iwehrman: bloop

iwehrman:
So you're sticking with the whole "do nothing if you try to collapse the last open panel" thing? At least in the case of two-panels-per-column, this seems worse than the current behavior. But if you and @placegraphichere are on the same page, I'll go along!

designedbyscience:
I said "more sane", not "completely sane"

iwehrman:
More sane it is.

Issue #2135, opened 2015-08-20
Stroke button in sampler HUD should set strokes, not fills


iwehrman:
@placegraphichere will correct me if I'm wrong, but I think that the stroke button in the sampler HUD should set the target stroke color (if possible) and not the target vector fill.

placegraphichere:
that is correct

iwehrman:
Fixed by #2139.

PR #2139, opened 2015-08-20
When sampling stroke, apply the whole stroke to shape layers

volfied:
Addresses #2135, @iwehrman please review.


iwehrman:
:+1::+1::+1:

Issue #1914, opened 2015-08-05
Color Picker has Opacity slider during modal type tool when it isn't applicable


pineapplespatula:
We cannot set the opacity of individual characters in text runs, ( because all we have is layer opacity to play with)
but when you open the color picker it lets you try and change the opacity 

we should disable it.

iwehrman:
Fixed by #2142.

PR #2142, opened 2015-08-20
Color picker opaque mode

iwehrman:
This adds an opaque mode to the `ColorInput` and `ColorPicker` that constrains selectable colors to be opaque and disables the opacity slider. The `ColorInput` used by the `Type` component now uses opaque mode when in a modal tool state.

This is definitely imperfect, but I think it's an improvement. Sometimes the non-opaque color flashes in the input before switching to an opaque color because it takes a while for PS to send us the modal state change notification. There's also nothing we can really do about people typing in non-opaque colors. 

Also, in opaque mode we force 100% opacity, but my first attempt just froze the existing opacity. After trying it out, it seemed like 100% opacity was preferable because it was more clear that the opacity was completely disabled.

Addresses #1914.

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -64,7 +64,7 @@ define(function (require, exports, module) {
             };
 
             if (!Immutable.is(this.state.postScriptMap, nextState.postScriptMap)) {
-                if (nextState.postMap) {
+                if (nextState.postScriptMap) {
                     // The list of all selectable type faces
                     var typefaces = nextState.postScriptMap
                         .entrySeq()
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -44,7 +44,7 @@ define(function (require, exports, module) {
         textLayer = require("adapter/lib/textLayer");
 
     var Type = React.createClass({
-        mixins: [FluxMixin, StoreWatchMixin("font")],
+        mixins: [FluxMixin, StoreWatchMixin("font", "tool")],
 
         shouldComponentUpdate: function (nextProps, nextState) {
             var getTexts = function (document) {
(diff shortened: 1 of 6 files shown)


src/style/shared/color-picker.less
```diff
@@ -223,6 +223,12 @@
     }
 }
 
+.color-picker-slider__disabled {
+    .color-picker-slider__pointer {
+        background: darken(whitesmoke, 20%);
```
iwehrman: FYI @placegraphichere

src/js/jsx/sections/style/Type.jsx
```diff
@@ -63,36 +63,55 @@ define(function (require, exports, module) {
                 return collection.pluck(document.layers.selected, "opacity");
             };
 
+            if (!Immutable.is(this.state.postScriptMap, nextState.postScriptMap)) {
+                if (nextState.postMap) {
+                    // The list of all selectable type faces
+                    var typefaces = nextState.postScriptMap
```
iwehrman: Moving this here so it only needs to be recalculated when the `postScriptMap` changes, instead of every time any flux state changes.

src/js/jsx/sections/style/Type.jsx
```diff
@@ -63,36 +63,55 @@ define(function (require, exports, module) {
                 return collection.pluck(document.layers.selected, "opacity");
             };
 
+            if (!Immutable.is(this.state.postScriptMap, nextState.postScriptMap)) {
+                if (nextState.postMap) {
```
volfied: you mean `nextState.postScriptMap`?

src/js/jsx/sections/style/Type.jsx
```diff
@@ -63,36 +63,55 @@ define(function (require, exports, module) {
...
+
+                            return {
+                                id: psName,
+                                title: fontObj.font
+                            };
+                        })
+                        .toList();
+
+                    this.setState({
+                        typefaces: typefaces
```
volfied: This is calling setState in shouldComponentUpdate!

volfied: I don't have a good advice on where to put this, `componentWillReceiveProps` only relies on `nextProps`. Feels like this logic should be part of the fontStore, instead of this view doing it.

volfied:
Got a likely typo, and a big React problem. @iwehrman

src/js/jsx/sections/style/Type.jsx
iwehrman: Yeah, I know what you mean. I could swear I thought about that and then decided against it for some reason... I'll try again.

iwehrman: (Though, in general, I don't think it's _necessarily_ wrong to have state changes in `shouldComponentUpdate`. But I agree it's often a bad code smell.)

iwehrman: Oh, I remember now why this isn't in the font store. It's because what we're building here is explicitly a view-only data structure: the options, in a particular format for `Datalist` and `Select`. It used to be that we recalculated this on every flux state update, and it seemed less bad to move it here. So, that's what I'm sticking with unless you want to argue about it more!

volfied: Yeah, but you're relying on the new state to compute the result of this function, and setState is async, isn't it?

iwehrman:
Typo fixed.

src/js/jsx/sections/style/Type.jsx
iwehrman: It was async previously too...

volfied: Considering this is a view only thing, and we only use it in render to pass it to a childProp. I recommend moving this logic to a private function or the `render` method, and not having this code smell!

volfied:
@iwehrman let me know if you'll make the change to move the typefaces calculation to render or leave it where it is.

src/js/jsx/sections/style/Type.jsx
iwehrman: But then you have to re-compute it on every render! This way you only have to re-compute it when the postScriptMap changes, which is very infrequent.

volfied:
Alright then!

Issue #2128, opened 2015-08-20
Guides: can't move them on Windows


chadrolfs:

chadrolfs:
you can create them by clicking in work area edge (blue highlight) but can't drag them

PR #2145, opened 2015-08-20
Hide the guide overlay on click so mouse events can go to PS

volfied:
Addresses #2128 

Problem was, because overlay was still showing, our synthesized mouse event was not going to Photoshop.Created a Watson bug for Fitz, since it goes to PS in Mac, and not in Windows, causing the issue.


iwehrman:
OK!

Issue #2137, opened 2015-08-20
Sampler HUD failed on multi-text-runs


chadrolfs:
create a multi-run text layer (contains 2 different Typefaces)
select the sampler tool
mouse over type layer and hit Space

Result: `Action sampler.showHUD failed: Error: The font for layer is not available!`

chadrolfs:
![screen shot 2015-08-20 at 12 10 05 pm](https://cloud.githubusercontent.com/assets/8442631/9393107/0c0b4d06-4735-11e5-8b9c-604080ac720b.png)

volfied:
This is not the multitext, but the fact that font is not being found in the postScript map.

Stupid question: Are you using a missing font?

chadrolfs:
I'm not but I am using the staging build instead of dev 1462

chadrolfs:
here are my exact steps:
cmd+n
hit the T key 
click in AB and type "abaasdfasdfasdfasdf" (default font)
(while still in edit mode) select a series of characters in the middle "abaa*sdfas*dfasdfasdf" and set it to Helvetica Neue Bold Italic
hit Enter (on Num Pad) to commit text
hit I for Sampler
mouse over Type layer on canvas and hit Space

iwehrman:
FBNC to @chadrolfs. See #2175 for future work.

PR #2146, opened 2015-08-20
Disable type sampling in multi-text-runs since they don't have type info

volfied:
Apparently it's a design choice. So we disable type sampling

Addresses #2137


iwehrman:
It would be better if `getTypeObjectFromLayer` always returned an object which just has some subset of the properties defined... But I'm guessing you don't want to deal with all that?

volfied:
@iwehrman Maybe later, but that's going to take longer than this fix.

iwehrman:
I filed #2175 to capture the future work.

PR #2148, opened 2015-08-21
Took away the blue hover on disabled inputs

DivyaPrabhakar:
Signifies they are not active
Regarding #1462

### DIFF
--- a/src/style/shared/number-input.less
+++ b/src/style/shared/number-input.less
@@ -30,6 +30,10 @@
     border: @hairline solid @highlight;
 }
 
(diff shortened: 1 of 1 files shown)


iwehrman:
Seems great, but I wonder if this addresses a particular issue? Hmmm...

DivyaPrabhakar:
😅

iwehrman:
Ah, #1462. I must have overlooked that! Anyway, the fix looks great. Thanks!

DivyaPrabhakar:
You are so irresponsible Ian.

Issue #2108, opened 2015-08-19
CC Libraries datalist arrow is confusing without an underline


iwehrman:
The Libraries datalist is confusing because there is no obvious association between the text and the drop-down arrow:
![image](https://cloud.githubusercontent.com/assets/1075154/9369144/3717b098-467c-11e5-8726-8a830c887b20.png)

The other datalists have an underline that connects the text to the arrow:
![image](https://cloud.githubusercontent.com/assets/1075154/9369162/4e4fc8ae-467c-11e5-8f95-df31e1b4107d.png)

iwehrman:
I am no longer confused.

PR #2151, opened 2015-08-21
Styling changes to the CC Libraries bar top

DivyaPrabhakar:
- Added an underline to "My Library" so that it signifies it is a datalist
Regarding #2108 

- Took away the bottom border for the title bar because it was looking too crowded with both the underline and the bottom border. Got designs from @davealonzo 

- Changed the height of the title to 5rem as per the design. Should we choose to add the bottom border again and take away the underline, the height of the top bar will still be the correct size. https://trello.com/c/jRlxiRui/491-design

### DIFF
--- a/src/style/sections/libraries/libraries-section.less
+++ b/src/style/sections/libraries/libraries-section.less
@@ -88,19 +88,27 @@
             justify-content: space-between;
             flex: none;
             height: 4rem;
-            margin: 0 2rem;
+            margin: 0 0rem;
             align-items: center;
-            border-top: @hairline solid @layer-divider;
-            border-bottom: @hairline solid @layer-divider;
         }
 
         &__content-input {
             border-bottom: @hairline solid @focus-highlight;
         }
+        display: flex;
+        flex-direction: row;
+        justify-content: space-between;
+        flex: none;
+        height: 5rem;
+        margin: 0 2rem;
+        align-items: center;
+        
         .drop-down {
             flex-grow: 1;
             border: 0;
-
+            width: 18.4rem;
+            border-bottom: @hairline solid @underlines;
+            
             input {
                 width: 100%;
                 font-size: 1.3rem;
@@ -110,7 +118,7 @@
 
         .libraries__split-button-list {
             flex: none;
-            width: 9rem;
+            width: 10rem;
             flex-shrink: 0;
             flex-grow: 0;
             justify-content: space-between;
@@ -152,7 +160,6 @@
         flex: none;
         height: 3rem;
         margin: 0 2rem;
-        border-top: @hairline solid @layer-divider;
 
         .libraries-bar__section__left {
             flex: none;
--- a/src/style/shared/select.less
+++ b/src/style/shared/select.less
@@ -25,7 +25,7 @@
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
-    padding: 0.5rem 0.8rem 0.5rem 0.8rem;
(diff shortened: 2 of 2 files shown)


shaoshing:
Thanks @DivyaPrabhakar . By removing the bottom border, It seems like unbalanced between the top border and list underline (is it just me?). We might have to also think about the design when creating / renaming a library. You can rebase master to import these new features into your branch. 

screen shot 2015-08-21 at 9 23 54 am

screen shot 2015-08-21 at 9 36 46 am

designedbyscience:
Can we just kill both of those hairlines? We are moving away from the hairlines in general. Maybe make the following 3 buttons have even space to that drop down (rather than align right?)

davealonzo:
@designedbyscience We can certainly can kill the hairline below the "LIBRARIES" title to match the Layers panel. But we should keep the blue one that invokes for create or rename library workflows

DivyaPrabhakar:
@davealonzo @designedbyscience @shaoshing  

So the consensus here is to:

1. Kill the top and bottom hairlines, and only have the underline that indicates a datalist

2. Add the underline to the bottom of My Library datalist 
alt text

3. About the evenly spacing the icons part, they are evenly spaced already because the width of the data list is pretty long. You can see that when we add the underline -- the underline goes all the way to the Share icon. Should I shorten the datalist width and then spread out the icons?

volfied:
@DivyaPrabhakar in case you're waiting on this, this PR needs rebase.

DivyaPrabhakar:
@davealonzo @designedbyscience @shaoshing Once I get a confirmation and answer on the comment above about the final changes, I will make the changes, rebase and have these changes be merged.

src/style/sections/libraries/libraries-section.less
```diff
@@ -90,17 +90,24 @@
...
+        flex: none;
+        height: 5rem;
+        margin: 0 2rem;
+        align-items: center;
+        
         .drop-down {
             flex-grow: 1;
             border: 0;
-
+           border-bottom: @hairline solid fade(@color-subsection-rule, 60%);
```
iwehrman: You appear to be missing exactly one space here :space_invader:

iwehrman:
Should the datalist text be flush left here? 
flush

DivyaPrabhakar:
Talked to @davealonzo about how the UI should look regarding @iwehrman's question about datalist alignments. He approved the changes that I made which now should look like this: 
 
alt text

- Removed hairlines
- Datalist flush left
- Extended width of datalist to reach icons
- Icons are flush right 
- Underlined the datalist

alt text

- Added padding to the bottom of the dropdown, under "Delete 'My Library'"

Changes made, back to you @iwehrman.

src/style/sections/libraries/libraries-section.less
```diff
@@ -88,19 +88,27 @@
...
+        height: 5rem;
+        margin: 0 2rem;
+        align-items: center;
+        
         .drop-down {
             flex-grow: 1;
             border: 0;
-
+            width: 18.4rem;
+            border-bottom: @hairline solid fade(@color-subsection-rule, 60%);
```
iwehrman: Should we be `fade`ing colors here? Is that kosher? Is there a better way that will make our lives easier later when trying to maintain different color themes? CC @placegraphichere

placegraphichere: lets not do this. I think @underline should work for this.

iwehrman:
Just one final question and then we're good to go.

iwehrman:
LGTM, thanks!

PR #2154, opened 2015-08-21
Export generator plugin connection improvements

mcilroyc:
- A few minor optimizations
- In debug mode, try to see if there is plugin websocket (briefly) *before* the usual routine of checking/starting generator.  This Allows us to use a remote generator in debug mode.
- Provide "maxConnectionAttempts to NodeDomain (this requires a new version of generator-connection.   PR forthcoming)

**This requires `bower install` to pick up the new generator-connection version specified in bower.json**

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -128,6 +128,10 @@ define(function (require, exports) {
         return descriptor.playObject(generatorLib.getGeneratorStatus())
             .then(function (status) {
                 return objUtil.getPath(status, "generatorStatus.generatorStatus") === 1;
+            })
+            .catch(function (e) {
+                log.error("Failed to determine generator status: %s", e.message);
+                return Promise.resolve(false);
             });
     };
 
@@ -140,7 +144,7 @@ define(function (require, exports) {
     var _setGeneratorStatus = function (enabled) {
         return descriptor.playObject(generatorLib.setGeneratorStatus(enabled))
             .catch(function (e) {
-                throw new Error("Could not enable generator", e);
+                throw new Error("Could not enable generator: " + e.message);
             });
     };
 
@@ -148,6 +152,7 @@ define(function (require, exports) {
      * Update the export store with the new service availability flag;
      *
      * @param {boolean} available
+     * @return {Promise}
      */
     var _setServiceAvailable = function (available) {
         return this.dispatchAsync(events.export.SERVICE_STATUS_CHANGED, { serviceAvailable: !!available });
@@ -465,8 +470,26 @@ define(function (require, exports) {
             return _exportService.init()
                 .bind(this)
                 .then(function () {
-                    log.info("Export: Plugin connection established");
+                    log.debug("Export: Generator plugin connection established");
                     return _setServiceAvailable.call(this, true);
+                });
+        };
+
+        // helper to enabled generator if necessary and then init exportService
+        var _enableAndConnect = function () {
+            return _getGeneratorStatus()
+                .then(function (enabled) {
+                    if (!enabled) {
+                        log.debug("Export: Starting Generator...");
+                        return _setGeneratorStatus(true);
+                    }
+                })
+                .then(function () {
+                    _exportService = new ExportService();
+                    return _initService.call(this)
+                        .catch(function (e) {
+                            throw new Error("ExportService.init explicitly returned: " + e.message);
+                        });
                 });
         };
 
@@ -474,7 +497,6 @@ define(function (require, exports) {
         // to see if the plugin is already running (perhaps on a remote generator connection)
         var preCheck;
         if (globalUtil.debug) {
-            log.debug("Doing a quick check to see if Export plugin is available");
             _exportService = new ExportService(true); // quickCheck mode
             preCheck = _initService.call(this)
                 .return(true)
@@ -487,27 +509,12 @@ define(function (require, exports) {
         }
 
         return preCheck
-            .bind(this)
             .then(function (preCheckResult) {
-                return preCheckResult || _getGeneratorStatus()
-                    .bind(this)
-                    .then(function (enabled) {
-                        if (!enabled) {
-                            log.info("Export: Starting Generator...");
-                            return _setGeneratorStatus(true);
-                        }
-                    })
-                    .then(function () {
-                        _exportService = new ExportService();
-                        return _initService.call(this)
-                            .catch(function (e) {
-                                throw new Error("ExportService.init explicitly returned: " + e);
-                            });
-                    })
-                    .catch(function (e) {
-                        log.error("Export Service failed to initialize correctly because: " + e);
-                        return Promise.resolve("Export Service not enabled, but giving up");
-                    });
+                return preCheckResult || _enableAndConnect();
+            })
+            .catch(function (e) {
+                log.error("Export Service failed to initialize correctly because: " + e.message);
+                return Promise.resolve("Export Service not enabled, but giving up");
             });
     };
     afterStartup.reads = [];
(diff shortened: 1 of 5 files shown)


src/js/actions/export.js
```diff
@@ -119,6 +120,40 @@ define(function (require, exports) {
...
     /**
+     * Get current status of generator
+     * @private
+     * @return {Promise.<boolean>}
+     */
+    var _getGeneratorStatus = function () {
+        return descriptor.playObject(generatorLib.getGeneratorStatus())
+            .then(function (status) {
+                return objUtil.getPath(status, "generatorStatus.generatorStatus") === 1;
+            });
```
iwehrman: Might wanna add a second function to this `then` handler to catch potential failures and return false?

src/js/actions/export.js
```diff
@@ -119,6 +120,40 @@ define(function (require, exports) {
...
+    /**
+     * Enable/Disable generator based on supplied parameter
+     * @private
+     * @param {boolean} enabled
+     * @return {Promise}
+     */
+    var _setGeneratorStatus = function (enabled) {
+        return descriptor.playObject(generatorLib.setGeneratorStatus(enabled))
+            .catch(function (e) {
+                throw new Error("Could not enable generator", e);
```
iwehrman: Unfortunately I believe the second argument will be lost. I think you either have to stringify it and append it to the message, or set `.cause = e`;

src/js/actions/export.js
```diff
@@ -119,6 +120,40 @@ define(function (require, exports) {
...
+            });
+    };
+
+    /**
+     * Update the export store with the new service availability flag;
+     *
+     * @param {boolean} available
+     */
+    var _setServiceAvailable = function (available) {
+        return this.dispatchAsync(events.export.SERVICE_STATUS_CHANGED, { serviceAvailable: !!available });
```
iwehrman: `@return {Promise}`?

src/js/actions/export.js
```diff
@@ -416,40 +454,60 @@ define(function (require, exports) {
...
-            })
-            .delay(500)
-            .then(function () {
-                _exportService = new ExportService();
+        // helper function to wait for the service init, and set the service "available" in our store
+        var _initService = function () {
+            return _exportService.init()
+                .bind(this)
+                .then(function () {
+                    log.info("Export: Plugin connection established");
```
iwehrman: I think this is too chatty by default.

src/js/actions/export.js
```diff
@@ -416,40 +454,60 @@ define(function (require, exports) {
...
+                    return _setServiceAvailable.call(this, true);
+                });
+        };
 
-                return _exportService.init()
+        // In debug mode we should first do an abbreviated test
+        // to see if the plugin is already running (perhaps on a remote generator connection)
+        var preCheck;
+        if (globalUtil.debug) {
+            log.debug("Doing a quick check to see if Export plugin is available");
```
iwehrman: Ditto. I think we should only add log messages like this if we think they will be useful for some engineers. It doesn't seem like this passes that test.

iwehrman: But, sweet feature.

src/js/actions/export.js
```diff
@@ -416,40 +454,60 @@ define(function (require, exports) {
...
+        }
+
+        return preCheck
+            .bind(this)
+            .then(function (preCheckResult) {
+                return preCheckResult || _getGeneratorStatus()
                     .bind(this)
-                    .then(function () {
-                        return this.dispatchAsync(events.export.SERVICE_STATUS_CHANGED, { serviceAvailable: true });
+                    .then(function (enabled) {
```
iwehrman: It would be a little nicer to just return the above promise and move the rest of this chain up to the outer promise?

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -161,7 +161,7 @@ define(function (require, exports, module) {
             this.setState({ fresh: false }, function () {
                 var exportActions = this.getFlux().actions.export,
                     document = this.state.document;
-                return exportActions.setAllAssetsRequested(document).then(function () {
+                exportActions.setAllAssetsRequested(document).then(function () {
```
iwehrman: Why wait for the next render (is that when these async callbacks happen?) to start this process?

iwehrman:
Done with a first pass. Back to @mcilroyc.

src/js/actions/export.js
mcilroyc: As written, if _getGeneratorStatus() rejects, it will result in giving up on connecting to the service.  I guess your point is that we should swallow these, and try to enable generator anyway.  I can go with that.

src/js/actions/export.js
mcilroyc: changed to debug.  common, lemme have that at least

src/js/actions/export.js
mcilroyc: removed this debug()

src/js/actions/export.js
mcilroyc: derp, yes.

src/js/jsx/sections/export/ExportAllPanel.jsx
mcilroyc: There seems not to be a reason.  flattened in next commit.  nice catch

src/js/actions/export.js
mcilroyc: I went the ` + e.message` route.  Are you down with that?

mcilroyc:
back to @iwehrman for a second pass.

mcilroyc:
Realized mistake, Taking back

mcilroyc:
Ok, re-back to you.  I had hastily refactored the "preCheck" logic incorrectly.  Fixed now.  Please check that again to make sure seems more readable than the previous version.  I introduced some extra "chop" but got rid of the nested `.then()` structure.  The challenge is that I do not want to do *anything* further if preCheck returns true, because the connection is already established.

iwehrman:
Needs moar rebase.

mcilroyc:
rebased

src/js/actions/export.js
```diff
@@ -119,6 +120,45 @@ define(function (require, exports) {
...
+     * @return {Promise.<boolean>}
+     */
+    var _getGeneratorStatus = function () {
+        return descriptor.playObject(generatorLib.getGeneratorStatus())
+            .then(function (status) {
+                return objUtil.getPath(status, "generatorStatus.generatorStatus") === 1;
+            })
+            .catch(function (e) {
+                log.error("Failed to determine generator status: %s", e.message);
+                return Promise.resolve(false);
```
iwehrman: For future reference, you can just return the value here instead of a Promise-boxed value.

src/js/actions/export.js
```diff
@@ -385,30 +428,30 @@ define(function (require, exports) {
         }
 
         if (!_exportService || !_exportService.ready()) {
-            Promise.reject("Export Service is not available");
+            return _setServiceAvailable.call(this, false)
+                .finally(function () {
+                    return Promise.resolve("Export Service is no longer available");
```
iwehrman: Ditto. Although this is kind of a weird way for the overall promise to resolve.

src/js/actions/export.js
```diff
@@ -416,39 +459,61 @@ define(function (require, exports) {
...
+        // In debug mode we should first do an abbreviated test
+        // to see if the plugin is already running (perhaps on a remote generator connection)
+        var preCheck;
+        if (globalUtil.debug) {
+            _exportService = new ExportService(true); // quickCheck mode
+            preCheck = _initService.call(this)
+                .return(true)
+                .catch(function () {
+                    _exportService = null;
+                    return Promise.resolve(false);
```
iwehrman: Ditto.

iwehrman:
Minor nits to be fixed in the next round. Geronimo!

Issue #2149, opened 2015-08-21
"+" button should be excluded from the region to collapse/expand Export panel


ktaki:

iwehrman:
Same for style copy/paste icons.

ktaki:
Verified the fix.

PR #2155, opened 2015-08-21
Suppress double click events in title header buttons

iwehrman:
Stop event propagation of double click events in title header buttons (Export add asset and Style copy/paste), even when disabled, the latter with new `onDisabledClick` and `onDisabledDoubleClick` event handlers on `Button`.

Addresses #2149.


volfied:
Alrighty then

PR #2161, opened 2015-08-21
Disable auto expand during swap

volfied:
Idea for this fix came to me during the design meeting... We can just disable auto expand during swap, so the PS canvas size does not shift underneath us in between swap operations!

Addresses #1532, so happy to finally fix it.

Requires https://github.com/adobe-photoshop/spaces-adapter/pull/136


iwehrman:
Very nice! It doesn't seem like general idea would _always_ be safe, but it seems safe here and maybe there are other uses for it. If there are, maybe we could have an "expansion dance" utility some day.

Issue #2174, opened 2015-08-22
Spell checker should be disabled?


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9425207/94f3f45a-48bc-11e5-8639-d5c71ceaded3.png)

Apparently this may be possible in HTML without any core changes: http://stackoverflow.com/a/4524179

iwehrman:
Fixed by #2176

PR #2176, opened 2015-08-22
Disable spell checker on text inputs

iwehrman:
What it says! Also, made proper use of `classnames` in `NumberInput` and deleted the dead `NumberArrayInput` component.

Addresses #2174.


volfied:
That's simple.

Issue #2177, opened 2015-08-23
Search not dismissed after placing CC Libraries graphic


iwehrman:
1. Search for a graphic in one of your CC Libraries
2. Press enter

Result: the object is placed on the canvas in the placement modal tool state but the search bar is not dismissed. It seems to be waiting for the underlying operation to complete before closing. Instead it should close immediately.

iwehrman:
This is partially due to a really terrible core issue in which CEF goes completely dormant while we're in the placement modal tool state.

iwehrman:
In more detail, here is what's happening: when the library search result is selected, a dialog-close action is executed and a libraries place-graphic action is executed. Currently, the latter does not attempt to wait for the former, but fixing that doesn't help because we have no good way of _fully_ waiting for the dialog-close action to finish. The dialog-close action can block until the dialog-store change event has been emitted, and `setStateFromFlux` has been called on all the relevant UI components, but React re-renders asynchronously and there is no good/apparent way to block the close-dialog action until this point. So, as a result, React is still in the process of re-rendering some of the dialogs (i.e., closing them) when the libraries place-graphic action is executed. This is a problem because CEF completely hangs once Photoshop enters the place-smart-object modal tool state. At this point, _nothing happens in CEF until the graphic has been placed_. This is a real nightmare. For now I don't see anything that we can do except for introducing an artificial delay between the close-dialog action and the place-graphic action.

jsbache:
The problem is that while the insertion grid is up we are still inside the play command of the original place  command (THandleDroppedLibraryAssetsCommand).
While executing our own Play commands, we suppress cef-timers & ignore further Play requests.
What is the preferred behavior?

iwehrman:
My super-hack does not seem to have solved the problem. Instead I think I need to work with @jsbache to figure out a better way to do this.

PR #2178, opened 2015-08-23
SuperHack: Fix graphic placement from search bar

iwehrman:
Work around a combined React-and-PS bug in which we can't block actions until React rendering has finished, and PS stops running CEF entirely during the place-graphic modal tool state. More details in #2177, which this "addresses."


volfied:
Super PR

Issue #2113, opened 2015-08-19
SuperSelect: AB: Selecting in the empty canvas will select images (not visible) over the actual canvas (no AB)


zoranagee:
Have a couple of AB with images that extend beyond the AB size. Click out in the white space - the hierarchy selects the image (not visible) over nothing. This differs from the way PS handles targeting. Where selecting the white space always selects outside the AB (the canvas) and not the actual hidden image.

iwehrman:
@volfied This sounds like a core Photoshop bug with hit testing and artboards?

zoranagee:
PS works as I would expect.

iwehrman:
Presumably PS does not use the same hit test routine that Design Space uses.

volfied:
This is on us, PS hitLayerTest is returning empty. The obvious fix will break couple other interactions, so let me sit on this a bit.

volfied:
Back in April, to allow selecting layers BEHIND the canvas (remember before artboards, we had a canvas?), I started not being restrictive with the PS hitTest.. it causes problems like this one. I'll fix it.

volfied:
@chadrolfs Similar to #2112, if a layer is extending over the edges of an artboard, and is invisible in that location, clicking it should no longer select it outside the artboard.

zoranagee:
I still see this in Sprint 29. - when doing marquee drag it highlights a bunch of hidden (not in the eyeball sense - in the clipped to other AB sense)

Here I'm marquee dragging the 5 squares and bounding boxes from all the other layers in AB to the left show up.
<img width="1440" alt="screen shot 2015-08-27 at 9 46 03 am" src="https://cloud.githubusercontent.com/assets/7904855/9526828/700dae96-4ca0-11e5-9107-cfe359b104bb.png">

chadrolfs:
@zoranagee ^^ that isn't this bug.  That is specific to marquee select and it may be as designed.

volfied:
It is as designed right now. However, we should track in a separate issue to change the design so marquee select doesn't affect layers in an artboard unless the marquee is intersecting with the artboard.

Issue #2112, opened 2015-08-19
SuperSelect: showing blue bounding of images not on targeted AB


zoranagee:
There might be some edge case around this that should work as is today but if you have two AB - and you have AB 1 selected and as you mouse through, it will in fact show the bounding edge of an image that is in AB2 (that happens to overlap - not visibly - with AB1). Too many distracting bounding boxes appearing and disappearing that aren't relevant to the task at hand.

(I have a video if this does't make sense)

placegraphichere:
is this part of the "don't show hidden layer" issue @iwehrman ?

iwehrman:
Hopefully not, because that is fixed.

iwehrman:
@zoranagee please post a link to the video and, ideally, also a link to the PSD if you/Adobe own the rights to it. Otherwise, just send links via email internally.

volfied:
Can you post the video @zoranagee ?

zoranagee:
The file is up on my transfer on smb://sjshare.corp.adobe.com/Photoshop/ under "DesignSpace"

iwehrman:
Bumping this to high priority because it really makes demoing hard. Oh, and did I mention I need to demo to the big boss this week? :grimacing:

volfied:
@chadrolfs To reproduce this bug, create two artboards (make sure artboard 2 is higher than artboard 1 in the layers panel), put a BIG rectangle in artboard 2, that also overlaps artboard 1.
Then put a small rectangle in artboard 1, and try to highlight it, big rectangle in artboard 2 will steal the highlight, after this fix, it will not highlight because we're not hovering over it in it's own artboard.

PR #2181, opened 2015-08-24
Improved select tool selection criteria

volfied:
Updated the rules for superselect a bit. Back in April, to make layers outside canvas selectable I made changes and used PS's hitTest only for z-order sorting. Well, it caused bugs like #2112 and #2113 where bigger layers in other artboards, that are higher in the z-order would highlight over the layer user was visibly hovering on.

Now we restrict superselect with artboards:

- A layer is highlighted only when the mouse is over it in the artboard. Highlight is still for the full layer bounds, but doesn't get lit unless the mouse is in the artboard.
- Superselect clicks can only select layers on visible pixels.

@iwehrman Please review, as I've bothered you the most about this issue.

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -36,6 +36,7 @@ define(function (require, exports) {
         system = require("js/util/system"),
         locks = require("js/locks"),
         events = require("js/events"),
+        Bounds = require("js/models/bounds"),
         documentActions = require("./documents"),
         layerActions = require("./layers"),
         toolActions = require("./tools"),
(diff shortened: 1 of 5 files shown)


src/js/actions/superselect.js
```diff
@@ -182,15 +192,15 @@ define(function (require, exports) {
...
      * @param {Immutable.List.<number>} ids
      * @return {Immutable.Iterable.<Layer>}
      */
-    var _removeArtboardIDs = function (layerTree, ids) {
-        return ids.filterNot(function (id) {
+    var _getArtboardIDs = function (layerTree, ids) {
+        return ids.filter(function (id) {
             var layer = layerTree.layers.get(id);
             
             return layer ? layer.isArtboard : true;
```
iwehrman: Is there some reason why you're expecting a null layer here?

volfied: I think this was added recently due to getting null layers during new layer creation / click happening too fast (the newly created layer wouldn't be in our models yet)

iwehrman: Mmmm, yeah. Love those races!

Issue #2079, opened 2015-08-17
Icons for Delete and Toggle Drop/Inner Shadow are not visible or incorrect


ktaki:
Only delete icons are visible

chadrolfs:
I actually think it's the delete icons that are not showing but the tooltips depending on where the mouse is can be confusing.   I think the toggle icons might be wrong for all the styles (fill, stroke and shadows) and the delete icons for shadows are missing.

iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9425199/480c8b66-48bc-11e5-8a6b-07eae3b3b5fe.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #2182, opened 2015-08-24
Fix for Fill, Stroke and Effects icons bug

designedbyscience:
Addresses #2079

This is really a temporary fix for this bug, as we are changing the looks for these panels right now. Also seeing if switching the visibility icons like this is any better...not sure it is


iwehrman:
Ok, sure!

Issue #2069, opened 2015-08-16
SHIFT-clicking to add to a selection within the selection bounding box fails


placegraphichere:
- Create 4 rects in a 2x2 grid
- Attempt to select all 4 rects
- The final rect will not select since the bounding box of the current selection overlaps 

GIF shows selecting rects individually, then tries to multi-select them

![multi-selection-bounding-interference](https://cloud.githubusercontent.com/assets/3229167/9291677/566a8d94-4393-11e5-8120-27653026ee30.gif)

iwehrman:
@DivyaPrabhakar to narrow down the problem.

DivyaPrabhakar:
EOD Friday, August 21 update

Our policy events are propagating correctly on click or SHIFT + CMD + click. In both of those cases, the selection tool works perfectly no matter where on the layer you click.'

It gets messed up within the red bounding box when only SHIFT + click are pressed. For some reason, the SHIFT only is not propagating events correctly (I don't want to say it nullifies it because we are registering the click in the browser, but it does not call the click function, virtually making the area within the red bounds the Bermuda triangle when SHIFT is held down.) My theory is that the policy action is changing when SHIFT is pressed down and maybe changing to ALPHA_PROPAGATE or something other than NEVER_PROPAGATE which is messing with selection. 

Further investigation on Monday. Need to ask @volfied a list of questions I have generated after looking into this (Don't tell me the answer just yet Barkin, I want to figure this one out myself :)) 

What is happening now when you have SHIFT pressed down and you click INSIDE the red box: 
![alt text](https://www.dropbox.com/s/gyd1qwrapyij07z/not_working_select.gif?raw=1 "Inside + shift")

What is happening now when you have SHIFT pressed down and you click OUTSIDE the red box: 
![alt text](https://www.dropbox.com/s/s996ce91jzuuiqu/working_select.gif?raw=1 "Inside + shift")

iwehrman:
![gifs](http://cdn.meme.am/instances2/500x/1502228.jpg)

volfied:
@DivyaPrabhakar You've figured it out already... but I'll answer any and all questions you have!

volfied:
Over to @chadrolfs to confirm!

PR #2186, opened 2015-08-24
Fixed the shift clicking to select any layer that is clicked on

DivyaPrabhakar:
Changed the event policy for SHIFT to send events to Ds. Regarding issue #2069

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -172,7 +172,7 @@ define(function (require, exports) {
                     height: psSelectionHeight + outset * 2
                 }
             ),
-            outsideShiftPolicy = new PointerEventPolicy(adapterUI.policyAction.ALWAYS_PROPAGATE,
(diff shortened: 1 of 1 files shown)


src/js/actions/tools.js
```diff
@@ -175,7 +175,7 @@ define(function (require, exports) {
             outsideShiftPolicy = new PointerEventPolicy(adapterUI.policyAction.ALWAYS_PROPAGATE,
                 adapterOS.eventKind.LEFT_MOUSE_DOWN,
                 {
-                    shift: true
+                    shift: false
```
volfied: So modifiers table is not a true/false table. It's more of an existence thing. 

In this bug's case, you don't want to disable shift, you want shift clicks to all come to us, which you do by changing the propagation action of this policy (to NEVER)

volfied:
@DivyaPrabhakar so close!

DivyaPrabhakar:
@volfied Got it, hope this is the correct change now.

volfied:
:signal_strength:

PR #2198, opened 2015-08-24
Safety checks for updateMouserOverHighlights calls during DOM updates

volfied:
Fixes issues that happened when opening new documents that DOM elements would be out of date and we'd be trying to retrieve their layer models from an invalid document layer tree.

Argh


iwehrman:
Looks good, thanks!

PR #2202, opened 2015-08-25
Export plugin init hot fix: Fixed binding in _initService

mcilroyc:
Fixes an issue with loading the export plugin in the situation where we must first start generator


iwehrman:
Three for one!

PR #2203, opened 2015-08-25
Mcilroy/29 hotfix export

mcilroyc:
This is the master-side of this hotfix.  Note that this seems like a messy merge, but this is what its supposed to be like, I think.


iwehrman:
If this screws something up, blame @mcilroyc.

PR #2204, opened 2015-08-25
null state improvements

placegraphichere:
- hover, select states for recent/new files
- removed section border on last child
- added column for exiting Ds from null state

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -180,7 +180,7 @@ define(function (require, exports, module) {
                             }),
                             panelTabBarClassNames = classnames({
                                 "panel__tab-bar": true,
-                                "panel__tab-bar_visible": true
+                                "panel__tab-bar__visible": true
                             }),
                             propertiesButtonClassNames = classnames({
                                 "toolbar-button": true,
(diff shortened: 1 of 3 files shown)


src/js/jsx/PanelSet.jsx
```diff
@@ -279,10 +279,21 @@ define(function (require, exports, module) {
                 );
             } else if (this.state.recentFilesInitialized) {
                 return (
-                    <div ref="panelSet" className="panel-set__container">
-                        <div className="panel panel__visible">
+                    <div className="panel-set panel-set__active null-state">
+                        <PanelColumn
```
mcilroyc: The neighboring elements, when they have only one attribute, are kept to one line.. so maybe do that here

src/js/jsx/PanelSet.jsx
```diff
@@ -279,10 +279,21 @@ define(function (require, exports, module) {
...
+                            visible="true">
                             <RecentFiles recentFiles={this.state.recentFiles} />
                             <ArtboardPresets />
+                        </PanelColumn>
+                        <div className="panel__tab-bar panel__tab-bar_visible">
```
mcilroyc: double underscore before visible

mcilroyc:
two nits - otherwise :white_check_mark:

placegraphichere:
back atcha

PR #2216, opened 2015-08-25
Allow the user to choose an export folder during each export

mcilroyc:
This PR contains two major things:

1. When a user initiates an export, they are presented with a "folder chooser" dialog to choose where the assets should be exported.  This starts in their home directory, but then should remember their choice during this session.  (improvements to follow).
1. Consistency with the new version of the generator-spaces plugin that went through a few breaking changes in order to clean things up.

Requires generator-spaces PR
merged ~~https://github.com/adobe-photoshop/generator-spaces/pull/3~~
min version:  jenkins playground-dev builds (~535win / 538mac) 

One particular caveat:
When in the OS folder chooser dialog, do NOT try to create a new folder.  Choose an existing one.

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -38,6 +38,7 @@ define(function (require, exports) {
         globalUtil = require("js/util/global"),
         objUtil = require("js/util/object"),
         collection = require("js/util/collection"),
+        strings = require("i18n!nls/strings"),
         log = require("js/util/log"),
         ExportAsset = require("js/models/exportasset"),
         ExportService = require("js/util/exportservice");
@@ -101,7 +102,8 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var _exportAsset = function (document, layer, assetIndex, asset, baseDir) {
-        return _exportService.exportLayerAsset(document, layer, asset, baseDir)
+        var fileName = (layer ? layer.name : strings.EXPORT.EXPORT_DOCUMENT_FILENAME) + asset.suffix;
+        return _exportService.exportLayerAsset(document, layer, asset, fileName, baseDir)
             .bind(this)
             .then(function (pathArray) {
                 // Do we need to be aware of exports that return >1 file path?
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -56,7 +56,12 @@ define(function (require, exports) {
      */
     var _exportService;
 
-    var _lastFolderPath;
+    /**
+     * The previous folder into which assets were saved.
+     * If populated, initialize the OS folder chooser here
+     * @type {?string}
+     */
+    var _lastFolderPath = null;
 
     /**
      * Fetch relevant data from both the document and export stores, and sync the metadata to
@@ -442,18 +447,11 @@ define(function (require, exports) {
     setAllNonABLayersExportEnabled.writes = [];
     setAllNonABLayersExportEnabled.transfers = [setLayerExportEnabled];
 
-    var promptForFolder = function () {
-        return _exportService.promptForFolder();
-    };
-    promptForFolder.reads = [];
-    promptForFolder.writes = [locks.GENERATOR];
-    promptForFolder.transfers = [];
-
     /**
      * Export all assets for the given document for which export has been enabled (layer.exportEnabled)
      *
      * @param {Document} document
-     * @return {Promise} resolves when all assets have been exported
+     * @return {Promise} Resolves when all assets have been exported, or if canceled via the file chooser
      */
     var exportAllAssets = function (document) {
         if (!document) {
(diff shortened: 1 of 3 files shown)


src/js/actions/export.js
```diff
@@ -55,6 +55,8 @@ define(function (require, exports) {
      */
     var _exportService;
 
+    var _lastFolderPath;
```
volfied: Need JSdoc for this

src/js/actions/export.js
```diff
@@ -438,6 +440,13 @@ define(function (require, exports) {
     setAllNonABLayersExportEnabled.writes = [];
     setAllNonABLayersExportEnabled.transfers = [setLayerExportEnabled];
 
+    var promptForFolder = function () {
```
volfied: Ditto.

src/js/actions/export.js
```diff
@@ -461,16 +470,30 @@ define(function (require, exports) {
...
+                                    return _exportAsset.call(this, document, layer, index, asset, baseDir);
+                                }, this);
+                        }, this);
+
+                    return Promise.all(exportArray.toArray());
+                },
+                function (err) {
+                    // promptForFolder rejected, probably just user-canceled
+                    if (err.message && err.message.startsWith("cancelError: cancel")) {
+                        log.warn("Prompt for folder failed, and it wasn't a simple 'cancel'");
```
volfied: you say it wasn't a simple cancel, but this boolean check is testing for `startsWith("cancelError")`

src/js/actions/export.js
```diff
@@ -461,16 +470,30 @@ define(function (require, exports) {
...
+                    }
+                    return Promise.resolve();
```
volfied: Maybe resolve to an empty array here would make the function more consistent.

volfied:
Did one pass, initially looks good except couple nits and missing JSDoc.

src/js/actions/export.js
shaoshing: We should definitely make JSCS checking jsdoc existence for us.

volfied: JSCS `jsDoc#enforceExistence` only applies to functions :/

src/js/actions/export.js
mcilroyc: Yeah... this is a little janky, but I think it is an acceptable hack for now. This has to do with the way generator "Domain" infrastructure is bundling the error.  I didn't want to dig too deep since this implementation is still somewhat tentative.

src/js/actions/export.js
mcilroyc: I feel that since the JSDoc makes no mention of what is being promised, it is OK as is.  I updated the JSDoc to be slightly more clear.  This promise gets dumped on the floor anyway.

mcilroyc:
@volfied I fixed up some JSDoc, and answered some other concerns (hopefully).  I don't want to merge this yet ... but I DO want to get it to the point that it can be merged in close concert with my pushing the generator plugin to perforce (so that our playground-dev jenkins builds are not DOA for very long).  So feel free to keep poking holes in this as you see fit.

src/js/actions/export.js
```diff
@@ -461,16 +481,30 @@ define(function (require, exports) {
...
+                                .map(function (asset, index) {
+                                    return _exportAsset.call(this, document, layer, index, asset, baseDir);
+                                }, this);
+                        }, this);
+
+                    return Promise.all(exportArray.toArray());
+                },
+                function (err) {
+                    // promptForFolder rejected, probably just user-canceled
+                    if (err.message && err.message.startsWith("cancelError: cancel")) {
```
iwehrman: Shouldn't you re-throw the error the error in this case?

src/js/actions/export.js
```diff
@@ -439,10 +448,21 @@ define(function (require, exports) {
     setAllNonABLayersExportEnabled.transfers = [setLayerExportEnabled];
 
     /**
+     * Open a dialog to allow the user to choose a directory
+     * @return {Promise.<string>} Resolves with pathName, or rejects if user-canceled
+     */
+    var promptForFolder = function () {
```
iwehrman: I can't find any place where this action is used. Am I just missing it?

mcilroyc: derp.  I think I originally was orchestrating this from outside the action (JSX), but now I'm just calling _exportService directly in `exportAllAssets`.  I'll remove this as I don't see any reason we would want to transfer to it right now.

iwehrman:
Just a couple Qs. Looks good to me otherwise!

src/js/actions/export.js
mcilroyc: well, I intentionally was not re-throwing, but now that I'm staring at it again I can see how it would be safer to be strict about this.  Will change.

mcilroyc:
Ian's nits addressed, and squashed into the previous "review comments" commit.  This is, afaict, ready to merge. I updated the PR description with the latest information.

iwehrman:
One more wish: file a bug for:
> When in the OS folder chooser dialog, do NOT try to create a new folder. Choose an existing one.
Otherwise, looks good!

Issue #2171, opened 2015-08-22
Series of errors resulted by creating and moving horizontal guides


ktaki:
Create a few horizontal guide at the top or bottom. Try moving them to the center. Cannot recover from the errors.

Action guides.setGuide failed: TypeError: Cannot read property 'orientation' of undefined
Uncaught TypeError: Cannot read property 'orientation' of undefined
Action tools.beforeStartup failed: Error: Unknown policy ID: 4
Uncaught TypeError: Cannot read property 'left' of null
Action export.onReset failed: TypeError: Cannot read property 'close' of null

volfied:
When you create them, are you just clicking in the area, or are you dragging them to the center after creation? 

At what point does the error happen? Is it after creating bunch of them, and the first move, or after couple moves? 

I need more details, and it would help a lot more if you posted the whole log instead of just the collected error lines.

ktaki:
It seems that the errors happen when clicking in the blue highlighted area at the top or bottom, and then try to move a guide created in the area on to the artboard. I will send a error log via email.

ktaki:
@volfied Can you look at this next? This is reproducible so often and consider to be a blocker since DS needs to be reset after the errors. Thanks.

volfied:
@ktaki Issue is addressed in #2217 yesterday, as soon as it's merged, you can test it.

ktaki:
This is fixed. Thanks!

Issue #2131, opened 2015-08-20
Guides are not deleted when dragged out of work area


chadrolfs:
drag guide to outside the work area
zoom out or hold down space and pan

Result: guide is still there, just moved.

ktaki:
This seems to be fixed, but testing is blocked by #2144.

chadrolfs:
When you do manage to drag them, you can delete them.

PR #2217, opened 2015-08-25
Add guide deletion by dragging outside, error checking for guide functions, better hover highlighting

volfied:
Depends on https://github.com/adobe-photoshop/spaces-adapter/pull/139

Addresses: #2131, #2171, and general guide edge performance.

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -117,6 +117,10 @@ define(function (require, exports) {
                 return guide && (guide.layerID === 0 || topAncestorIDs.has(guide.layerID));
             });
 
+        if (!canvasBounds) {
+            return Promise.resolve();
+        }
+
         // Each guide is either horizontal or vertical with a specific position on canvas space
         // We need to create a rectangle around this guide that fits the window boundaries
         // that lets the mouse clicks go to Photoshop (ALWAYS_PROPAGATE)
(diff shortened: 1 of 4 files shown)


src/js/stores/ui.js
```diff
@@ -266,6 +266,10 @@ define(function (require, exports, module) {
                 windowWidth = window.document.body.clientWidth,
                 windowHeight = window.document.body.clientHeight;
 
+            if (!centerOffsets) {
```
iwehrman: I wonder if this will solve the occasional cloaking weirdness I've seen recently.

src/js/actions/guides.js
```diff
@@ -114,9 +114,13 @@ define(function (require, exports) {
             topAncestors = currentDocument.layers.selectedTopAncestors,
             topAncestorIDs = collection.pluck(topAncestors, "id"),
             visibleGuides = guides.filter(function (guide) {
-                return guide.layerID === 0 || topAncestorIDs.has(guide.layerID);
+                return guide && (guide.layerID === 0 || topAncestorIDs.has(guide.layerID));
```
iwehrman: I guess there was a good reason for this, but it is a little odd that the store is holding a reference to null guides.

src/js/actions/guides.js
```diff
@@ -196,47 +200,101 @@ define(function (require, exports) {
...
      * @param {number} payload.index Index of the edited guide
-     * @param {string} payload.orientation New orientation of the guide
-     * @param {number} payload.position New position of the guide
+     * @param {boolean=} sendChanges If true, will call the action descriptor to delete the guide from PS
      *
      * @return {Promise}
      */
-    var setGuide = function (payload) {
-        return this.dispatchAsync(events.document.history.nonOptimistic.GUIDE_SET, payload)
+    var deleteGuide = function (payload, sendChanges) {
```
iwehrman: Out of curiosity, why go with the "payload" pattern here instead of a more typical `document model, index, flag` parameter list?

iwehrman:
Just a few tiny questions. Looks fine generally.

src/js/stores/ui.js
volfied: It shouldn't solve the cloaking weirdness... but it should solve the errors `left of undefined` we've been seeing onReset and sometimes on startup.

src/js/actions/guides.js
volfied: Guides are stored by index, and we listen to `set` events from Photoshop to change our guide structure.

If the user is clicking fast enough we can get into a situation where we get the guide index n+1 before we get guide index n's set event. In these cases, because we store guides in an indexed list, index n will be undefined until that event is processed, but guide policies may need updating between those two moments, necessitating the null check.

src/js/actions/guides.js
volfied: You're right... I should switch it to the more universal pattern we're trying to push. Changes incoming.

volfied: It started with the fact that initially I was immediately dispatching the information to the stores, so I didn't feel the need to unwrap things and wrap them back again. But now that we're calling adapter functions, I should make the change.

volfied:
@iwehrman I've updated my code to be more universal. Can you check?

src/js/actions/guides.js
```diff
@@ -270,15 +335,17 @@ define(function (require, exports) {
...
-                var payload = {
-                    documentID: target[0]._id,
-                    layerID: event.layerID,
-                    index: target[1]._index - 1, // PS indices guides starting at 1
-                    orientation: event.orientation._value,
-                    position: event.position._value
-                };
-
-                this.flux.actions.guides.setGuide(payload);
+                var mockDocument = {
```
iwehrman: Why not just fetch the document model?

src/js/actions/guides.js
```diff
@@ -270,15 +335,17 @@ define(function (require, exports) {
...
+                        id: target[0]._id
+                    },
+                    mockGuide = {
```
iwehrman: Also weird? Seems like the canonical parameter list ought to be: `Document, Layer, orientation, position, index`.

volfied: There is no layer involved with guides, the owner layer ID can be 0 (which means it's not targetting any layers), I figured in this case the variable list should be Document, Guide (or an equivalent model), index (which is not part of the guide, but it's key)

volfied:
@iwehrman Addressed your concerns, didn't agree with the setGuide parameter list comment.

Issue #2098, opened 2015-08-19
Actions should use the options pattern


iwehrman:
Currently many of our actions have variadic signatures to accommodate some number of optional parameters. For example, [`type.setColor`](https://github.com/adobe-photoshop/spaces-design/blob/535b942799f406a20f13bf5a0d7edf1f75f95a4c/src/js/actions/type.js#L234-L246) takes two optional booleans to control whether or not opacity should be taken into account, and whether or not the history state should be coalesced. Using variadic signatures for this is bad for a few reasons:

1. It breaks down as soon as there is more than one or two optional arguments because it's hard to figure out which of the parameters were omitted.
2. The order of arguments is important to the function, but opaque to the caller. 
3. Adding additional arguments frequently requires them to be re-arranged, which requires all the call sites to be updated.
4. The optional parameters are typically hard to understand (consider, e.g., all the action calls that end with `, true, true);`). 

Instead, we should thoroughly convert all of our actions to make use of the "options" pattern, in which all optional arguments are collected into a final, optional object parameter. This would solve all of the above problems, and also make it easier to extend the action system with more advanced features. For example, I would like to add a notion of transaction to the action system, but doing so requires actions to carry around a transaction ID which propagates into the `play` calls. If we already had options arguments which were passed from action to action, and from action to `play` calls, then this would be very straightforward to implement.

This is probably a post-M3 change, though I've already worked through some of the [major](https://github.com/adobe-photoshop/spaces-adapter/tree/iwehrman/transactions) [issues](https://github.com/adobe-photoshop/spaces-design/tree/iwehrman/transactions), and it might be necessary to partially implement for #1947.

volfied:
Unassigning myself as it is code clean up and someone else may want to take it.

Issue #1947, opened 2015-08-07
Multiple history states are recorded for pasting a style


ktaki:
Undo reveals middle states that show inconsistency between actual canvas drawing and the state of the style panel. Either removing this inconsistency (i.e., update the style panel when Undo'ed) or record just one history state so that one undo takes care of going back to the original state.

New copy/paste style commands are great!

volfied:
Thanks for tracking this, @iwehrman and I have some crazy ideas to implement to fix it, which is lower priority currently than the Sampler HUD feature. I will get to fixing this after the HUD is complete.

volfied:
FBNC to @ktaki

ktaki:
Undo is successful, but subsequent apply of a style causes error.

Create a shape layer
Apply a style, say, a drop shadow from  the libraries panel
Undo
Apply the same style again from the libraries panel

![screen shot 2015-08-27 at 2 41 07 pm](https://cloud.githubusercontent.com/assets/2029597/9533878/bcb1a1e8-4cc9-11e5-866e-5ff6e85efade.png)

volfied:
libraries.applyLayerStyle does not create any history state in our world, so undo removes the layer from our model. Looking into it.

volfied:
@ktaki Now i'm extremely confused by this bug. Is this a libraries style apply bug, or paste layer style?

Can you please open new bugs for new issues instead of recycling old issues? If the paste layer style issue, which this bug title and description states is fixed, then it should be closed.

ktaki:
Oops. Sorry.

ktaki:
Verified the fix.

Issue #1896, opened 2015-08-02
Eliminate redundant InCurrentDocument actions


iwehrman:
We tried this pattern for a while early on, but in hindsight it seems better (i.e., much less boilerplate code) to just make the document parameter optional.

PR #2218, opened 2015-08-25
Transactions for sampler actions

volfied:
Also clean up for shapes actions to accept the {Document, Layers, Model, Options} variables so they're more unified. Cleans up couple other actions that sampler relied on for same reasons.

Also got rid of the ...selectedInCurrentDocument actions that were used for the menu items. With the unified variable set, it's easy to do so.

This will make undoing in one step possible for sampler actions,  #1947 
Also helps with clean up issues #2098 and #1896 

samplerundo


shaoshing:
Wow, I need to upgrade the libraries#applyColor to adopt the transaction feature!

volfied:
@shaoshing This PR takes care of that as well, sorry I forgot to mention it.

shaoshing:
Brilliant! :+1:  thank you

src/js/actions/libraries.js
```diff
@@ -606,14 +606,19 @@ define(function (require, exports) {
...
+                ignoreAlpha: false
+            };
 
         // FIXME: Setting font color and fill color at the same time will result in two histroy states.
         //        We should merge the two states when transaction is supported.
         var setTextColorPromise = textLayers.isEmpty() ? Promise.resolve() :
                 this.transfer(typeActions.setColor, currentDocument, textLayers, color, true, true),
             setShapeFillColorPromise = vectorLayers.isEmpty() ? Promise.resolve() :
-                this.transfer(shapeActions.setFillColor, currentDocument, vectorLayers, color, true, true, true);
+                this.transfer(shapeActions.setFillColor, currentDocument, vectorLayers, color, shapeFillOptions);
```
mcilroyc: at a glance, this looks like `true, true, true` is replaced with `false, true, false` (via shapeFillOptions), but I may be missing something.

src/js/actions/sampler.js
```diff
@@ -261,12 +289,17 @@ define(function (require, exports) {
...
+            },
+            shapePromise = shapeLayers.isEmpty() ? Promise.resolve() :
+                this.transfer(shapeActions.setStroke, doc, shapeLayers, stroke, actionOpts),
             textPromise = textLayers.isEmpty() ? Promise.resolve() :
-                this.transfer(typeActions.setColor, doc, textLayers, color, false, false);
+                this.transfer(typeActions.setColor, doc, textLayers, color, actionOpts),
+            transactionPromise = descriptor.endTransaction(transaction);
 
-        return Promise.join(shapePromise, textPromise);
+        return Promise.join(shapePromise, textPromise, transactionPromise);
```
mcilroyc: I guess in practice this may not matter, but I think you should wait for the shape and text promises to resolve before "ending" the transaction.

src/js/actions/shapes.js
```diff
@@ -42,17 +42,18 @@ define(function (require, exports) {
...
      * @private
+     * @param {object} options
      * @param {object} documentRef  a reference to the document 
      * @param {string} name localized name to put into the history state
-     * @param {boolean=} coalesce Whether to coalesce this operations history state
      * @return {object} options
      */
-    var _options = function (documentRef, name, coalesce) {
-        return {
+    var _mergeOptions = function (options, documentRef, name) {
```
mcilroyc: This function weird me out a little.  I feel like this is the translation between the "options" being passed around by our methods, and the "options" sent to photoshop.  If that's the case, then maybe it should really "merge" at all.

src/js/actions/shapes.js
```diff
@@ -160,27 +161,27 @@ define(function (require, exports) {
...
-        if (enabled === undefined || enabled === null) {
-            enabled = true;
+        var eventName;
+
+        if (options.enabled === undefined || options.enabled === null) {
+            options.enabled = true;
             eventName = events.document.history.optimistic.STROKE_CHANGED;
         } else {
             eventName = events.document.STROKE_ENABLED_CHANGED;
-            enabledChanging = true;
```
mcilroyc: Since a this if/else is applying the defaulting logic, doesn't it still need to set the enabledChanging bit?

src/js/actions/type.js
```diff
@@ -240,20 +241,21 @@ define(function (require, exports) {
...
      */
-    var setColor = function (document, layers, color, coalesce, ignoreAlpha) {
+    var setColor = function (document, layers, color, options) {
         var layerIDs = collection.pluck(layers, "id"),
             layerRefs = layerIDs.map(textLayerLib.referenceBy.id).toArray(),
             normalizedColor = color.normalizeAlpha(),
             opaqueColor = normalizedColor.opaque(),
             playObject = textLayerLib.setColor(layerRefs, opaqueColor),
-            typeOptions = _getTypeOptions(document.id, strings.ACTIONS.SET_TYPE_COLOR, coalesce);
+            typeOptions = _.merge(options,
```
mcilroyc: similar to my comment earlier, I don't immediately understand why the result of `_getTypeOptions` is being merged into `options`

mcilroyc:
1st pass complete.  good stuff.  see comments inline.

src/js/actions/libraries.js
volfied: I'm going to refer to @shaoshing here, but it felt like ignoreAlpha and coalesce should have been false to begin with (we do not want to coalesce this action)

src/js/actions/shapes.js
volfied: It's making sure the transactionID is in all batchPlay options.

src/js/actions/shapes.js
volfied: `enabledChanging` was never used in this one, and it was a copy/paste artifact :)

src/js/actions/type.js
volfied: We need to pass the transaction ID in the batchPlay options so the descriptor.batchPlay knows not to play the descriptor playObjects but instead append them to the transaction list.

volfied:
@mcilroyc Addressed your concerns and answered questions. Thanks for pointing out the endTransaction promise. Back to you

src/js/actions/type.js
mcilroyc: sorry to beat a dead horse, but...
the weird side effect of this is that the descriptor ALSO gets the "ignoreAlpha" property which is meaningless.  I think a more "strict" approach would be to pass the `options` in to _getTypeOptions as a parameter and "cherry pick" the props into their appropriate places in the final typeOptions that will be sent to the descriptor (coalesce, transaction).   

basically, I'm worried that there isn't a clear delineation between the options that gets passed in to our actions, and those that get passed in to descriptor.  Maybe I'm overreacting. 

Related: transaction should probably be specified in JSDoc if that's a meaningful option to `setColor`.

mcilroyc:
merging now.  I'm recommending a follow-up task to clarify the point at which method "options" are mapped to "play options", in particular `.transaction`, and to prevent unintended/extraneous options from being leaked into the play options.

PR #2221, opened 2015-08-25
Increase contrast between primary and secondary search info

placegraphichere:
addresses #1992 
some search tweaks 

- larger font sizes
- brighter text overall
- removed parens on keyboard shortcut
- removed italics on sub-option info
- hover state brighter for sub-option info
- small padding change
- some clashes with the new colors fixup


iwehrman:
For the merely curious:
image

iwehrman:
Lovin' it :hamburger: :fries:

PR #2222, opened 2015-08-26
Localized resources update.

fzhu1978:
Newline added at the end of the file.
New strings.json was introduced and mocked.


iwehrman:
Over to @volfied since he started the review on the original PR.

src/nls/de/strings.json
```diff
@@ -1,5 +1,491 @@
 {
+    "ACTIONS": {
+        "ADD_STROKE": "[8413887] ü_Add Stroke",
```
volfied: @fzhu1978 Are these intentionally left untranslated?

fzhu1978: Yes, once a new string is added, the process is:
1. ALF scans and harvests.
2. Use localization mock string instead English string, as shown in this line.
3. Send Lockit which contains all untranslated strings to linguists.
4. After translation comes back, ALF use real localized string substitutes the mock string.

fzhu1978:
Oh no, there is a new changes in this pull request that WALF merge from master into feature/localization.

@volfied Can I merge this pull request now?  To keep it clean, I have to create a new request which only contains localization changes.

volfied:
Thanks for the response @fzhu1978. Usually the reviewer merges the Pull Request once all the questions and concerns have been addressed.

In this case, they are, so I'm hitting the green button. Thank you for all your work.

fzhu1978:
Thanks Barkin!

Because ALF will create pull request which contains a lot of unrelated changes, so I am going to create pull request by hand once new translation comes back.

Regards,
Feng

PR #2227, opened 2015-08-26
Adding artboards to a document without freezing

DivyaPrabhakar:
Create Artboard and Create Artboard from Template was freezing the UI after the second or third artboard was created and no other artboard was selected. Narrowed down the problem to a bug in Immutable that sent it into infinite loop. This PR accounts for that bug and now allows the creation of an artboard to handle both situations, when an artboard is selected and when it is not. This fixes the bug and prevents the freezing when creating an artboard. 

Regarding issue #1905

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -1201,36 +1201,45 @@ define(function (require, exports, module) {
      */
     LayerStructure.prototype.createGroup = function (documentID, groupID, groupEndID, groupName,
         isArtboard, boundsDescriptor) {
+        // Creates the group head and end, needed for any type of group 
         var groupHead = Layer.fromGroupDescriptor(documentID, groupID, groupName, false,
                 isArtboard, boundsDescriptor),
             groupEnd = Layer.fromGroupDescriptor(documentID, groupEndID, "", true),
-            layersToMove = this.selectedNormalized.flatMap(this.descendants, this).toOrderedSet(),
-            layersToMoveIndices = layersToMove.map(this.indexOf, this),
-            layersToMoveIDs = collection.pluck(layersToMove, "id"),
-            groupHeadIndex = this.layers.size - layersToMoveIndices.last(),
-            newGroupIDs = Immutable.Seq([groupEndID, layersToMoveIDs, groupID]).flatten().reverse(),
-            removedIDs = collection
-                .difference(this.index, layersToMoveIDs) // Remove layers being moved
-                .reverse(), // Reverse because we want to slice from the end
+            groupHeadIndex,
+            newIDs;
+        
+        if (isArtboard) {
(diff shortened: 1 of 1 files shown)


volfied:
Drive-by comment: It's better for commits to describe what they've changed, and not the why/what they've fixed it.

iwehrman:
You should also describe what the problem was, and how your change addresses it in the PR summary. This makes it easier for reviewers to understand your changes.

DivyaPrabhakar:
Improving the PR now

src/js/models/layerstructure.js
```diff
@@ -1201,36 +1201,57 @@ define(function (require, exports, module) {
...
-                .reverse(), // Reverse because we want to slice from the end
-            newIDs = removedIDs
-                .slice(0, groupHeadIndex) // First chunk is all layers up to top most selected one
-                .concat(newGroupIDs) // Then our new group
-                .concat(removedIDs.slice(groupHeadIndex)), // Then the rest
+            groupHeadIndex,
+            updatedLayers,
+            newLayerStructure;
+        
+        if (isArtboard) {
```
iwehrman: There is a lot of duplicated code across these two branches. Is it possible to have, e.g., two if statements so that we don't have any duplicated code?

iwehrman: For example, this is duplicated:
```javascript
            updatedLayers = this.layers.withMutations(function (layers) {
                layers.set(groupID, groupHead);
                layers.set(groupEndID, groupEnd);
            });

            newLayerStructure = this.merge({
                layers: updatedLayers
            });
            
            newLayerStructure = newLayerStructure
                .updateSelection(Immutable.Set.of(groupID));
```

iwehrman:
Back to @DivyaPrabhakar for a little more tightening.

src/js/models/layerstructure.js
```diff
@@ -1201,36 +1201,48 @@ define(function (require, exports, module) {
...
+                    .reverse(); // Reverse because we want to slice from the end
             newIDs = removedIDs
                 .slice(0, groupHeadIndex) // First chunk is all layers up to top most selected one
                 .concat(newGroupIDs) // Then our new group
-                .concat(removedIDs.slice(groupHeadIndex)), // Then the rest
-            updatedLayers = this.layers.withMutations(function (layers) {
+                 .concat(removedIDs.slice(groupHeadIndex)); // Then the rest
+        }
+
+        updatedLayers = this.layers.withMutations(function (layers) {
```
iwehrman: Since updatedLayers and newLayerStructure aren't defined until here, it would be nicer to wait to declare them until here as well.

iwehrman: Final complaint :)

iwehrman:
Perfecto.

Issue #2228, opened 2015-08-26
Cannot change alpha and size values of fill and stroke


ktaki:
Create a shape. Change stroke size. 
Action shapes.setStrokeWidth failed: TypeError: Cannot read property 'coalesce' of undefined

![screen shot 2015-08-26 at 2 11 27 pm](https://cloud.githubusercontent.com/assets/2029597/9506403/d287ea04-4bfc-11e5-838f-533b1f3118cf.png)

ktaki:
Verified the fix.

PR #2229, opened 2015-08-26
Initialize shapes options to empty object if null

volfied:
If it wasn't provided, we were trying to access values from within it, and _.merge would fail.

Addresses #2228

### DIFF
--- a/src/js/actions/shapes.js
+++ b/src/js/actions/shapes.js
@@ -53,6 +53,10 @@ define(function (require, exports) {
      * @return {object} options
      */
     var _mergeOptions = function (options, documentRef, name) {
+        options = options || {
+            coalesce: false
+        };
+
         return _.merge(options, {
             paintOptions: {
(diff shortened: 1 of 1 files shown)


src/js/actions/shapes.js
```diff
@@ -520,18 +523,18 @@ define(function (require, exports) {
      */
     var setFillOpacity = function (document, layers, opacity, options) {
         // dispatch the change event
```
mcilroyc: comment seems a bit stale

src/js/actions/shapes.js
```diff
@@ -53,6 +53,10 @@ define(function (require, exports) {
      * @return {object} options
      */
     var _mergeOptions = function (options, documentRef, name) {
+        options = options || {
```
mcilroyc: this function still grosses me out : ) but I know there's a bug for it

Issue #2170, opened 2015-08-21
Guides: unable to move guided that are created under edge highlight


chadrolfs:
new doc
mouse over work area edge to reveal blue highlight and click
pan doc so that guide is not under edge and try to move

Result: won't move

volfied:
Can you try to repro this with policy frames shown in a non-compiled build? It feels like a timing issue where policies are having a hard time catching up.

volfied:
That's the only time I can repro this, but I want to make sure I got the steps right.

chadrolfs:
I can repro this with current resources (dev 522 and master @ ffa24f9c7c730c0a4dd74c14f378fbfcbf57c7c9)

chadrolfs:
still broken

chadrolfs:
https://www.dropbox.com/s/quxeztd3tl4m5cz/guide-underHiLite_noDrag.mov?dl=0

volfied:
The newer issue is due to us calling descriptors to create the guide, and then not moving it, so we don't receive a notification for it. It's fixed on core side: CL 1038728

Will be available Jenkins Dev Mac - #530

chadrolfs:
This is still failing on Windows.

volfied:
Fitz and I looked into it, and we found a bug in Windows side that's causing this bug. He's looking into it.

volfied:
@chadrolfs With latest Jenkins, @jfitz-adobe fixed this, please confirm.

PR #2232, opened 2015-08-26
Reset guide policies AFTER border policies are installed

volfied:
We were installing border policies before guide policies, which gave them precedence. So a guide within the selected layers' border would not be clickable (clicks would come to us). #2170 

Now we install guide policies AFTER border policies so they are higher priority.


iwehrman:
Makes sense!

PR #2236, opened 2015-08-26
lintspaces: set maximum newline to one.

shaoshing:
I am getting tired of @volfied's nits :godmode:


iwehrman:
Yeah, @volfied is real stickler.

Issue #2234, opened 2015-08-26
Sampler HUD fails to apply stroke to Type


chadrolfs:
Create a Type and Shape layer
select the Type layer
(with Sampler) mouse over Shape, hit Space and click on Stroke

Result: `Action sampler.applyColor failed: TypeError: undefined is not a function`

chadrolfs:
happens somewhat intermittently under specific circumstances.  Sent you video via slack

chadrolfs:
apparently has to do with where you click on the stroke icon in the hud.  
when you hit it, you see applyColor
when you don’t you see applyStroke

iwehrman:
FBNC to @chadrolfs.

Issue #2233, opened 2015-08-26
Color picker: errors attempting to set color of Type (modal or otherwise)


chadrolfs:
`Action type.setColor failed: TypeError: Cannot assign to read only property 'interactionMode' of true`

PR #2237, opened 2015-08-27
Transactionize libraries.setColor, and fix shape actions

volfied:
Addresses #2228, #2234, #2233 
Also sets libraries.applyColor to use transactions (seems like I actually forgot to do this @shaoshing!)

@shaoshing can you review?

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -40,6 +40,7 @@ define(function (require, exports) {
 
     var events = require("../events"),
         locks = require("../locks"),
+        strings = require("i18n!nls/strings"),
         pathUtil = require("js/util/path"),
         collection = require("js/util/collection"),
         layerActionsUtil = require("js/util/layeractions"),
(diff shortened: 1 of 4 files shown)


src/js/actions/libraries.js
```diff
@@ -607,20 +608,29 @@ define(function (require, exports) {
...
+            },
+            setTextColorPromise = textLayers.isEmpty() ? Promise.resolve() :
+                this.transfer(typeActions.setColor, currentDocument, textLayers, color, actionOpts),
             setShapeFillColorPromise = vectorLayers.isEmpty() ? Promise.resolve() :
-                this.transfer(shapeActions.setFillColor, currentDocument, vectorLayers, color, shapeFillOptions);
+                this.transfer(shapeActions.setFillColor, currentDocument, vectorLayers, color, actionOpts);
 
-        return Promise.join(setTextColorPromise, setShapeFillColorPromise);
+        return Promise.join(setTextColorPromise, setShapeFillColorPromise)
+            .then(function () {
```
shaoshing: Shall we use `finally` here (and other actions using transaction) to ensure terminating the transaction, in case any of the joined promises failed?

shaoshing:
Reviewed and tested; it works great . I have one question about ending the transaction. Back to you @volfied

src/js/actions/libraries.js
volfied: endTransaction is going to be the one that calls all the batchPlays, so in cases where one of the actions fail, the failure will not be due to the descriptor, but some other reason. We probably don't want to play anything in those cases, and this will provide a nice stopgap.

So I think `.then` is the right way to go.

Issue #2209, opened 2015-08-25
Coordinates for new artboards are not correct


chadrolfs:
cmd/ctrl + n
cmd/ctrl + - to zoom out
select Artboard 1
click the plus sign to the left

Result: Artboard 2 is created but coordinates and plus overlay is not correct

chadrolfs:
this is on current staging build with sprint 29.3 release html:

![screen shot 2015-08-25 at 11 22 02 am](https://cloud.githubusercontent.com/assets/8442631/9475399/512c93c4-4b1b-11e5-8fe8-de761bff094b.png)

volfied:
Something is going on with artboards, since Divya is working on other artboard bugs, I'm going to wait on this one a bit.

DivyaPrabhakar:
@volfied I am working on this now, will update you within an hour or so either way.

iwehrman:
Here is my theory, as articulated to @DivyaPrabhakar:
![image](https://cloud.githubusercontent.com/assets/1075154/9509559/03dc487a-4c16-11e5-8913-e8b3695717d6.png)

chadrolfs:
It would be great if we could fix this issue this sprint (30)

DivyaPrabhakar:
I agree with Ian regarding his theory and apparently this race between CEF and PS is a commonly occurring bug that we need a more consistent way of solving. This was discussed at stand-up -- passing it back to @volfied.

PR #2241, opened 2015-08-27
On autoCanvasResizeShift events, enqueue the dispatch asynchronously …

volfied:
…through an action

Addresses #2209 
Also a bug in `GuidesOverlay.jsx` that would sometimes cause a throw.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1724,6 +1724,30 @@ define(function (require, exports) {
     duplicate.transfers = ["documents.updateDocument", addLayers, select];
 
     /**
+     * Dispatches a layer translate for all layers in the document
+     *
+     * @param {object} event Action Descriptor of autoCanvasResizeShift event
+     * @return {Promise}
+     */
+    var handleCanvasShift = function (event) {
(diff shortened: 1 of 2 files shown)


src/js/actions/layers.js
```diff
@@ -1840,20 +1861,17 @@ define(function (require, exports) {
...
-                        y: event.to.vertical
-                    }
+                currentDocument = applicationStore.getCurrentDocument(),
+                position = {
+                    x: event.to.horizontal,
+                    y: event.to.vertical
                 };
 
-                this.dispatch(events.document.TRANSLATE_LAYERS, payload);
+            if (!currentDocument) {
```
iwehrman: It seems like this should be handled in the action too?

iwehrman: Maybe the whole event should just go into the action?

volfied: I was trying to fit into our variable model.. but yeah, for event handler actions we can probably be outside the norm. Will make the changes.

iwehrman:
Perfect.

Issue #1880, opened 2015-07-31
Type: Initial type settings shown as "mixed" when creating new text layer


ktaki:
Instead of saying "mixed" for all the type settings, show current font settings that are either default or inherited from the previous creation of a text layer.

chadrolfs:
what's interesting is if you click, type and then delete one char, then typeface and settings populate correctly.

pineapplespatula:
@this needs some more communication with core, I'll chat with @isonno about some of our type issues.

iwehrman:
The problem seems to be that the text descriptor simply has the wrong information in it right after creating the text layer.

iwehrman:
https://watsonexp.corp.adobe.com/#bug=4038142

iwehrman:
FBNC to @ktaki.

ktaki:
Now the font name is not "mixed" but it always stay with default font and attributes, "Myriad Pro", "Regular", black, rgb(0,0,0), Size 16, etc., regardless of previous setting used to create a text layer.

iwehrman:
I spent some time understanding this yesterday. It's a race condition that starts with the fact that even after we get a valid `updatedTextProperties` event with the correctly color, the `textKey` property of the layer descriptor has invalid style information. I have a workaround in mind though...

volfied:
Over to @ktaki to verify.

ktaki:
Verified the fix.

PR #2242, opened 2015-08-27
Fix initial type properties in modal state

iwehrman:
There are a variety of problems when creating a new type layer in a modal state. After we learn about the new layer from the `createTextLayer` event, we call `addLayers` to initialize the model, but that model does not yet have any style information due to a PS bug. However, we do get an `updateTextProperties` event with the initial style information, but we were previously losing that because it came before the `addLayers` call had finished. This works around those problems by saving the style properties from this event and applying it to the new layer model as soon as it has been created.

Note that, in a perfect world, all these handlers would be actions managed by the flux controller. Unfortunately I think a bigger refactoring job will be needed to convert all these handlers to actions because of the way they're intertwined (see, e.g., the `_layersReplaced` variable). So I'm going to leave that to a different day, and instead use this single-purpose fix to work around the race and address the open issue.

Unrelated change: `addLayers` is now marked as safe to execute in modal states. Previously it wasn't, and in some cases this caused the modal state to be killed by the controller right after clicking to create the type layer. That bug has caused a lot of consternation over the last few weeks. To help ensure that this does not happen again, I've upgraded the "Killing modal tool state" log message emitted by the controller from "debug" to "warning" status, since this is probably a condition we want to know about and explicitly address if it happens again.

Addresses #1880.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -341,6 +341,7 @@ define(function (require, exports) {
                 return nextLayerCount !== naiveLayerCount;
             });
     };
+    addLayers.modal = true;
     addLayers.reads = [locks.PS_DOC];
     addLayers.writes = [locks.JS_DOC];
     addLayers.post = [_verifyLayerIndex, _verifyLayerSelection];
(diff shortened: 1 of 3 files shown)


volfied:
Yuck, but it works.

Issue #1491, opened 2015-06-12
KBSC for Ungroup could be Shift+Cmd+G instead of Opt+Cmd+G to match standard Ps


DesignSpaceBot:
So that I have one less thing to remember.

-- @ktaki at [2015-05-29T22:08:57Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1491)

placegraphichere:
@DivyaPrabhakar you may want to reference this one as well with keyboard shortcuts

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #2245, opened 2015-08-27
KBSC change for Ungroup layers on Mac and Windows

DivyaPrabhakar:
Changed the KBSC for Ungroup to CMD + SHIFT + g (Mac) and CTL + SHIFT + g (Windows). Note, the previous KBSC on Windows was ALT + CTL + g. Not sure if that was on purpose but changed it to SHIFT + CTL + g for now, let me know if that is not correct. 

Regarding issue #1491

### DIFF
--- a/src/static/menu-mac.json
+++ b/src/static/menu-mac.json
@@ -466,7 +466,7 @@
                         "keyChar": true,
                         "modifiers": {
                             "command": true,
-                            "option": true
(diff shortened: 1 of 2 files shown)


iwehrman:
Do these match the corresponding shortcuts in PS?

iwehrman:
LGTM.

Issue #1877, opened 2015-07-31
Focus is lost when tabbing out of a datalist


chadrolfs:
select a text layer, enter X (transform field), tab (repeat)
or set focus in Letter and Shift + tab (repeat)

Result: focus drops after blend mode or Weight if you are reverse tabbing

(also happens in edit mode)

pineapplespatula:
this may be a good candidate for @DivyaPrabhakar,

iwehrman:
This has nothing to do with type. I suspect that one of @hamburge's PRs broke this. @chadrolfs: if you have time, can you `git bisect` this failure?

chadrolfs:
92856812c5c0b72c2294ac29e7f228164466b35e is the first bad commit
commit 92856812c5c0b72c2294ac29e7f228164466b35e
Author: Jeffrey Cohen <jecohen@adobe.com>
Date:   Tue May 12 11:04:27 2015 -0700

    Allow text tool to work in modal state

pineapplespatula:
it looks like its related to my addition or _releaseFocus which it seems too heavy handedly gives keyboard focus back to PS so that we can be back in the modal editing state. We may need to work with @jfitz-adobe  on this one to figure out what to do about keyboard focus, or is there a way we can only call this when we exit via enter and not via tab ?

pineapplespatula:
or @iwehrman is there a way we can only call this when we exit via enter and not via tab ?

volfied:
FBNC to @chadrolfs

chadrolfs:
Not fixed.  Gets stuck on Typeface.

iwehrman:
You have to hit tab twice currently; the second time opens the select menu.

chadrolfs:
That wasn't what I was seeing in previous build but with current sources, this looks fixed so closing this.

PR #2249, opened 2015-08-27
Do not return keyboard focus to PS on blur

iwehrman:
Revert focus-handling changes from 9285681 so that keyboard focus isn't returned to PS after tabbing away from a datalist. This broke tabbing through datalists, as described in #1877.

There are still issues with focus in modal type states, but @jfitz-adobe is working on a different solution for that. There are also still issues with keyboard navigation and `Datalist`, which this does not address. If you find other concrete issues, please file them separately.

Addresses #1877.


volfied:
LGTM

Issue #2246, opened 2015-08-27
Typing in Stroke color field fails


chadrolfs:
create a rect shape
click in stroke color field and type "red" and hit enter

Result:  `Action shapes.setStrokeColor failed: Error: value conversion failed`

iwehrman:
FBNC to @chadrolfs.

PR #2251, opened 2015-08-27
Set stroke options correctly and apply stroke

volfied:
One more for the count... Stroke.jsx was not assigning some options in certain cases, causing an undefined value to go to PS. #2246

Also, the inside stroke in Sampler HUD was calling applyColor instead of applyStroke, so fixed that too.
#2234

Also copies stroke into the clipboard, instead of just strokeColor (which was no longer used)


iwehrman:
Looks great!

Issue #1936, opened 2015-08-06
Type tool deselect should not call resetLayers


volfied:
If there is extra core work that needs to be done to prevent calling reset on type tool deselect, than please track it with a Watson bug.

pineapplespatula:
has watson bug 4032729

iwehrman:
@pineapplespatula sez:

    when you exit the modal state, our model has be modified to reflect the current modal selection. When
    you escape without reseting we dont refelct the whole layer state. so we wont update to a mixed 
    selection in the property panel

iwehrman:
I think this can be deferred until after M3.

iwehrman:
After re-reading my comment on the above Watson bug, I now think I was confused previously about the nature of this bug. I thought it was about the fact that we call `resetLayers` after editing a type layer. (Which, if that's even a bug, it's a rather benign one.) But I now realize it's actually about the `resetLayers` call on deselect of the type tool. I have no idea why we're doing that, especially since we get a `set` or `make` event whenever a text layer is committed or created, and that calls `resetLayers`. The condition under which we call it also doesn't make much sense. So, I'm just going to rip it out unless somebody objects so that we can close this bug.

iwehrman:
Wait, maybe this was supposed to handle the case in which you _cancel_ a modal edit? But switching tools should commit the current modal state, not cancel it AFAIK. This might make sense in the case of the superselect type sub-tool, for which deselection and exiting the modal state should be co-occurring, but it ain't there. Also, canceling out of the type subtool seems to incorrectly leave you in the type tool anyway...

PR #2258, opened 2015-08-28
Type tool API update

iwehrman:
1. Parse style properties from `createdTextLayer` event as well as the `updateTextProperties` event, instead of having to back-patch the properties from the latter event after calling `addLayers`. This reverts the garbage from #2242.
2. Removes the `resetLayers` call from the type tool deselect handler because I have no idea why it does that. Addresses #1936. 

Requires Mac Jenkins build pgdev.541 from Thursday evening.

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -522,7 +522,7 @@ define(function (require, exports) {
             properties: properties
         };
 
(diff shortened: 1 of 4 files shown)


mcilroyc:
I added a small fix.  merging.

Issue #549, opened 2015-06-12
Mysterious document marquee after creating a text layer


DesignSpaceBot:
![marquee](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @iwehrman at [2015-02-04T19:56:08Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/549)

PR #2262, opened 2015-08-28
Use the new guide kind descriptor enum

volfied:
Requires pg-dev @1039042  #549. Jesper is reviewing the CL and will integrate to staging when it's done.

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -51,6 +51,15 @@ define(function (require, exports) {
         "orientation",
         "position",
         "itemIndex",
+        "kind"
+    ];
+    
+    /** 
+     * Properties that are not in all guides
+     *
+     * @type {Array.<string>}
+     */
+    var _optionalGuideProperties = [
         "layerID"
     ];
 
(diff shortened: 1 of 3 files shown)


src/js/actions/guides.js
```diff
@@ -66,13 +75,20 @@ define(function (require, exports) {
...
                 index: 1,
                 count: -1
-            },
-            getOpts = {
-                failOnMissingProperty: true
             };
 
-        // FIXME: Should we reverse these?
-        return descriptor.getPropertiesRange(docRef, rangeOpts, _guideProperties, getOpts);
+        var requiredPromise = descriptor.getPropertiesRange(docRef, rangeOpts,
```
iwehrman: I wonder if it would make sense to factor this pattern out into a helper function? I assume this is basically copied and pasted from the layer range properties code?

src/js/models/guide.js
```diff
@@ -39,6 +39,13 @@ define(function (require, exports, module) {
         documentID: null,
 
         /**
+         * True if the guide belongs to the document, false if it's an artboard guide
+         *
+         * @type {boolean}
+         */
+        isGlobal: null,
```
iwehrman: Is there a better name for this? Global is probably not the right word here. How about `isLayerGuide` or `isDocumentGuide`?

iwehrman:
Nits :tm:

src/js/actions/guides.js
volfied: There were enough differences that I didn't want to attempt to separate them at this point (extensions, start indices being different, etc.)

iwehrman: 10-4.

volfied:
@iwehrman Addressed one nit, not sure I want to tackle the other one here.

iwehrman:
LGTM.

Issue #1860, opened 2015-07-29
CC Lib: adding a vector smart object imports into library incorrectly


chadrolfs:
open a file that has an embedded vector smart object layer (basically an illustrator file) and select it
click the Add Graphic icon at the bottom of the Lib panel

Result: layer is added to library but it has no icon.  Switching to standard PS reveals that it was created as PS instead of AI.

Attempting to add it to a doc in DS results in:
`Action libraries.createLayerFromElement failed: Error: unknown exception type
    at eval (eval at undefined, <anonymous>:1:5)
    at eval (native)`

chadrolfs:
![screen shot 2015-07-29 at 11 49 40 am](https://cloud.githubusercontent.com/assets/8442631/8966691/8d0d7e74-35e7-11e5-907f-e90d8c7e4b83.png)

shaoshing:
@volfied could you check this out?

volfied:
We no longer see the error, but I will fix the Ai/PS source issue.

chadrolfs:
@volfied @shaoshing Here's the error I get now:
``` Action libraries.createElementFromSelectedLayer failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -1715 Photoshop message: error: -1715 message: Could not create a library graphic because of a program error.
    at eval (eval at undefined, <anonymous>:1:5)
    at eval (native)
From previous event:
    at Descriptor.batchPlay (file:///Users/crolfs/Documents/git-repos/spaces-design/bower_components/spaces-adapter/src/ps/descriptor.js:373:14)
    at Descriptor.batchPlayObjects (file:///Users/crolfs/Documents/git-repos/spaces-design/bower_components/spaces-adapter/src/ps/descriptor.js:418:21)
    at Descriptor.playObject (file:///Users/crolfs/Documents/git-repos/spaces-design/bower_components/spaces-adapter/src/ps/descriptor.js:352:21)
    at Object.<anonymous> (file:///Users/crolfs/Documents/git-repos/spaces-design/src/js/actions/libraries.js:156:35)```

chadrolfs:
hmmm, CNR now - PS may have been in a bad state after a previous crash.

chadrolfs:
Looks like it may be the first time you attempt to add it that you get that error.  Just repro'd again.

shaoshing:
[...]

Issue #2230, opened 2015-08-26
Positions of icons are off from clickable regions in Libraries panel header


ktaki:
Icons for show menu (down arrow), Share, Send Link, and View on web site are off to the right compared to actual clickable regions. Trying to open libraries dropdown often causes sharing web site to open in browser.
![share](https://cloud.githubusercontent.com/assets/2029597/9507893/2c8713f0-4c06-11e5-9e02-b63a9af01a54.png)

ktaki:
Verified the fix.

Issue #2201, opened 2015-08-25
Stale empty layer remains after placing cloud-linked smart object in new artboard


iwehrman:
1. Create a new artboard document
2. Drag a graphic out of your CC library into the empty artboard.

Result:
The layers panel shows two layers: one for the new CLSO, and one for the original empty child layer. This is wrong because the original empty child layer is deleted in the PS document when adding the CLSO layer, so the layers panel is inconsistent with the PS document model.

Expected:
The empty child layer should be replaced by the CLSO layer.

iwehrman:
Calling all `layers.addLayers` fans, CC @mcilroyc @volfied.

volfied:
@iwehrman can you confirm this fixed?

PR #2268, opened 2015-08-28
CC Libraries - improvements and bug fixes

shaoshing:
## Improvements

* Make sure to update asset's preview image when it is edited outside of DS (e.g. Ai, or win Ps)
* Remove the last "updateDocument" in libraries actions.
* Prevent long asset title from scrolling horizontally when scrolling the asset list.

## Resolved issues

#1860 CC Lib: adding a vector smart object imports into library incorrectly (hopefully)
#2230 Positions of icons are off from clickable regions in Libraries panel header
#2201 Stale empty layer remains after placing cloud-linked smart object in new artboard

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -404,7 +404,7 @@ define(function (require, exports) {
     };
     disposeDocument.reads = [];
     disposeDocument.writes = [locks.JS_DOC, locks.JS_APP];
-    disposeDocument.transfers = [layerActions.resetLinkedLayers, application.updateRecentFiles, ui.updateTransform];
+    disposeDocument.transfers = ["layers.resetLinkedLayers", application.updateRecentFiles, ui.updateTransform];
     disposeDocument.lockUI = true;
 
     /**
@@ -605,7 +605,7 @@ define(function (require, exports) {
     };
     selectDocument.reads = [locks.JS_TOOL];
     selectDocument.writes = [locks.JS_APP];
-    selectDocument.transfers = [layerActions.resetLinkedLayers, historyActions.queryCurrentHistory,
+    selectDocument.transfers = ["layers.resetLinkedLayers", historyActions.queryCurrentHistory,
         ui.updateTransform, toolActions.select, ui.cloak, guideActions.queryCurrentGuides];
     selectDocument.lockUI = true;
     selectDocument.post = [_verifyActiveDocument];
--- a/src/js/jsx/sections/libraries/Library.jsx
+++ b/src/js/jsx/sections/libraries/Library.jsx
@@ -69,6 +69,18 @@ define(function (require, exports, module) {
 
         componentWillMount: function () {
             this._setTooltipThrottled = synchronization.throttle(os.setTooltip, os, 500);
+            this._libraryLastModified = this.props.library.modified;
+        },
+        
+        shouldComponentUpdate: function (nextProps) {
+            // Library's modified time reflects itself and its elements, so there is no need to check its element's 
+            // modified time.
+            return this._libraryLastModified !== nextProps.library.modified ||
+                !_.isEqual(this.state, nextState);
+        },
+        
+        componentWillUpdate: function () {
+            this._libraryLastModified = this.props.library.modified;
         },
 
         /**
--- a/src/js/jsx/sections/libraries/Library.jsx
+++ b/src/js/jsx/sections/libraries/Library.jsx
@@ -25,7 +25,8 @@ define(function (require, exports, module) {
     "use strict";
 
     var React = require("react"),
-        classnames = require("classnames");
+        classnames = require("classnames"),
+        _ = require("lodash");
 
     var os = require("adapter/os"),
         synchronization = require("js/util/synchronization"),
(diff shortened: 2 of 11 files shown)


src/js/actions/documents.js
```diff
@@ -605,7 +605,7 @@ define(function (require, exports) {
     };
     selectDocument.reads = [locks.JS_TOOL];
     selectDocument.writes = [locks.JS_APP];
-    selectDocument.transfers = [layerActions.resetLinkedLayers, historyActions.queryCurrentHistory,
+    selectDocument.transfers = ["layers.resetLinkedLayers", historyActions.queryCurrentHistory,
```
shaoshing: Change to string type action name to resolve circular dependencies between the layers and documents action.

src/js/models/layerstructure.js
```diff
@@ -822,10 +822,13 @@ define(function (require, exports, module) {
...
-                // otherwise, replace the selected layer
-                var selectedLayers = document.layers.selected;
-                replaceLayer = selectedLayers && selectedLayers.size === 1 && selectedLayers.first();
+            } else if (this.selected && this.selected.size === 1) {
+                // otherwise, replace the selected layer if it has the same index with the new layer.
+                var selectedLayer = this.selected.first(),
+                    selectedLayerIndex = this.indexOf(selectedLayer),
+                    newLayerIndex = descriptors[0].itemIndex;
+                    
+                replaceLayer = selectedLayerIndex === newLayerIndex && selectedLayer;
```
shaoshing: turns out that PS does not replace the single-empty-selected layer after placing an embedded layer, so we need to check the new layer's index to determine whether we should do a replacement.

src/js/jsx/sections/libraries/Library.jsx
```diff
@@ -70,6 +71,18 @@ define(function (require, exports, module) {
...
+        
+        shouldComponentUpdate: function (nextProps, nextState) {
+            // Library's modified time reflects itself and its elements, so there is no need to check its element's 
+            // modified time.
+            if (this._libraryLastModified === nextProps.library.modified &&
+                _.isEqual(this.state, nextState)) {
+                return false;
+            }
+            
+            this._libraryLastModified = nextProps.library.modified;
```
volfied: This update should be done in componentDidUpdate. It's not React-like to write into things in SCU calls.

src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
```diff
@@ -43,30 +43,60 @@ define(function (require, exports, module) {
         
         getInitialState: function () {
             return {
-                loading: true,
-                renditionPath: null
+                loading: false,
+                renditionPath: null,
+                lastModified: null
```
volfied: shouldn't this be `this.props.element.modified` ?

volfied: I see the null helps with `_updateRenditionPath` starting on init... not really happy with the whole logic here... but oh well!

src/js/models/layerstructure.js
volfied: This is a really hairy function... one of our weakest points. Did you test this with other replace layer interactions (like replacing a background layer, NOT replacing a background layer etc.)

volfied:
Got couple questions @shaoshing

src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
shaoshing: I moved `lastModified` out of the state as it is not related to component updates, also made the code more clear by moving the `lastModified` assignment to the `updateRenditionPath` function.

src/js/models/layerstructure.js
shaoshing: The change here depends on the descriptor attribute `itemIndex` from PS, so it should be reliable to do a replacement when the new layer has the same index with the background layer ? Here are some scenarios I tested and passed:

* select bg layer, draw a rectangular, bg layer replaced
* selecte multiple layers (including bg layer), draw a rectangular, bg layer NOT replaced
* select bg layer, drag asset (except for Ai asset, which will create a new layer) from libraries, bg layer replaced
* deselect bg layer, draw a rectangular, bg layer NOT replaced

Please let me know if I miss anything here.

src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
```diff
@@ -49,24 +49,54 @@ define(function (require, exports, module) {
...
-            var element = this.props.element;
+            this._updateRenditionPath();
+        },
+        
+        componentWillUpdate: function () {
+            this._updateRenditionPath();
+        },
+        
+        shouldComponentUpdate: function (nextProps, nextState) {
+            return this._lastModified !== nextProps.element.modified ||
```
volfied: `_lastModified` is not declared anywhere, can you declare it with a JSDoc at the top of the component?

volfied:
One more, @shaoshing

shaoshing:
Thank you for reviewing @volfied . I moved the `_lastModified` variable to a more significant location and added doc. Please check it out.

volfied:
:+1: nice work

Issue #2255, opened 2015-08-28
Sampling multi-run text color has unexpected result


chadrolfs:
Not sure what is expected.  Maybe this is as designed?  Colors swap.

https://www.dropbox.com/s/6di6bramif3whh8/sample-multirun-textcolor.mov?dl=0

volfied:
Doesn't work with libraries/applyStyle as well.. seems like something changed with descriptors.

volfied:
Also will fix the leading issue where text bunches up at the top...

iwehrman:
FBNC to @chadrolfs.

PR #2269, opened 2015-08-28
Get lineHeight from a text style only if it's positive

volfied:
On autoLeading, leading would be -1, and `if (leading)` would return true causing the multi line text cramming issue.

### DIFF
--- a/src/js/stores/font.js
+++ b/src/js/stores/font.js
@@ -210,7 +210,7 @@ define(function (require, exports, module) {
                 };
             }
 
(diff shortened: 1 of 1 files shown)


volfied:
Tracked in #2255

iwehrman:
I'm open to suggestions about a better way to represent this. Previously there was a crazy encoding in which `null` meant one thing and `undefined` meant another. Switching `leading` to always be a number seemed like an improvement, but I'm sure it could be better.

PR #2270, opened 2015-08-28
Type modal history fixes

mcilroyc:
undo/redo
- While in modal state* the UNDO action will now play the actual 101 command in photoshop.  This is OK because photoshop deals with history differently in the text world.  More improvements possible, but this is a decent first step.
- Still has minor bug with showing menu item in the correct situations

updating text properties while modal will **NOT**:
- Send `historyStateInfo` as a play option to photoshop
- Amend history (admittedly I was confused on this matter coming in to this bug work)
- Reset bounds (which, awkwardly, is the thing that *correctly* creates history state when non-modal)

Committing an existing text layer will now AMEND history (via the "set" event).  As an example, this means that the correct "mixed" state will be captured after a multi-color-text run is committed.  So that an undo/redo pair will result in "mixed".

this should address #1848

*as ian pointed out, the definition of "modal state" may need to be tightened up... although some immediate tests have been positive.

I started a wiki on the subject: https://github.com/adobe-photoshop/spaces-design/wiki/History-&-modal-text-editing

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -57,24 +57,30 @@ define(function (require, exports) {
      * @private
      * @param {number} documentID
      * @param {string} name localized name to put into the history state
+     * @param {boolean} modal is the app in a modal state
      * @param {boolean=} coalesce Whether to coalesce this operations history state
      * @return {object} options
      */
-    var _getTypeOptions = function (documentID, name, coalesce) {
-        return {
+    var _getTypeOptions = function (documentID, name, modal, coalesce) {
+        var options = {
             paintOptions: {
                 immediateUpdate: true,
                 quality: "draft"
             },
-            historyStateInfo: {
+            canExecuteWhileModal: true,
+            ignoreTargetWhenModal: true
+        };
+
+        if (!modal) {
+            options.historyStateInfo = {
                 name: name,
                 target: documentLib.referenceBy.id(documentID),
                 coalesce: !!coalesce,
                 suppressHistoryStateNotification: !!coalesce
-            },
-            canExecuteWhileModal: true,
-            ignoreTargetWhenModal: true
-        };
+            };
+        }
+
+        return options;
     };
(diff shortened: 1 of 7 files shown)


src/js/actions/type.js
```diff
@@ -57,24 +57,55 @@ define(function (require, exports) {
...
+     * @param {Promise} psPromise The promise of the photoshop update operation
+     * @param {Promise} jsPromise The promise of the flux dispatch
+     * @param {Document} document
+     * @param {Immutable.List<Layer>} layers
+     * @param {boolean} modal in modal text edit state?
+     * @return {Promise} Resolves when both input promises are resolved,
+     *                            and after the resetBounds promise, if necessary
+     */
+    var _resetBoundsAfter = function (psPromise, jsPromise, document, layers, modal) {
+        return Promise.join(psPromise, jsPromise,
```
iwehrman: I don't like passing the promises into this function. I realize that there is otherwise the duplication of clients needing to `join` the promises on their own. I can't clearly articulate why I don't like this. It just "feels" like the wrong separation for a very small payoff. :walking:

iwehrman:
My only complaint is with the helper function. This PR will conflict with the changes in #2276, but the resolution should be straightforward.

src/js/actions/type.js
mcilroyc: I was laying in bed saturday morning letting my mind drift as I am wont to do, slowly processing the circumstances of my life and past week, and I suddenly had the realization that it was mildly gross to pass promises into that function.  true story.

mcilroyc:
I removed the awkward helper function and pushed the changes on top of the third commit in this PR.  I would like to refactor all these type property actions since they all follow the same basic pattern (except color) but that'll be for another day.

iwehrman:
> The unexamined life is not worth living.

-- @mcilroyc

Issue #1882, opened 2015-07-31
Text Edit: Alignment changes does not work in text edit mode


ktaki:
Click Align icons in text editing mode. No changes shown on canvas.

pineapplespatula:
The modification that allows us to handle modal commands doesn't seem to support paragraph styles. 
@chadrolfs  would you be able to log a watson bug about it?

chadrolfs:
@ktaki originally logged this so I'll let him do the honors.

ktaki:
Here it is. https://watsonexp.corp.adobe.com//#bug=4032644

iwehrman:
@pineapplespatula There's a typo in the `alignAssigned` event property name:
![image](https://cloud.githubusercontent.com/assets/1075154/9296258/423dae46-443d-11e5-8515-d009dd9dedab.png)

Also 1) what is the point of these `Assigned` property names? I can't figure out how to get in a state in which they are meaningful. 

Also 2) why can't the event just _omit_ the relevant property names if they aren't specified in the event?

In any case, some JS-side change relevant to this are FIP by me, but we should still talk about potential changes to the event interface.

pineapplespatula:
they were a cross over from the way that PS represents its character style model, but you are right that it makes more sense in a javascript world to just omit them from the descriptor. I'll look into making the core changes needed for that.

iwehrman:
Can you give instructions for how to reproduce an event with, say, `sizeAssigned === false`?

pineapplespatula:
not anymore, because i took out assigned keys in the descriptors. 
but you would have done it by making a text layer with two characters, changing on the of the characters sizes and then selecting both characters

iwehrman:
So now it just omits `size`?

pineapplespatula:
yes

iwehrman:
As of CL 1038558, the alignment commands seem to work, although the changes don't take effect on the canvas until after the modal state is committed. @jsbache to address!

mcilroyc:
The latest PR 2276 should address this issue.  FBNC to @ktaki

ktaki:
Verified the fix.

PR #2276, opened 2015-08-29
Fix modal text alignment

iwehrman:
The problem was that we were trying to merge type options into another `undefined` set of options, which results in undefined. That is, `_.merge(undefined, x) === undefined`. This broke the alignment action because the `canExecuteWhileModal` option is required in the modal state. This fixes the problem by having `_getTypeOptions` perform the merge into a fresh object.

Addresses #1882.

**Update:** Also pushed dccdd8b to address #2277.

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -59,10 +59,11 @@ define(function (require, exports) {
      * @param {string} name localized name to put into the history state
      * @param {boolean} modal is the app in a modal state
      * @param {boolean=} coalesce Whether to coalesce this operations history state
+     * @param {object=} options Inherited into the type options returned, if present
      * @return {object} options
      */
-    var _getTypeOptions = function (documentID, name, modal, coalesce) {
-        var options = {
+    var _getTypeOptions = function (documentID, name, modal, coalesce, options) {
+        var typeOptions = {
             paintOptions: {
                 immediateUpdate: true,
                 quality: "draft"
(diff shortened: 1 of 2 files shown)


src/js/actions/type.js
```diff
@@ -487,8 +490,7 @@ define(function (require, exports) {
             layerRefs = layerIDs.map(textLayerLib.referenceBy.id).toArray();
 
         var setAlignmentPlayObject = textLayerLib.setAlignment(layerRefs, alignment),
-            typeOptions = _.merge(options,
-                _getTypeOptions(document.id, strings.ACTIONS.SET_TYPE_ALIGNMENT)),
+            typeOptions = _getTypeOptions(document.id, strings.ACTIONS.SET_TYPE_ALIGNMENT, options),
```
mcilroyc: Is this missing the coalesce arg?

mcilroyc:
One concern, otherwise I'll merge this and rebase my possibly-conflicting PR.

src/js/actions/type.js
iwehrman: Mmmm, yeah. Will fix.

iwehrman:
`coalesce` argument folded into first commit; branch rebased against master to incorporate new `modal` parameter to `_getTypeOptions`. Please double-check my rebase skillz. Back to @mcilroyc.

Issue #2257, opened 2015-08-28
Canceling type edit does not update style panel


iwehrman:
1. Create a text layer with multi-colored text.
2. Commit the text.
3. Click or double-click (depending on the tool) to begin editing the text. 
4. Place your selection in exactly one text-color run. Observe that color in the style panel.
5. Cancel the modal state with `cmd-.`.

Result: the style panel lists the same color.
Expected: the style panel lists "mixed".

iwehrman:
Observed in #1936.

iwehrman:
This bug is now blocked on core Watson issue 4045762.

isonno:
Core watson bug 4045762 fixed in playground-dev branch

iwehrman:
Thanks @isonno! FBNC to @chadrolfs.

PR #2278, opened 2015-08-30
Update type properties after canceling edits

iwehrman:
This uses the new `toolModalStateChanged` event's `reason` property to handle canceled type modal states. The goal is to reset the layers, but that should _only_ happen if there was no `deleteTextLayer` event, which handles the exit from the modal state specially by removing the layer models.

Partially addresses #2257, but note that there is [a core bug](https://watsonexp.corp.adobe.com/#bug=4045762) which notes that there is no `deleteTextLayer` event when canceling out of a modal tool state for a layer that has never been committed. This causes a Photoshop error in that case because `resetLayers` is unable to reset models for layers that have already been deleted in the Photoshop document. But, when that core bug is addressed, no further JavaScript changes should be required.

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -50,6 +50,7 @@ define(function (require, exports, module) {
         tools: require("./tools"),
         transform: require("./transform"),
         type: require("./type"),
+        typetool: require("./typetool"),
         ui: require("./ui")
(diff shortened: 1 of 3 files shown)


volfied:
Ok!

Issue #2256, opened 2015-08-28
Canceling edits in type sub-tool does not return to select tool


iwehrman:
1. Make a text layer.
2. Switch to the select tool.
3. Double-click the text layer. Feel free to make some changes.
4. Cancel the edits with `cmd-.`.

Result: the type subtool is still active.
Expected: the select tool is active.

iwehrman:
Observed in #1936.

PR #2279, opened 2015-08-30
Activate select tool after canceling edits from type subtool

iwehrman:
What it says! Previously this only happened when the edits were committed, but not when canceled with `cmd-.`. 

Addresses #2256.


volfied:
:+1:

Issue #2253, opened 2015-08-27
Guides: cursor and drag area are not in sync: Clicking to the left of a guide calls marquee


chadrolfs:
drag out a guide, move the mouse from left to right towards the guide until the cursor changes, then click and drag to the left

Result: marquee select.

mcilroyc:
to @chadrolfs

PR #2280, opened 2015-08-31
Floor and ceil the guide policy areas so they sit pixel perfect

volfied:
Addresses #2253, in certain guide locations, the edge of the policy would be at the wrong place and mouse events would stay with us.


mcilroyc:
I've provided my famous weird javascript ceiling caution: `Math.ceil(100 * .55)`, and so I'm merging.

Issue #1767, opened 2015-07-20
Layers highlight over the resize/rotate areas, they shouldn't.


volfied:
Fitz is looking into this, this is here to keep track.

We still get highlights on our resize/rotate areas, and clicking them doesn't really select the highlighted layer.

We should not be getting external adapter mouse move events if the area we're on is on ALWAYS_PROPAGATE.

volfied:
After discussing this with Fitz, I realized that mouse down policy areas should not affect mouse move events. So we're going to check for policy areas on our end to make sure our mouse moves are not over an ALWAYS_PROPAGATE area. This simplifies the API.

iwehrman:
LGTM!

iwehrman:
Over to @chadrolfs to verify. Thanks!

PR #2281, opened 2015-08-31
Don't highlight superselect layers under always_propagate policies

volfied:
With all these policies flying around, it's hard to know when to highlight a layer or not... so I wrote this PR.

policyhighlight

Addresses #1767

### DIFF
--- a/src/js/jsx/tools/SuperselectOverlay.jsx
+++ b/src/js/jsx/tools/SuperselectOverlay.jsx
@@ -31,10 +31,12 @@ define(function (require, exports, module) {
         d3 = require("d3"),
         _ = require("lodash");
 
-    var OS = require("adapter/os");
+    var UI = require("adapter/ps/ui"),
+        OS = require("adapter/os");
 
     var system = require("js/util/system"),
         mathUtil = require("js/util/math"),
+        keyUtil = require("js/util/key"),
         uiUtil = require("js/util/ui");
 
     // Used for debouncing the overlay drawing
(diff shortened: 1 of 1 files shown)


mcilroyc:
Looks reasonable to me.  @iwehrman this seems like something you'll want to at least be aware of.  I'll merge soon if I hear nothing.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -400,6 +402,53 @@ define(function (require, exports, module) {
...
+         */
+        _positionUnderAlwaysPropagatePolicy: function (x, y) {
+            var policyStore = this.getFlux().store("policy"),
+                modifierStore = this.getFlux().store("modifier"),
+                modifiers = modifierStore.getState(),
+                modifierBits = keyUtil.modifiersToBits(modifiers),
+                pointerPolicies = policyStore.getMasterPointerPolicyList(),
+                underAlways = false;
+
+            _.forEachRight(pointerPolicies, function (policy) {
```
iwehrman: Would this be better as a `some` loop instead? It seems like you're really just searching for the first time `underAlways === true`.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -400,6 +402,53 @@ define(function (require, exports, module) {
...
+                        area.top < y && area.bottom > y;
+
+                if (intersects) {
+                    if (policy.action === UI.policyAction.ALWAYS_PROPAGATE) {
+                        underAlways = true;
+                    } else {
+                        underAlways = false;
+                    }
+                }
+            }.bind(this));
```
iwehrman: We always prefer the receiver argument of explicit binds when available, as it is here.

iwehrman:
2 nits. Otherwise, I like it.

src/js/jsx/tools/SuperselectOverlay.jsx
volfied: Nope. For example, with our border policies, we have concentric rectangles, and the center one is a never propagate (because we can click layers "underneath" multiple selected recrtangles etc)

So when we click the center, we'll hit the bigger ALWAYS_PROPAGATE rectangle first, and then hit the NEVER_PROPAGATE one and know that at the top lies a NEVER_PROPAGATE rectangle and we should highlight.

volfied:
@iwehrman addressed your nits.

src/js/jsx/tools/SuperselectOverlay.jsx
volfied: Thought about this a bit more, and simplified it, thank you!

iwehrman:
LGTM!

PR #2284, opened 2015-08-31
Allow document-level exports

mcilroyc:
Two functional changes:
- Now supports document-level export, for NON AB documents, using the Export Panel with *no layers selected*
- An :arrow_down: icon in the export panel which performs an export of the selected layer(s) or document.
 
Reviewer notes:
- Added `EXTENSION_DATA_NAMESPACE` to global.js (we discussed this on a recent engineering powow, which mixed agreement)
- I started the process of handling multiple layer selection in export panel, but stopped short in order to keep this PR of reasonable size.  A PR is forthcoming that will take it further.
- Moved the step of setting assets to "requested" status as part of the action, instead of it being orchestrated within react
- Renamed the react component to `ExportList` since it now handles document level exports, and will soon handle multi-layer

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -71,7 +71,7 @@ define(function (require, exports) {
      *
      * @private
      * @param {number} documentID
-     * @param {number=} layerID optional, if supplied, sync doc-level metadata
+     * @param {number=} layerID Optional. If not supplied, sync doc-level metadata
      * @return {Promise}
      */
     var _syncExportMetadata = function (documentID, layerID) {
@@ -94,7 +94,7 @@ define(function (require, exports) {
 
             exportsMetadata = {
                 exportAssets: layerExportsArray,
-                exportEnabled: exportEnabled === undefined ? false : exportEnabled
+                exportEnabled: !!exportEnabled
             };
 
             // set layer extension data
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -29,6 +29,7 @@ define(function (require, exports) {
 
     var descriptor = require("adapter/ps/descriptor"),
         layerLib = require("adapter/lib/layer"),
+        documentLib = require("adapter/lib/document"),
         generatorLib = require("adapter/lib/generator"),
         preferenceLib = require("adapter/lib/preference");
 
@@ -40,14 +41,9 @@ define(function (require, exports) {
         collection = require("js/util/collection"),
         strings = require("i18n!nls/strings"),
         log = require("js/util/log"),
+        global = require("js/util/global"),
         ExportAsset = require("js/models/exportasset"),
         ExportService = require("js/util/exportservice");
-
-    /**
-     * Export metadata is stored in the "extension" data structure in Ps, using this as the namespace
-     * @type {string}
-     */
-    var EXTENSION_DATA_NAMESPACE = "designSpace";
 
     /**
      * An instance of the ExportService utility used to communicate with the generator plugin
@@ -65,7 +61,8 @@ define(function (require, exports) {
 
     /**
      * Fetch relevant data from both the document and export stores, and sync the metadata to
-     * the photoshop layer "extension data"
+     * the photoshop "extension data".  If layerID is not supplied, then only document level 
+     * metadata is synced
      *
      * Note that the underlying adapter API to setExtensionData requires the full data structure to be supplied
      * which is why we are always fetching directly from both stores prior to calling setExtensionData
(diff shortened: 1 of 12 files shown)


src/js/actions/export.js
```diff
@@ -74,41 +71,82 @@ define(function (require, exports) {
      *
      * @private
      * @param {number} documentID
-     * @param {number} layerID
+     * @param {number=} layerID optional, if supplied, sync doc-level metadata
```
volfied: Do you mean, if not supplied?

src/js/actions/export.js
```diff
@@ -74,41 +71,82 @@ define(function (require, exports) {
...
-            exportEnabled = document && layer && layer.exportEnabled;
-
-        var exportsMetadata = {
-            exportAssets: layerExportsArray,
-            exportEnabled: exportEnabled === undefined ? false : exportEnabled
-        };
+            exportsMetadata,
+            playObject;
+
+        if (layerID || layerID === 0) {
```
volfied: Confused by the `layerID === 0` line here

src/js/actions/export.js
mcilroyc: yes, thanks

src/js/actions/export.js
mcilroyc: I wasn't sure if layerIDs could legitimately be zero, but if so, then a simple `if layerID` test will return false.  because javascript.

iwehrman: Since the param is marked as optional (`number=`) it should be sufficient to test for `layerID !== undefined`.

src/js/actions/export.js
```diff
@@ -74,41 +71,82 @@ define(function (require, exports) {
...
+                throw new Error("Could not sync export metadata for doc:" + documentID + ", layer:" + layerID);
+            }
+
+            exportEnabled = layer.exportEnabled;
 
-        var playObject = layerLib.setExtensionData(documentID,
-                layerID, EXTENSION_DATA_NAMESPACE, "exportsMetadata", exportsMetadata);
+            exportsMetadata = {
+                exportAssets: layerExportsArray,
+                exportEnabled: exportEnabled === undefined ? false : exportEnabled
```
iwehrman: or just `!!exportEnabled`

src/js/actions/export.js
```diff
@@ -467,38 +498,93 @@ define(function (require, exports) {
...
+                        }, this);
+                }, this);
+
+                return Promise.all(exportArray.toArray());
+            });
+        }.bind(this);
+
+        // prompt for folder and then export to the result.
+        // resolve immediately if no folder is returned
+        return _exportService.promptForFolder(_lastFolderPath || "~")
```
iwehrman: This probably isn't new, but: does is default "~" string work on Windows?

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -108,16 +109,45 @@ define(function (require, exports, module) {
...
+                exportsList = documentExports && documentExports.rootExports;
+            }
+                
+            var existingScales = (exportsList && collection.pluck(exportsList, "scale")) || Immutable.List(),
                 remainingScales = collection.difference(ExportAsset.SCALES, existingScales),
                 nextScale = remainingScales.size > 0 ? remainingScales.first() : null,
-                nextAssetIndex = (layerExports && layerExports.size) || 0;
+                nextAssetIndex = (exportsList && exportsList.size) || 0;
+
+            var layers = layer ? Immutable.List([layer]) : null;
```
iwehrman: (I know this is temporary, but I can't help myself.) The following is better: `Immutable.list.of(layer)`.

src/js/actions/export.js
mcilroyc: Will clean up in next PR, since this code is shifting slightly therein

src/js/actions/export.js
mcilroyc: I think I tested this, but I took a note to double check.

src/js/stores/export.js
```diff
@@ -128,15 +128,22 @@ define(function (require, exports, module) {
...
 
             if (!documentID) {
                 throw new Error ("Can not update an asset without a valid documentID (%s)", documentID);
             }
 
             var curDocumentExports = this.getDocumentExports(documentID) || new DocumentExports(),
-                nextDocumentExports = curDocumentExports.mergeLayerAssets(layerIDs, payload.assetPropsArray);
+                nextDocumentExports;
 
+            if (layerIDs || layerIDs === 0) {
```
iwehrman: Ditto with the === 0 here.

iwehrman: Er, maybe not. I'm not clear why layerIDs would be equal to 0 here.

iwehrman:
Just a few surface-level comments. I have some concerns that I've expressed to @placegraphichere about disabling document export for AB documents. My current opinion is definitely that the cons of enabling AB document export are outweighed by the cons of disabling it.

src/js/stores/export.js
mcilroyc: sorry, but would like to defer this.  In the next PR, this handler accepts an iterable of layerIDs.

mcilroyc:
I fixed a few nits, but also requested deferment of others

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -134,29 +164,37 @@ define(function (require, exports, module) {
...
                 containerContents = (<div>{strings.EXPORT.SELECT_SINGLE_LAYER}</div>);
+                disabled = true;
+            } else if (document.layers.backgroundSelected) {
+                // don't support exports of the background
+                containerContents = null;
+                disabled = true;
+            } else if (document.layers.selected.size === 0 && document.layers.hasArtboard) {
+                // don't support document exports when artboards
+                containerContents = null;
+                disabled = true;
```
iwehrman: After discussion with @placegraphichere, I believe the consensus is that we _should_ allow exporting documents that contain artboards, assuming it is at most a small amount of work to do so.

mcilroyc: Ok I'll capture this as a trello card.

mcilroyc: Correction: I just pushed a commit that removes the restriction.  If the asset generation is too slow or broken we can open bugs against it.  @iwehrman

src/js/actions/export.js
iwehrman: @ktaki When you test this, can you verify that with clean preferences the default folder-chooser location is the user's home directory on Windows?

iwehrman:
done

Issue #2130, opened 2015-08-20
Dragging mouse from outside edge blinks the create guide area


chadrolfs:
http://cl.ly/1h1V0q2F0c3p

iwehrman:
I strongly disagree that this is a low priority bug. This really makes the UI feel bad.

volfied:
This is hairy... externalMouseMove isn't always sending us the correct coordinates. I'll work with Fitz to figure out why.

iwehrman:
FBNC to @chadrolfs. Nice work, @volfied.

chadrolfs:
Looks fixed on Mac but not on Win

volfied:
@chadrolfs #2288 will fix this, however, during testing, please keep in mind that guide creation is still not working correctly on Windows, so please test this bug/fix only for edge blinking.

PR #2288, opened 2015-09-01
ShouldComponentUpdate for GuidesOverlay

volfied:
This prevents component updates occurring due to mouse moves in Guides Overlay, which caused the blink. 

Addresses #2130

### DIFF
--- a/src/js/jsx/tools/GuidesOverlay.jsx
+++ b/src/js/jsx/tools/GuidesOverlay.jsx
@@ -24,7 +24,8 @@
 define(function (require, exports, module) {
     "use strict";
 
-    var React = require("react"),
+    var Immutable = require("immutable"),
+        React = require("react"),
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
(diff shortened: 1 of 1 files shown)


volfied:
Please keep in mind that this will not fix #2170

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -60,17 +61,27 @@ define(function (require, exports, module) {
...
+                uiState = uiStore.getState(),
                 currentTool = toolStore.getCurrentTool(),
                 currentDocument = applicationStore.getCurrentDocument();
 
             return {
                 document: currentDocument,
                 tool: currentTool,
-                modalState: modalState
+                modalState: modalState,
+                uiState: uiState
```
mcilroyc: It seems weird that you are adding these to two state properties just to be used in SCU.  How are these two states accessed within the rendering?

volfied: Because of the React/D3 dance... I want any ui store and tool modal state changes to cause redraw (scrolling, zooming etc. all change window to canvas transform matrices, and we don't wanna show these during modal states)

Look at drawOverlay method for modalState being used, and any uiStore. functions within the file for how the ui store state is being used.

Issue #2285, opened 2015-08-31
Pinched Adobe Stock icon


iwehrman:
Retina:
![image](https://cloud.githubusercontent.com/assets/1075154/9592806/9be622b0-4ffb-11e5-9116-2fe7cca4ef86.png)

PR #2291, opened 2015-09-01
Fix Adobe Stock icon so it's not squished

placegraphichere:
fix for issue #2285


volfied:
Yoink, and looks good, but please avoid using the word "fix" or the issue number in your commit message next time. It automatically closes the bug which is undesired behavior.

Issue #2265, opened 2015-08-28
Toolbar is not included in cloaked UI bounds


iwehrman:
![cloak](https://s3.amazonaws.com/f.cl.ly/items/1f3h0c1i0p42271Q152X/Screen%20Recording%202015-08-28%20at%2011.59%20AM.gif)

Issue #2214, opened 2015-08-25
After switching displays, guide zones are the wrong size


iwehrman:
1. Launch on monitor A
2. Move window to monitor B
3. Drag out a guide:

![image](https://cloud.githubusercontent.com/assets/1075154/9478034/6a37a692-4b2a-11e5-9ea4-39b50fc79d45.png)

iwehrman:
We should be able to use the new `displayConfigurationChanged` OS event to address this.

volfied:
Giving this one to @chadrolfs as it's guides related.

Issue #2156, opened 2015-08-21
Unhandled exception clicking toolbar button during startup


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9411607/d4d552f8-47db-11e5-8617-bb96805f05fd.png)

This is happening during `afterStartup`. It's not clear why we haven't selected a tool by this point.

iwehrman:
Fixed by #2294.

PR #2294, opened 2015-09-01
Update panel sizes when display configuration changes

iwehrman:
1. Updates panel sizes (and hence the adapter's "overlay offset" state) when the display configuration changes, using the new `displayConfigurationChanged` os event. Addresses #2214.
2. Reduces `updateToolbarWidth` to a helper function that immediately transfers to `updatePanelSizes`. Removes the `TOOLBAR_PINNED` event.
3. Averts an unhandled exception when clicking on the toolbar before any tool is known to be active. Addresses #2156.
4. Correctly updates the toolbar width even when unpinned. Addresses #2265.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -236,6 +236,7 @@ define(function (require, exports) {
      * Parse the panel size information and dispatch the PANELS_RESIZED ui event
      *
      * @private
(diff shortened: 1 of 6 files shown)


iwehrman:
Also pushed a commit which ensures that the panel sizes are updated when changing the column configuration.

volfied:
Looks good to me.

Issue #2175, opened 2015-08-22
Better support for sampling mixed text layers


iwehrman:
Text layers with mixed style ranges can't currently be sampled. We should at least support sampling those properties that are consistent throughout the text layer. For example, if the colors are different but everything else is the same, we should be able to sample everything but the color. 

For context see #2146 and #2137.

iwehrman:
Tossing this over to @chadrolfs for verification.

PR #2296, opened 2015-09-01
Create only the most common type properties for getTypeObjectFromLayer

volfied:
Also fixes applyTextStyle to do selection dance.

Addresses #2175,

Requires https://github.com/adobe-photoshop/spaces-adapter/pull/142

@iwehrman please review alongside 142

### DIFF
--- a/src/js/actions/sampler.js
+++ b/src/js/actions/sampler.js
@@ -42,6 +42,7 @@ define(function (require, exports) {
         layerActions = require("./layers"),
         typeActions = require("./type"),
         shapeActions = require("./shapes"),
+        toolActions = require("./tools"),
         layerActionsUtil = require("js/util/layeractions"),
         strings = require("i18n!nls/strings");
 
@@ -129,11 +130,7 @@ define(function (require, exports) {
                 if (source.isTextLayer() && allTextLayers) {
                     var fontStore = this.flux.store("font");
 
-                    try {
-                        typeStyle = fontStore.getTypeObjectFromLayer(source);
-                    } catch (err) {
-                        // For mixed type styles, this function throws, so we ignore!
-                    }
+                    typeStyle = fontStore.getTypeObjectFromLayer(source);
                 }
 
                 result.push({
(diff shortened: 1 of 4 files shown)


src/js/stores/font.js
```diff
@@ -138,27 +138,41 @@ define(function (require, exports, module) {
...
-            }
+            // If it's mixed, or there is no fontObj, we skip adding these values to the type object
+            // adbeFont, fontFamily, fontStyle, fontWeight
+            if (fontObj) {
+                var family = fontObj.family,
+                    fontRecord = this._familyMap.get(family),
+                    psObj = fontRecord.get(fontObj.font, null);
+
+                if (!psObj) {
+                    throw new Error("The font for layer is not available!");
```
iwehrman: I'm curious as to why you chose to throw here. This will cause `duplicateTextStyle` to reject. Is that what you want? Would it be better to just log a warning, return null, and then make sure callers handle that scenario?

iwehrman: Also `libraries.createCharacterStyleFromSelectedLayer` and `sampler.copyLayerStyle` will reject, causing a controller reset.

iwehrman:
One question about error handling. Back to @volfied.

src/js/stores/font.js
volfied: I don't want this function to return null, I want it to return whatever is available in that layer's characterstyle in an object. For mixed state layers, this will only be the common properties. I'll at least log a warning there and skip that property.

volfied:
back to you @iwehrman

src/js/actions/sampler.js
```diff
@@ -554,15 +551,23 @@ define(function (require, exports) {
                 textAlignmentPromise,
                 textStylePromise,
                 effectsPromise)
+            .bind(this)
             .then(function () {
                 return descriptor.endTransaction(transaction);
+            })
+            .then(function () {
+                return this.transfer(layerActions.resetLayers, document, targetLayers);
```
volfied: Added these, because during hte transactions, our resetLayers / resetBorderPolicies do nothing (as the local data structures don't get updated)

iwehrman: That's scary.

iwehrman: Fine for now, but we should definitely talk more about this.

PR #2303, opened 2015-09-01
Draw canvas bounds with policy frames

iwehrman:
When showing policy frames, also show the canvas bounds (as defined by the UI edges) to make it easier to debug issues with the overlay offsets, etc. It looks like this:
image


volfied:
What could go wrong!

iwehrman:
Already found one more canvas bounds bug :grimacing:

Issue #2144, opened 2015-08-20
Unable to move guides attached to Artboards


chadrolfs:
in classic PS, have a doc with an Artboard
with the AB targeted (focused in layers panel)
(with Rulers shown) drag out a couple guides into the AB
Switch to DS and try to move the guides

Note: you can only see the guides if the AB has focus

volfied:
Will be fixed by #2158

ktaki:
This is still be the case with merged PR #2158.
Create a new artboard. Drag a guild on to the artboard.
Deselect artboard.
Select artboard.
The guide cannot be moved.

volfied:
I can not repro, @ktaki. Can you provide video with the latest build of full steps please?

chadrolfs:
Here are the breakage scenarios:

Open a doc that has doc guides and one AB with guides attached to it.
- if you move the guides in the AB first, you will not be able move the doc guides
- if you move the doc guides first, you won't be able to move the AB guides
- if you move the actual AB before you touch either type of guide, you end up not being able to move any.

volfied:
Oops, I forgot to remove the reverse call, so index i in n guides was ending up editing n-i th guide

iwehrman:
FBNC to @chadrolfs.

PR #2305, opened 2015-09-01
Get guides in correct order from PS and ask for them on layer moves

volfied:
My recent core related guide changes, I introduced a `reverse` call to getGuidesForDocumentRef call... so editing one guide was causing another guide to change.

I also noticed that artboards move their guides with them, but we don't do anything to update the guides. So for now I'm querying all guides on resetBounds calls. In the future this can be done smarter by just asking for that artboards guides if any is moved (I'd like to unblock guides first, and then fix this in a separate issue)

Addresses new gripes in #2144

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -86,7 +86,6 @@ define(function (require, exports) {
             function (required, optional) {
                 return _.chain(required)
                     .zipWith(optional, _.merge)
-                    .reverse()
                     .value();
             });
     };
@@ -316,10 +315,16 @@ define(function (require, exports) {
     /**
      * Re-gets the guides of the given document and rebuilds the models
      *
-     * @param {Document} document
+     * @param {Document=} document Default is active document
      * @return {Promise}
      */
     var queryCurrentGuides = function (document) {
+        var appStore = this.flux.store("application");
+
+        if (document === undefined) {
+            document = appStore.getCurrentDocument();
+        }
+        
         var docRef = documentLib.referenceBy.id(document.id);
 
         return _getGuidesForDocumentRef(docRef)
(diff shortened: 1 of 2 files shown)


src/js/actions/guides.js
```diff
@@ -86,7 +86,6 @@ define(function (require, exports) {
             function (required, optional) {
                 return _.chain(required)
                     .zipWith(optional, _.merge)
-                    .reverse()
```
iwehrman: ¯\\\_(ツ)\_/¯

src/js/actions/guides.js
```diff
@@ -320,6 +319,10 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var queryCurrentGuides = function (document) {
+        var appStore = this.flux.store("application");
+
+        document = document || appStore.getCurrentDocument();
```
iwehrman: Need a JSDoc fix up if this param is optional. Also, it would be better to do this the right way, testing for `document === undefined`.

src/js/actions/layers.js
```diff
@@ -487,12 +488,14 @@ define(function (require, exports) {
                     this.dispatch(events.document.history.nonOptimistic.RESET_BOUNDS, payload);
                 }
             }).then(function () {
+                return this.transfer(guides.queryCurrentGuides);
```
iwehrman: Both of these `.then`s should be on their own line.

iwehrman: Also, do we really need to delay `resetBorderPolicies` until _after_ `queryCurrentGuides` is complete?

Vaguely related gripe: I really want to clean up our action terminology. Quick, what's the difference between "setX", "updateX", "resetX" and "queryX"??

volfied: Yes, `resetBorderPolicies` transfers to `resetGuidePolicies` so we need the latest guides.

volfied:
Nits addressed, @iwehrman

iwehrman:
Okie doke.

Issue #2317, opened 2015-09-02
Guides zones get stuck on when moving cursor from canvas to UI


iwehrman:
![stuck](https://s3.amazonaws.com/f.cl.ly/items/2g012l1c3i420F160K0e/Screen%20Recording%202015-09-02%20at%2012.24%20PM.gif)

iwehrman:
Fixed by #2318.

PR #2318, opened 2015-09-02
Listen to html mousemove events for guides overlay

volfied:
@iwehrman please review since you opened the issue :)

Addresses #2317


iwehrman:
OK, sure.

Issue #2275, opened 2015-08-29
Selecting artboard after dismissing Export dialog causes error


ktaki:
Create an artboard. Add a shape. Add export setting to the artboard. Open Export dialog. Do export. Dismiss dialog. Click the artboard name to select the artboard. This often (not always) fails with the following error. (This is a fatal error. DS needs to be reset.)

![screen shot 2015-08-28 at 5 23 28 pm](https://cloud.githubusercontent.com/assets/2029597/9559546/f343c46c-4da9-11e5-9b99-448a66e6a7da.png)

iwehrman:
I'm hitting something similar with the following steps:

1. Create a new non-AB document. Unlock the background layer. Configuring an export for that layer.
2. Repeat.
3. Select first document.

Result:
```
Error: Failed to register dialog: datalist-layerExportAsset-scale2-0 already exists.
    at Fluxxor.createStore.registerDialog (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/stores/dialog.js:131:23)
    at React.createClass.componentWillMount (file://./js/jsx/shared/Dialog.jsx:109:44)
    at chainedFunction [as componentWillMount] (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6044:9)
    at ReactCompositeComponentMixin.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6884:14)
    at ReactDefaultPerf.measure (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:10068:19)
    at ReactPerf.measure.wrapper [as mountComponent] (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13968:31)
    at Object.ReactReconciler.mountComponent (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:14781:35)
[...]

Issue #2153, opened 2015-08-21
Export assets are given "@nullx" suffix


iwehrman:
Click the Export + button a few times and you start getting an `@nullx` suffix after running out of other suffixes like `@1x`, `@2x`, etc.

iwehrman:
Fixed by #2319.

PR #2319, opened 2015-09-02
Mcilroy/export multi select

mcilroyc:
1. Quick add (IOS, HDPI) asset sets (via @placegraphichere)
1. Manage export assets when multiple layers are selected

When multiple layers are selected, we now attempt to find a uniform set of assets shared between them.  This is not particularly smart because it relies on the indexes of each being equal.  Improving this should be possible, but saving for future work.

For example:
{a, b, c}
{a, b, d}
 --> {a, b}

but 
{a, b, c}
{z, a, b, d}
 --> {}

Furthermore, when you ADD new asset(s) when multiple layers are selected, it will *inject* the new asset(s) *after* the last uniform value.  Consider inserting {x, y} in the following...
{a, b, c}
{a, b, c, d, e}
-->
{a, b, c, x, y}
{a, b, c, x, y, d, e}

Notes:

1. ps metadata sync will now batch the updates into a single play call (natch)
1. Use a unique key for ExportList which should fix the occasional dialog register/deregister bugs like #2275
1. Addresses #2153
1. Includes some icon cleanup, including the :heavy_plus_sign: used by Shadow


iwehrman:
Both the iOS and the hdpi buttons add an extra 1x SVG export. That isn't intended, right?

iwehrman:
Also, the iOS button is not adding a suffix to the 1x asset. Is that intended?

mcilroyc:
yes, yes
https://github.com/adobe-photoshop/spaces-design/pull/2319/files#diff-80e393b81d94d82d78d6e321768db693R191

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -188,28 +195,36 @@ define(function (require, exports, module) {
...
+                    });
+                    
+                    assetGroups = collection.zip(allExports);
+                } else {
+                    assetGroups = collection.zip(Immutable.List.of(documentExports.rootExports));
+                }
             }
 
-            if (!exportsList || exportsList.size < 1) {
+            if (!assetGroups || assetGroups.size < 1) {
```
iwehrman: Wait, I thought that `.zip` returned a Seq which does not have a `size` property?

mcilroyc: Well, it *does* unless you post-filter it.  I should probably be safer here, true.  This is not the instance of zip that was causing the issue that led me to realize its a seq.

src/js/models/documentexports.js
```diff
@@ -202,6 +267,56 @@ define(function (require, exports, module) {
...
+            var newAssetArray = assetProps.map(function (props) {
+                var newAsset = new ExportAsset(props);
+                if (!_.has(props, "suffix")) {
+                    newAsset = newAsset.deriveSuffix();
+                }
+                return newAsset;
+            });
+
+            var assetList = this.layerExportsMap.get(layerID) || Immutable.List(),
+                nextAssetList = assetList.splice.apply(assetList, [index, 0].concat(newAssetArray));
```
iwehrman: Brain freeze.

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -229,8 +222,26 @@ define(function (require, exports, module) {
...
-                                    viewbox="0 0 12 12"
-                                    CSSID="plus" />
+                                    viewbox="0 0 16 16"
+                                    CSSID="add-new" />
+                            </Button>
+                            <Gutter />
+                            <Button
+                                className="button-iOS"
+                                title={strings.TOOLTIPS.EXPORT_IOS_PRESETS}
+                                onClick={addIOSAssetClickHandler || _.noop}>
```
iwehrman: Pretty sure these `|| _.noop`s are not necessary. If it's null or undefined, React does the right thing.

src/js/jsx/sections/export/ExportList.jsx
mcilroyc: Furthermore, some of the above can be refactored to use the cached prop in documentExports, so I'll do that.

src/js/actions/export.js
```diff
@@ -67,53 +67,54 @@ define(function (require, exports) {
...
-        } else {
-            var rootExports = documentExports && documentExports.rootExports,
-            rootExportsArray = rootExports && rootExports.toJS() || [];
+        };
 
-            exportsMetadata = {
-                exportAssets: rootExportsArray
+        // prepare play objects for document or layer level, based on existence of layerIDs
+        if (layerIDs) {
+            playObjects = layerIDs.map(_buildPlayObject).toArray();
```
iwehrman: At some point, we should just push the Immutable dependency into the adapter to avoid all this `toArray` nonsense.

src/js/actions/export.js
```diff
@@ -163,11 +166,44 @@ define(function (require, exports) {
...
+                return this.transfer(updateExportAsset, document, _layers, assetIndex, assetProps);
             }) ;
     };
 
     /**
+     * Insert an asset (or assets) into a document or set of layers, by inserting into the existing list
+     * at the given index.
+     * Sync to PS metadata afterwards
+     *
+     *
```
iwehrman: @shaoshing is gonna smack you if you try to check in two empty lines in a row here.

src/js/actions/export.js
```diff
@@ -163,11 +166,44 @@ define(function (require, exports) {
...
+     * @param {object|Array.<object>} props ExportAsset-like properties to be merged, or an array thereof
+     * @return {Promise}
+     */
+    var _insertAssetsAtIndex = function (document, layers, assetIndex, props) {
+        var documentID = document.id,
+            layerIDs = layers && collection.pluck(layers, "id") || undefined,
+            assetPropsArray = Array.isArray(props) ? props : [props],
+            payload;
+
+        payload = {
```
iwehrman: I think the preferred style is just to declare payload on this line instead of declaring it and then defining it.

iwehrman:
For the record, I think the new icons are in the wrong order. I think it should be: iOS, hdpi, :arrow_down: :heavy_plus_sign:. CC @placegraphichere

iwehrman:
That's all I got. Lookin' good. Back to @mcilroyc.

mcilroyc:
I fixed a few things and pushed a new commit.  Back to @iwehrman, unless he's waiting on feedback from @placegraphichere

iwehrman:
I'll complain more to @placegraphichere offline. Here goes nothin'!

PR #2321, opened 2015-09-02
Minor dependency version updates

iwehrman:
In honor of facebook/immutable-js#602, here are some minor updates for third-party dependencies. I did some light testing, and I'm not aware of any changes in functionality relevant to our project.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,27 +24,28 @@
     "tests"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


iwehrman:
The only point of concern is the major-version upgrade of mathjs, but everything seems to be working as before.

volfied:
None of the breaking changes listed in mathjs HISTORY page apply to us.

Issue #2301, opened 2015-09-01
marquee select flag not set correctly: highlights not shown (Sampler, HUD, etc)


chadrolfs:

chadrolfs:
Gets you into a bad state where bad things happen (like the type datalist getting hosted).

volfied:
I don't see the typeface datalist being broken, even when I repro this bug.

chadrolfs:
I can't either with updated jenkins build and master

chadrolfs:
although I just repro'd in a different scenario.  I'm not sure what the reliable repro case is.

Issue #2150, opened 2015-08-21
Very slow "nudge" transform needs to be investigated


tiburbage:
Not sure if we intend to support arrow keys nudge action for layers/artboards, but our response when you do this is very slow. Naturally, this is most noticeable for heavy artboards or layer groups.

1 - Open VermilionArtboards.psd in standard ui
2 - With the Move Tool (or Artboard Tool) enabled, select one of the Artboards
3 - Press any of the "arrow" keys to nudge the content around. It works very quickly.
4 - Switch to Design Space
5 - Repeat the nudge. I'm seeing 20-30 sec wait times to complete.

Since we seem to be able to drag the same content around will low latency, hoping a more efficient update code path can be applied to nudging.

iwehrman:
Each instance of `transform.nudgeLayers` takes ~100ms in a debug build in Vermilion for me. Are you seeing the problem when you hold an arrow key down?

iwehrman:
CC @volfied because I'm pretty sure we discussed this at some point. I suspect the problem is with how we're throttling this command. I also suspect that the fix is not terribly obvious with our infrastructure. What we really need is something like a nudge tracker...

tiburbage:
When you say "hold an arrow key down", do you mean as opposed to a single tap? I'm just giving it a single keydown/up, but I think it gets worse if I hold it down. The processing time I cited is on Windows. On my MacBook Pro, it is faster but still a long spin.

tiburbage:
Logged in Watson: https://watsonexp.corp.adobe.com/#bug=4045560
Jesper thinks we need to provide access to a faster transform method.

jsbache:
yes: DS seems to be using transform and we need to provide access to "nudgeLayers" from the move tool implementation

jsbache:
CL: 1039134 add a nudgeSelectedLayers command to the document element. This can be used to move layers similar to what happens when the user press the arrow key in standard.
See http://sol.corp.adobe.com/Photoshop/PlaygroundActionDescriptor#Nudging_selected_layers
Sample:
var reference = {_ref: "document", _enum: "ordinal", _value: "targetEnum"}
var desc = {"null":reference, "offset":{"vertical":10, "horizontal":0}}
_playground.ps.descriptor.play("nudgeSelectedLayers", desc, {}, callback)

jsbache:
Required support in Photoshop is available via eventMove and eventCopy.
[...]

PR #2322, opened 2015-09-02
React to Nudge, and fix marquee

volfied:
f9a3db3 addresses #2150, and it works. However, PS doesn't always send us ALL the nudge events, so if you want to test it, please download latest Jenkins of pg-dev to get ALL the nudge events.

40c6003 addresses #2301, problem was, I was only finishing marquee select if there was a rectangle being drawn, so for small enough distances / fast mouse events / action not finishing on time, the marquee flag would remain up, and overlays would be broken. Now it's more stable, and we reset the flag on tool select as well (so user can't switch tools while marquee selecting to leave the flag up, darn user!)

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -697,7 +697,7 @@ define(function (require, exports) {
      * Otherwise will add/transfer selection to layers
      *
      * @param {Document} doc Owner document
-     * @param {Array.<number>} ids Layer IDs
+     * @param {?Array.<number>} ids Layer IDs
      * @param {boolean} add Flag to add to or replace selection
      * @return {Promise}
      */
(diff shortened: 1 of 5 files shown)


src/js/actions/superselect.js
```diff
@@ -703,6 +703,10 @@ define(function (require, exports) {
      */
     var marqueeSelect = function (doc, ids, add) {
         this.dispatch(events.ui.SUPERSELECT_MARQUEE, { enabled: false });
+
+        if (!ids) {
```
iwehrman: If `ids` can be null, the JSDoc param needs to be updated with `?Array...`

iwehrman:
Just the JSDoc nit.

iwehrman:
Oh, wait, also: don't you want to remove the `nudgeLayers` action?

volfied:
Thanks @iwehrman, made changes.

iwehrman:
:+1:

Issue #2287, opened 2015-09-01
Info message for empty export and libraries should match visually


placegraphichere:
![screen shot 2015-08-31 at 10 40 09 pm](https://cloud.githubusercontent.com/assets/3229167/9595829/4369fb58-5031-11e5-9119-aa555fc7eaa8.png)
![screen shot 2015-08-31 at 10 40 06 pm](https://cloud.githubusercontent.com/assets/3229167/9595828/43699d5c-5031-11e5-948f-d67e8216f107.png)

iwehrman:
FBNC to @placegraphichere. I believe this message no longer appears in the panel at all.

placegraphichere:
should be handled with #2323

PR #2323, opened 2015-09-03
unify panel messaging appearance

placegraphichere:
fix for #2287, even though we won't use the messaging on export, we'll need this sometime, somewhere.


iwehrman:
It seems like this _should_ be used in case a layer is selected for which export is not supported, e.g. a background. (CC @mcilroyc) So, let's do it.

PR #2325, opened 2015-09-03
match Ps layer nesting readability

### LINKED ISSUE #2286: Cannot display deeply nested groups in layers panel
Practically, we only can support 6 levels plus artboard (total 7 levels of nested groups) currently.

screen shot 2015-08-31 at 5 00 56 pm

In standard Ps,

screen shot 2015-08-31 at 5 01 23 pm

placegraphichere:
fixes #2286


shaoshing:
:+1:

Issue #2187, opened 2015-08-24
Takes 2 clicks to select/move an artboard if another layer is selected


chadrolfs:
new AB doc, drag out some shapes, Select tool (V), click and move shapes (expected workflow).  With a shape selected, click on AB title and drag

Result: first click drops selections.  it takes a second click to select and move the AB

iwehrman:
FBNC to @chadrolfs.

PR #2327, opened 2015-09-03
Always include artboards in superSelectable

volfied:
Will help address #2187, we basically were not including owner artboard in superselectable layers, so now artboards are special cased, so they can be selected/dragged in one go.


iwehrman:
Yes.

iwehrman:
I think super select is done now.

PR #2331, opened 2015-09-03
Toggle overlay cleanup

volfied:
Back in #1960, when we had on canvas transform, we had to toggle the overlays for a lot of actions ourselves. But with the invention of cloaking technology, and us drawing transform through Photoshop, most of those toggles became obsolete...

Simple code clean up, removed it where it didn't make a difference.


mcilroyc:
hoo ray

Issue #2168, opened 2015-08-21
Hard to grab transformation handles in skew or distort mode


ktaki:
With a shape layer press Cmd (mac) or Ctrl (win) to be in skew/distort mode. Drag transformation handles. The first time works, but subsequent tries may or may not work. This is because as soon as mouse is released, the bounding box is redrawn to be a rectangle in DS while the transformation mode continues in Ps and the transformation bounding box returned from Ps may be different.

chadrolfs:
Using modifier while transform dragging.

mcilroyc:
FBNC to @ktaki   -- this passed my cursory test, but admittedly I may not be attuned to some of the nuances.

ktaki:
It works better than before, transformation handles are now not easy to miss. However, true free transformation which standard Ps offers with modal state cannot be achieved here. Once you release the mouse button, the bounding box and handles will reset to be a rectangle. This prevent users from adjusting details of the transformation. I accept this as a limitation of current design, but will open a separate issue for this hoping to be improved in the future.

PR #2332, opened 2015-09-03
Add a always-propagate policy for command modifier (distort) on border

volfied:
This makes cmd dragging easier on transform handles. Before, you had to click, and then hold down cmd to distort.

Addresses #2168


mcilroyc:
seems to work!

Issue #2231, opened 2015-08-26
Swap performance is slow


chadrolfs:
On my windows machine, it's in the 2553ms range.  On my mac pro it's 1600ms.

chadrolfs:
with the App Ideas.psd file swapping each artboard.  Let me know if you need the file.

volfied:
Seems like our extensive playLayerActions that calculates move for each child layer in the artboard takes 2.3 seconds in this file. We should first check and see if we can swap artboards without referring to their children.

iwehrman:
dupe of #1513?

volfied:
Not a dupe, #1513 was due to us having to wait for models to update so we can call swap in the next model (instead of immediately calling swap twice on the same model). This is an actual performance hit that happens during busy artboard calculations.

volfied:
Note to self: if swapping artboards, don't break it down to layers and see if just swapping the artboard locations is faster.

mcilroyc:
@chadrolfs FBNC via the above PR.

chadrolfs:
Still slow.  (on mac) using dev build 586, master @ 312f909b.
` Finished action transform.swapLayersCurrentDocument in 6528ms with RTT 6533ms; 1/1`

volfied:
At this point we're sending 5 commands to PS in total, so something in PS is slowing down. Sent a profile report to Jesper to see how we can make it fast on native side.

chadrolfs:
now 159ms to 249ms.  So much better, thank you!

PR #2334, opened 2015-09-03
Don't child calculate for artboard swap so it's faster

volfied:
For swap layers, we have an algorithm that calculates the offset of the two layers to be swapped, and applies that offset to EVERY layer that's child of the swapped layer. This takes quite a while, because we're asking PS to transform a whole bunch of layers. For artboards, we can skip this by just translating the artboard.

This helps swapping two artboards go from 2500 ms to 170 ms. Of course, if you swap two busy groups, it'll still take a long time, but that's much less common/expected than swapping two artboards.

Addresses #2231


mcilroyc:
looks good - and seems to work!

Issue #1081, opened 2015-06-12
New document is zoomed incorrectly


DesignSpaceBot:
New documents should start in the position that results from cmd-0.
![big](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @iwehrman at [2015-04-09T23:52:44Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1081)

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=3981864

volfied:
Fixed by #2339

volfied:
New documents now start on the correct location, zoom level is different from cmd+0, because new doucments start at 66% zoom, and cmd+0 is fit to window, which will vary the zoom level. I'm closing this.

Issue #2206, opened 2015-08-25
Zoom shortcuts don't function correctly


chadrolfs:
cmd/ctrl + n
drag out some rect shapes inside AB
layer > create new artboard
cmd/ctr + 0 (fit in window)

Result: only artboard 1 is set to fit zoom
even if you add shapes to artboard 2 and then hit cmd/ctrl + 0 again, it zooms to fit only artboard 1

iwehrman:
FBNC to @chadrolfs!!!!!

Issue #2169, opened 2015-08-21
Zoom out after creating a new doc shifts to the right under panels


chadrolfs:
https://www.dropbox.com/s/6wmo0kui308kwch/ZoomOut.mov?dl=0

placegraphichere:
Additional 
- Center a document artboard
- select none
- zoom out
-  artboard will shift up and to the right oddly

https://www.dropbox.com/s/hkemdvftzffapgn/zoom-out-centers-things-strangely.mov?dl=0

chadrolfs:
It's really bad on Windows.  It zooms out to the right and then pops back to center.  I'll try to post a video

iwehrman:
FBNC to @chadrolfs!!!!!!

chadrolfs:
Fixed on Mac.  Still looks weird on Win.

https://www.dropbox.com/s/0svfs7hheqgwh1i/capture_Tue_Sep_08_07.41.24.mkv?dl=0

mcilroyc:
take two, @chadrolfs

Issue #1709, opened 2015-07-09
Fit to Window should not use "infinite canvas" artboard bounds


iwehrman:
1. Create a new artboard-preset document
2. Move the artboard down and to the right.
3. View > Fit to Window

Result: find the artboard!

The problem is that we're using the document bounds, but... we shouldn't be. We probably need to compute our own bounds as the union of all top-level layers, and the centering code should use those bounds instead. These virtual bounds should be a derived property of the layer structure or document model.

shaoshing:
@iwehrman  The centering position is correct and the art board does centered. However, if you move the art board fast with the track pad, it will be still in the "moving" state after centered, and this cause the shifting problem. I confirmed standard PS has the same problem, so this is a core bug. Shall I report in watson?

![offset-centering](https://cloud.githubusercontent.com/assets/118264/9066544/3633aa24-3a8d-11e5-8d95-0e05be98102c.gif)

volfied:
This is also fixed by #2339

iwehrman:
Fixed either by #2339 or #1999.

PR #2339, opened 2015-09-03
On no-selection, don't shift zoom center.

volfied:
If there is no selection during zoom in/out, we keep the center fixed and zoom out/in correctly.

Addresses #661, and possibly couple other zoom bugs that I will add as I see them.
#2206
#2169
#1081
#1709
@iwehrman You sic'ed me on this, you review it :)

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -370,14 +370,13 @@ define(function (require, exports) {
      * Sets zoom to the value in the payload
      * Centering on the selection, pan is on by default
      *
-     * @param {{zoom: number, pan: boolean}} payload
+     * @param {{zoom: number}} payload
      * @return {Promise}
      */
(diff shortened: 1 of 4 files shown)


volfied:
With 889fc9c, also addresses #2206

volfied:
Also addresses #2169

volfied:
Also addresses #1081

volfied:
Also addresses #1709

iwehrman:
:moneybag: :moneybag: :moneybag:

iwehrman:
This doesn't quite fix #1081 because of the manual re-positioning that we're doing. This is related to #1999. Imma gonna fix that by just ripping out our re-centering code, since that should no longer be necessary given that our overlay offsets are fixed.

PR #2341, opened 2015-09-04
Styling for the exports panel

DivyaPrabhakar:
With reference to the designs here: https://trello.com/c/KR5wYVkW/494-design

*Spacing for the formlines in the export panel:* 
- I used the new stylings written for `.formline` and `.formline__space-between` to have the elements span the entire width of the Export Panel, as the designs show. 

- In order to have the formlines span the entire width of the Export Panel, I had to take away some of the modularity that `containerContents` provided so that an extra div would not be inserted. Then I could have dictate that the width of the `<ExportList>` will span the container. 

- I took out the `<Gutter>` components and replaced them by applying the `.control-group__vertical` to every element in a formline.  I have taken out all of the `<Gutter>` components `ExportList.jsx` and `ExportPanel.jsx`

- Added a dummy icon to the header formline in the Export Panel so that the labels would align properly with the inputs below. 

*Underlines for inputs*
- I added a div around the  `TextInput` component because all the other components within the formline were surrounded by divs. This meant when I added the underline, the `TextInput` line would be higher than the underlines for the datalists. 

- I also had to change the font size of the `TextInput` to match the font size in the datalists. This again helped even out the underlines between the different types of inputs. 

*Hover states* 
- On hover, each element in the formline changes color to indicate the hover state 

Final look of the Export Panel: 

alt text

alt text


designedbyscience:
1) What a great write-up. 
2) Failing on some `grunt` task, so you'll need to clean that up

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -226,28 +226,29 @@ define(function (require, exports, module) {
...
                                 title={strings.EXPORT.TITLE_SETTINGS}
                                 size="column-4"
                                 className="label__medium__left-aligned">
                                 {strings.EXPORT.TITLE_SETTINGS}
                             </Label>
-                            <Gutter />
+                            <Button
+                                className="layer-exports__delete-button">
+                                <SVGIcon CSSID="dummy" />
+                            </Button>
```
designedbyscience: So instead of actually making a dummy icon, you can just make a div with class column-xx (whatever size is needed) to fill the horizontal space. You should be able to get rid of the dummy SVG altogether, and not use the Button component here (using components that are at all complex has a penalty)

src/js/jsx/shared/Datalist.jsx
```diff
@@ -671,6 +671,7 @@ define(function (require, exports, module) {
 
             var dropDownClasses = classnames({
                 "drop-down": true,
+                "control-group__vertical": true,
```
designedbyscience: Remove this (we may not want EVERY datalist to have this class). Instead, you can wrap that component in a div that has that class

designedbyscience:
Design changes:
untitled-1___49_6___artboard_1__rgb_8____

src/style/sections/exports/exports.less
```diff
@@ -165,6 +165,33 @@
...
 }
 
+.layer-exports__header {
+   width: 100%;
+
+   .formline.formline__space-between {
+        & > div {
+            border-bottom: @hairline solid @underlines;
+            
+            & > input[type="text"] {
```
designedbyscience: Do you need the `>` selector? Best to have less specific selectors when possible

src/style/sections/exports/exports.less
```diff
@@ -165,6 +165,33 @@
     flex-direction: row;
 }
 
+.layer-exports__header {
+   width: 100%;
+
+   .formline.formline__space-between {
```
designedbyscience: It feels like (which is a terrible phrase to use) that there is a cleaner way to do this...I'm looking into it

designedbyscience: Okay, in main.less:157 and in drop-down.less:61, I added .export to the selector. That adds the underlines to the inputs. Then, you should only need to keep the font-size property that you have in this block

DivyaPrabhakar:
All changes made except the decision to have x's in the Scale column. Waiting for @placegraphichere to comment on that.

mcilroyc:
drive by comment: are you intentionally merging this in to a non-master branch?  I'll assume yes.

iwehrman:
I think the idea was to be able to review it (only) against the @designedbyscience's PR branch. Now that that branch has landed, this PR will have to be closed and a new PR opened against master. (There's no way to re-target a PR, right?) Luckily, PRs are free.

Issue #1999, opened 2015-08-11
New artboard document repositions itself


iwehrman:
New documents should not require repositioning. They should be positioned correctly by default. This should happen automatically as a result of the `setOverlayOffsets` calls. If the new document is not being  positioned correctly on creation, then either a) the bounds we're giving to `setOverlayOffsets` is incorrect, b) we are somehow re-positioning the artboard differently after creation; or c) there is a bug in Photoshop.

![jump](https://s3.amazonaws.com/f.cl.ly/items/0l2i2h2d0E1S3u1G3213/Screen%20Recording%202015-08-11%20at%2004.52%20PM.gif)

ktaki:
Could we improve the behavior of the first doc opened  (#2043) ?

zoranagee:
This is really jarring when you are going back and forth - check it out.
![documentzoom](https://cloud.githubusercontent.com/assets/7904855/9591542/e6f45c28-4ff0-11e5-9fa0-e0e0fd10e701.gif)

zoranagee:
We need to fix this one for the MAX keynote.

iwehrman:
The tricky thing here is that when you switch to standard PS, it moves the document into the center of the window. So we almost certainly want to do _some_ kind of repositioning when entering Design Space because the center of the window is probably not the best placement for Design Space. My current best guess is that we just want to _center_ all the already-open documents when they're first activated in Design Space. In particular, we should not be changing the zoom level at all, and we certainly should not be doing any repositioning of documents opened from within Design Space because we've already made it possible for PS to position new documents correctly when Design Space is active.

iwehrman:
Offline, @placegraphichere has suggested that, in fact, we do not want to even just center documents that were already open when switching to Design Space. I guess the thought is that having them off-center is better than the alternative of panning the canvas to a possibly distant position. So, OK!

PR #2343, opened 2015-09-04
Don't reposition newly selected documents

iwehrman:
1. Do not manually reposition documents on first selection. Instead, either leave the documents where they already are, or let Photoshop position the canvas using the overlay offsets.
2. Enable Zoom to Selection and Center Selection for the document when no layers are selected.

Addresses #1999 and #1081.

Note that a document opened in Design Space when there are currently no open documents will be positioned incorrectly because the panel bounds aren't currently updated until _after_ the first document is open. I intend to address that in this PR as well, but it doesn't work yet. So, I'm putting this up for review/testing first (by @placegraphichere) to make sure that, besides the first-doc behavior, this is an improvement.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -348,18 +348,25 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
+        var selection;
         switch (payload.on) {
-            case "selection":
-                targetBounds = currentDoc.layers.selectedAreaBounds;
-                if (!targetBounds || targetBounds.empty) {
-                    targetBounds = currentDoc.bounds;
-                }
-                break;
-            case "document":
+        case "selection":
+            selection = true;
+            break;
+        case "document":
+            selection = false;
+            break;
+        default:
+            selection = !currentDoc.layers.selected.isEmpty();
+        }
+
+        if (selection) {
+            targetBounds = currentDoc.layers.selectedAreaBounds;
+            if (!targetBounds || targetBounds.empty) {
                 targetBounds = currentDoc.visibleBounds;
-                break;
-            default:
-                throw new Error("Unexpected 'on' value");
+            }
+        } else {
+            targetBounds = currentDoc.visibleBounds;
         }
 
         return this.transfer(centerBounds, targetBounds, payload.zoomInto);
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -257,6 +257,43 @@ define(function (require, exports) {
     updatePanelSizes.modal = true;
 
     /**
+     * Set the overlay offsets in PS in anticipation of opening/creating the
+     * first document (i.e., from a state in which there are no open documents).
+     * This is used, e.g., to ensure that the offsets account for the UI columns
+     * that will be shown once the document is open. See #1999 for more details.
+     *
+     * @return {Promise}
+     */
+    var setOverlayOffsetsForFirstDocument = function () {
+        var flux = this.flux,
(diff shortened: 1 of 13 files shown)


iwehrman:
This ready for review. I've added a second commit to ensure that, when the first document is opened/created (i.e., from the no-doc state), the overlay offsets from the some-doc state are anticipated and set correctly in PS before the action document open/create action is played. The result is that it should always be the case that if you create or open a document in design space, PS will correctly place it in the center of our panel bounds without additional repositioning or scripting.

Major changes needed to accomplish this:

1. The individual panel components now responsible for calling `ui.updatePanelSizes` to declare _just_ their own panel size.
2. An `IconBar` component was factored out of the `PanelSet` component.
3. The `PanelSet` and `IconBar` report their UI sizes independently. The former also reports the number of columns that the given size corresponds to.
4. A new `ui.setOverlayOffesetsForFirstDocument` action is called before opening or creating documents. This action calculates the anticipated panel boundaries and sets the overlay offsets accordingly.
5. `File > Open` now executes `documents.open` (without a file path) instead of a native menu command. `File > New...` now executes `documents.createNewExtended` instead of a native menu commands. This was necessary so that those commands could first call `ui.setOverlayOffsetsForFirstDocument` before opening/creating the document. (Note that we can't do anything about the cases in which PS opens a document and only tells us about it afterward, like when you drag a file from the Finder into the empty window.)

src/js/jsx/DocumentHeader.jsx
```diff
@@ -57,6 +59,26 @@ define(function (require, exports, module) {
...
+         * Update the sizes of the panels.
+         *
+         * @private
+         */
+        _updatePanelSizes: function () {
+            var node = React.findDOMNode(this.refs.tabContainer),
+                headerHeight;
+
+            if (node) {
+                headerHeight = node.getBoundingClientRect().height;
```
volfied: Would clientHeight be faster here?

src/js/jsx/IconBar.jsx
```diff
@@ -0,0 +1,102 @@
...
+         * Update the sizes of the panels.
+         *
+         * @private
+         */
+        _updatePanelSizes: function () {
+            var node = React.findDOMNode(this),
+                iconBarWidth;
+
+            if (node) {
+                iconBarWidth = node.getBoundingClientRect().width;
```
volfied: ditto

src/js/jsx/DocumentHeader.jsx
iwehrman: Yes, but it's not accurate. It doesn't get the border, which adds non-zero width. This change was intentional. Also, since we're not trying to animate with this in a tight loop, I don't think the speed difference will be measurable.

src/js/actions/ui.js
```diff
@@ -257,6 +257,43 @@ define(function (require, exports) {
...
+        if (applicationStore.getDocumentCount() > 0) {
+            return Promise.resolve();
+        }
+
+        var preferencesStore = flux.store("preferences"),
+            uiStore = flux.store("ui"),
+            preferences = preferencesStore.getState(),
+            columnCount = 0;
+
+        if (preferences.get(uiStore.components.LAYERS_LIBRARY_COL), true) {
```
iwehrman: I amended the commit here and just below to use `true` as the default if the key is not found in the preferences table. This is how it works in `PanelSet`.

volfied:
Ok then, tested it a bit, and had one go-round with Ian.

PR #2345, opened 2015-09-04
CC Libraries - support open-for-edit for graphic assets.

### LINKED ISSUE #1408: Closing Library Linked Smart Object blinks zero-doc state
Open a file containing a LLSO
dclick on the LLSO layer in layers panel to open
cmd/ctrl + w to close

Result: blink zero-doc UI

-- @chadrolfs at [2015-05-07T23:59:12Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1408)

shaoshing:
**require changes from the adapter and generator-space plugin, see bottom for setting up generator-space plugin**

This PR implements open-for-edit for graphic assets. You can double click on graphic asset's thumbnail to open the asset for editing. Some notes:

* currently it only supports edit PSD file. The other file types require features that are not supported in DS yet.
* switching back to standard PS will break the upload-after-save feature, but switch from standard PS to DS will maintain the functionality.
* Close document will delete the temp files.

It also fix #1408 Closing Library Linked Smart Object blinks zero-doc state.

lib-edit

@volfied please review. 
@mcilroyc I made the change to import the file system commands, so could you review this PR?

#### setup the generator-space plugin

```
Run latest PS dev build (> 594)
```
or
```
cd PROJECTS
git clone git@github.com:adobe-photoshop/generator-spaces.git
cd generator-spaces
git checkout shao/filesystem
npm install

cd PS-DEV/Adobe\ Photoshop\ CC\ 2015.app
rm -rf Contents/Required/Plug-Ins/Generator/spaces.generate/
ln -s PROJECTS/generator-spaces Contents/Required/Plug-Ins/Generator/spaces.generate
RESTART PS-DEV
```

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -676,6 +676,40 @@ define(function (require, exports) {
     exportDocumentAssets.reads = [locks.JS_DOC, locks.JS_EXPORT];
     exportDocumentAssets.writes = [locks.GENERATOR];
     exportDocumentAssets.transfers = [promptForFolder, updateExportAsset];
+    
+    /**
+     * Copy file from one location to another.
+     *
+     * @param {string} sourcePath
+     * @param {string} targetPath
+     * @return {Promise} 
+     */
+    var copyFile = function (sourcePath, targetPath) {
+        if (!_exportService || !_exportService.ready()) {
+            return _setServiceAvailable.call(this, false);
+        }
+
+        return _exportService.copyFile(sourcePath, targetPath);
+    };
+    copyFile.reads = [];
+    copyFile.writes = [locks.GENERATOR];
+    
+    /**
+     * Remove files at specific locations.
+     *
+     * @param {Array.<string>} filePaths
+     * @return {Promise}
+     */
+    var removeFiles = function (filePaths) {
+        if (!_exportService || !_exportService.ready()) {
+            return _setServiceAvailable.call(this, false);
+        }
+
+        return _exportService.removeFiles(filePaths);
+    };
+    removeFiles.reads = [];
+    removeFiles.writes = [locks.GENERATOR];
+>>>>>>> Add export actions for libraries: copyFile & removeFiles.
 
     /**
      * After start up, ensure that generator is enabled, and then initialize the export service
@@ -796,5 +830,9 @@ define(function (require, exports) {
     exports.exportLayerAssets = exportLayerAssets;
     exports.exportDocumentAssets = exportDocumentAssets;
     exports.afterStartup = afterStartup;
+    
+    exports.copyFile = copyFile;
+    exports.removeFiles = removeFiles;
+
     exports.onReset = onReset;
 });
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -695,21 +695,20 @@ define(function (require, exports) {
     copyFile.writes = [locks.GENERATOR];
     
     /**
-     * Remove files at specific locations.
+     * Delete files at specific locations.
      *
      * @param {Array.<string>} filePaths
      * @return {Promise}
      */
-    var removeFiles = function (filePaths) {
+    var deleteFiles = function (filePaths) {
         if (!_exportService || !_exportService.ready()) {
             return _setServiceAvailable.call(this, false);
         }
 
-        return _exportService.removeFiles(filePaths);
+        return _exportService.deleteFiles(filePaths);
     };
-    removeFiles.reads = [];
-    removeFiles.writes = [locks.GENERATOR];
->>>>>>> Add export actions for libraries: copyFile & removeFiles.
+    deleteFiles.reads = [];
+    deleteFiles.writes = [locks.GENERATOR];
 
     /**
      * After start up, ensure that generator is enabled, and then initialize the export service
@@ -832,7 +831,7 @@ define(function (require, exports) {
     exports.afterStartup = afterStartup;
     
     exports.copyFile = copyFile;
-    exports.removeFiles = removeFiles;
+    exports.deleteFiles = deleteFiles;
 
     exports.onReset = onReset;
 });
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -45,18 +45,14 @@ define(function (require, exports) {
         layerActionsUtil = require("js/util/layeractions"),
         collection = require("js/util/collection"),
         layerActions = require("./layers"),
+        exportActions = require("./export"),
+        documentActions = require("./documents"),
         searchActions = require("./search/libraries"),
         shapeActions = require("./shapes"),
         typeActions = require("./type"),
+        preferencesActions = require("./preferences"),
         policyActions = require("./policy"),
-        preferencesActions = require("./preferences");
-
-    /**
-     * Preference name for storing the ID of the last selected library.
-     *
-     * @private
-     */
-    var _LAST_SELECTED_LIBRARY_ID_PREF = "lastSelectedLibraryID";
+        LibrarySyncStatus = require("js/models/library_sync_status");
 
     /**
      * For image elements, their extensions signify their representation type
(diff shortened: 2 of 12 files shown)


src/js/actions/libraries.js
```diff
@@ -499,6 +497,145 @@ define(function (require, exports) {
...
     renameAsset.writes = [locks.CC_LIBRARIES, locks.JS_LIBRARIES];
+    
+    /**
+     * Open the specified graphic asset for edit by creating a temp copy of the asset's document and open
+     * in Photoshop. The library store will listen to the document save and close event. 
+     * 
+     * @param {AdobeLibraryElement} element
+     * @return {Promise}
+     */
+    var openGraphicForEdit = function (element) {
```
volfied: You also need to transfer to this from `actions/superselect.js#editLayer`, when it's a library linked smart object. Double clicking a such layer will work that way.

src/js/actions/export.js
```diff
@@ -643,6 +643,39 @@ define(function (require, exports) {
...
+    copyFile.reads = [];
+    copyFile.writes = [locks.GENERATOR];
+    
+    /**
+     * Remove files at specific locations.
+     *
+     * @param {Array.<string>} filePaths
+     * @return {Promise}
+     */
+    var removeFiles = function (filePaths) {
```
volfied: Can you call this delete file? I don't think remove is right for a FS API.

src/js/actions/libraries.js
```diff
@@ -499,6 +497,145 @@ define(function (require, exports) {
...
+        if (representation.type !== _REP_PHOTOSHOP_TYPE) {
+            // Due to the limitation in DS, we only support editing psd file.
+            log.debug("[CC Lib] unsupported graphic type: " + representation.type);
+            return Promise.resolve();
+        }
+        
+        var tempFilePath,
+            tempPreviewPath;
+        
+        return Promise
```
volfied: Seems like an unnecessary wrap with Promise here.

src/js/actions/libraries.js
```diff
@@ -499,6 +497,145 @@ define(function (require, exports) {
...
+                if (graphicEditStatus) {
+                    tempFilePath = graphicEditStatus.documentPath;
+                    tempPreviewPath = graphicEditStatus.previewPath;
+                    return;
+                }
+                
+                // Otherwise, create a temp copy of the element for edit.
+                return _getTempPaths()
+                    .bind(this)
+                    .then(function (paths) {
```
volfied: yuck, nesting promises is bad.

src/js/actions/libraries.js
```diff
@@ -499,6 +497,145 @@ define(function (require, exports) {
...
+            .then(function () {
+                var previewSetting = {
+                    path: tempPreviewPath,
+                    width: RENDITION_GRAPHIC_SIZE,
+                    height: RENDITION_GRAPHIC_SIZE
+                };
+                
+                // Open the graphic asset's temp file and tell PS to generate an updated 
+                // preview whenever the file is saved
+                return this.flux.actions.documents.open(tempFilePath, {
```
volfied: You want to transfer here, enqueueing one action from another is a good way to get deadlocks.

volfied: Also, with enqueuing, by the time next .then rolls around, you're not guaranteed that the document will be open.

src/js/actions/libraries.js
```diff
@@ -499,6 +497,145 @@ define(function (require, exports) {
...
+                        documentID: documentID,
+                        documentPath: tempFilePath,
+                        previewPath: tempPreviewPath,
+                        element: element
+                    };
+
+                return this.dispatchAsync(events.libraries.OPEN_GRAPHIC_FOR_EDIT, payload);
+            })
+            .catch(function (e) {
+                log.debug("[CC Lib] " + e);
```
volfied: Maybe a log.warn instead, and a better `[header]` letting us know it's the openGraphicForEdit function.

src/js/stores/library.js
```diff
@@ -214,6 +253,144 @@ define(function (require, exports, module) {
...
+                return;
+            }
+            
+            var element = editStatus.element,
+                documentPath = editStatus.documentPath,
+                previewPath = editStatus.previewPath;
+            
+            editStatus.isUpdatingContent = true;
+            
+            this.flux.actions.libraries.updateGraphicContent(element, documentPath, previewPath)
```
volfied: Enqueueing actions from stores are a bad idea, can you transfer to this from the save action somehow? Paging @iwehrman for his opinion/idea as well.

src/js/stores/library.js
```diff
@@ -214,6 +253,144 @@ define(function (require, exports, module) {
...
+         * Handle save document. If an element is associated with the document, we update the element's 
+         * content and preview image.
+         *
+         * @private
+         * @param {object} payload
+         * @param {number} payload.documentID 
+         * @param {string} payload.path - document path
+         */
+        _handleSaveDocument: function (payload) {
+            var documentID = payload.documentID,
```
volfied: Do these need to `waitFor` the document store?

src/js/stores/library.js
```diff
@@ -270,6 +447,31 @@ define(function (require, exports, module) {
...
+         * Find element in all libraries by its ID.
+         *
+         * @private
+         * @param {string} elementID
+         * @return {?AdobeLibraryElement}
+         */
+        _findElementByID: function (elementID) {
+            var libraries = this._libraryCollection.libraries;
+            
+            for (var i = 0; i < libraries.length; i++) {
```
volfied: if you pass in libraryID as well, you can use `AdobeLibraryCollection.getLibraryByID` and `AdobeLibraryComposite.getElementByID`

Although, what's the best to do is, instead of the ID, store the element reference, and use `CCLibraries.ResolveElementReference` to directly get the element.

volfied:
Hey @shaoshing, this is impressive work! I have couple nits and worries. After oyu address them, I'll give it one more review.

src/js/actions/libraries.js
volfied: Nesting promises isn't that bad, from waht I've learned... but I don't like this all `then` column here, becauseit does not return anything, and only acts in side effects. So that means paths can be returned to the next then in the chain, and be handled in there synchronously.

src/js/actions/libraries.js
shaoshing: Added support for cloud linked smart object.

src/js/actions/libraries.js
shaoshing: I agree with you, and I updated the nested promise to reduce unnecessary `then` calls.

src/js/stores/library.js
shaoshing: we don't need a `waitFor` here as it does not rely on the document store

src/js/stores/preferences.js
```diff
@@ -141,17 +141,24 @@ define(function (require, exports, module) {
...
+            this.setPreference(payload.key, payload.value);
+        },
 
+        /**
+         * Set a single preference.
+         *
+         * @param {sring} key
+         * @param {*} value
+         */
+        setPreference: function (key, value) {
```
shaoshing: I open the access to the Preference#setPreference, so that we can avoid calling preference actions in the library store. Please let me know if there is a better way to achieve this.

shaoshing:
Hi @volfied , I made some changes to address your comments, including store edit status in preference. Please review again. You can test this with latest jenkin build (> 594).

src/js/actions/superselect.js
```diff
@@ -309,23 +310,31 @@ define(function (require, exports) {
...
+                    
+                resultPromise = this.transfer(libraryActions.openGraphicForEdit, element);
+            } else {
+                // For linked smart objects, this option shows the fix broken link dialog if the link is broken
+                
+                var editOptions = {
+                    interactionMode: descriptor.interactionMode.DISPLAY
+                };
+
+                _logSuperselect("edit_smart_object");
```
volfied: This line should be called for all smart object types (or better yet, add a `"edit_cloud_object"` logging for CLSOs)

src/js/stores/preferences.js
volfied: Paging @iwehrman as he'll know preferences code better.

volfied: Reading your code a bit more, I've noticed that you're calling this function from the library store. Those store handlers are reacting to some dispatch, so what you should do instead is install handlers for those events in this store as well, and edit this store's own data structures only through those handlers.

Or you can enqueue/transfer to setPreference action from the actions that are dispatching things (if you need library store to react first, use dispatchAsync, then grab whatever data you need from the library store and enqueue a set preference action.

shaoshing:
Hi @volfied , I've refactored the implementation to achieve one-way data flow between action and store. Now all the actions calls in the library store has been removed. Please review again and let me know your thoughts. Thank you.

src/js/actions/superselect.js
shaoshing: Added log for CLSO.

PR #2346, opened 2015-09-04
Implement clear guides menu item and action

volfied:
Requires https://github.com/adobe-photoshop/spaces-adapter/pull/145

Adds View>Clear Guides button that clears out all guides in current document.

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -313,6 +313,36 @@ define(function (require, exports) {
     setGuide.transfers = [resetGuidePolicies, deleteGuide];
 
     /**
+     * Clears all the guides in the given document
+     *
+     * @param {{id: number}} document Document model or an object containing document ID
+     * @return {Promise}
+     */
+    var clearGuides = function (document) {
+        if (document === undefined) {
+            var appStore = this.flux.store("application");
+
+            document = appStore.getCurrentDocument();
+        }
+
+        var payload = {
+                documentID: document.id
+            },
+            clearObj = documentLib.clearGuides(documentLib.referenceBy.id(document.id)),
+            dispatchPromise = this.dispatchAsync(events.document.history.nonOptimistic.GUIDES_CLEARED, payload),
+            clearPromise = descriptor.playObject(clearObj);
+
+        return Promise.join(dispatchPromise, clearPromise)
+            .bind(this)
+            .then(function () {
+                return this.transfer(resetGuidePolicies);
+            });
+    };
+    clearGuides.reads = [];
+    clearGuides.writes = [locks.JS_DOC];
+    clearGuides.transfers = [resetGuidePolicies];
+
+    /**
      * Re-gets the guides of the given document and rebuilds the models
      *
      * @param {Document=} document Default is active document
@@ -416,6 +446,7 @@ define(function (require, exports) {
     exports.deleteGuide = deleteGuide;
     exports.resetGuidePolicies = resetGuidePolicies;
     exports.queryCurrentGuides = queryCurrentGuides;
+    exports.clearGuides = clearGuides;
 
     exports.beforeStartup = beforeStartup;
     exports.onReset = onReset;
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -315,7 +315,7 @@ define(function (require, exports) {
     /**
      * Clears all the guides in the given document
      *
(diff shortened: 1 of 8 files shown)


src/js/actions/guides.js
```diff
@@ -313,6 +313,36 @@ define(function (require, exports) {
     setGuide.transfers = [resetGuidePolicies, deleteGuide];
 
     /**
+     * Clears all the guides in the given document
+     *
+     * @param {{id: number}} document Document model or an object containing document ID
+     * @return {Promise}
+     */
+    var clearGuides = function (document) {
+        if (document === undefined) {
```
mcilroyc: this is purely a style nit, but seems like this could simplified to doc = doc || this.flux....getCurrentDOcument()

src/js/actions/guides.js
```diff
@@ -313,6 +313,36 @@ define(function (require, exports) {
     setGuide.transfers = [resetGuidePolicies, deleteGuide];
 
     /**
+     * Clears all the guides in the given document
+     *
+     * @param {{id: number}} document Document model or an object containing document ID
```
mcilroyc: JSDoc should reflect this param's optional nature

src/js/actions/guides.js
```diff
@@ -313,6 +313,36 @@ define(function (require, exports) {
...
+
+            document = appStore.getCurrentDocument();
+        }
+
+        var payload = {
+                documentID: document.id
+            },
+            clearObj = documentLib.clearGuides(documentLib.referenceBy.id(document.id)),
+            dispatchPromise = this.dispatchAsync(events.document.history.nonOptimistic.GUIDES_CLEARED, payload),
+            clearPromise = descriptor.playObject(clearObj);
```
mcilroyc: another pure style nit: seems weird that the play-related lines are adjacent.

src/js/actions/guides.js
```diff
@@ -313,6 +313,36 @@ define(function (require, exports) {
...
+
+        return Promise.join(dispatchPromise, clearPromise)
+            .bind(this)
+            .then(function () {
+                return this.transfer(resetGuidePolicies);
+            });
+    };
+    clearGuides.reads = [];
+    clearGuides.writes = [locks.JS_DOC];
```
mcilroyc: this should lock PS_DOC too?  (strictly speaking, even though the transfer probably handles glosses over this fact)

src/js/stores/document.js
```diff
@@ -1094,6 +1095,21 @@ define(function (require, exports, module) {
...
+         * Clears all the guides
+         *
+         * @private
+         * @param {{documentID: number}} payload
+         */
+        _handleGuidesCleared: function (payload) {
+            var documentID = payload.documentID,
+                document = this._openDocuments[documentID];
+
+            var nextDocument = document.set("guides", Immutable.List());
```
mcilroyc: this could throw if the doc isn't found?  Maybe `setIn` would obviate this?

src/js/actions/guides.js
mcilroyc: Furthermore, why not just `Document=` ?   It currently isn't even being used this way, right?  And in the "future" we'd probably have a full blown Document model.  I dunno.

mcilroyc:
One actual JSDoc nit (`=`), one possible bug, and some uber-nit questions. back to @volfied

src/js/actions/guides.js
volfied: I had a line in a past PR that way and @iwehrman suggested I use === undefined check instead...

I like the explicitness of this style.

src/js/actions/guides.js
volfied: I was matching the other guide functions (where we receive the document ID from Photoshop and don't bother to grab the model), but this can be changed for clearGuides since it's an action and not just a handler.

src/js/actions/guides.js
volfied: I don't see anything wrong with that.

src/js/actions/guides.js
volfied: Good catch.

src/js/stores/document.js
volfied: All the functions in this store can throw if doc isn't found. I think it is safe to assume we will not intentionally send null documents to the store.

volfied:
@mcilroyc Thank you for review, addressed some nits, and answered your questions.

Issue #2273, opened 2015-08-29
Exporting empty artboard or empty layer causes error


ktaki:
Export Failed for asset 0 of layerID 3, documentID 401, with error Cannot call method 'info' of undefinedTypeError: Cannot call method 'info' of undefined

mcilroyc:
The terribleness of the above error is being addressed in a generator plugin fix: https://github.com/adobe-photoshop/generator-spaces/commit/2ef03036af896fd55edc795b68719bf9349d7bf5

A in-progress DS PR will fix the way this is handled (or better yet, prevented).

mcilroyc:
Note to self: empty artboards, and empty text layer (eg., a single whitespace text) is reported by our internal model as non-empty, yet generator complains in both cases that they are empty.

mcilroyc:
As of yesterday (jenkins dev mac 594), when an export fails, the underlying message from generator plugin is more meaningful.  This, combined with the above-merged PR which prevents empty artboards and layer from *attempting* to export, addresses this issue.

There may be situations where the export fails and we can not have predicted it.  We should at least be handling it somewhat correctly now, and I'd like to log those issues individually as they occur.

ktaki:
I still see the sam error. Create a new artboard doc. Select the Artboard 1 in layers panel. Click on export icon, then export it.

mcilroyc:
The case you just described *seems to me* to be a very unlikely edge case that I will have to handle separately.  Namely the first artboard with the empty placeholder layer.  (Who's trying to export an empty artboard!?)

mcilroyc:
note to future self: I think the method `layerstructure.isEmptyGroup` could be safely modified to check for groups that (recursively) contain empty (zero bounds area) layers.  But this touches several features, so I don't think it is worth the risk pre-MAX

iwehrman:
The console error is really more of a warning w.r.t. Design Space at this point, and since we can't fully predict Generator errors I don't think this is really a bug. However, I agree with @mcilroyc's wish to extend `isEmptyGroup` to recursively check for empty subgroups. I also agree this isn't worth trying to do for M3, so I'm kicking it to M4.

iwehrman:
Actually, can't you just do `bounds = document.layers.childBounds(layer); empty = !bounds || bounds.empty`? That's recursive as all get-out.

PR #2348, opened 2015-09-04
Export panel and its operations should filter out unsupported layers

mcilroyc:
- Only show and operate on "supported" subset of selected layers
- Currently unsupported: background layers, empty artboards, empty groups, zero-bound layers
- If layers are selected, but none are supported, a message is shown
- If layers are selected, but there are no common assets, a message is shown
- Uses @placegraphichere 's Panel Message pattern
- Includes some minor react housekeeping
- Review note: Added new function `filterExportable` to `layerstructure` (instead of documentExport) so that it would have easy access to `isEmptyGroup`

This addressees part of #2273
(A separate plugin change will make the error message better when we DO request export of layers that generator pukes on)

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -375,9 +375,15 @@ define(function (require, exports) {
      * Adds an asset, or assets to the end of the document asset list, or that of a layer or layers
      * If props not provided, choose the next reasonable scale and create an otherwise vanilla asset
      *
+     * Recognizes an empty list of layers as implying doc-level export
+     *
+     * Layer-level exports:
+     * Will only add assets to those layers which are supported.
+     * Adds assets after the last uniform asset shared by the given set of layers.
+     *
      * @param {Document} document
      * @param {DocumentExports} documentExports
-     * @param {Immutable.List.<Layer>=} layers
+     * @param {Immutable.List.<Layer>} layers
      * @param {object|Array.object=} props asset-like object, or an array thereof.  if not supplied, picks next scale
      * @return {Promise}
      */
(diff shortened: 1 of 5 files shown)


src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -200,56 +200,50 @@ define(function (require, exports, module) {
...
+                    <ExportAssetFace
+                        document={document}
+                        layers={layers}
+                        index={k}
+                        key={key}
+                        faceKey={key}
+                        exportAssets={i} />
                 );
-            }
+            }, this).toArray();
```
iwehrman: No need to convert into an array if it's just going into a React component.

src/nls/root/strings.json
```diff
@@ -379,7 +379,9 @@
         "TITLE_SCALE": "Scale",
         "TITLE_SUFFIX": "Suffix",
         "TITLE_SETTINGS": "Settings",
-        "EXPORT_DOCUMENT_FILENAME": "document"
+        "EXPORT_DOCUMENT_FILENAME": "document",
+        "ONLY_UNSUPPORTED_LAYERS_SELECTED": "Unsupported Layer",
+        "NO_ASSETS": "Nothing configured"
```
iwehrman: At the very least I'd change this to "No exports configured." But, ideally, this text would also explain to the user _how_ to configure some exports. I'm unsure about how to word this in a completely satisfactory way, though.

src/nls/root/strings.json
```diff
@@ -379,7 +379,9 @@
         "TITLE_SCALE": "Scale",
         "TITLE_SUFFIX": "Suffix",
         "TITLE_SETTINGS": "Settings",
-        "EXPORT_DOCUMENT_FILENAME": "document"
+        "EXPORT_DOCUMENT_FILENAME": "document",
+        "ONLY_UNSUPPORTED_LAYERS_SELECTED": "Unsupported Layer",
```
iwehrman: Maybe "No supported layers selected" or "No exportable layers selected"?

src/js/models/layerstructure.js
```diff
@@ -801,6 +801,19 @@ define(function (require, exports, module) {
...
     /**
+     * Given a set of layers, return only those layers which will support being exported
+     * This excludes background layers and empty layers
+     *
+     * @param {Immutable.Iterable.<Layer>} layers
+     * @return {Immutable.Iterable.<Layer>}
+     */
+    Object.defineProperty(LayerStructure.prototype, "filterExportable", objUtil.cachedLookupSpec(function (layers) {
+        return layers.filterNot(function (layer) {
+            return !layer.bounds || layer.bounds.empty || layer.isBackground || this.isEmptyGroup(layer);
```
iwehrman: Are adjustment layers supported? I'd actually explicitly list all the layer kinds that aren't supported here. We're going to have to do something better than we're currently doing overall for unsupported documents...

src/js/actions/export.js
```diff
@@ -374,9 +374,15 @@ define(function (require, exports) {
      * Adds an asset, or assets to the end of the document asset list, or that of a layer or layers
      * If props not provided, choose the next reasonable scale and create an otherwise vanilla asset
      *
+     * Recognizes an empty list of layers as implying doc-level export
```
iwehrman: Not a huge fan of using an empty list to denote doc-level exports, but maybe it makes sense in the bigger picture.

mcilroyc: yes, I know what you mean.  Although I tried to minimize the places that uses this ... and I think it makes it easier here for React to just shovel the selected layers in here and let the action understand "empty selected" .

src/js/jsx/sections/export/ExportList.jsx
mcilroyc: johnny-carson

src/js/models/layerstructure.js
mcilroyc: good calll - I'll add adj layers here.  I'm not aware at this juncture of others to add.

mcilroyc:
Fixed a few of the nits.  I assume @placegraphichere may also want to massage the wording of the two special-case messages in ExportPanel.  back to @iwehrman

mcilroyc:
oh, just FYI, I did rebase to master and fixed a minor conflict in classnames being added to some sub-components in ExportList

Issue #2347, opened 2015-09-04
Unhandled exception in search for current documents


iwehrman:
1. Open a new document
2. cmd-F to search
3. type "doc"
4. Hit enter to select "Filter by current documents"
5. Unhandled exception:

```
TypeError: Cannot read property 'split' of undefined
    at Fluxxor.createStore.handleExecute (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/stores/search.js:177:33)
    at eval (file://./js/jsx/Search.jsx:76:33)
From previous event:
    at React.createClass._handleOption (file://./js/jsx/Search.jsx:74:18)
    at React.createClass._handleChange (file://./js/jsx/search/SearchBar.jsx:360:32)
    at React.createClass._handleSelectChange (file://./js/jsx/shared/Datalist.jsx:302:48)
    at React.createClass._handleSelectClick (file://./js/jsx/shared/Datalist.jsx:328:22)
    at React.createClass._handleSelectClose (file://./js/jsx/shared/Datalist.jsx:347:22)
    at React.createClass.close (file://./js/jsx/shared/Select.jsx:338:24)
    at React.createClass._handleInputKeyDown (file://./js/jsx/shared/Datalist.jsx:270:28)
    at React.createClass._handleKeyDown (file://./js/jsx/shared/TextInput.jsx:274:28)
```

iwehrman:
Fixed by #2349.

PR #2349, opened 2015-09-04
Always set the ID on Select#setSelected

volfied:
Problem: updateAutoFill would not match any filter, and set the `state.id` of Datalist to null... Then we would use selectExtreme to select the first result. And as far as Select knew, it was already at that ID, so it didn't need to set it's own state again. So that's why it would work when we typed 'do', but not 'doc'.

Always selecting it makes sure that states are correct.

Addfresses #2347

### DIFF
--- a/src/js/jsx/shared/Select.jsx
+++ b/src/js/jsx/shared/Select.jsx
@@ -230,7 +230,7 @@ define(function (require, exports, module) {
          * @param {string} id
          */
         _setSelected: function (id) {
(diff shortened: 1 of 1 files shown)


mcilroyc:
yowzers.  I spent some time a few days ago staring at this code wrt to the DataLists in export panel.  There's a chance this affects a tweak I made in DataList in a local branch.  I'll dig a little deeper, this is just a quick note to that effect.

mcilroyc:
Ok.  seems compatible with my change and that's all that matters!  (seem OK in general)

PR #2350, opened 2015-09-04
Toggle overlays on activationChanged events

volfied:
@jsbache pointed this out. When PS loses focus, we should toggle our overlays so the guides overlay doesn't just stay lit. We re-enable them on app focus.

activationchanged


mcilroyc:
@volfied am i doing it wrong? http://cl.ly/image/2Q2w1E2L2D2x

volfied:
Thank you for the catch @mcilroyc, patched it.

PR #2351, opened 2015-09-05
Disable keyboard policies while OS folder chooser is open

mcilroyc:
...and then re-enabled them afterwards.

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -551,12 +551,12 @@ define(function (require, exports) {
                 }
             )
             .finally(function () {
-                return this.transfer(policyActions.reenabledKeyboardPolicies);
+                return this.transfer(policyActions.reenableKeyboardPolicies);
             });
     };
     promptForFolder.reads = [];
     promptForFolder.writes = locks.ALL_LOCKS;
-    promptForFolder.transfers = [policyActions.disableKeyboardPolicies, policyActions.reenabledKeyboardPolicies];
+    promptForFolder.transfers = [policyActions.disableKeyboardPolicies, policyActions.reenableKeyboardPolicies];
 
     /**
      * Export all layer assets for the given document for which export has been enabled (layer.exportEnabled)
@@ -662,13 +662,14 @@ define(function (require, exports) {
 
         // prompt for folder and then export to the result.
         // resolve immediately if no folder is returned
-        return _exportService.promptForFolder(_lastFolderPath || "~").then(function (baseDir) {
-            return baseDir ? exportToDir(baseDir) : Promise.resolve();
-        });
+        return this.transfer(promptForFolder)
+            .then(function (baseDir) {
+                return baseDir ? exportToDir(baseDir) : Promise.resolve();
+            });
     };
     exportDocumentAssets.reads = [locks.JS_DOC, locks.JS_EXPORT];
     exportDocumentAssets.writes = [locks.GENERATOR];
-    exportDocumentAssets.transfers = [updateExportAsset];
+    exportDocumentAssets.transfers = [promptForFolder, updateExportAsset];
 
     /**
      * After start up, ensure that generator is enabled, and then initialize the export service
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -43,7 +43,8 @@ define(function (require, exports) {
         log = require("js/util/log"),
         global = require("js/util/global"),
         ExportAsset = require("js/models/exportasset"),
-        ExportService = require("js/util/exportservice");
+        ExportService = require("js/util/exportservice"),
+        policyActions = require("js/actions/policy");
 
     /**
      * An instance of the ExportService utility used to communicate with the generator plugin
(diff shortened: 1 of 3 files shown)


src/js/actions/policy.js
```diff
@@ -189,6 +213,61 @@ define(function (require, exports) {
...
+    };
+    disableKeyboardPolicies.reads = [];
+    disableKeyboardPolicies.writes = [locks.PS_APP, locks.JS_POLICY];
+
+    /**
+     * Re-enabled cached keyboard policies
+     *
+     * @return {Promise}
+     */
+    var reenabledKeyboardPolicies = function () {
```
iwehrman: Actions are verbs, so I think this should be called `reenableKeyboardPolicies`! That would also match `disableKeyboardPolicies` better.

src/js/actions/policy.js
```diff
@@ -35,40 +35,73 @@ define(function (require, exports) {
...
+     * The previous policies during temporary policies disabling
      *
-     * @private 
-     * @param {string} kind A value defined in PolicyStore.eventKind
-     * @param {Array.<KeyboardEventPolicy>} policies
-     * @return {Promise}
+     * @type {?EventPolicySet}
      */
-    var _addPolicies = function (kind, policies) {
+    var _cachedKeyboardPolicies = null;
```
iwehrman: This state needs to be reset to null in `onReset`. Sometimes I think we should actually bar state in actions entirely... Another day!

src/js/actions/policy.js
```diff
@@ -35,40 +35,73 @@ define(function (require, exports) {
...
+     * Install a new policy list.
+     *
+     * @private 
+     * @param {string} kind A value defined in PolicyStore.eventKind
+     * @param {Array.<KeyboardEventPolicy>} policies
+     * @return {Promise}
+     */
+    var _addPolicies = function (kind, policies) {
+        if (_cachedKeyboardPolicies && kind === PolicyStore.eventKind.KEYBOARD) {
+            var msg = "Trying to add keyboard policies while there is a cached state: " + kind ;
```
iwehrman: Typo :tractor: `there`. Also, weird semicolon space.

src/js/actions/policy.js
```diff
@@ -82,24 +115,15 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var _removePolicies = function (kind, id, commit) {
+        if (_cachedKeyboardPolicies && kind === PolicyStore.eventKind.KEYBOARD) {
+            var msg = "Trying to remove keyboard policies while there is a cached state: " + kind + " : " + id;
```
iwehrman: Ditto.

src/js/actions/policy.js
```diff
@@ -189,6 +213,61 @@ define(function (require, exports) {
...
+            kind = PolicyStore.eventKind.KEYBOARD;
+
+        _cachedKeyboardPolicies = policyStore.getPolicies(kind);
+
+        return adapterUI.getKeyboardPropagationMode()
+            .bind(this)
+            .then(function (mode) {
+                _cachedKeyboardMode = mode;
+                policyStore.clearPolicies(kind);
+            }).then(function () {
```
iwehrman: `.then` on a new line.

src/js/actions/policy.js
```diff
@@ -189,6 +213,61 @@ define(function (require, exports) {
...
     /**
+     * Temporarily disable keyboard policies, caching the previous state
+     *
+     * @return {Promise}
+     */
+    var disableKeyboardPolicies = function () {
+        var policyStore = this.flux.store("policy"),
+            kind = PolicyStore.eventKind.KEYBOARD;
+
+        _cachedKeyboardPolicies = policyStore.getPolicies(kind);
```
iwehrman: Maybe in the spirit of making this hard to screw up, check here to make sure there aren't already cached keyboard policies? I.e., no double disabling.

src/js/actions/policy.js
```diff
@@ -189,6 +213,61 @@ define(function (require, exports) {
...
+                .bind(this)
+                .then(function () {
+                    return _syncPolicies.call(this, kind);
+                })
+                .finally(function () {
+                    _cachedKeyboardPolicies = null;
+                    _cachedKeyboardMode = null;
+                });
+        } else {
+            return Promise.resolve();
```
iwehrman: I'd call this an error. Also, aesthetically, I'd get this error case out of the way early on in the method.

src/js/stores/policy.js
```diff
@@ -89,6 +89,35 @@ define(function (require, exports, module) {
...
+         */
+        clearPolicies: function (kind) {
+            this._policySets[kind] = new EventPolicySet();
+        },
+
+        /**
+         * Get the entire PolicySet of the given kind
+         *
+         * @param {string} kind
+         * @return {?EventPolicySet} [description]
```
iwehrman: `[description]

src/js/actions/export.js
```diff
@@ -533,6 +534,31 @@ define(function (require, exports) {
...
+                function (e) {
+                    throw new Error("Failed to stash keyboard policies prior to opening OS dialog: " + e.message);
+                }
+            )
+            .finally(function () {
+                return this.transfer(policyActions.reenabledKeyboardPolicies);
+            });
+    };
+    promptForFolder.reads = [];
+    promptForFolder.writes = locks.ALL_LOCKS;
```
iwehrman: This makes me a little uncomfortable, but I guess I can't think either of a problem it will cause or a "better" solution. It does seem like if keyboard policies are disabled we shouldn't let other actions run concurrently. So, I guess :+1:

iwehrman:
Just a few nits. Looks good! Thanks again!

mcilroyc:
Fixed nits.  Removed "review only" tag.  Back to @iwehrman

src/js/actions/policy.js
```diff
@@ -189,6 +213,66 @@ define(function (require, exports) {
...
+        var policyStore = this.flux.store("policy"),
+            kind = PolicyStore.eventKind.KEYBOARD;
+
+        _cachedKeyboardPolicies = policyStore.getPolicies(kind);
+
+        return adapterUI.getKeyboardPropagationMode()
+            .bind(this)
+            .then(function (mode) {
+                _cachedKeyboardMode = mode;
+                policyStore.clearPolicies(kind);
```
iwehrman: No reason for this to be in it's own `.then` handler.

src/js/actions/policy.js
```diff
@@ -204,11 +288,28 @@ define(function (require, exports) {
     beforeStartup.reads = [];
     beforeStartup.writes = [locks.PS_APP, locks.JS_POLICY];
 
+    /**
+     * Reset
+     *
+     * @return {Promise}
+     */
+    var onReset = function () {
+        this._cachedKeyboardPolicies = null;
```
iwehrman: These should just be variable assignments, not property assignments.

iwehrman:
One nit, one bug, one beer.

src/js/actions/policy.js
mcilroyc: good thing it's the weekend so you can't yell at me :sunny:

mcilroyc:
fixed new nit, and :bug: ... squashed over previous commit.

PR #2352, opened 2015-09-05
Place CC Libraries graphic assets asynchronously

iwehrman:
This changes the execution of the CC Libraries `placeElement` action to be considered "asynchronous" by the adapter, which allows CEF to continue to run while the command is being played, even while in the smart-object placement tracker. A side effect of playing commands in this state is that they can generate additional events, and so it was also necessary to convert the `placeEvent` handler into an action which calmly, safely declines to add layers to the model which have already been added. It also seems to be necessary to explicitly set the required key propagation policies when playing `placeElement` asynchronously.

This fully addresses #2177, and also removes a few `Promise.delay` superhacks that only worked if you were sufficiently lucky.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -731,6 +731,38 @@ define(function (require, exports) {
     toggleSmartGuidesVisibility.writes = [locks.JS_DOC, locks.PS_DOC];
 
     /**
+     * Handler for the placeEvent notification. If the event contains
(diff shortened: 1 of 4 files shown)


iwehrman:
Requires a pgdev Jenkins build from at least the late afternoon of Friday the 4th.

mcilroyc:
looks good to my eye, but assigning it to @volfied for library / policy opinion.

volfied:
LGTM

PR #2357, opened 2015-09-08
Set correct policy modifier for borders on Win

volfied:
Realized this breaks things on Windows.

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -39,6 +39,7 @@ define(function (require, exports) {
         locks = require("js/locks"),
         policy = require("./policy"),
         shortcuts = require("./shortcuts"),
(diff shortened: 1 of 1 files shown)


src/js/actions/tools.js
```diff
@@ -152,7 +153,8 @@ define(function (require, exports) {
             // The resize rectangles are roughly 8 points radius
             inset = 4,
             // In case of artboards, we have no rotate, so we can stay within the border
-            outset = artboards ? inset : 27;
+            outset = artboards ? inset : 27,
+            distortModifier = system.isMac ? { command: true } : { control: true };
```
mcilroyc: seems like this pattern gets repeated a few times throughout DS, so in the **future** we may consider abstracting to ~ `getSuper()`

Issue #2295, opened 2015-09-01
Type datalist sometimes doesn't populate or expand


chadrolfs:
Datalist doesn't populate.

iwehrman:
WFM at f39ac6691e40bef6321a052c3ef4f3b83f755c8f with Mac pgdev.560. Anything in the console?

iwehrman:
Also: can you `git bisect` it?

volfied:
I don't see this either. Over to @chadrolfs to re-confirm.

chadrolfs:
The cause is #2301.  When you get into this state, the datalist doesn't populate.  Otherwise, it functions correctly.

chadrolfs:
I've been hitting this from time to time.  Not sure what the reliable repro case is.  Here's an instance I ran into it but repeating the steps didn't repro:
https://www.dropbox.com/s/equl4pg2c7dfhnc/Type-Commit-Datalist.mov?dl=0
could maybe related to #2313?  I dunno

chadrolfs:
seems easier to repro on Win

iwehrman:
Nothing in the console?

ktaki:
No errors reported on console. With #2338, I placed the full log at:
smb://sjshare/photoshop/Transfer/ktaki/For_IanW/DevToolConsoleLog.zip

Repeating "Enqueuing/Executing/Finished action policy.addKeyboardPolicies" when I was trying to click on the menu. It looks like stuck in the editing of the font name whenever click on the typeface menu.

iwehrman:
Here are some concrete repro steps:
1. Open this file from @zoranagee: http://cloud.wehrman.org/0T2e0d3m1k0u
2. Make sure the empty text layer is selected under Artboard 1 > Header > Text > ...
3. Edit the text layer. (I did so by double-clicking that layer's text icon).
4. Clicking the Typeface datalist does not open the the drop-down.

I also don't see anything on the console.

Also: I see the same behavior with the other text layers in that folder, so I don't think it has anything to do with the empty text layer in particular.

volfied:
Question to all reproducers: Do you have font sync enabled?

ktaki:
Font Sync? Yes.

In my case, it can be reproduced with the following steps.
1. New doc, select type tool, click on canvas, click on Tyepface menu. The menu appears. Commit edit.
2. New doc, select type tool, click on canvas, click on Tyepface menu. The menu does not appear.
[...]

PR #2366, opened 2015-09-08
Move typefaces for Datalist calculation to the font store

volfied:
Because we were setting `state.typefaces` in the shouldComponentUpdate of Type.jsx, if the two postScriptMaps between two instances of a Type.jsx were equal, we wouldn't recompute the new state.typefaces, and it would be passed `undefined` into the Typeface datalist, causing #2295, not rendering the Dialog.

Now we calculate the typefaces in the font store, this is not ideal, as they're very UI specific... but at the same time this is the sanest way of doing this.

@iwehrman Since we chatted about this, please review the fix.


iwehrman:
:+1::+1::+1:

Issue #2207, opened 2015-08-25
Previously saved docs with shapes open with Stroke: mixed.


chadrolfs:
attempting to change stroke results in failure and `Unhandled rejection Error: Unable to set stroke properties of layer: inner shadow`

chadrolfs:
let me know if you need a test file.

chadrolfs:
This occurs on release so it is not a result of the new color picker design

volfied:
I reproed it with one of my old documents, `AGMStrokeStyleInfo` property is not available for some layers!

volfied:
Looks like AGMStrokeStyleInfo is not available when the stroke is disabled. Will fix our data structures

volfied:
I messed up during singularizing strokes and fills... fixing.

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
You can now change Stroke attributes without errors but previously saved doc with shapes open with Stroke: mixed

Leaving bug open and lowering priority

mcilroyc:
@chadrolfs please see barkin's write-up in #2502

PR #2372, opened 2015-09-09
Re-add addStroke and wire it correctly to stroke being added

volfied:
I messed up during singularizing strokes and fills and removed these necessary methods and wired them to the wrong store handler. This undoes the damage.

Addresses #2207

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -28,6 +28,7 @@ define(function (require, exports, module) {
 
     var Layer = require("./layer"),
         LayerNode = require("./layernode"),
+        Stroke = require("./stroke"),
         Bounds = require("./bounds"),
         Radii = require("./radii");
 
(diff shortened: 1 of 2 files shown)


src/js/models/layerstructure.js
```diff
@@ -1317,6 +1318,37 @@ define(function (require, exports, module) {
...
+    LayerStructure.prototype.addStroke = function (layerIDs, strokeStyleDescriptor) {
+        var isList = Immutable.List.isList(strokeStyleDescriptor);
+       
+        var getStroke = function (index) {
+            return isList ?
+                Stroke.fromStrokeStyleDescriptor(strokeStyleDescriptor.get(index)) :
+                Stroke.fromStrokeStyleDescriptor(strokeStyleDescriptor);
+        };
+       
+        var nextLayers = Immutable.Map(layerIDs.reduce(function (map, layerID, index) {
```
mcilroyc: side note: I now believe (with very little actual evidence to back it up) that it is more efficient to map things to an array of two-element arrays, then create an Immutable.Map (which accepts that as a constructor format), and then merge that in to the main Map.  See here: https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/models/documentexports.js#L261    food for thought anyway.

src/js/stores/document.js
```diff
@@ -742,6 +742,24 @@ define(function (require, exports, module) {
         },
 
         /**
+         * Adds a stroke to the specified document and layers
+         * This also handles updating strokes where we're refetching from Ps
+         *
+         * @private
+         * @param {{documentID: !number, layerStrokes: {layerID: number, strokeStyleDescriptor: object}}} payload
```
mcilroyc: stale JSDoc.

mcilroyc:
One JSDoc nit.  Sorry, but back to @volfied.

Issue #2370, opened 2015-09-09
Select hover shows incorrect hit-test


placegraphichere:
- Create shape or text with holes
- hover over the layer
- blue hover shows up
- hover over hole area within the blue bounding box
- click to select
- layer behind will be selected
![select-hit-test-issue](https://cloud.githubusercontent.com/assets/3229167/9753461/c8ddd032-5682-11e5-8d0d-08c6b68ef87d.gif)

placegraphichere:
![select-hit-test-issue-2](https://cloud.githubusercontent.com/assets/3229167/9753498/653df40c-5683-11e5-8f00-036cba3d3c9e.gif)

iwehrman:
I bet @volfied wants to look into this!

volfied:
This is a technical limitation of our own bounds checking and hitTest being asynchronous. It's a problem we've been aware of since first implementation. I'm not sure what we can do to fix it, we had a whole different slew of bugs and usability issues when we turned off hitTest and relied solely on our own bounds (whatever is highlighted will be selected)

Short of getting a pixel map from PS every time for the topmost layer ID at a pixel, I don't see a solution to this. I'll try hooking up hitTest to highlighting so with a slight delay we highlight the layer that'll be selected, but I have my doubts on how that'll look.

mcilroyc:
@placegraphichere check master to see if this is working the way you'd expect.

mcilroyc:
also heads up @chadrolfs and @ktaki - this is a moderatly significant change to the way hover-highlighting works in super-select tool

chadrolfs:
This works now as I would expect but introduced another issue.

PR #2373, opened 2015-09-09
Use hitTest instead of bound comparison for superselect overlay

volfied:
I always shied from doing this because I thought an async function in a mousemove handler would not look good...
Well, it's quite fast and works!

Adresses #2370

### DIFF
--- a/src/js/jsx/tools/SuperselectOverlay.jsx
+++ b/src/js/jsx/tools/SuperselectOverlay.jsx
@@ -492,34 +492,38 @@ define(function (require, exports, module) {
                 }
             });
 
-            // Yuck, we gotta traverse the list backwards, and D3 doesn't offer reverse iteration
-            _.forEachRight(d3.selectAll(".layer-bounds")[0], function (element) {
-                var layer = d3.select(element),
-                    layerID = mathUtil.parseNumber(layer.attr("layer-id")),
-                    layerSelected = layer.attr("layer-selected") === "true",
-                    layerModel = layerTree.byID(layerID);
+            uiUtil.hitTestLayers(this.state.document.id, canvasMouse.x, canvasMouse.y)
+                .bind(this)
+                .then(function (hitLayerIDs) {
+                    return hitLayerIDs.findLast(function (id) {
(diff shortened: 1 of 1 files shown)


src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -492,34 +492,38 @@ define(function (require, exports, module) {
...
-                } else {
-                    layer.classed("layer-bounds-hover", true)
-                        .style("stroke-width", 0.0);
-                }
-            });
+            uiUtil.hitTestLayers(this.state.document.id, canvasMouse.x, canvasMouse.y)
+                .bind(this)
+                .then(function (hitLayerIDs) {
+                    return hitLayerIDs.findLast(function (id) {
+                        return !this.state.document.layers.byID(id).isArtboard;
```
mcilroyc: nit: I think you could use `layerTree` (defined above) and avoid these binds?

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -492,34 +492,38 @@ define(function (require, exports, module) {
...
+                            layerID = mathUtil.parseNumber(layer.attr("layer-id")),
+                            layerSelected = layer.attr("layer-selected") === "true",
+                            layerModel = layerTree.byID(layerID);
+
+                        // Sometimes, the DOM elements may be out of date, and be of different documents
+                        if (!layerModel) {
+                            return;
+                        }
+                        
+                        if (!underPolicy && !marquee && !highlightFound && layerID === topID) {
```
mcilroyc: I may be way off here, but now that we know about a definitive `topID` I have the vague feeling that we could avoid iterating over all the D3 elements, and instead do: "set all to no highlighted state" and then "grab just the topID element and apply some logic".  You don't even have to respond to this if I'm crazy.

mcilroyc:
I'll merge this soon - my nits/questions are insignificant

mcilroyc:
Actually, @volfied I may not understand the referenced issue, but when i try to repro this, I see it as still broken.  http://cl.ly/021w3a3T421Y

Issue #2302, opened 2015-09-01
Export header icons need a disabled state


iwehrman:
These guys:
![image](https://cloud.githubusercontent.com/assets/1075154/9617660/4ac1e308-50b7-11e5-8ada-02482c772bf8.png)
For example, the "export these assets" button should be disabled if there are no configured assets.

iwehrman:
Fixed by #2380.

Issue #2212, opened 2015-08-25
Swapping artboards should also swap position in the Layers panel


davealonzo:
Swap two artboards using the "swap position" tool in the properties panel.

Result: Artboards (and contents) swap successfully as expected on canvas, but not in the layers panel.

This can potentially be confusing when trying to export artboards in a certain order based on how they  are laid out on the canvas with the wrong correlating layer position.

volfied:
Swap position swaps layers in X and Y dimensions. Layers panel lists them in Z-order. Mixing the dimensions together will not produce nice results, and is not a well defined behavior. Layers are never meant to be ordered in some way in the layers panel tied to their position on canvas.

This is not a bug.

davealonzo:
But doesn't export-all currently determine the order of the list it generates in the modal based on how AB layers are ordered in the Layers panel? @mcilroyc Can you verify this?

The current design for the "add prefix" feature in export-all is dependent on generating the intended ordered list of ABs. So, It would be great if the export-all list was determined by how users typically order their ABs on canvas rather than by the Z-order from the Layers panel.

iwehrman:
I believe the layers in the export dialog are ordered alphanumerically by layer name.

mcilroyc:
Yes, *export-all* orders the artboards by index (z-order).

mcilroyc:
@davealonzo after you have had a chance to review the new Export All dialog which sorts artboards by canvas layout, can you let us know if this issue is still outstanding?   That is: do we ALSO want to swap the order in the layers panel?  I'm hoping you say NO, of course : )

davealonzo:
@mcilroyc Just caught this one (sorry)...

This is no longer an issue with the new sorting mechanism. So NO, we don't need to swap layer order  (thank you! :-)

PR #2374, opened 2015-09-09
Export grab bag; history, busy state, sorting, auto-asset creation...

mcilroyc:
A handful of export-related changes...

1. History store now tracks the export store' state, and modifications to export assets create history states.
1. Pursuant to the above, DataList and TextInput were modified so that selecting the same (default) value will not call its `onChange` (thus not causing frivolous history states when tabbing through export configuration values)
1. Artboards in the Export dialog are now sorted via their "grid" layout on canvas as opposed to by Z-index in the layer's panel (this partially mitigates #2212, I think)
1. Artboards in the Export dialog: optionally prefix the filename with a three digit value, generated on the fly by the grid-sorted order of artboards
1. The above "prefix" option is defaulted to false, but the user's choice is sticky, via DS preferences.
1. Export Panel: :arrow_down: icon will auto-generate a default asset if none exist, making simple export a one-click operation
1. Creating a new artboard will auto-populate a simple 1x asset configuration.  Note that it will not be fully supported until the artboard has at least one visible component (otherwise generator will not export it, so it is marked as not export-supported initially)
1. Exports are now done without blocking the UI, and the "busy" state is reflected in both Export dialog and panel.  This UI will need to improve, but this is a start.
1. The folder prompt process will throw a custom error if the user cancels, which allowed for a more readable (IMO) refactoring of the export* actions.
1. Document-level exports will use the document's name as the base file name
1. Addresses disabled state export button icon bug #2302
1. Addresses non-empty-group support bug #2376

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -487,6 +487,34 @@ define(function (require, exports) {
     addAsset.transfers = [];
 
     /**
+     * Add a default document to the layer specified by layerID, if it doesn't already have any assets
+     * Does not create a history state
+     *
+     * @param {number} documentID
+     * @param {number} layerID
+     */
+    var addDefaultAsset = function (documentID, layerID) {
+        var documentExports = this.flux.stores.export.getDocumentExports(documentID, true),
+            document = this.flux.stores.document.getDocument(documentID);
+
+        if (!document) {
+            return Promise.reject(new Error("Can not find document " + documentID + " to addDefaultAsset"));
+        }
+
+        var layer = document.layers.byID(layerID),
+            layerExports = documentExports.getLayerExports(layerID);
+
+        if (!layerExports || layerExports.isEmpty()) {
+            return this.transfer(updateExportAsset, document, Immutable.List.of(layer), 0, {}, true);
+        } else {
+            Promise.resolve();
+        }
+    };
+    addDefaultAsset.reads = [locks.JS_DOC, locks.JS_EXPORT];
+    addDefaultAsset.writes = [];
+    addDefaultAsset.transfers = [updateExportAsset];
+
+    /**
      * Delete the Export Asset configuration specified by the given index
      *
      * If layers is empty, or not supplied, delete a document-level asset
@@ -911,6 +939,7 @@ define(function (require, exports) {
     exports.updateLayerAssetSuffix = updateLayerAssetSuffix;
     exports.updateLayerAssetFormat = updateLayerAssetFormat;
     exports.addAsset = addAsset;
+    exports.addDefaultAsset = addDefaultAsset;
     exports.deleteExportAsset = deleteExportAsset;
     exports.setLayerExportEnabled = setLayerExportEnabled;
     exports.setAllArtboardsExportEnabled = setAllArtboardsExportEnabled;
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -79,9 +79,10 @@ define(function (require, exports) {
      * @private
      * @param {number} documentID
      * @param {Immutable.Iterable.<number>=} layerIDs Optional. If not supplied, sync doc-level metadata
+     * @param {boolean=} suppressHistory Optional, if truthy then do not supply photoshop with historyStateInfo
      * @return {Promise}
      */
-    var _syncExportMetadata = function (documentID, layerIDs) {
+    var _syncExportMetadata = function (documentID, layerIDs, suppressHistory) {
         var documentExports = this.flux.stores.export.getDocumentExports(documentID),
             document = this.flux.stores.document.getDocument(documentID),
             playObjects;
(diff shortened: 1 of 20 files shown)


mcilroyc:
I will tidy up the commits prior to merging, but left it in for granularity of review.

src/js/actions/export.js
```diff
@@ -158,7 +182,7 @@ define(function (require, exports) {
                     filePath: pathArray[0],
                     status: ExportAsset.STATUS.STABLE
                 };
-                return this.transfer(updateExportAsset, document, _layers, assetIndex, assetProps);
+                return this.flux.actions.export.updateExportAsset(document, _layers, assetIndex, assetProps, true);
```
pineapplespatula: why are we not transferring here?

mcilroyc: see comment at line 158

pineapplespatula: the commend does not say why.

src/js/actions/export.js
```diff
@@ -167,11 +191,33 @@ define(function (require, exports) {
...
     };
 
     /**
+     * Store a promise in state that will resolve when all given export promises
+     * are resolved or any are rejected
+     * 
+     * @param {Immutable.Iterable.<Promise>} exportList
+     */
+    var _batchExports = function (exportList) {
+        if (_activeExports) {
```
pineapplespatula: where does _activeExports get set to a non null value?

src/js/actions/export.js
```diff
@@ -440,6 +501,34 @@ define(function (require, exports) {
...
+            return Promise.reject(new Error("Can not find document " + documentID + " to addDefaultAsset"));
+        }
+
+        var layer = document.layers.byID(layerID),
+            layerExports = documentExports.getLayerExports(layerID);
+
+        if (!layerExports || layerExports.isEmpty()) {
+            return this.transfer(updateExportAsset, document, Immutable.List.of(layer), 0, {}, true);
+        } else {
+            Promise.resolve();
```
pineapplespatula: should this return here?

src/js/actions/export.js
mcilroyc: I've elaborated in the comment, see next commit.

src/js/actions/export.js
mcilroyc: 4 lines below this one

src/js/actions/export.js
mcilroyc: yep, good catch

src/js/util/gridsort.js
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
pineapplespatula: 2015

src/js/actions/export.js
```diff
@@ -137,19 +153,27 @@ define(function (require, exports) {
...
+     * When the export service's promise is resolved, a fresh flux action is called
+     * to perform the metadata sync (to wit: it does not transfer to updateExportAsset)
      *
      * @private
      * @param {Document} document
      * @param {Layer=} layer If not supplied, this is a document-level asset
      * @param {number} assetIndex index of this asset within the layer's list
      * @param {ExportAsset} asset asset to export
+     * @param {string=} baseDir Optional directory path in to which assets should be exported
+     * @param {string=} prefix
```
pineapplespatula: Optional? description?

src/js/util/gridsort.js
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    /**
+     * Recursively search rows to determine where the given layer's bounds fit
+     * And then merge it in.
+     *
+     * @private
+     * @param {Array.<object>} rows
+     * @param {Layer} layer
+     * @param {number=} index if not supplied, start at zero
```
pineapplespatula: you usually say "Optional" i'm not sure its needed, but i noticed given the "="  in the type but i thought i would mention it

src/js/util/gridsort.js
```diff
@@ -0,0 +1,120 @@
...
+        index = index || 0;
+
+        var row = rows[index],
+            bounds = layer.bounds,
+            newIndex;
+
+        if (!row) {
+            newIndex = index;
+        } else if (bounds.bottom < row.top) {
+            newIndex = index - 1;
```
pineapplespatula: why doesnt this case need a recursive call?

pineapplespatula:
@mcilroyc  just a few questions and nits. feel free to rebase

src/js/util/gridsort.js
mcilroyc: Since we're always starting at the "top", I believe this is safe.

mcilroyc:
@pineapplespatula I addressed your comments in my latest commit and/or answered questions above.  I've rebased partially.  I this new level of granularity is what I want.  It is still a "lot of commits" but this is a big PR.  OK?

src/js/util/gridsort.js
pineapplespatula: will we ever not start from the "top"? it seems like an easy fix.

pineapplespatula:
I am curious why these commits weren't individual PRs.

pineapplespatula:
Good enough, with such a large PR i cant promise i didn't miss anything, but I'm going to go ahead and merge it now

src/js/util/gridsort.js
mcilroyc: I hear what you're saying, but 1) this passed all my tests and I'd rather not have to retest, because 2) this feature will probably have some tweaks after dave/seth get their hands on it anyway, so don't want to over do it just yet.

mcilroyc:
rebased, fixed one conflict.  back to @pineapplespatula.

The reason this has so much in one PR: I did it over labor day weekend and there were a few times where fixes got intermingled amongst my various local branches, and THEN there were some rebases due to other open PRs.... and well, this was just safer.  won't happen again.  well, it probably will, but at least I'll try not to.

Issue #2377, opened 2015-09-09
Export panel can be collapsed/expanded by clicking on "+", "ios", and "hpdi"


ktaki:
Those icon regions should be excluded.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

Issue #2371, opened 2015-09-09
Export "+" and other icons are visible but non-functional when panel is collapsed


placegraphichere:
needs design input, but it's an issue

iwehrman:
Thus buttons probably just need a disabled state.

mcilroyc:
fyi, disabled state is now working (on my branch), and I will set these to be disabled if collapsed.  good start?

placegraphichere:
I think that's a great start. I think at the end of the day, it'd be best to open the panel and "do" the action clicked. Lets close when your branch lands @mcilroyc?

mcilroyc:
@placegraphichere  take a look at my branch `mcilroy/export-dbl-click-header` (the above-referenced PR).  This uses your "end of the day" suggestion of expanding the panel upon button click.

Alternately, for comparison, the first commit in that branch ( `git checkout HEAD~1` ) has it the simpler way: buttons are disabled when collapsed.

iwehrman:
Over to @ktaki for verification.

ktaki:
Verified expanding the panel upon button click. Clicking export button adds an asset if none added and opens folder chooser. Good!

PR #2380, opened 2015-09-09
Mcilroy/export dbl click header

mcilroyc:
This should be dealt with *after* #2374
It addresses 
#2377, #2371, #2302, and #2397


mcilroyc:
@iwehrman I've rebased this and, as JC indicated with the label swap, this is ready to go.  However, you'll see my note in the related 2371 that each commit here has a slightly different functionality.  If we can all agree on one, I'll squash the commits accordingly.

iwehrman:
LGTM.

Issue #2382, opened 2015-09-10
Error when setting stroke color for an empty layer or artboard


ktaki:
Create a new artboard. With layer 1 selected, click on a box for setting stroke color and select a color.

![screen shot 2015-09-09 at 5 35 18 pm](https://cloud.githubusercontent.com/assets/2029597/9777450/7b68d152-5719-11e5-816a-a611ba9001b1.png)

The same error happens when an artboard (with some shape layers) is selected and stroke color is set.

mcilroyc:
note: I think the desc should read "Create a new artboard *document* ..."
indeed it seems that for "empty" layers such as the placeholder dummy layer, the stroke buttons (inner, outer...) should be disabled.

volfied:
Will fix this as part of my color picker work.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix. No color setting is allowed.

PR #2387, opened 2015-09-10
Color Picker grabbag

volfied:
- Moved coalescing functionality into color picker
 - Add focus state highlight to color picker swatch so when it's tabbed to, it's highlighted, and toggles dialog with Enter/Esc
 - On dialog open, focus on the ColorInput, and tab becomes locked into the color picker dialog. Tab on Opacity input goes to Color input, and shift tab on color input goes to opacity input.
 - Fix a stroke issue #2382

### DIFF
--- a/src/js/jsx/shared/ColorPicker.jsx
+++ b/src/js/jsx/shared/ColorPicker.jsx
@@ -307,9 +307,9 @@ define(function (require, exports, module) {
             });
 
             var overlay;
-            if (this.props.hasOwnProperty("hue")) {
+            if (this.props.hasOwnProperty("color")) {
                 // this is an alpha slider
-                var bgColor = tinycolor({ h: this.props.hue, s: 1, v: 1 }).toHexString(),
+                var bgColor = tinycolor(this.props.color.toJS()).toHexString(),
                     bgGradient = "linear-gradient(to right, rgba(1, 1, 1, 0) 0%, " + bgColor + " 100%)";
 
                 overlay = (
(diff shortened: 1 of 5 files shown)


iwehrman:
Looks like the opacity input isn't hooked up?

iwehrman:
The text input doesn't preserve the format:
purple

iwehrman:
When you dismiss the picker with enter/escape, focus seems to return to the swatch but it does _not_ have the focused style.

iwehrman:
Also, the swatch itself might need to be broken out into it's own component so that it can use the `Focusable` mixin. We need to make sure keyboard focus is returned to PS if you hit escape when the swatch is focused, for example.

iwehrman:
Something weird is happening where if the swatch has focus and shows the focus ring, and then you cmd-tab away and back to PS, the focus ring disappears. This is probably related to the comment above about the focus ring not showing up when the dialog is dismissed.

src/js/jsx/shared/ColorPicker.jsx
```diff
@@ -658,29 +813,91 @@ define(function (require, exports, module) {
...
+         * Focuses on the ColorType component
+         */
+        focusInput: function () {
+            this.refs.input.focus();
+        },
+
+        /**
+         * Focuses on the opacity input component
+         */
+        focusOpacityInput: function () {
```
iwehrman: Should be private?

iwehrman:
Looks great besides a few issues with the swatch and the opacity input. Also note that I am very happy to merge this sooner rather than later and to deal with issues in follow up PRs. Just let me know.

iwehrman:
In particular, I would merge this if you _only_ fixed the opacity input. But, if you'd rather try to fix up the swatch focus issues now, that's fine too.

volfied:
@iwehrman Fixed the opacity box... As far as the focus goes, that is the problem I asked about this morning in the chat room. Using Focusable might be a viable option, but I'd like to keep that a separate issue as it's not a usability blocker.

volfied:
Added two more changes based on @placegraphichere's comments.

iwehrman:
OK, looks great! I'll file a few follow-up for some remaining issues that can wait until next week.

PR #2388, opened 2015-09-10
UI Mixed State changes

designedbyscience:
Here are some updates for Mixed States.

Additionally, some small layout adjustments and design changes.

Notably, the font preview swatch no longer reflect the size of the font

Mixed Vector Layers:
untitled-1___51_6___rgb_8____

Mixed Vector and Type:
untitled-1___51_6___rgb_8____

Mixed Type, same color, different faces
untitled-1___51_6___rgb_8____

Mixed Type Layers
untitled-1___51_6___rgb_8____

### DIFF
--- a/src/js/jsx/sections/style/AppearancePanel.jsx
+++ b/src/js/jsx/sections/style/AppearancePanel.jsx
@@ -31,15 +31,18 @@ define(function (require, exports, module) {
         Immutable = require("immutable"),
         classnames = require("classnames");
 
-    var os = require("adapter/os");
+    var os = require("adapter/os"),
+        Layer = require("js/models/layer");
 
     var TitleHeader = require("jsx!js/jsx/shared/TitleHeader"),
         Button = require("jsx!js/jsx/shared/Button"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
+        VectorFill = require("jsx!./VectorFill"),
         Type = require("jsx!./Type"),
         VectorAppearance = require("jsx!./VectorAppearance"),
         strings = require("i18n!nls/strings"),
-        synchronization = require("js/util/synchronization");
+        synchronization = require("js/util/synchronization"),
+        collection = require("js/util/collection");
 
     var AppearancePanel = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("style")],
@@ -84,12 +87,12 @@ define(function (require, exports, module) {
                 this.props.hidden !== nextProps.hidden) {
                 return true;
             }
-            
+
             if (!Immutable.is(this.props.document.layers.selected,
                 nextProps.document.layers.selected)) {
                 return true;
             }
-            
+
             if (!nextProps.visible && !this.props.visible) {
                 return false;
             }
@@ -152,13 +155,29 @@ define(function (require, exports, module) {
                 "section-container": true,
                 "section-container__collapsed": !this.props.visible
             });
-            
+
+            var layerKinds = collection.pluck(this.props.document.layers.selected, "kind"),
+                uniformLayerKind = collection.uniformValue(layerKinds, false),
+                hasSomeTextLayers = Layer.layerKinds.TEXT === uniformLayerKind,
+                hasSomeVectorLayers;
+
+            if (!uniformLayerKind) {
+                hasSomeTextLayers = layerKinds.some(function (layerKind) {
+                    return Layer.layerKinds.TEXT === layerKind;
+                });
+                
+                hasSomeVectorLayers = layerKinds.some(function (layerKind) {
+                    return Layer.layerKinds.VECTOR === layerKind;
+                });
+            }
+
             var sectionClasses = classnames({
                 "section": true,
                 "appearance": true,
                 "section__active": this.props.active,
                 "section__sibling-collapsed": !this.props.visibleSibling,
-                "section__collapsed": !this.props.visible
+                "section__collapsed": !this.props.visible,
+                "appearance__mixed": !uniformLayerKind && hasSomeTextLayers && hasSomeVectorLayers
             });
 
             var copyStyleDisabled = !(this.props.document && this.props.document.layers.selected.size === 1),
@@ -176,8 +195,18 @@ define(function (require, exports, module) {
 
             var containerContents = this.props.document && this.props.visible && !this.props.disabled && (
                 <div>
-                    <VectorAppearance document={this.props.document} />
-                    <Type document={this.props.document} />
+                    <VectorFill
+                        uniformLayerKind={uniformLayerKind}
+                        hasSomeTextLayers={hasSomeTextLayers}
+                        document={this.props.document} />
+                    <Type
+                        document={this.props.document}
+                        uniformLayerKind={uniformLayerKind}
+                        hasSomeTextLayers={hasSomeTextLayers} />
+                    <VectorAppearance
+                        uniformLayerKind={uniformLayerKind}
+                        hasSomeTextLayers={hasSomeTextLayers}
+                        document={this.props.document} />
                 </div>
             );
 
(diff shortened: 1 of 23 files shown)


shaoshing:
Hey @designedbyscience, you may already noticed, but here are some issues I encountered when testing the mix state:

1. I think we should remove these attributes for empty layers and art boards? (and they are not working)
 screen shot 2015-09-10 at 2 49 03 pm

2. Shall we show text preview (and text color) for mixed text and shape layers? I am asking this because I found myself looking for text button and realized it got merged into the stroke color button. 
screen shot 2015-09-10 at 2 55 52 pm
 screen shot 2015-09-10 at 2 55 58 pm

3. How about mixed font style?
 screen shot 2015-09-10 at 3 04 06 pm

4. Stroke color and border attributes should not shown as mixed, if I select an empty layer and a shape layer.
screen shot 2015-09-10 at 3 20 11 pm

5. I am unable to set stroke and border position (maybe there are more) when select text and shape layers
ui-mix-update

designedbyscience:
In general, I think these questions bring up some questions about consistency in behavior which I'll discuss with @placegraphichere (some might be bugs, some might be things we want to change)

iwehrman:
Also note that there may be some actions that didn't work correctly previously with mixed selections. In that case, we should file bugs about those issues instead of blocking this PR. Someone other than @designedbyscience should be tossed onto those non-UI issues.

designedbyscience:
Okay...we changed a bunch of things in this.

Big change/rule to be aware of:

For mixed vector/type selections, we are disabling any control that doesn't apply to all
Disabled inputs appear with NO content

Mixed but not disabled inputs should have -- in them

@shaoshing back to you

src/js/jsx/sections/style/AppearancePanel.jsx
```diff
@@ -152,13 +154,23 @@ define(function (require, exports, module) {
                 "section-container": true,
                 "section-container__collapsed": !this.props.visible
             });
-            
+
+            var selectedLayers = this.props.document.layers.selected,
+                mixedVectorType = selectedLayers.some(function (layer) {
```
shaoshing: `mixedVectorType` does not sound like a boolean. Maybe call it `hasMixedVectorType`?

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -272,18 +288,11 @@ define(function (require, exports, module) {
...
-                        <ToggleButton
-                            title={strings.TOOLTIPS.TOGGLE_STROKE}
-                            name="toggleStrokeEnabled"
-                            buttonType="layer-not-visible"
-                            selected={stroke.enabledFlags}
-                            selectedButtonType={"layer-visible"}
-                            onClick={!this.props.disabled ? this._toggleStrokeEnabled : _.noop}
-                            size="column-2"
-                        />
+                            {toggleVisibilityButton}
```
shaoshing: This line has extra indent :beetle:

src/js/jsx/sections/style/Type.jsx
```diff
@@ -506,26 +517,53 @@ define(function (require, exports, module) {
...
-                        Aa
-                    </div>
-                );
+                    return (
+                        <div
+                            className="fill__preview"
+                            style={fillStyle}/>
+                    );
+                } else {
+                    typeStyle = {
```
shaoshing: Looks like `typeStyle` is only used in this scope and its initial value will always get overwritten. Maybe consider remove the outer `typeStyle` assignment?

src/js/jsx/sections/style/Type.jsx
```diff
@@ -506,26 +517,53 @@ define(function (require, exports, module) {
...
-                    <div className="formline">
+                weightListID = "weights-" + this.props.document.id,
+                fillClassNames = classnames("formline", {
+                    "mixed-faces": familyName === strings.STYLE.TYPE.MIXED
+                }),
+                fillBlendFormline;
+            
+            if (this.props.mixedVectorType) {
+                fillBlendFormline = null;
+            } else {
```
shaoshing: Assigning `fillBlendFormline` to null seems not necessary here, as the default value `undefined` will do the job.

src/js/jsx/sections/style/Type.jsx
```diff
@@ -597,31 +641,31 @@ define(function (require, exports, module) {
...
                                 <SplitButtonItem
-                                    disabled={this.props.disabled}
+                                    disabled={locked}
                                     iconId="text-right"
                                     className={"split-button__item__fixed"}
                                     selected={alignment === "right"}
                                     onClick={this._handleAlignmentChange.bind(this, textLayer.alignmentTypes.RIGHT)}
                                     title={strings.TOOLTIPS.ALIGN_TYPE_RIGHT} />
                                 <SplitButtonItem
+                                    disabled={locked || !box}
```
shaoshing: `box` is only used here but is defined at line 455 (200 lines away). can you move it closer to where it is used?

src/js/jsx/sections/style/VectorAppearance.jsx
```diff
@@ -44,19 +43,16 @@ define(function (require, exports, module) {
         },
 
         render: function () {
-            var onlyTextLayers = this.props.document.layers.selected.every(function (layer) {
-                return layer.kind === layer.layerKinds.TEXT;
-            });
-
-            if (onlyTextLayers) {
+            if (this.props.uniformLayerKind === this.props.document.layers.byIndex(0).layerKinds.TEXT) {
```
shaoshing: Do you think we can import these constants from the adapter and refer to TEXT directly, instead of reading it from layer instance? For example:

``` 
var LAYER_KINDS = require("adapter/lib/layer").layerKinds
if (this.props.uniformLayerKind === LAYER_KINDS.TEXT) {
```

Or expose them to the Layer class 

```
var Layer = require("js/models/layer");
if (this.props.uniformLayerKind === Layer.KINDS.TEXT) {
```

src/js/jsx/sections/style/VectorFill.jsx
```diff
@@ -111,20 +111,22 @@ define(function (require, exports, module) {
                 enabledFlags: enabledFlags
             };
         },
-
```
shaoshing: line break please :bee:

src/js/jsx/sections/style/VectorFill.jsx
```diff
@@ -111,20 +111,22 @@ define(function (require, exports, module) {
...
         render: function () {
-            var onlyTextLayers = this.props.document.layers.selected.every(function (layer) {
-                return layer.kind === layer.layerKinds.TEXT;
-            });
-
-            if (onlyTextLayers) {
+            if (this.props.uniformLayerKind === this.props.document.layers.byIndex(0).layerKinds.TEXT) {
```
shaoshing: same here for the layer kind constants.

shaoshing:
Should the leading and tracking icons be greyout as well?
screen shot 2015-09-18 at 4 06 29 pm

These don't seem right for mixed text & smart object, and text & artboard
screen shot 2015-09-18 at 4 08 16 pm
screen shot 2015-09-18 at 4 09 37 pm

shaoshing:
Done reviewing! I have couple nits about the code, and two questions about the mix state, otherwise looks awesome! Back to you @designedbyscience

src/js/jsx/sections/style/Type.jsx
designedbyscience: Gonna push back on this one: I think that grouping `box` with the other collection operations makes more sense than moving it someplace closer

src/js/jsx/sections/style/VectorAppearance.jsx
designedbyscience: This seems like a reasonable idea, however, I think you should file that as a separate code cleanup task (as, the pattern I used here is the typical pattern for JSX throughout the rest of the code)

designedbyscience:
Rewrote a lot of mixed state detection code, seems cleaner now (more information passed through props, less looking at the layer set inside of components)

Also, addressed the visual quirks above

shaoshing:
Bravo :beers:

Issue #2308, opened 2015-09-01
Discoverability: asset links/icons for delete, send link, view on website


chadrolfs:
This needs some XD input.  Currently, you have to click below or above the asset name field which isn't terribly intuitive or discoverable.

davealonzo:
Clicking on the Library name area to select the item doesn't seem to respond. But if you click slightly above or below the name within the hover area, it easily selects as expected. This might be contributing to the issue.

The expected behavior is to click anywhere in the hover area (including the name or thumbnail) to select the item and subsequently invoke the icons.

iwehrman:
XD input received! Back over to @shaoshing.

chadrolfs:
Great!  Simply I think that means this: single click = show icons, double-click = rename

Issue #2248, opened 2015-08-27
Error applying style from CC lib after undo


ktaki:
Create a shape layer
Apply a style, say, a drop shadow from the libraries panel
Undo
Apply the same style again from the libraries panel

![screen shot 2015-08-27 at 2 41 07 pm](https://cloud.githubusercontent.com/assets/2029597/9534983/da3703e0-4cd1-11e5-8a5b-c4bcd30831d0.png)

chadrolfs:
@shaoshing you might need to work with the history king @mcilroyc on this.

mcilroyc:
@shaoshing I recently did some history work w.r.t. exports, so this is somewhat fresh in my mind.  If you're stuck on this, or need history support, let me know.

ktaki:
Verified the fix.

Issue #2197, opened 2015-08-24
CC Lib: creating a new library should enter name field immediately upon creation and hide previous library assets


chadrolfs:
click on Libraries drop down: Create New Library
start typing

Result: nothing.  
Expect: 1.) user should immediately have focus in Library Name field with that text selected so user can name their library 2.) assets from current library should be hidden until user clicks Create (dumping them into an empty library) or Cancel (reverting to previous library and assets.

chadrolfs:
Looks fixed on Win - after Create New Library, I can start typing immediately to name, but not on Mac.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
partially fixed.  assets from current library should be hidden.  see related bug #2864

mcilroyc:
FBNC to @chadrolfs, again.

PR #2390, opened 2015-09-10
CC Libraries - bug fixes

shaoshing:
This PR resolved the following issues:

##### #2308 Discoverability: asset links/icons for delete, send link, view on website

Now you click single click on title to toggle the asset buttons. Note that single click on title will have a delay before showing the buttons, as we have to wait and make sure the single click is not triggered by a double click event (https://github.com/adobe-photoshop/spaces-design/compare/shao/libraries-fixes?diff=unified&expand=1&name=shao%2Flibraries-fixes#diff-9c5b0e09defc4119032323e45a76eab1R128). 

The clickable area for toggling the buttons is also expanded for easier control. 

ui-lib-buttons

##### #2248 Error applying style from CC lib after undo

This is caused by the out dated history state because applying text / layer style does not emit new history event. I added the `historyStateInfo` and it is working fine. @mcilroyc can you help me confirm this is the right solution?

##### #2197 CC Lib: creating a new library should enter name field immediately upon creation and hide previous library assets

Improved the UI to highlight the new library input and buttons.

ui-lib-new

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -359,7 +359,7 @@ define(function (require, exports) {
                 
                 currentLibrary.beginOperation();
                 
-                var newElement = currentLibrary.createElement(currentLayer.name, ELEMENT_CHARACTERSTYLE_TYPE),
+                var newElement = currentLibrary.createElement("", ELEMENT_CHARACTERSTYLE_TYPE),
                     representation = newElement.createRepresentation(_REP_CHARACTERSTYLE_TYPE, "primary"),
                     imageRepresentation = newElement.createRepresentation(_REP_PNG_TYPE, "rendition");
                     
@@ -814,9 +814,11 @@ define(function (require, exports) {
      */
     var applyLayerStyle = function (element) {
         var appStore = this.flux.store("application"),
-            currentDocument = appStore.getCurrentDocument();
+            currentDocument = appStore.getCurrentDocument(),
+            selectedLayers = currentDocument ? currentDocument.layers.selected : Immutable.List(),
+            selectedUnlockedLayers = selectedLayers.filter(function (l) { return !l.locked; });
 
-        if (!currentDocument || currentDocument.layers.selected.isEmpty()) {
+        if (selectedUnlockedLayers.isEmpty()) {
             return Promise.resolve();
         }
 
@@ -827,7 +829,8 @@ define(function (require, exports) {
             })
             .bind(this)
             .then(function (path) {
-                var layerRef = layerEffectAdapter.referenceBy.current,
+                var layerIDs = collection.pluck(selectedUnlockedLayers, "id"),
+                    layerRef = layerIDs.map(layerEffectAdapter.referenceBy.id).toArray(),
                     placeObj = layerEffectAdapter.applyLayerStyleFile(layerRef, path),
                     options = {
                         historyStateInfo: {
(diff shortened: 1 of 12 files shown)


volfied:
@shaoshing Needs rebase, I'll start reviewing.

src/js/actions/libraries.js
```diff
@@ -620,9 +621,15 @@ define(function (require, exports) {
             .bind(this)
             .then(function (path) {
                 var layerRef = layerEffectAdapter.referenceBy.current,
-                    placeObj = layerEffectAdapter.applyLayerStyleFile(layerRef, path);
-
-                return descriptor.playObject(placeObj);
+                    placeObj = layerEffectAdapter.applyLayerStyleFile(layerRef, path),
```
volfied: Let's change this to playSimpleLayerActions call, so locked layers can be overridden. Right now we throw an error with this pure call if the layer is locked.

src/js/actions/libraries.js
```diff
@@ -659,16 +666,21 @@ define(function (require, exports) {
...
+            applyObj = textLayerAdapter.applyTextStyle(layerRef, styleData),
+            options = {
+                historyStateInfo: {
+                    name: strings.ACTIONS.APPLY_TEXT_STYLE,
+                    target: documentLib.referenceBy.id(currentDocument.id)
+                }
+            };
 
-        return layerActionsUtil.playSimpleLayerActions(currentDocument, textLayers, applyObj, true)
+        return layerActionsUtil.playSimpleLayerActions(currentDocument, textLayers, applyObj, true, options)
```
volfied: Just like here!

src/js/jsx/sections/libraries/Library.jsx
```diff
@@ -186,6 +186,14 @@ define(function (require, exports, module) {
 
             return elements;
         },
+        
+        /**
+         *
```
volfied: Maybe complete the JSDoc by explaining what this does?

src/js/jsx/sections/libraries/assets/CharacterStyle.jsx
```diff
@@ -84,21 +86,20 @@ define(function (require, exports, module) {
...
-            var fontSizeAndColorStr = _.remove([fontSizeStr, fontColorHex], null).join(", ");
-
-            var classNames = classnames("libraries__asset", {
-                "assets__graphic__dragging": this.props.isDragging,
-                "libraries__asset-selected": this.props.selected
-            });
-
-            var displayName = element.displayName !== "" ? element.displayName : font.name + " " + font.style,
+            var fontSizeAndColorStr = _.remove([fontSizeStr, fontColorHex], null).join(", "),
+                displayName = element.displayName !== "" ? element.displayName : font.name + " " + font.style,
```
volfied: This is confusing to me, is it either the display name, or font name + " " + style...

or is it either the display name or the font name , followed by space and font style. Can you add parentheses to make it clear.

volfied:
@shaoshing Got couple issues, and needs rebase, but looks good otherwise!

src/js/actions/libraries.js
volfied: On second thought, please run these by Seth and Dave to see if we should apply library assets to locked layers, in which case these would be no-ops.

src/js/jsx/sections/libraries/assets/CharacterStyle.jsx
shaoshing: empty string is considered as `false` in JS, so I simplified this line to `element.displayName ||  (font.name + " " + font.style)`

src/js/actions/libraries.js
shaoshing: According the Seth, layer / text style should not apply to locked layers, so I switched to `playSimpleLayerActions` with overwrite lock turned off. Also filtered out locked layers before apply the style, so that this will works smoothly on multi selection of locked and unlocked layers.

shaoshing:
Thanks @volfied , I've addressed your comments. Back to you.

volfied:
Boom!

Issue #2211, opened 2015-08-25
Search should only return libraries results if there is an open doc


iwehrman:
Trying to do anything with the results yields an unhandled exception.

placegraphichere:
is this a technical limitation? There may be reasons to open library items with an empty doc state.

chadrolfs:
If it's easy to do something useful with this, great, if not, disable it so it doesn't break.

volfied:
When #2345 is merged, I will make changes to open graphics for edit, and filter out other assets from search.

shaoshing:
In no-doc mode, search graphic asset and hit return will open the asset for edit. Please confirm @iwehrman

iwehrman:
Wellll, it's almost fixed, except that we still return results in the search dialog for which we apparently do not support direct editing?
```
Executing action libraries.openGraphicForEdit after waiting 6ms; 1/0
[CC Lib] unsupported graphic type: image/png
```
How would it be hard to filter the results so that only the editable assets are listed?

iwehrman:
The good news is that this is clearly no longer a high-priority issue, since there is no error and it works correctly much more often. Demoting to medium priority.

iwehrman:
Fixed by #2427.

PR #2395, opened 2015-09-10
Only allow library graphics to be listed/opened in no doc state

volfied:
Now that we have `openGraphicForEdit` action, we can actually quickly open library graphic assets to be edited in no-doc state. Other assets don't get listed.

Addresses #2211


shaoshing:
Found some redundant checking (not introduce from this PR) that I will address in a separate PR, otherwise, looks good to me. Thank you @volfied

PR #2403, opened 2015-09-11
Export Panel: Asset creation should be throttled, and the export button should be immediately disabled upon click

mcilroyc:
Both the export panel, and the export all dialog...


iwehrman:
Just to be clear, you want this to become re-enabled whenever there is _any_ change event from the export store?

iwehrman:
That seems fine for now, but it also does seem rather brittle...

mcilroyc:
Changed it to be more explicit in un-disabling export button (not listening to export state changes), and I extended the pattern to the Export All Panel

iwehrman:
:sheep: :it:

PR #2404, opened 2015-09-11
Export dialog styling changes only, not positioning changes 

DivyaPrabhakar:
As per the designs found here: https://trello.com/c/KR5wYVkW/494-design (@placegraphichere said we are following the Stripped down version). Position changes will be separate PR.

- Created a new class to style the types of headers that we use in the dialog (.dialog-header) 
- Created a new class to denote left and right column of the export dialog to add different widths 
- Took away all <Gutter> elements in ExportAllPanel.jsx and replaced with the new class .control-group__vertical
- Add correct colors for the header, export asset icons, export asset names, and checkboxes
- Adjusted the positioning of the "Cancel | Export" buttons
- Add correct scrolling properties and colors for the scroll bar 

Current look after these changes: 
Export Dialog

### DIFF
--- a/src/js/jsx/sections/export/ExportAllPanel.jsx
+++ b/src/js/jsx/sections/export/ExportAllPanel.jsx
@@ -32,7 +32,6 @@ define(function (require, exports, module) {
         _ = require("lodash");
 
     var Button = require("jsx!js/jsx/shared/Button"),
-        Gutter = require("jsx!js/jsx/shared/Gutter"),
         CheckBox = require("jsx!js/jsx/shared/CheckBox"),
         TitleHeader = require("jsx!js/jsx/shared/TitleHeader"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon");
@@ -40,7 +39,6 @@ define(function (require, exports, module) {
     var strings = require("i18n!nls/strings"),
         svgUtil = require("js/util/svg"),
         collection = require("js/util/collection"),
-        gridSort = require("js/util/gridsort"),
         ExportAsset = require("js/models/exportasset");
 
     /**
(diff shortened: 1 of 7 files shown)


src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -40,7 +39,6 @@ define(function (require, exports, module) {
     var strings = require("i18n!nls/strings"),
         svgUtil = require("js/util/svg"),
         collection = require("js/util/collection"),
-        gridSort = require("js/util/gridsort"),
```
mcilroyc: :red_circle: red alert   -- I think we've got some merge/rebase issues

DivyaPrabhakar: @mcilroyc I think this change came when I rebased against master? I didn't delete gridSort so maybe @designedbyscience did for a reason? Then we can get more clarity about this and whether it is necessary or not, thus may not be an issue.

src/js/jsx/shared/TitleHeader.jsx
```diff
@@ -38,6 +39,12 @@ define(function (require, exports, module) {
         render: function () {
             var workingTitle = this.props.title;
 
+            var classNameSet = {
+                "section-title": true
+            };
+
+            var className = classnames(classNameSet, this.props.className);
```
shaoshing: you can merge these lines into one line `var className = classnames("section-title", this.props.className)`

src/style/sections/exports/exports.less
```diff
@@ -40,37 +41,57 @@
...
+.export-dialog-right_column {
+    width: 40%;
+}
+
 .exports-panel__asset-list__container {
     display: flex;
     flex-direction: column;
     flex-grow: 1;
+    margin-left: 1.5rem;
+
```
shaoshing: Found extra line break.

shaoshing:
@DivyaPrabhakar since you are touching the styles, can you also align the text and icons?

68747470733a2f2f7777772e64726f70626f782e636f6d2f732f6b757268626b616439636d306e35612f53637265656e73686f74253230323031352d30392d313125323030392e33362e31312e706e673f7261773d31

PR #2407, opened 2015-09-11
remove menu item for vector mask editing.

pineapplespatula:


volfied:
What about shortcuts-mac and shortcuts-win.json files?

pineapplespatula:
@iwehrman  said not to worry about the strings and their localizations for MAX

pineapplespatula:
@iwehrman  is this what you meant, or did you just mean the non root languages?

iwehrman:
Well, I meant that you could leave all the strings in there. You removed one string, but not others. I guess I'd either remove none of them or all of them?

pineapplespatula:
ok, they are all ripped out now

iwehrman:
LGTM.

Issue #2405, opened 2015-09-11
Once a font at the top of the font menu is selected, subsequent changing of font becomes difficult


ktaki:
Select a font at the top of the menu. Try select another font. Hovering a mouse over other fonts causes flickering the font name field between the font hovering and the font at the top of the menu. On Mac, it becomes nearly impossible to change the font selection. On Windows, it behaves the same way, but other fonts can be selected by careful positioning of the mouse.

volfied:
Definitely related to https://github.com/adobe-photoshop/spaces-design/issues/2391, and probably related to my recent DataList changes.

I'm looking at it.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

Issue #2391, opened 2015-09-10
Unable to use mouse or arrows to select Type Weight datalist items


chadrolfs:

volfied:
Steps to reproduce: 

1 - Create a type layer
2 - Click the weight box (default: Regular)
3 - Mouse over other options

Weight keeps bouncing back to Regular.

ktaki:
Verified the fix.

PR #2408, opened 2015-09-11
Fix binary search in Select edge case for index 0

volfied:
If key we're looking for was at index 0, the binary search algorithm would fail. 

Causing index to become -1, and reset to the first item, which would cause the bounces in #2405 and #2391


iwehrman:
Seems to work!

PR #2410, opened 2015-09-11
Reduce UI size on retina displays

iwehrman:
For some reason the retina UI sizes are all a little bigger than non-retina UI sizes, which doesn't make any sense. They all also feel a little too big to me. This reduces the sizes across the board on retina. I think it feels better on a big 5k iMac and on a 15" MBP. If anybody can test on smaller retina devices, that would be helpful.

Before:
before

After:
after

### DIFF
--- a/src/style/main.less
+++ b/src/style/main.less
@@ -252,23 +252,23 @@ input {
 
 @media (-webkit-min-device-pixel-ratio: 2) and (min-device-width: 1281px) and (max-device-width: 1440px) {
 
-  html { font-size: @base-9; }
+  html { font-size: @base-8; }
 
 }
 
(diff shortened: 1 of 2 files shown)


designedbyscience:
HOW COOL IS IT THAT THIS CHANGE IS SO FEW LINES OF CODE??

iwehrman:
hi

iwehrman:
Actually though, I'm not really sure if this is sufficient. We have similar breakpoints set in a few different places. Do those need to be updated manually?

iwehrman:
Commit amended with some mirror changes in the color picker.

placegraphichere:
I thought I had opinions on this, turns out I was all wrong about opinions.

Issue #2369, opened 2015-09-08
The Choose Folder dialog unresponsive when superselect tool is selected


ktaki:
Create a doc with a shape layer and set export option, select superselect tool, and do Cmd-E or Ctrl-E.
Click Export. The Choose Folder dialog appears, but it is not responsive to any mouse click. Only an exception is the dismiss button "X" at the to right corner. If you move the dialog to outside of the application window, it will respond to mouse click. Windows only.

mcilroyc:
My hunch is that this is the same root cause as #2363   - did I hear you volunteer to create a watson issue for this/these?  Or shall I? @ktaki

ktaki:
I opened https://watsonexp.corp.adobe.com/#bug=4052784 for #2363.

ktaki:
The watson bug 4052784 has been fixed. According to Jesper, this should be fixed by javascript disabling policies around showing the export dialog. @mcilroyc

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #2411, opened 2015-09-11
Disabling policies (ie, temporarily during modal) now includes pointer policies

mcilroyc:
addresses #2369

### DIFF
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -35,18 +35,16 @@ define(function (require, exports) {
         KeyboardEventPolicy = EventPolicy.KeyboardEventPolicy;
 
     /**
-     * The previous policies during temporary policies disabling
-     *
-     * @type {?EventPolicySet}
+     * The previous policies and default modes during temporary policies disabling
+     * {
+     *     keyboardPolicies: EventPolicySet,
+     *     pointerPolicies: EventPolicySet,
+     *     keyboardMode: number,
+     *     pointerMode: number
+     * }
+     * @type {?object}
      */
-    var _cachedKeyboardPolicies = null;
-
-    /**
-     * The previous default keyboard mode during temporary policies disabling
-     *
-     * @type {?number}
-     */
-    var _cachedKeyboardMode = null;
+    var _cachedPolicyState = null;
 
     /**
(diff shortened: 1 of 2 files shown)


src/js/actions/policy.js
```diff
@@ -213,62 +210,80 @@ define(function (require, exports) {
...
             return Promise.reject(new Error("Can not disable keyboard policies if a cached state exists"));
         }
 
-        var policyStore = this.flux.store("policy"),
-            kind = PolicyStore.eventKind.KEYBOARD;
-
-        _cachedKeyboardPolicies = policyStore.getPolicies(kind);
+        var policyStore = this.flux.store("policy");
+        
+        _cachedPolicyState = {};
```
iwehrman: Seems like it would match existing style better and otherwise be no worse to just do this in one fell swoop with an object literal?

src/js/actions/policy.js
```diff
@@ -213,62 +210,80 @@ define(function (require, exports) {
...
-                return adapterUI.setKeyboardPropagationMode(
-                    { defaultMode: adapterUI.keyboardPropagationMode.FOCUS_PROPAGATE });
+                return Promise.join(
+                    adapterUI.setKeyboardPropagationMode(
+                        { defaultMode: adapterUI.keyboardPropagationMode.FOCUS_PROPAGATE }),
+                    adapterUI.setPointerPropagationMode(
+                        { defaultMode: adapterUI.pointerPropagationMode.ALPHA_PROPAGATE }));
+            })
+            .catch(function (err) {
+                _cachedPolicyState = null;
```
iwehrman: I think this is unnecessary since, if action fails, we'll reset state anyway?

mcilroyc: I guess I was guarding against this exception being trapped by the caller

iwehrman:
Looks great! Two tiny Qs. Back to @mcilroyc for those.

mcilroyc:
I squashed over the commit with the object literal suggestion.  I feel my explanation for the `catch()` is sufficient, but if you feel strongly I can certainly remove it.

iwehrman:
WFM.

Issue #2416, opened 2015-09-11
Color picker solid-color swatch should be solid


iwehrman:
Instead of duplicating the outer context-specific swatch:
![swatch](https://s3.amazonaws.com/f.cl.ly/items/0d2K3k2J1y2s442U1V3a/Image%202015-09-11%20at%202.42.04%20PM.png)

iwehrman:
The text input also is aligned and sized strangely. I also think we should ditch the swatch underline.

iwehrman:
CC @placegraphichere @volfied.

volfied:
Doh, I was passing the swatchOverlay directly.. maybe I shouldn't be :)

iwehrman:
Fixed by #2419.

PR #2419, opened 2015-09-11
Render the same thumb overlay for all color pickers

volfied:
I was mistakenly passing the swatch overlay of ColorInput to the colortype thumbnail... but we should just render the pure color instead.

Addresses #2416


iwehrman:
Looks good. I'll open different issues for the related bugs.

PR #2426, opened 2015-09-11
Show the document header in the no-doc state

iwehrman:
This makes the document header visible in both the doc and no-doc states to reduce UI flashing when going back and forth between those states. I also think it looks less weird, given that the toolbar is also enabled in the no-doc state.

no-doc header

There's no technical reason for this change. I just think it looks nicer. Over to @placegraphichere to decide whether he agrees or not.


placegraphichere:
tumblr_inline_nfkmhcya3z1qkf9i9

PR #2427, opened 2015-09-12
CC Libraries - bug fixes and improvements

shaoshing:
This PR includes the following fixes / improvements:

1. make applying non-RGB color asset works
2. Mouse over character style title will show font info, instead of display name. This allow users to check the font name and style, when they overwrite the default display name.
3. Mouse over graphic asset will show `display name` + `file extension`
4. Add new graphic asset will reveal the new graphic asset by scrolling to top of the graphic list automatically.
5. show only editable graphics in search result when there is no document. Fix for #2211
6. Resolved warning in Type: ```Failed propType: Invalid prop `value` of type `boolean` supplied to `Datalist`, expected `string`. Check the render method of `Type`.```

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -102,9 +102,10 @@ define(function (require, exports) {
     /**
      * List of acceptable image representations that PS can place as
      *
-     * @type {Array}
+     * @private
+     * @type {Set}
      */
-    var _EDITABLE_IMAGE_REPRESENTATIONS = [
+    var _PLACEABLE_GRAPHIC_REPRESENTATION_TYPES = new Set([
         "image/png",
         "image/jpeg",
         "image/jpg",
@@ -115,7 +116,20 @@ define(function (require, exports) {
         "application/photoshop.large",
         "application/illustrator",
         "application/pdf"
-    ];
+    ]);
+    
+    /**
+     * List of graphic representations types that are editable in DS.
+     * Due to the limitation in DS, we only support editing psd file.
+     * 
+     * TODO we should support editing images (png, jpg, etc.) by running the 
+     * flatten command.
+     *
+     * @type {Set}
+     */
+    var EDITABLE_GRAPHIC_REPRESENTATION_TYPES = new Set([
+        _REP_PHOTOSHOP_TYPE
+    ]);
     
     /**
      * Dimention of asset's preview image. Content is guaranteed to fit into a square of `size` x `size` pixels.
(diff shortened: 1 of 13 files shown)


src/js/actions/libraries.js
```diff
@@ -118,6 +118,17 @@ define(function (require, exports) {
     ];
     
     /**
+     * List of graphic representations types that are editable in DS.
+     * Due to the limitation in DS, we only support editing psd file.
+     * 
+     * TODO we should support editing images (png, jpg, etc.) by running the 
+     * flatten command.
+     */
+    var EDITABLE_GRAPHIC_REPRESENTATION_TYPES = [
```
iwehrman: It looks like this is only used as a set. Would it make more sense for it to be a `window.Set` object?

src/js/jsx/sections/libraries/LibraryBar.jsx
```diff
@@ -47,7 +47,12 @@ define(function (require, exports, module) {
          * @private
          */
         addGraphic: function () {
-            this.getFlux().actions.libraries.createGraphicFromSelectedLayer();
+            this.getFlux().actions.libraries.createGraphicFromSelectedLayer()
+                .then(function () {
+                    // Scroll to the top of the graphics list to reveal the newly added graphic.
+                    var graphicsListEle = window.document.getElementsByClassName("libraries__assets__graphic")[0];
```
iwehrman: Is it the graphic _guaranteed_ to be in the DOM at this point? Just want to make sure there's no chance of dereferencing past the end of the array her.

src/js/jsx/sections/libraries/assets/Color.jsx
```diff
@@ -36,25 +35,36 @@ define(function (require, exports, module) {
...
+        /**
+         * Find color asset's RGB representation and then return instance of Color model.
+         *
+         * @private
+         * @return {ColorModel}
+         */
+        _getColor: function () {
+            var representations = this.props.element.representations;
+            
+            for (var i = 0; i < representations.length; i++) {
```
iwehrman: It would be better to use a library function for this instead of a `for` loop. Maybe something like [`_.find`](https://lodash.com/docs#find). Also, it's unclear to me, but it seems possible that this could return null/undefined. If so, the JSDoc should be amended appropriately.

src/js/jsx/sections/libraries/assets/Color.jsx
```diff
@@ -89,25 +100,25 @@ define(function (require, exports, module) {
...
-
-            this.getFlux().actions.libraries.applyColor(color);
+            this.getFlux().actions.libraries.applyColor(this._getColor());
         },
 
         render: function () {
             var element = this.props.element,
-                displayName = element.displayName !== "" ? element.displayName : this.state.hexValue;
+                title = this._getPrimaryColorStringValue(),
+                hextValue = this._getColor().toTinyColor().toHexString().toUpperCase(),
```
iwehrman: Typo :tractor:: should be `hexValue`.

iwehrman:
A few minor questions and suggestions. Thanks for the quick fix of #2211! Back to @shaoshing.

shaoshing:
Hey @iwehrman I made some changes to address your comments. Like you `Set` hint; I found it easier and more readable to use `Set#has` than `Array.indexOf() !== -1`. Also improve the scroll behavior so that the asset list will scroll for all types of asset, not just graphics. Now back to you.

iwehrman:
Looks good! `Set#has` is generally faster than `Array.indexOf` too!

PR #2432, opened 2015-09-12
Initialize the window transform on startup

iwehrman:
With #2343, we no longer zoom the document on startup. But that means that we also don't get a `scroll` event from Photoshop to initialize our window transform. This makes it impossible to select objects on the canvas until the first scroll event (e.g., when panning the document). This fixes the problem by explicitly setting the transform in `ui.afterStartup`. It has to be _after_ startup because `ui.updateTransform` short circuits if there is no active document.

Addresses: my complaints in Slack.

Thanks to @jsbache for helping me narrow down the issue!


volfied:
Thanks for fixing this!

PR #2435, opened 2015-09-12
Reset cursor when updating Guard

iwehrman:
It looks like the cursor doesn't automatically reset itself when changed until the after mouse moves next. The `Guard` component frequently changes the cursor when the UI is disabled, but it can get stick around in the disabled state for an unwelcome amount of time if you're not moving the mouse. This fixes the problem by calling `os.resetCursor` when the `Guard` switches between enabled and disabled states.

Addresses: concerns.


volfied:
Ok!

PR #2439, opened 2015-09-13
Delay loading inessential layer properties

iwehrman:
This PR delays requests for "inessential" properties on layer descriptors until layers are first selected. This is designed to improve startup performance by reducing the amount of information fetched from Photoshop. On VermilonArtboards.psd, for example, startup time is reduced by about 25% (from 7.1s to 5.3s).

This change is accomplished first by splitting off from the `_optionalLayerProperties` array in `actions.layers` another array, `_lazyLayerProperties`. When a document is initialized with `layers._getLayersForDocument`, the (remaining) optional properties are fetched for all layers, while the lazy properties are only fetched for the selected layers. Layer models for which the full suite of properties have not been fetched have their `initialized` property set to false until fully initialized. 

The behavior of sibling helper function `layers._getLayersByRef` remains unchanged; it always fetches _all_ of the properties. This function is used to reset or initialize subsets of layers after the document has been initialized. This makes it easy to be sure that selected layers are always initialized without making major architectural changes. For example, `addLayers` just works. Note also that there are no JSX changes whatsoever here; the bulk of the work is in the mildly elaborated `layers._getLayersForDocument` as well as `layers.select`. Easy peasy!

Future work: do not load lazy properties at all for inactive documents, and load them for selected layers in those documents upon first activation.

**Review only** until after the MAX deadline. In the meantime, please help test this! Potential problems might involve the panels not showing up-to-date information when the layer selection changes, as well as errors resulting in other components attempting to access null properties of uninitialized models.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -74,37 +74,62 @@ define(function (require, exports) {
         "itemIndex",
         "background",
         "boundsNoEffects",
-        "opacity",
-        "layerFXVisible",
         "mode"
     ];
 
     /**
-     * Properties to be included if present when requesting
-     * layer descriptors from Photoshop.
+     * Basic optional properties to request of layer descriptors from Photoshop.
+     * When initializing documents, these are the only properties requested for
+     * non-selected layers.
      * 
      * @private
      * @type {Array.<string>} 
      */
     var _optionalLayerProperties = [
-        "adjustment",
-        "AGMStrokeStyleInfo",
-        "textKey",
         "boundingBox",
         "layerKind",
-        "keyOriginType",
-        "fillEnabled",
-        "fillOpacity",
-        "layerEffects",
-        "proportionalScaling",
         "artboard",
         "artboardEnabled",
-        "pathBounds",
         "smartObject",
-        "globalAngle",
         "layerSectionExpanded",
         "vectorMaskEnabled"
     ];
+
+    /**
+     * Inessential optional properties to request of layer descriptors from Photoshop.
+     * When initializing documents, these properties are NOT requested for non-selected
+     * layers.
+     *
+     * @private
+     * @type {Array.<string>}
+     */
+    var _lazyLayerProperties = [
+        "layerID", // redundant but useful for matching results
+        "globalAngle",
+        "pathBounds",
+        "proportionalScaling",
+        "adjustment",
+        "AGMStrokeStyleInfo",
+        "textKey",
+        "fillEnabled",
+        "fillOpacity",
+        "keyOriginType",
+        "layerEffects",
+        "layerFXVisible", // the following are required but this is not enforced
+        "opacity"
+    ];
+
+    /**
+     * All optional properties to request of layer descriptors from Photoshop.
+     * When resetting any layer(s), these properties are requested. This is the
+     * union of _optionalLayerProperties and _lazyLayerProperties.
+     *
+     * @private
+     * @type {Array.<string>}
+     */
+    var _allOptionalLayerProperties = _lazyLayerProperties
+        .slice(1)
+        .concat(_optionalLayerProperties);
 
     /**
      * Get layer descriptors for the given layer references. Only the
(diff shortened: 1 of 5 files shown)


iwehrman:
@mcilroyc: Unsurprisingly, there is also some chance that I have screwed something up w.r.t. history states. See in particular my choice to `amendHistory` when calling `resetLayers`. I might like to chat with you 1:1 about how this all works so that you can help me be sure I'm doing the right thing.

src/js/actions/layers.js
```diff
@@ -684,15 +758,18 @@ define(function (require, exports) {
...
         var dispatchPromise,
-            revealPromise;
+            revealPromise,
+            initializePromise;
             
         if (!modifier || modifier === "select") {
             payload.selectedIDs = collection.pluck(layerSpec, "id");
             dispatchPromise = this.dispatchAsync(events.document.SELECT_LAYERS_BY_ID, payload);
             revealPromise = this.transfer(revealLayers, document, layerSpec);
+            initializePromise = this.transfer(initializeLayers, document, layerSpec);
```
mcilroyc: are you intending for this promise not to be waited-on by the resultant `select` promise?

src/js/stores/document.js
```diff
@@ -169,7 +169,17 @@ define(function (require, exports, module) {
...
             this._openDocuments[nextDocument.id] = nextDocument;
 
-            if (!suppressChange) {
+            if (suppressChange) {
+                return;
+            }
+
+            // If some selected layer remains uninitialized, a new document will
+            // come along shortly. Wait until then to trigger the change event.
+            var initialized = nextDocument.layers.selected.every(function (layer) {
```
mcilroyc: so far this is the only thing that creeps me out.  Is it not possible to wait on all the layers being initialized before calling setDocument?  Is it somehow not desirable to do so?

src/js/actions/layers.js
iwehrman: Whoops, this is a mistake. I went back and forth on how to structure this a number of times and forgot to add it back into the `join` below. Will fix.

src/js/stores/document.js
iwehrman: I understand your concern, but it was intentional. The problem is that it's very difficult to ensure that we don't emit a change event between the time when the selection has changed and when the properties have been initialized. If we don't limit the change events here, then we get a) a small number of errors in the JSX due to accessing uninitialized properties (which isn't a big deal) and b) more importantly, a lot of blinking in the UI due to drawing panels without initialized properties. If I remove this, there has to be a lot more `shouldComponentUpdate` code overhead. I decided instead to do it here to codify an assumption that the UI can make: if it's selected, then it's initialized. It solves all the above problems in two lines of code. Still, I can see why you wouldn't like it and I'm happy to discuss it further.

mcilroyc: I definitely understand/undertsood  *why* we wanted to limit the change events, I just wondered if there was a way to handle this in the document/layer action dances instead of inside the store.  But, I won't press too hard.  If there are issues, I'm sure they'll be immediate and obvious :eyes:

mcilroyc:
regarding "amendHistory" ... TBH, I don't explicitly see where your "choice to amendHistory when calling resetLayers".  But, in spirit I agree that these background elaboration of layers with the non-essential properties should "amend" history (to wit: update the current history state's document model with the new information).  This might cause us to "throw away" this new information if the user undos to a previous state... but this should be of negligible performance impact, and I can't imagine a process by which we would be more clever without being overly complicated.

mcilroyc:
When selecting a type layer, I get a react warning, which feels like it could be related:
http://cl.ly/image/3J1K3h3i1d30

mcilroyc:
I suppose I'll assign this back to you check on the above-referenced react warning.  Otherwise, I think this is merge-able now that MAX dust is settling.

iwehrman:
Pretty sure that React warning is due to an unrelated issue that @shaoshing was fixing. I'm going to take a second pass on some of the concurrency choices in the `select` action. The history amendment choice is in the `initializeLayers`' call to `resetLayers`.

iwehrman:
I've made a few slight tweaks:

1. Correctly and consistently only call `initializeLayers` after the selection change has finished. This calls `resetLayers` with history amendment enabled, which to the best of my knowledge is the correct thing to do here.
2. Moved more properties (`layerFXVisible` and `opacity`) into the "lazy" set.
3. Eliminated the optimization by which some (namely, non-vector and non-type) layers were marked initialized by default. This was a faulty optimization because any kind of layer can have a layer effect applied to it. And, of course, this is required now that opacity is a lazy property as well. So, basically, all layers are uninitialized until they are explicitly initialized via selection.

Besides the aforementioned future work of only loading lazy properties of selected layers on the _active_ document (instead of for all documents), we may also consider some "idle" loading strategy, wherein we load lazy properties of unselected layers when we think the application is otherwise idle so that it doesn't have to happen at the same time as the first selection.

Marking as ready for review.

Issue #2438, opened 2015-09-13
Infinite loop in Select binary search


iwehrman:
The relevant slice of state that results in the divergence is here: https://gist.github.com/iwehrman/d26909880ea138323ac1

The cause is #2408 commit a600cac719904490879cef085b37d8f3683d91d1.

iwehrman:
This happened when typing in the typeface datalist, by the way.

volfied:
Back to @iwehrman

iwehrman:
Fixed by #2440, AFAICT.

PR #2440, opened 2015-09-13
Fix binary search from infinite loop

volfied:
@iwehrman Please review...

I created a snippet to test this as well, and tested with: empty list, one element list with existing, one element list with an non existing, your font list with repro case, run it 0-n for ranges of n...

Will address #2438 and hopefully put out the fire.

### DIFF
--- a/src/js/jsx/shared/Select.jsx
+++ b/src/js/jsx/shared/Select.jsx
@@ -180,7 +180,7 @@ define(function (require, exports, module) {
             var size = options.size,
                 low = 0,
                 high = size - 1,
-                middle = Math.floor(high / 2),
+                middle,
                 value;
 
             while (low <= high) {
--- a/src/js/jsx/shared/Select.jsx
+++ b/src/js/jsx/shared/Select.jsx
@@ -179,20 +179,20 @@ define(function (require, exports, module) {
             // binary search for the position
             var size = options.size,
                 low = 0,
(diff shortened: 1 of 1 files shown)


src/js/jsx/shared/Select.jsx
```diff
@@ -179,20 +179,20 @@ define(function (require, exports, module) {
             // binary search for the position
             var size = options.size,
                 low = 0,
-                high = size,
+                high = size - 1,
                 middle = Math.floor(high / 2),
```
mcilroyc: maybe don't need this anymore?

mcilroyc:
for kicks, would this also work:
`_.indexOf(collection.pluck(options, "id").toArray(), key, true)`
(where the last `true` arg instructs lodash to use binary search)
maybe that's only OK if we cache the plucked array.  Or maybe I'm totally crazy.

src/js/jsx/shared/Select.jsx
volfied: True dat.

volfied:
Yeah without caching of that collection we would be spending time with pluck (which I think is an expensive operation).

Removed the unnecessary init @mcilroyc, thanks for reviewing.

mcilroyc:
seems like we *could* cache that array, and use a standardized lodash search.  Anyway, merging.

Issue #2421, opened 2015-09-11
Color format is not preserved when dialog is closed


iwehrman:
![format](https://s3.amazonaws.com/f.cl.ly/items/2S070h1I3g0u3e1Q3g0y/Screen%20Recording%202015-09-11%20at%2003.35%20PM.gif)

iwehrman:
As of #2444, this works while the picker is open. It should _also_ persist even after the dialog is closed. I think it would be fine to just stick this format value in the user preferences and attempt to use that value whenever the picker is opened.

iwehrman:
Fixed by #2499.

PR #2444, opened 2015-09-14
Preserve color format in color picker lifetime

volfied:
If you type color name, until you change color/alpha it should stay that name. If you type HSV(...) it should be shown in HSV format.

Addresses #2421


iwehrman:
This is definitely an improvement, but I also think the preference should persist even after the dialog is closed. I'll elaborate #2421 accordingly, but merge this now.

PR #2445, opened 2015-09-14
Export 'loader' icon should be animated

mcilroyc:
@designedbyscience says he understands why this fix works.  So I guess I'll assign it to him.  I think this is risk-free enough to make it in to the MAX build.


designedbyscience:
This is actually using the work that @shaoshing had previously implemented for the Libraries loader, so we're good to go

shaoshing:
If you are interested in the fix, the short answer is Chrome does not support animated SVG imported externally. Here is the details https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/jsx/Main.jsx#L42

Issue #2422, opened 2015-09-11
Color picker text is small and misaligned


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9827783/e96e49de-589a-11e5-9162-2b41faaf0fb1.png)

Input from @placegraphichere is required here, so assigning him first.

placegraphichere:
this is progress on this:
<img width="323" alt="screen shot 2015-09-14 at 3 51 08 am" src="https://cloud.githubusercontent.com/assets/3229167/9856507/cd88a43e-5ada-11e5-91b6-b77ee3ce26bf.png">

iwehrman:
Fixed by #2446.

PR #2446, opened 2015-09-14
color picker tweaks

placegraphichere:
bug ref #2422 

- increased font size
- alignment for color input value
- spacing around swatches, maps and sliders
- spacing for RGB/HSB
- dialog border-radius tweak
- dialog arrow tweak for retina
- added box-shadow on color sliders/maps to reduce white glow on corners
- start of map pointer fitting on the edges of slider better
- removed some old MQ settings

- removed some math, and unnec css
- Note: this does not fix the color slider min-max range issues

### DIFF
--- a/src/style/search/search-bar.less
+++ b/src/style/search/search-bar.less
@@ -23,7 +23,7 @@
 
 @search-bar-width: 50rem;
 @drop-down-width: 46rem;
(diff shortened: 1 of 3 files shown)


iwehrman:
Better!

Issue #2450, opened 2015-09-14
Type swatch overlay should have fixed font size


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9857833/db007f50-5ad1-11e5-8e7f-efacb5a0986a.png)

iwehrman:
Fixed by #2452.

PR #2452, opened 2015-09-14
Set font size as fixed for type swatch, so we can always see it

designedbyscience:
Addresses #2450


iwehrman:
2 rems oughta be enough for anyone.

PR #2455, opened 2015-09-14
Unchecked checkboxes should be 'transparent' with a grey border

mcilroyc:
This is my "best attempt" at styling the unchecked state of checkboxes to match the spectrum standards.  
This is either:

1. "good enough"
1. Needs a ninja like @designedbyscience to carry it forward with CSS tweaks
1. Needs some SVG magic instead of a pure CSS treatment, since my current belief is that the minor alignment issues are not solvable with CSS hackery.

cc: @davealonzo @placegraphichere @DivyaPrabhakar 

example

### DIFF
--- a/src/style/shared/button.less
+++ b/src/style/shared/button.less
@@ -195,27 +195,27 @@ button {
 }
 
 .button-checkbox {
-    width: 1.5rem;
-    height: 1.5rem;
+    width: 1.3rem;
+    height: 1.3rem;
     position: relative;
     label {
-        width: 1.5rem;
-        height: 1.5rem;
+        width: 1.3rem;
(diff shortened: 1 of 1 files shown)


mcilroyc:
In case anyone checked this PR within the first 3 minutes, I did just push a small alignment fix for the "mixed" status.

mcilroyc:
@designedbyscience I just want your rough opinion on this.  If it is better than before, merge it.  If it is worse, then let please provide your opinion as to whether or not we need graphic support for these bad boys.

designedbyscience:
Made some more adjustments, added some pixel fixes for Retina, merging now

Issue #2456, opened 2015-09-14
Unhandled exception in SuperSelectOverlay


iwehrman:
This happened when I opened Vermilion from another document:
```
TypeError: Cannot read property 'isArtboard' of null
    at eval (file://./js/jsx/tools/SuperselectOverlay.jsx:499:68)
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:4342:23
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:1223:57
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:997:34
    at List.__iterate (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:2756:13)
    at ToKeyedSequence.__iterate (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:993:25)
    at KeyedIterable.reversedSequence.__iterate (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:1223:23)
    at KeyedIterable.mixin.findEntry (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:4341:12)
    at KeyedIterable.mixin.find (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:4335:24)
    at List.mixin.findLast (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/immutable/dist/immutable.js:4465:42)
    at eval (file://./js/jsx/tools/SuperselectOverlay.jsx:498:40)
From previous event:
    at React.createClass.updateMouseOverHighlights (file://./js/jsx/tools/SuperselectOverlay.jsx:497:18)
    at React.createClass.mouseMoveHandler (file://./js/jsx/tools/SuperselectOverlay.jsx:176:22)
[...]

PR #2458, opened 2015-09-14
Add a null check for layer highlighting

volfied:
Addresses #2456, it must have been a timing issue where the OS.mouseMoveHandler got a layer ID from PS that's not in our models.


iwehrman:
LGTM.

Issue #1681, opened 2015-07-06
modal dialog alert: Could not use pen tool because of a program error.


chadrolfs:
New doc, select Rect tool and drag out a shape within AB, select the Pen tool and click once to set an anchor point, click on the Rectangle 1 shape in layers panel, (pen tool still selected) click on canvas.

Result: modal dialog alert: Could not use pen tool because of a program error.

on Win you may get this as well when you try to set focus the Rect in the layers panel: 
`Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined` [Edit - Barkin: This is unrelated and fixed]

chadrolfs:
After you click OK to the modal error dialog, if you set focus back to Shape 1 and click again, it will connect the first anchor point to the one you just set, essentially letting you continue to draw out the shape you started with the pen tool.

volfied:
Ah, this has nothing to do with artboards. Our swap algorithm does one thing for the first situation, and another thing for the second situationa fter they're swapped, due to size differences in the layers.

I'll look into it, but we never tried to make our swap algorithm symmetric.

volfied:
`barkin/1681`, Not a fan of the transfer skip FIXME comment in there so leaving this Fix in Progress for now.

volfied:
I've fixed this in my branch... However, I think a better solution would be to start new pen layers instead of complaining about lack of use, which would make this a core bug. I'm still leaving my PR open to remedy the situation, but we should look into not needing that fix.

mcilroyc:
I think @volfied is saying that this is fix fa realz now.  @chadrolfs

chadrolfs:
Following original steps in issue, I still get modal error dialog.

volfied:
Will be fixed in pgdv.max.672 with CL 1044284. 

As a side effect, any time layer selection changes, we erase the current sub path index now, so to be able to continue an open path in the selected layer, you'll need to click one of it's open anchors to activate that sub path.

chadrolfs:
Apparently the core fix broke things on the other side of the fence.  They are reverting so this bug will once again re-occur.  Thanks to @fosterbrereton for the heads up.  Here is the core issue:
http://moriarty.corp.adobe.com/bug?bug=4071562

chadrolfs:
Yep, it's back and I don't think we have time to come up with a solution.  For discussion.  I'm recommending kick to M4.

chadrolfs:
[...]

PR #2460, opened 2015-09-14
Reset from pen tool to default when selected layer is not pen friendly

volfied:
Addresses #1681, if we select a shape layer or a completed pen path layer, clicking with the pen tool causes an error.

A better fix would be to change behavior in core and in those situations start a new path layer with the pen tool. That can be done on core side and will be tracked as such.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -714,13 +714,16 @@ define(function (require, exports) {
                     }
                 }).then(function () {
                     return this.transfer(tools.resetBorderPolicies);
+                }).then(function () {
+                    return this.transfer(tools.resetPenTool);
                 });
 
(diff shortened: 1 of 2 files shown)


src/js/actions/layers.js
```diff
@@ -714,13 +714,16 @@ define(function (require, exports) {
                     }
                 }).then(function () {
                     return this.transfer(tools.resetBorderPolicies);
+                }).then(function () {
+                    return this.transfer(tools.resetPenTool);
```
iwehrman: Does this need to wait for `resetBorderPolicies`? Or could these both be `join`d together?

volfied: Funny thing is, resetBorderPolicies won't do anything in this bug's case, because it only works on superselect tool. So they can be joined. I'm going to sit on this fix for a while.

PR #2462, opened 2015-09-14
Reset guide policy ID on reset

volfied:
This was causing issues where an error would put us in a constant loop of resets, because we forgot to nullify the guide policy ID between failures.


iwehrman:
Sometimes I wonder if we should completely ban action module state...

PR #2463, opened 2015-09-14
hover accuracy, focus state spectrum look

placegraphichere:
- hover for layers on item box, not just title input
- rename look following new libraries tweaks and spectrum spec


iwehrman:
OK!

PR #2467, opened 2015-09-14
Global export button added to document header

DivyaPrabhakar:
Adds an "open export dialog" to the document header 

Things to note: 

1.) Despite some of the comps regarding the UI (example: https://trello.com/c/zD4DGHqz/452-epic-ui-overhaul), I was told to only insert "open export dialog" icon for the time being. 

2.) @designedbyscience and I talked about the DOM structure and he said it is best to keep the new class .generator-header as a sibling to .document-header for CSS stability. 

This is what the UI looks currently:

### DIFF
--- a/src/js/jsx/DocumentHeader.jsx
+++ b/src/js/jsx/DocumentHeader.jsx
@@ -31,8 +31,11 @@ define(function (require, exports, module) {
         Immutable = require("immutable");
 
     var os = require("adapter/os");
-    
-    var DocumentHeaderTab = require("jsx!js/jsx/DocumentHeaderTab");
+
+    var DocumentHeaderTab = require("jsx!js/jsx/DocumentHeaderTab"),
+        Button = require("jsx!js/jsx/shared/Button"),
+        SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
+        strings = require("i18n!nls/strings");
 
     var DocumentHeader = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("application", "document")],
@@ -112,7 +115,7 @@ define(function (require, exports, module) {
             
             window.addEventListener("resize", this._handleWindowResize);
         },
-        
+
         componentWillUnmount: function () {
             os.removeListener("displayConfigurationChanged", this._updatePanelSizes);
             window.removeEventListener("resize", this._handleWindowResize);
@@ -130,7 +133,7 @@ define(function (require, exports, module) {
                 headerWidth: React.findDOMNode(this).clientWidth
             });
         },
-        
+
         /** @ignore */
         _handleTabClick: function (documentID) {
             var selectedDoc = this.getFlux().store("document").getDocument(documentID);
@@ -139,15 +142,24 @@ define(function (require, exports, module) {
             }
         },
 
+        /**
+         * Opens the export dialog 
+         */
+        _openExportPanel: function () {
+            this.getFlux().actions.export.openExportPanel();
+        },
+
         render: function () {
             var documentStore = this.getFlux().store("document"),
                 document = this.state.document,
                 smallTab = this.state.headerWidth / this.state.documentIDs.size < 175;
             // Above: This number tunes when tabs should be shifted to small tabs
 
+            var exportDisabled = !document || document.unsupported;
+
             var documentTabs = this.state.documentIDs.map(function (docID) {
                 var doc = documentStore.getDocument(docID);
-                
+
                 if (doc) {
                     return (
                         <DocumentHeaderTab
@@ -166,6 +178,18 @@ define(function (require, exports, module) {
                 <div className="document-container">
                     <div className="document-header" ref="tabContainer">
                             {documentTabs}
+                    </div>
+                    <div className="export-header">
+                        <div className="export-header-buttons">
+                            <Button
+                                className="button-plus button-simple"
+                                title={strings.TOOLTIPS.EXPORT_DIALOG}
+                                disabled={exportDisabled}
+                                onClick={this._openExportPanel}>
+                                <SVGIcon
+                                    CSSID="extract-all" />
+                            </Button>
+                        </div>
                     </div>
                 </div>
             );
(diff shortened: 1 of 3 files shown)


iwehrman:
The icon is both visible and active when there is no document open. At the very least it should be disabled. I'm not sure whether it should be visible or not. My hunch is that it should remain visible but inactive.

davealonzo:
Would recommend changing the hover title to "Export..." to reflect the same title from the File menu.

src/style/document-header.less
```diff
@@ -59,13 +59,26 @@
...
     position: relative;
     overflow: hidden;
     display: flex;
 }
 
+.generator-header {
+    height: @toolbar-button-size;
+    line-height: @toolbar-button-size;
+    width: 7%;
+    max-width: 7%;
```
iwehrman: I think this needs a fixed width. If the screen is very wide, the icon drifts away from the right edge:
image

src/js/jsx/DocumentHeader.jsx
```diff
@@ -167,6 +178,15 @@ define(function (require, exports, module) {
                     <div className="document-header" ref="tabContainer">
                             {documentTabs}
                     </div>
+                    <div className="generator-header">
+                        <SplitButtonList>
+                            <SplitButtonItem
```
mcilroyc: I won't make a federal case out of this, but I don't see why this should be a split button as opposed to a regular button.  I could be wrong about this, but I think SplitButton is sort of a historical relic and should be used sparingly, and only for sets of buttons that are functionally and perhaps interactively joined somehow.

mcilroyc:
I have one engineery nit (split button), and then both Ian and dave have suggestions.  In particular, I'll reiterate that the button should be disabled when no document is open.

DivyaPrabhakar:
Changes made: 

1. Changed the tooltip as per @davealonzo suggestion
2. Made the button active only when a document is open 
3. Used the <Button> component instead of <SplitButton>

src/style/document-header.less
```diff
@@ -59,13 +59,54 @@
...
     line-height: @toolbar-button-size;
-    max-width: 100%;
-    width: 100%;
+    flex-grow: 10;
     position: relative;
     overflow: hidden;
     display: flex;
 }
 
+.generator-header {
```
mcilroyc: this is a nit that I should have caught before, but I don't really like the use of `generator`.  I'd prefer `export-`.   I'm on the fence as to whether or not this is a must-fix.

src/style/document-header.less
```diff
@@ -59,13 +59,54 @@
...
+        &.button-simple__disabled {
+            fill: @item-inactive;
+
+            &:hover {
+                fill: @item-inactive;
+            }
+        }
+}
+
+.generator-header-buttons .button-simple.button-simple__disabled.button-plus svg {
```
mcilroyc: sweet mother of mercy that's a lot of selectors.  It seems like we should have a far simpler way to make the SVG disabled for disabled buttons.  just sayin.  maybe a future cleanup task.

src/js/jsx/DocumentHeader.jsx
```diff
@@ -31,14 +31,19 @@ define(function (require, exports, module) {
...
+        SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
+        strings = require("i18n!nls/strings");
 
     var DocumentHeader = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("application", "document")],
 
         getInitialState: function () {
-            return {};
+            return {
+                exportDisabled: true
```
mcilroyc: I think you can actually just remove exportDisabled from the react state (here, and `componentWillReceiveProps` in its entirety).  It isn't actually being used (you are deriving it within `render()` separately,  and that's totally sufficient)

src/js/jsx/DocumentHeader.jsx
```diff
@@ -167,6 +193,18 @@ define(function (require, exports, module) {
...
                             {documentTabs}
                     </div>
+                    <div className="generator-header">
+                        <div className="generator-header-buttons">
+                            <Button
+                                className="button-plus button-simple"
+                                title={strings.TOOLTIPS.EXPORT_DIALOG}
+                                disabled={exportDisabled}
+                                onClick={this._openExportPanel}>
+                                <SVGIcon
```
mcilroyc: thanks, I like this better as a Button (even though  you have to deal with SVGIcon separately ... and we might oughta make Button handle this for us like SplitButton does) ... but I'm just rambling.  Ignore all this.

src/style/document-header.less
```diff
@@ -59,13 +59,54 @@
...
+}
+
+.generator-header-buttons .button-simple.button-plus {
+        margin-right: 2rem;
+
+        & svg:hover {
+            fill: @item;
+        }
+
+        &.button-simple__disabled {
```
mcilroyc: related to my other comment a few lines down ... isn't this redundant to that separate disabled-svg block?

mcilroyc:
Back to @DivyaPrabhakar.  I think the state cleanup is important (to avoid future confusion), and replacing the term "generator" with "export" would be nice, since you have to dig back in.  Streamlining the CSS for disabling the button is optional.

DivyaPrabhakar:
@mcilroyc Made the changes you pointed out and tried to clean up the CSS a bit. The reason that block of CSS is so big is because it handles both disabled and the enabled state, because the difference is just one class. I could make it two different blocks, one for enabled, the other for disabled, if you think readability will improve?

src/js/jsx/DocumentHeader.jsx
```diff
@@ -97,7 +100,8 @@ define(function (require, exports, module) {
             return this.state.count !== nextState.count ||
                 this.state.headerWidth !== nextState.headerWidth ||
                 !Immutable.is(this.state.documentIDs, nextState.documentIDs) ||
-                !Immutable.is(this.state.document, nextState.document);
+                !Immutable.is(this.state.document, nextState.document) ||
+                this.state.exportDisabled !== nextState.exportDisabled;
```
mcilroyc: one vestigial tail of the old state

Issue #2454, opened 2015-09-14
Type: Esc to exit Typeface datalist results in policy error and bad state


chadrolfs:
New doc
select Type tool, click on canvas and type
while still in edit mode, select all on canvas text you just typed
click in Typeface drop down to expand it
hit Esc twice

`Action policy.removeKeyboardPolicies failed: Error: Unknown policy ID: 11`

iwehrman:
I've had a branch in the cooker for a while now that should address this. I was just waiting for @jfitz-adobe's focus changes to land. So, I'll have a look.

ktaki:
Hasn't this been fixed with the #2472?

PR #2472, opened 2015-09-14
Modal type keyboard policy fixes

iwehrman:
1. Don't install keyboard policies on focus unless we're in a modal state.
2. Don't install keyboard policies more than once.

Fixes the extreme brokenness in #2454, although not _all_ the focus issues are resolved.


volfied:
Will need more unraveling later, but that's going to take much longer than just fixing the extreme brokenness.

Issue #2466, opened 2015-09-14
Export panel does not have a scroll bar


ktaki:

ktaki:
Can we fix this for MAX?

DivyaPrabhakar:
@ktaki @mcilroyc I can take this. I would have fixed it in this: https://github.com/adobe-photoshop/spaces-design/issues/2152 but because of the time crunch, I talked to @ktaki and he said a PR for just the scrollbar is fine for now, so doing that now.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #2475, opened 2015-09-14
Added scrollbar to exports panel

DivyaPrabhakar:
Added scrollbar to the exports panel when layer is selected and when nothing is selected. References issue #2466.


iwehrman:
It's a scrollbar! 
image

Issue #2447, opened 2015-09-14
No doc state shows doc title bar (and it's doesn't align with Recent Files divider)


chadrolfs:
this happened sometime between master pushed to P4 on Friday 9/11 and now.  84cd55642d1501cafdbb627f91382481337ec523  didn't show title bar in no doc state.
![ice_screenshot_20150914-105715](https://cloud.githubusercontent.com/assets/8442631/9857541/f5415c6a-5acf-11e5-868c-319a733f926f.png)

iwehrman:
Showing document header in no-doc UI is intentional. We just need to tweak the Recent Files header so that, when it's empty, it has the right height.

iwehrman:
CC @placegraphichere

placegraphichere:
see #2476

iwehrman:
FBNC to @chadrolfs.

PR #2476, opened 2015-09-15
align recent files, do not add empty recent file html content

placegraphichere:
addresses #2447

### DIFF
--- a/src/js/jsx/sections/nodoc/RecentFiles.jsx
+++ b/src/js/jsx/sections/nodoc/RecentFiles.jsx
@@ -73,16 +73,24 @@ define(function (require, exports, module) {
                     );
                 }, this);
 
-            return (
-                <section className="recent-files section section__active">
-                    <TitleHeader title={strings.NO_DOC.RECENT_FILES_TITLE} />
-                    <div className="section-container">
-                        <ul className="link-list__list">
-                            {recentFilelinks}
-                        </ul>
-                    </div>
-                </section>
-            );
+            if (recentFilesLimited.isEmpty()) {
+                return (
+                    <section className="recent-files section section__active">
+                        <TitleHeader title={strings.NO_DOC.RECENT_FILES_TITLE} />
+                    </section>
+               );
+            } else {
+                return (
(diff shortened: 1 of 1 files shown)


src/js/jsx/sections/nodoc/RecentFiles.jsx
```diff
@@ -61,6 +61,7 @@ define(function (require, exports, module) {
         render: function () {
             var recentFilesLimited = this.props.recentFiles.slice(0, MAX_RECENT_FILES),
                 shortenedPaths = pathUtil.getShortestUniquePaths(recentFilesLimited),
+                recentFilesCount = recentFilesLimited.length,
```
iwehrman: If `this.props.recentFiles` and hence `recentFilesLimited` are `Immutable.List`s instead of `Array`s then this will cause a warning in the console because `length` is not the correct property. Instead, you want to use `size`. http://facebook.github.io/immutable-js/docs/#/List/size

src/js/jsx/sections/nodoc/RecentFiles.jsx
```diff
@@ -72,17 +73,25 @@ define(function (require, exports, module) {
...
-                    <TitleHeader title={strings.NO_DOC.RECENT_FILES_TITLE} />
-                    <div className="section-container">
-                        <ul className="link-list__list">
-                            {recentFilelinks}
-                        </ul>
-                    </div>
-                </section>
-            );
+                
+            if (recentFilesCount >= 1) {
```
iwehrman: And, if you gotta dive back in anyway, please change this to use a strict inequality for better consistency with our style `recentFilesCount > 0`. Alternatively, you can eliminate the count variable and just use the `isEmpty()` method. http://facebook.github.io/immutable-js/docs/#/List/isEmpty

iwehrman:
Back to @placegraphichere to fix some little Immutable issues.

iwehrman:
Rebased and fixed my own nits. Back to @placegraphichere to review and merge.

placegraphichere:
thanks for fixing my mistakes

Issue #2461, opened 2015-09-14
Attempting to add guides when AB is beneath highlight area moves the AB instead of creating guide


chadrolfs:
cmd/ctrl + n
select and drag AB so that it is under the work area bounds
hover over area where guide creation highlight and AB over lap, click and try to drag out a guide

Result: AB is moved and a guide may be created.

chadrolfs:
then subsequent attempts to add a guide either creates a guide below click area or fails to create one or moves the AB.

chadrolfs:
https://www.dropbox.com/s/uexe1ifp92b1h2m/Guide-moveAB.mov?dl=0

chadrolfs:
https://www.dropbox.com/s/uexe1ifp92b1h2m/Guide-moveAB.mov?dl=0

mcilroyc:
to @chadrolfs

PR #2477, opened 2015-09-15
For artboard guides, change position to relative mouse location

volfied:
Depends on https://github.com/adobe-photoshop/spaces-adapter/pull/148 to work fully. Basically, PS code changed so when we're creating artboard guides, we gotta make up for the relative positioning PS does for our window to canvas transforms. 

Will address #2461


mcilroyc:
don't  -  go  - chaaasin'  ...

Issue #2471, opened 2015-09-14
Sampler HUD: Use keyboard+click combo as an alternative to SPACEBAR


placegraphichere:
Slack user feedback has come in a few times asking for a keyboard+click alternative to the spacebar to activate the HUD. I think we should listen at this point and give it a try. @volfied I'd like your input on this too.

Suggestions include: 
- OPT+SHIFT+click
- CMD+SHIFT+click

Also frees up spacebar for panning while in eyedropper tool in the future

DivyaPrabhakar:
@bradee was mentioning the same thing to me after Demo Days -- considering even bringing up the HUD when eyedropper selected + click, since that seems like the only reason you would have picked the eyedropper in the first place.

volfied:
I'd definitely say cmd+shift+click, option+shift is really awkward to press.

chadrolfs:
only one modifier is preferred.

volfied:
At this point, this issue is waiting on #2480 to be merged, or be closed where we keep spacebar.

volfied:
Passing to @placegraphichere to make a call. Seth, please test #2480 and see if it's something you like to merge.

PR #2480, opened 2015-09-15
Show sampler HUD on shift+click instead of spacebar

volfied:
Implements enhancement #2471. Wrote this out for @placegraphichere to test and get the feel of it.

### DIFF
--- a/src/js/tools/sampler.js
+++ b/src/js/tools/sampler.js
@@ -29,7 +29,6 @@ define(function (require, exports, module) {
         util = require("adapter/util");
 
     var events = require("js/events"),
-        system = require("js/util/system"),
         Tool = require("js/models/tool"),
         EventPolicy = require("js/models/eventpolicy"),
         PointerEventPolicy = EventPolicy.PointerEventPolicy,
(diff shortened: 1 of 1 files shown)


iwehrman:
We're abandoning the spacebar shortcut??

volfied:
I believe so, this is to get a feel for cmd+shift+click. People except space to be a pan at all times.

chadrolfs:
:cry: Space makes it so easy.  Cmd + Shift + click makes it feel more clunky and less easy to access.

volfied:
Updated this to be shift+click, since shift+click was FX, and it's available in the HUD as well. Please test and tell me about your experience here

iwehrman:
I prefer the spacebar shortcut because it works without having to click with a tool. I'd like to use the same shortcut with the vector mask create-mask-with-shape HUD, and I'm not sure shift-click really works there?

chadrolfs:
Issues:  
- Space pan doesn't work
- Sampling Type only samples Typeface (doesn't include color/align/size)

volfied:
Closing this as we've decided to stay with spacebar.

Issue #2453, opened 2015-09-14
Sampler: sampling Text attributes does not include alignment


chadrolfs:

chadrolfs:
create 2 type layers.
layer 1: black type, align left
layer 2: red type, align center
select layer 1, hit i for Sampler
mouse over layer 2, hit space and click on Type icon
result: alignment stays left.  should change to center align

chadrolfs:
argh, was on release branch instead of master.

PR #2481, opened 2015-09-15
Add text alignment to type sampling/type style application

volfied:
Addresses #2453


shaoshing:
The code looks good, but I found the target layer's align icon does not reflect its new state. Back to you @volfied 

sampler-text-style

shaoshing:
merge this PR as the above issue no longer exist in latest Jenkin build (635)

Issue #2430, opened 2015-09-12
Libraries datalist should be dismissed when clicking outside of it


iwehrman:
![click](https://s3.amazonaws.com/f.cl.ly/items/111H2T0n0E1a2y1s0L2P/Screen%20Recording%202015-09-11%20at%2005.33%20PM.gif)

iwehrman:
Fixed by #2486.

Issue #2312, opened 2015-09-02
CC Libraries panel scroll position changes with active document


iwehrman:
To reproduce:

1. Create a new document, reveal the Libraries panel, scroll it to the bottom.
2. Create a new document. Observe that the Libraries panel has jumped back to the top.

This is because we're keeping N copies of the Libraries panel in the DOM, just like all the other panels. This makes a lot more sense for the other panels, but not for the Libraries panel because it is not document-specific. Potential workarounds:

1. Just save the last scroll position and set it when switching documents.
2. Create just one Libraries panel at the `PanelSet` level and pass it as a prop to child `Document`-specific `Panel` instances.

designedbyscience:
Fixed by 2486

iwehrman:
Fixed by #2486, indeed.

PR #2486, opened 2015-09-15
Render one libraries panel for all documents to maintain consistent state

shaoshing:
Currently each document has its own panel set in the DOM, which is reasonable as documents have different attributes and state. However, this cause problems to the library panel as switching between document should maintain the same state (scroll position, selected asset, dialog, etc.). The problem is originally reported in #2312. This PR changed the behavior by rendering only one libraries panel for all documents, and still keep separate copy of other panels. 

(Note that I tried and it is not working by creating a top level libraries panel instance and pass it as a prop to the panel set, because the instance will have different component key (final key = parent keys + component key) when rendered and will result in DOM replacement.)

ui-separate-libraries

This PR also addressed #2430 by adding the onBlur handler back. 

ui-datalist

@designedbyscience since you are the master of UI, I am assigning this to you :dancers:

### DIFF
--- a/src/js/jsx/PanelColumn.jsx
+++ b/src/js/jsx/PanelColumn.jsx
@@ -28,10 +28,8 @@ define(function (require, exports, module) {
         classnames = require("classnames");
 
     var PanelColumn = React.createClass({
-
         shouldComponentUpdate: function (nextProps) {
-            // The document is inactive
-            if (!nextProps.current) {
+            if (!this.props.visible && !nextProps.visible) {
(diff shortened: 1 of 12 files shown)


shaoshing:
Hmm, Travis is unable to complete `npm install`. My local test shows all passed.

designedbyscience:
Looks good so far...and has the added benefit of speeding up switching between documents

designedbyscience:
Alright, looks good.

PR #2487, opened 2015-09-15
Spaced out buttons in the exports panel header

DivyaPrabhakar:
@designedbyscience Suggested this tweak to the Exports Panel 

Before:

After:


iwehrman:
LGTM!

PR #2492, opened 2015-09-16
Fixed export panel input configurations - fixing a recent regression

mcilroyc:
These changes to the inputs were accidentally overwritten by a recent commit.  @placegraphichere this should fix the suffix input disabled issue you mentioned.  It also restores the ability to tab through the inputs without making changes.


iwehrman:
LGTM.

Issue #2441, opened 2015-09-14
Clear Guides enabled with no guides


chadrolfs:
Clear Guides command should be greyed out if the active document doesn't contain any guides.

iwehrman:
FBNC to @chadrolfs.

PR #2498, opened 2015-09-16
Disable clear guides menu item if there are no guides in current doc

volfied:
Addresses #2441


iwehrman:
Have guides, will travel.

PR #2499, opened 2015-09-16
Add last used color picker format to preferences, so it's persistent

volfied:
Addresses second part of #2421


iwehrman:
Brilliant.

Issue #2375, opened 2015-09-09
Opening a very large file throws an exception


zoranagee:
open image found at https://www.dropbox.com/sh/h9zbex06an7s8nv/AACLTiX9CGJqXC4_FPRcSfWna?dl=0
RESULT:
Unrecoverable error: Error: Unexpected paragraph alignment value: justifyLeft

chadrolfs:
You don't need to download the big doc to repro.  Here are steps:
new doc in classic PS
select Type tool and click-drag to create a text/paragraph box
populate it with text, select all the text
open the paragraph panel (Window > Paragraph), click on the Justify last left icon
press numPad enter to commit the text
enter DS

chadrolfs:
When fixing, make sure all paragraph settings are handled (justify, hyphenation, etc)

volfied:
Back over to @chadrolfs

PR #2500, opened 2015-09-16
Allow all text justification values 

iwehrman:
This relaxes an overly strict check in the `ParagraphStyle` construction function that would throw if it found a justification value it didn't understand. As it turns out, there are more valid justification values than those previously checked for! Namely, `justifyLeft`, `justifyCenter` and `justifyRight`. This just lets anything pass through into the model as long as it's not falsey.

Addresses #2375. Instructions for testing are in that issue.


volfied:
Ok!

PR #2502, opened 2015-09-16
When stroke is in empty color state, initialize it at default color

volfied:
If we had an image coming from PS, where stroke was set to empty color (red line crossing white square), we would get "null" as the color, and downsample would accept that as mixed. Instead, we consider those layers (where stroke is null) to be default color.

Reason for this is: if any property of stroke other than color is set (or the eye icon is clicked to enable stroke), it'll start with the default color anyways, so this removes the ambiguity of the mixed stroke for no-stroke layers.

Addresses #2207


mcilroyc:
fair enough.  Seems like there is a chance this could have confusing side-effects, but only one way to find out!

Issue #2490, opened 2015-09-16
Focus is not coming back to canvas after changing color in the text editing mode


ktaki:
Windows only. After or during changing the color of font with the color picker tool, you cannot continue typing. Closing the color picker, reopening it, and closing it again will regain the focus on the type layer on canvas. On Mac, you can continue typing while the color picker UI is open.

A comment by John F in https://watsonexp.corp.adobe.com/#bug=4051807

NotThisProduct

This looks to be a Javascript issue.
The JS font color picker will correctly invoke keyboardFocusAcquire / keyboardFocusRelease when it is click-activated "from the FontIconButton" and a color is selected. However, the JS font color picker also will remain persistently open in the DS UI until dismissed with another click on the FontIconButton. When the JS font color picker is in this "persistently open state", it fails to call keyboardFocusAcquire / keyboardFocusRelease which causes loss of focus issues described by Taki.

volfied:
Bounced the bug back, we're not getting the identical set of events from the adapter. I don't see how this can be fixed on the JS side.

volfied:
Passing to @tiburbage while it's being fixed on Watson side.

iwehrman:
I don't think I understand the bug. I don't see anything out of the ordinary on Windows. If there is a subtle focus transfer problem, can you try to give very concrete repro steps? Also, are we sure this is still a high-priority bug?

iwehrman:
Also, is this different from #2753?

ktaki:
I can close this as fixed. #2753 remains as an issue.

PR #2503, opened 2015-09-16
Focusable color swatch

iwehrman:
This factors out a `ColorSwatch` component from the `ColorInput` component and makes it a `Focusable component. Once you focus the swatch by tabbing to it, you can hit enter to open the color picker. After the picker closes, the swatch regains focus. From there, you can keep tabbing or hit escape to return focus to PS. 

It's possible, though perhaps not likely, that this addresses #2490.

### DIFF
--- a/src/js/jsx/shared/ColorInput.jsx
+++ b/src/js/jsx/shared/ColorInput.jsx
@@ -34,7 +34,8 @@ define(function (require, exports, module) {
 
     var os = require("adapter/os");
 
-    var Dialog = require("jsx!js/jsx/shared/Dialog"),
+    var Focusable = require("js/jsx/mixin/Focusable"),
+        Dialog = require("jsx!js/jsx/shared/Dialog"),
         ColorPicker = require("jsx!js/jsx/shared/ColorPicker"),
         Color = require("js/models/color"),
         strings = require("i18n!nls/strings"),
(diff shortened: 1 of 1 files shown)


iwehrman:
@ktaki if you have a chance, would you regress #2490 against this branch?

ktaki:
Works fine on Mac. On Windows, you can change the color once (before, you could not), but second attempt of changing color fails and typing cannot be continued.
With type tool, start typing on canvas. Open color picker, change the color, don't close the color picker. You can continue typing with the new color. Click on color picker to select another color. At this point, you are no longer able to type on canvas. Closing the color picker by clicking the SET TEXT COLOR box will exit editing mode.

iwehrman:
Bummer. In any case, I think that is a separate issue and that this should still be merged. Reviewers may decide otherwise, however!

src/js/jsx/shared/ColorInput.jsx
```diff
@@ -52,6 +53,60 @@ define(function (require, exports, module) {
...
+                case "Escape":
+                    this.releaseFocus()
+                        .bind(this)
+                        .then(function () {
+                            React.findDOMNode(this).blur();
+                        });
+                    break;
+            }
+
+            this.props.onKeyDown(event);
```
mcilroyc: part of me thinks this should validate the existence of this prop before calling it, since it is not otherwise required by react.  But I realize this is a "private" component, so I'll leave it up to you.

iwehrman: That was my thinking... I could be convinced otherwise, though.

mcilroyc:
LGTM, and I agree that the issue taki raised is likely independent of this change.

Issue #2484, opened 2015-09-15
Dragging an asset off canvas bounds, nests incorrectly and doesn't expand canvas bounds leaving it unselectable


chadrolfs:
cmd/ctrl + n
zoom out
drag an asset onto workarea well outside AB bound, hit enter to place
attempt to click and drag it

Result: you can't select it on canvas.  Classic PS updates correctly and you're able to select and move it.

https://www.dropbox.com/s/vi77lfc098l6hcw/asset-offcanvas.mov?dl=0

chadrolfs:
@volfied can probably help here if not take this issue

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
This is still broken but may be a result of other issues that are in the way( #2522 #2523 ).  Sometimes it works, sometimes it doesn't.  For example it doesn't work with svg or illustrator assets.

iwehrman:
FBNC to @chadrolfs.

PR #2506, opened 2015-09-16
Don't replace the empty layer while creating layers from library elem…

volfied:
So creating a new layer works right.

@iwehrman, in #2201 thought that the empty layer left behind in our model was a bug. But PS actually leaves this empty layer when we do a library place. So I'm reverting @shaoshing's changes from #2268 (bcbb1ed) so we don't break replace algorithm for add layers in other places.

Will address #2484 alongside correct layer positioning.

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -874,7 +874,7 @@ define(function (require, exports, module) {
                 var selectedLayers = document.layers.selected;
                 replaceLayer = selectedLayers && selectedLayers.size === 1 && selectedLayers.first();
             }
-            
+
             // The selected layer should be empty and a non-background layer unless replace is explicitly provided true
             replace = replaceLayer &&
                 (replace ||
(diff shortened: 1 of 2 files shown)


iwehrman:
Uhhh, #2201 was definitely a bug. Our layers panel was different from the one in standard PS. I'll test it all again though...

volfied:
This also addresses #2484

src/js/models/layerstructure.js
```diff
@@ -869,15 +869,12 @@ define(function (require, exports, module) {
...
-                    newLayerIndex = descriptors[0].itemIndex;
-                    
-                replaceLayer = selectedLayerIndex === newLayerIndex && selectedLayer;
+            } else {
+                // otherwise, replace the selected layer
+                var selectedLayers = document.layers.selected;
+                replaceLayer = selectedLayers && selectedLayers.size === 1 && selectedLayers.first();
             }
-
+
```
iwehrman: :space_invader:

iwehrman:
LGTM!

PR #2508, opened 2015-09-17
Fix error loading documents with a background layer

iwehrman:
@volfied pointed out this morning that he was unable to load a picture of his puppy dog. This was due to my recent change in #2439 to `layers._getLayersForDoc` handling layer indexing incorrectly for documents with background layers. In my defense, the way Photoshop does this is completely :banana: :banana:.

CC @mcilroyc since you reviewed #2439 and CC @volfied because it was your puppy.

### DIFF
--- a/src/style/shared/number-input.less
+++ b/src/style/shared/number-input.less
@@ -32,10 +32,6 @@
     border-bottom: @hairline solid @item-active;
 }
 
-.number-input:disabled:hover {
-    border: none;
-}
-
 .number-input__dirty:focus,
 .number-input__clean:focus {
     color: @item;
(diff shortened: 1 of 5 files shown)


iwehrman:
I pushed another commit to also disable the blend mode and opacity inputs for background layers. I put it in this PR because you need the fix in the first commit to even load documents with background layers. Addresses #2507.

src/style/shared/number-input.less
```diff
@@ -32,10 +32,6 @@
     border-bottom: @hairline solid @item-active;
 }
 
-.number-input:disabled:hover {
```
mcilroyc: wut

mcilroyc:
ok

src/style/shared/number-input.less
iwehrman: It looked bad. I'll alert @placegraphichere in case he wants to change the styling more, but without this it seems fine.

placegraphichere: This will probably get covered in my spectrum color branch. Pretty sure I already altered it there

Issue #2516, opened 2015-09-17
Error on changing the color of font in editing mode


ktaki:
With type tool, click on canvas and start typing. Click on the set color box to invoke color picker.
(pgDev Mac 648 and the latest master)

Action policy.addKeyboardPolicies failed: Error: Trying to add new policies but a cached state exists

![screen shot 2015-09-17 at 2 24 31 pm](https://cloud.githubusercontent.com/assets/2029597/9946465/471cb668-5d48-11e5-8e36-005508fa5a4b.png)

pineapplespatula:
i backed out my text policy change, this bug should be gone for now... but i'm not sure what the longer term plan is for modal tool states and keyboard policy

ktaki:
No longer happening.

PR #2518, opened 2015-09-17
revert disablePolicy for modal type tool

pineapplespatula:
pull this until we have a real plan


pineapplespatula:
fixes issue #2516

Issue #2399, opened 2015-09-11
When Export Dialog is open, hovering over the guides area still works 


DivyaPrabhakar:
![Guides active when export dialog open](https://www.dropbox.com/s/uc2lgsdjxg9cznr/guides_active_dialog.gif?raw=1)

mcilroyc:
Dup of #2138, or at least closely related.

mcilroyc:
assigning this to @DivyaPrabhakar for a discount double check

DivyaPrabhakar:
Looks good on my end

Issue #2138, opened 2015-08-20
Clicks in the ExportAll dialog can grab an underlying guide


mcilroyc:
This is also possible in the Design Space Intro screen.  This is only particularly troubling when the guide happens to be below a checkbox.

iwehrman:
The dialog infrastructure should ensure that the pointer propagation policy is NEVER_PROPAGATE while a modal dialog is open.

mcilroyc:
Note this happens as well in the style panel, if you can find an otherwise non-clickable area. 

also dragging @volfied in to this - he had a concern with ian's never-propagate suggestion.

volfied:
Actually, I rescind my concern. It should be safe to add a NEVER_PROPAGATE policy on top when dialogs are open, as it will z-order override the other area policies.

Policy overlay won't visualize it though, I use it only for area policies.

mcilroyc:
booya

PR #2519, opened 2015-09-17
During modal JS dialogs, disable scrim interactions, hide overlays, a…

volfied:
…nd keep mouse events to us

Addresses #2399 , #2138

### DIFF
--- a/src/js/actions/dialog.js
+++ b/src/js/actions/dialog.js
@@ -24,8 +24,23 @@
 define(function (require, exports) {
     "use strict";
 
+    var Promise = require("bluebird");
+
+    var adapterUI = require("adapter/ps/ui"),
+        adapterOS = require("adapter/os");
+
     var events = require("../events"),
-        locks = require("js/locks");
+        policy = require("./policy"),
+        locks = require("js/locks"),
+        EventPolicy = require("js/models/eventpolicy"),
+        PointerEventPolicy = EventPolicy.PointerEventPolicy;
+
+    /**
+     * Never propagate policy ID of each open dialog
+     *
+     * @type {{string: number}}
+     */
+    var _policyMap = {};
 
     /**
      * Open a dialog with a given ID and optional dismissal policy.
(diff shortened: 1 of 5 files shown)


mcilroyc:
I think I found an issue.  When the color picker is open, you can't drag guides (even thought cursor changes indicating that you ought to be able to).

volfied:
@mcilroyc: Cursor problem can't be fixed in JS, and is a prevalent problem with most of our never_propagate policies.

However, I made another change to add never_propagate only for modal dialogs, so color pickers won't cause never_propagate overlay. Back to you

src/js/actions/dialog.js
```diff
@@ -38,14 +53,28 @@ define(function (require, exports) {
...
+            ),
+            isModal = this.flux.store("dialog").isModalDialog(id),
+            dispatchPromise = this.dispatchAsync(events.dialog.OPEN_DIALOG, payload),
+            policyPromise = isModal ?
+                this.transfer(policy.addPointerPolicies, [neverPropagatePolicy]) :
+                Promise.resolve();
 
-        return this.dispatchAsync(events.dialog.OPEN_DIALOG, payload);
+        return Promise.join(dispatchPromise, policyPromise, function (dispatchResult, policyResult) {
+            if (isModal) {
```
mcilroyc: nit/suggestion: this might look better if you just attach this as a `.then()` of the policyPromise.  Along the same lines, I think this might be more readable (above), and avoid a never-used instantiation of the neverPropagatePolicy when not modal,  if you separate out one single branch for the `isModal` code path that sets up the policyPromise accordingly, or `resolve` if not.

src/js/actions/dialog.js
```diff
@@ -24,8 +24,23 @@
...
+
+    /**
+     * Never propagate policy ID of each open dialog
+     *
+     * @type {{string: number}}
+     */
+    var _policyMap = {};
 
     /**
      * Open a dialog with a given ID and optional dismissal policy.
```
mcilroyc: might be worth noting the new policy magic here in the JSDoc

mcilroyc:
@volfied thanks - I think this clears up the issue I described.  I have two nits/suggestions, but otherwise this is merge-able.

volfied:
@mcilroyc back to you, thanks for reviewing.

PR #2527, opened 2015-09-18
On place events, replace if layer already exists

volfied:
We used to just skip calling addLayers, but this caused problems like #2484.

I tested this with couple places/drags, and it seems to work fine, and not break other place events.


iwehrman:
You break it, you bought it!

Issue #2494, opened 2015-09-16
Inconsistent delete icon hover state


iwehrman:
With layer effects, the delete icon only appears on hover. With export, it's always visible.
![image](https://cloud.githubusercontent.com/assets/1075154/9914603/7467ac22-5c67-11e5-8892-7baefe875d1e.png)
My preference is definitely to have the icons in the right-most column only visible on hover.

iwehrman:
Fixed by #2528.

PR #2528, opened 2015-09-18
Made trash/visibility icons consistent

designedbyscience:
Now the effects icons for trash/visibility appear all the time...Sorry @iwehrman 

Addresses #2494


iwehrman:
I shall retreat from this battle, but the war is not over! However, these effects icons should have the same hover state highlighting as in the exports panel. So, if you fix that, I'll merge this now and then try to sneakily change the overall visibility behavior at a later date.

designedbyscience:
Okay! Changes made

1funny191

iwehrman:
:walking:

PR #2532, opened 2015-09-18
Fetch extension data during resetLayers

mcilroyc:
This addresses an unreported, occasional bug.  Sometimes when you move an artboard, the subsequent resetLayers was discarding the "exportEnabled" flag for the layer.


iwehrman:
LGTM.

Issue #2473, opened 2015-09-14
Prefix mismatch in Export


ktaki:
The settings below will generate 003Artboard 1@1x.png and 005Artboard 2@1x.png. The prefix numbering will change depending on the number of assets (the right pane) generated.

![screen shot 2015-09-14 at 3 58 47 pm](https://cloud.githubusercontent.com/assets/2029597/9863988/e15bd35e-5af9-11e5-93e2-97d8c9d5cf26.png)

ktaki:
Verified the fix.

PR #2533, opened 2015-09-18
Better artboard export prefix handling

mcilroyc:
This fixes a pretty brain-dead mistake in the way artboard prefixes were being derived.  Now, in the react render() method after sorting artboards, a map of layer>prefix is created (if necessary).  That map is provided to the "export all" action.  This maintains consistency of prefixes across the UI, and the actual export process.... but the prefixes are still transient (not persisted to PS metadata, nor maintained in our store which would require a lot more overhead of trying to keep up with AB positions on canvas).  Addresses #2473

bonus: Also add the "Use Prefix" label as a translatable string

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -176,7 +176,7 @@ define(function (require, exports) {
             fileName = baseName + asset.suffix,
             _layers = layer ? Immutable.List.of(layer) : null;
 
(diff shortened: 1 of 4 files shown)


mcilroyc:
FYI - I snuck in a last minute change per @davealonzo  to add a dash after the prefix (in the filename)

volfied:
LGTM

PR #2535, opened 2015-09-18
CC Libraries - improve character style titles in the panel and search result

shaoshing:
This PR displays leading and tracking of character styles in the libraries panel, and fix their title in the search result. It also extracts the code of formatting character style into `util/libraries` to avoid duplicates in search and libraries panel, and make them less error prone..

screen shot 2015-09-18 at 11 27 14 am

screen shot 2015-09-18 at 11 27 36 am


volfied:
Clean.

Issue #2512, opened 2015-09-17
Unable to select Libraries datalist items; can't change libraries


chadrolfs:
![screen shot 2015-09-17 at 1 54 49 pm](https://cloud.githubusercontent.com/assets/8442631/9945745/f632f8a6-5d43-11e5-8509-d51356c0847b.png)

iwehrman:
It looks like somehow the blur handler of the text input inside the libraries datalist is executing before the select onClick handler can run...

iwehrman:
Here's the cause from #2486: https://github.com/adobe-photoshop/spaces-design/commit/a6f60d9e0f9d15229ad83bd98be994a856155824#diff-2679935a0e56611617e48e9185145f97R696

Looking for a "good" solution to the original problem he found...

mcilroyc:
FBNC to @chadrolfs

Issue #2409, opened 2015-09-11
Blend mode for type layer is not updated till other field is touched


ktaki:
Create a type layer and change the blend mode. It does not change till you change one of other fields such as opacity, font name, or font size.

mcilroyc:
'twas merely a flesh wound.  FBNC to @ktaki

ktaki:
Verified the fix.

Issue #2266, opened 2015-08-28
Clicking off the datalist will select whatever is hovered on


placegraphichere:
When live datalists are used (export datalists, font weight datalist maybe others) 

Rolling off the datalist entry and clicking outside the datalist will still apply the hovered item in the datalist. Expected would be to cancel the datalist and return to the selection before it was clicked.

chadrolfs:
@shaoshing fixed a bug exactly like this with CC Libraries drop down.

chadrolfs:
reviewed - not M3

iwehrman:
I think if #3667 is addressed (e.g., by #3787) then this is not a ship blocker.

PR #2536, opened 2015-09-18
Various Datalist event handling fixes

iwehrman:
1. Ensure that tabbing away from a focused `Datalist` does not open the `Datalist`. Once a `Datalist` is focused, start typing, or use the up/down arrows to open the `Datalist`.
2. Ensure that changing the blend mode updates the `Datalist` in `Type` component.
3. Ensure that the selections can be made in the `LibraryList` `Datalist` can be made either by pressing enter or or clicking to commit.
4. Ensure that tabbing away from the `LibraryList` `Datalist` releases focus back to PS completely. This is the last focusable element in the UI, and otherwise the app ends up in an indeterminate focus state which causes problems with shortcuts, etc.

Remaining known issues:

1. Committing a `Datalist` selection with enter should leave the `Datalist` focused, but it doesn't.
2. Shift-tabbing away from the `LibraryList` `Datalist` should probably _not_ release focus entirely, but it's also not clear that the user would really intend for a shift-tab to do something meaningful in this case...
3. There are lots of weirdnesses with navigating the `LibraryList` `Datalist` due to the distinct sections of the menu.

Addresses #2512 and #2409.

### DIFF
--- a/src/js/jsx/sections/libraries/LibraryList.jsx
+++ b/src/js/jsx/sections/libraries/LibraryList.jsx
@@ -289,6 +289,7 @@ define(function (require, exports, module) {
 
             return (<div className="libraries__bar__top__content">
                 <Datalist
+                    ref="list"
                     list={listID}
                     className="dialog-libraries"
                     options={listOptions}
@@ -296,6 +297,7 @@ define(function (require, exports, module) {
                     live={false}
                     autoSelect={false}
                     onChange={this._handleChangeLibrary}
+                    releaseOnBlur={true}
                     defaultSelected={selectedLibraryID}
                     disabled={this.props.disabled} />
                 <SplitButtonList className="libraries__split-button-list">
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -65,7 +65,9 @@ define(function (require, exports, module) {
             // IDs of items that, when selected, won't close the dialog
             dontCloseDialogIDs: React.PropTypes.arrayOf(React.PropTypes.string),
             // SVG class for an icon to show next to the Text Input
-            filterIcon: React.PropTypes.string
+            filterIcon: React.PropTypes.string,
+            // Release keyboard focus when the text input is blurred.
+            releaseOnBlur: React.PropTypes.bool
         },
 
         getDefaultProps: function () {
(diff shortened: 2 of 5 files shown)


placegraphichere:
Any chance this will address #2266 ?

iwehrman:
This will not address #2266. That behavior is as-designed (originally) for "live" datalists, which include type and blend modes, but not Libraries. We could make those datalists not "live", or we could make live datalists more complicated so that the changes that are effected before committing are in some sense just previewing the changes. The former is trivial and latter is non-trivial.

mcilroyc:
Feel free to say this is totally unrelated, but when you choose "Create New Library" it stinks that you can't immediately start typing the new name

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -296,6 +307,7 @@ define(function (require, exports, module) {
                     live={false}
                     autoSelect={false}
                     onChange={this._handleChangeLibrary}
+                    onBlur={this._handleLibraryListBlur}
```
mcilroyc: seems slightly odd that you have to register a handler, just to in turn call a method on itself.  Like, maybe this would also work as an optional property ~ `releaseOnBlur`.  If there's one thing Datalist needs more of, its optional parameters.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -345,9 +347,15 @@ define(function (require, exports, module) {
...
+        /**
+         * Deactivates the datalist when the select menu is closed.
+         *
+         * @private
+         * @param {SyntheticEvent} event
+         * @param {string} action Either "apply" or "cancel"
+         */
         _handleSelectClose: function (event, action) {
-            if (this.props.autoSelect) {
+            if (this.props.autoSelect || action === "apply") {
```
mcilroyc: I bet this was the most fun part of this PR

mcilroyc:
Assigning back to wehrman for a few followup questions, otherwise this LGTM

iwehrman:
Agree that the "Create New Library" is a separate issue. I'll file it.

src/js/jsx/sections/libraries/LibraryList.jsx
iwehrman: OK, sure. Commit amended.

src/js/jsx/shared/Datalist.jsx
iwehrman: You don't even know.

iwehrman:
Back 2 @mcilroyc.

iwehrman:
See #2540 for the related "Create New Library" issue.

PR #2542, opened 2015-09-18
UI: Zoom status in toolbar

placegraphichere:
- Visuals ONLY, needs to wired up
- added Zoom.jsx for feature
- added zoom to the toolbar
- styled it up
- ready for @volfied to wire up/review

### DIFF
--- a/src/js/jsx/Zoom.jsx
+++ b/src/js/jsx/Zoom.jsx
@@ -42,6 +42,10 @@ define(function (require, exports, module) {
             };
         },
 
+        shouldComponentUpdate: function (nextProps, nextState) {
+            return (this.state.zoom !== nextState.zoom);
+        },
+
         /**
          * Handle the change of zoom input, the actions will take care
          * of making sure the zoom is at an acceptable range
@@ -58,13 +62,30 @@ define(function (require, exports, module) {
             this.getFlux().actions.ui.zoom(payload);
         },
 
+        /**
+         * If tab key is pressed, we don't want focus to go to some other bar/panel
+         * so we blur this element and send focus back to PS
+         *
+         * @private
+         * @param {SyntheticEvent} event
+         */
+        _handleKeyDown: function (event) {
+            if (event.key === "Tab") {
+                React.findDOMNode(this.refs.input).blur();
+                event.preventDefault();
+            }
+        },
+
         render: function () {
             return (
                 <div className="zoom">
                     <NumberInput
+                        tabIndex="-1"
+                        ref="input"
                         suffix="%"
                         value={this.state.zoom}
                         onChange={this._handleZoomChange}
+                        onKeyDown={this._handleKeyDown}
                     />
                 </div>
             );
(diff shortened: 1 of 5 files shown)


iwehrman:
If you swap that `input` with a `NumberInput` in the JSX, it will be easier to style it in a way that is consistent with the other number inputs, since that's the next thing that has to happen anyway.

Also, we'll probably have to elaborate `NumberInput` to always tack on a `%` at the end. A long time ago we had support for that...

volfied:
I'll take care of the NumberInput as part of my work, and implement suffix as necessary

volfied:
@iwehrman Can you please review, I hooked the button up, added `suffix` to NumberInput, and fixed some JSDocs I saw on the way.

iwehrman:
Some complaints:

1. I suspect that this should be styled using the existing `number-input` class + a class specific to the zoom input instead of with a selector that refers directly to an `input`?
2. The styling of the selected state is inconsistent with the other `NumberInput` instances. The selected state in particular doesn't look quite right.
3. We probably should be more aggressive than usual usual about blurring and releasing keyboard focus with this input. I don't think we should support tabbing to a different input. Or, at the very least, we should make sure that the tab behavior makes sense and doesn't, e.g., end up somewhere unexpected. Right now it ends up in the Libraries `Datalist`, which seems worse than having tab just blur and release focus when it loses focus. 
4. When you change the value, the old value comes back before being replaced again by the new value. Our other `NumberInputs` don't suffer this issue.
5. Typing in a number with a trailing % results in a failure.
6. The `NumberInput` seems to appear in the lower left even when the toolbar is not pinned. (Maybe that's a feature?)

But, besides that, it's great! :sunglasses:

volfied:
@placegraphichere I will address 3,4 and 5. Can you please address 1, 2 and 6 here?

volfied:
@iwehrman I've fixed the behavior issues, for styling, please continue this with @placegraphichere

placegraphichere:
@volfied I'm wondering if we can avoid showing "0%" on document load?

placegraphichere:
I've fixed 1 and 6. I think 2 should be covered in the spectrum branch

volfied:
@iwehrman Ready for your reviewing pleasure. I've disabled the zoom widget when zoom is 0 (document load, no document)

src/js/jsx/Zoom.jsx
```diff
@@ -0,0 +1,100 @@
...
+        /**
+         * If tab key is pressed, we don't want focus to go to some other bar/panel
+         * so we blur this element and send focus back to PS
+         *
+         * @private
+         * @param {SyntheticEvent} event
+         */
+        _handleKeyDown: function (event) {
+            if (event.key === "Tab") {
+                React.findDOMNode(this.refs.input).blur();
```
iwehrman: Does focus need to be returned to PS here?

iwehrman:
I saw a problem with focus once, but I can't repro. So, let's just do it be on the lookout for focus-related issues. Thanks @volfied and @placegraphichere!

Issue #2434, opened 2015-09-12
Selecting layer ranges is slow


iwehrman:
1. Open VermilionArtboards.psd.
2. Collapse all layer groups.
3. Deselect all layers.
4. Click the first layer to select it.
5. Shift-click the last layer to select the range of four artboards.

This takes ~4s (debug) on a 5k iMac, which feels like an eternity. The time is all spent in the view layer handling the `SELECT_LAYERS_BY_INDEX` event. It seems related to the large number of ancestors of the selected layers; selecting ranges of leaves is fast by comparison.

PR #2543, opened 2015-09-19
Speed up layers.selectedNormalized

iwehrman:
In #2434 I observed that selection of layer ranges with many ancestors was very slow. A quick profiler run showed that almost all of the time was being spent in the `Flip` component's `componentWillReceiveProps` method, and in particular that method's call to `LayerStructure.prototype.selectedNormalized`. That method has an `O(m^2*n)` algorithm, for `n` layers and `m` selected layers, where an `O(m*log(m)*n)` algorithm suffices. Making this small change improves performance significantly.

However, I found it mysterious that this should be the case, since in #2434 `m` is should be very small. It turns out that there is a _different_ bug wherein PS is interpreting range selection to include all descendants, even hidden ones, in the range. This seems wrong to me, but it does explain why the change makes a pretty significant improvement. Seperately, however, I'm looking into why PS is doing this. In the meantime, this simple change stanches the bleeding somewhat.

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -426,7 +426,7 @@ define(function (require, exports, module) {
          */
         "selectedNormalized": function () {
             // For each layer, remove all it's descendants from the group
-            var selected = this.selected;
+            var selected = this.selected.toSet();
             return selected.filterNot(function (layer) {
                 return this.strictAncestors(layer)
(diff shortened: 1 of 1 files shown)


designedbyscience:
LGTM (and seems a bit faster, based on the simple test of selecting all layers in Vermilion a few times in a row)

Issue #2468, opened 2015-09-14
With Export panel expanded, collapsed Effects panel cannot be expanded


ktaki:
When Export panel is full (meaning that the contents reaches to the bottom), double clicking Effects panel header won't expand the Effects panel. You need to collapse Export panel first.

ktaki:
It seems that the height of Export dialog depends on the contents of it. I think it should expand to a half of the Effects panel. The picture below shows when 2 assets exist in Export dialog and the dialog was clicked to expand.

![screen shot 2015-09-22 at 5 53 03 pm](https://cloud.githubusercontent.com/assets/2029597/10035784/d7ce4de4-6152-11e5-9de9-0a58dabb1771.png)

placegraphichere:
see #2693

ktaki:
Verified the fix.

PR #2544, opened 2015-09-21
apply min-height on effects panel

placegraphichere:
initial fix for #2468

### DIFF
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -246,14 +246,22 @@
     flex-shrink: 1;
 }
 
+.effects.section__active {
+    min-height: 24rem;
+}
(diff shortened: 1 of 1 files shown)


src/style/panel.less
```diff
@@ -246,14 +246,22 @@
     flex-shrink: 1;
 }
 
+.effects:not(.section__collapsed) {
```
designedbyscience: You shouldn't need the :not selector here.

placegraphichere:
back to @designedbyscience 

- removed :not in favor of .section__active class

designedbyscience:
Better!

Issue #2534, opened 2015-09-18
Hitting Esc after dragging asset onto canvas removes from canvas but not layers panel


chadrolfs:
drag an asset onto canvas, hit Esc
note: asset is removed from canvas but still shows in layers panel

attempts to select it in the layers panel results in:
`Action layers.select failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Select” is not currently available.`

Issue #2524, opened 2015-09-17
Attempting to Delete a LLSO immediately after placing results in many errors


chadrolfs:
drag an asset onto canvas and immediately hit Delete

Result: ` Action layers.deleteSelected failed: Error: Host is in a modal state and the request cannot be executed.`

You end up in a bad state (but you can right-click and place)

https://www.dropbox.com/s/64kamnhhz3ckcy1/LLSO-DeleteAfterPlace.mov?dl=0

Issue #2523, opened 2015-09-17
LLSO Library assets are placed as embedded instead of linked 


chadrolfs:
Somewhat inconsistent depending on the tool or asset you are dragging onto canvas.  Sometimes it's linked, sometimes it's embedded.

https://www.dropbox.com/s/fh1edlcgcqohow4/LLSO-Embedded.mov?dl=0

chadrolfs:
`GET file:///Users/crolfs/playground-repos/spaces-design/src/img/ico-layer-5-linked-mask.svg net::ERR_FILE_NOT_FOUND`

chadrolfs:
@designedbyscience ^^  This your bug?

chadrolfs:
oh nevermind, I got that file not found on a different branch.  This bug is against master.  sry, carry on, nothing to see here

PR #2546, opened 2015-09-21
CC Libraries - add new CLSO layer through placeEvent handle

shaoshing:
The PR change the behavior on handling new CLSO layer and expanded graphic layers (drop graphic asset with ALT modifier).

For CLSO, the new layer is added through the "placeEvent" handler, to keep consistent with embedded layer (show the new layer after transform mode). It also help address issues #2534,  #2524, and #2523 (press delete / esc during transform mode)

lib-so

For expanded graphic layers, since there is no "placeEvent" emitted, we will use the previous way by calling "getLayerIDs" and "addLayers" to discover the new layers.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -783,7 +783,8 @@ define(function (require, exports) {
                     .bind(this)
                     .then(function (policyID) {
                         var docRef = docAdapter.referenceBy.id(currentDocument.id),
-                            placeObj = libraryAdapter.placeElement(docRef, element, path, location);
+                            placeObj = libraryAdapter.placeElement(docRef, element, path, location),
+                            hasAlt = this.flux.stores.modifier.getState().alt;
 
                         // This command is played "asynchronously", which means that Photoshop will
                         // allow CEF to run while the command is being played (i.e., while in the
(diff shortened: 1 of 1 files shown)


src/js/actions/libraries.js
```diff
@@ -793,28 +794,35 @@ define(function (require, exports) {
...
+                        // Dropping graphic asset with ALT/OPT modifier will expand the asset (instead of adding 
+                        // CLSO layer) and result in creating multiple layers in Photoshop. Unlike regualr drop 
+                        // event (without modifier), this will not trigger the "placeEvent" handler, so we need 
+                        // to get the new layer IDs through the difference between the previous layer IDs and the
+                        // next layer IDs.
+                        //
+                        // FIXME: we should instead get IDs back from Photoshop when layers are placed with modifier, 
+                        //        so we don't have to get all the layer IDs.
+                        if (hasAlt) {
+                            placeElementPromise.then(function () {
```
volfied: Not a fan of the structure here. If alt is being pressed, you wait for that promise to fdinish, and then update the document, but you return the placeElement promise already.

What you can do instead is: 

`placeElementPromise = placeElementPromise.then(function ...`

and `return placeElementPromise` as usual at the end.

volfied:
Got one nit @shaoshing, looks good otherwise.

shaoshing:
Thanks @volfied that's a good one! I've made the change so back to you.

volfied:
:+1:

Issue #2521, opened 2015-09-17
First artboard missing a default asset


ktaki:
Create a new artboard doc. Select Artboard 1.
    RESUTL: The export panel says "No exports configured"
    EXPECTED: The export panel says "No supported layers selected"
Add a rectangle to Artboard 1. Select Artboard 1
    RESULT: The export panel says "No exports configured"
    EXPECTED: A default asset is added

This works artboards added to the doc later.

ktaki:
The first AB created has Layer 1 as default. Therefore, an asset is added automatically. However, this empty layer causes an error when exported if no other layers were added.

New AB doc, select AB, Export.

![screen shot 2015-09-22 at 9 58 08 am](https://cloud.githubusercontent.com/assets/2029597/10026492/f1bf787c-6116-11e5-8774-baa2d3b8cd04.png)

mcilroyc:
@ktaki we've actually discussed this scenario before in a previous bug. I don't see a convenient way to guard against this, and it seems not worth trying.  The app handles this appropriately (console error, but otherwise recovers) ... and any user trying to export an artboard with the default placeholder layer deserves what they get!

mcilroyc:
@ktaki see: https://github.com/adobe-photoshop/spaces-design/issues/2273#issuecomment-139668932

ktaki:
OK. No objections. Closing.

PR #2549, opened 2015-09-21
Add default asset to artboards on new doc creation, and other artboard export conveniences

mcilroyc:
This addresses #2521 by adding a default asset during new doc creation (w/ artboard).

Furthermore, to make artboard export more convenient, especially for docs created in standard PS:
- Export All panel shows artboards even if they have no assets configured
- "Checking" an artboard that has no assets (including the "All artboards" mass checkbox at the top) will auto-add a 1x asset where needed.


volfied:
LGTM

Issue #2504, opened 2015-09-16
ALT+drag to duplicate multiple objects does not update layers panel


placegraphichere:
- Select multiple objects
- ALT+drag to duplicate them
- Layers panel will only show one layer selected
- Canvas continues to show multiple layers selected

expected: layers panel should show all layers selected

shaoshing:
The cause is at https://github.com/adobe-photoshop/spaces-design/commit/4729d94cd3309c9756ff1d7b9164192f838325a6#diff-cad1f13dd4227353281fce9ef1e32a58R690 Changing line 690 to `isNewSelected = selected` will fix the problem but I think there must be a reason for adding `i + 1 === layerIDs.length`. Summoning @mcilroyc since he is the author of the commit.

placegraphichere:
fixed

PR #2550, opened 2015-09-21
Relaxed the condition for isNewSelected in layerstructure's addLayers

mcilroyc:
This addresses #2504.  @shaoshing passed this to me because he was assuming I remembered the purpose of the second conjunct at line 891.  Alas, I do not.  Removing it fixes the issue (as he suggested) and seems to work correctly for other code paths:

1. dragging a file on to the PS icon
1. new shape layers
1. new text layers

I'd love for someone besides @shaoshing to look at this to double check both of our work.

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -888,7 +888,7 @@ define(function (require, exports, module) {
                 nextIndex = layerStructure.index,
                 descriptor = descriptors[i],
                 layerIndex = descriptor.itemIndex - 1,
-                isNewSelected = selected && i + 1 === layerIDs.length,
+                isNewSelected = selected,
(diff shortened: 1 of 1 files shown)


iwehrman:
I too cannot remember exactly what that disjunct was for. One additional testing scenario you might want to include is pasting _multiple_ layers into a situation that will cause replacement.

shaoshing:
Tested with pasting multiple layers with and without replacement; the behaves is the same with master. However, I think the result is incorrect and should file separate issues for them:

layers-paste-replace

layers-paste

shaoshing:
I am merging this and will create separate issues for the problems above. Thank you @mcilroyc

PR #2553, opened 2015-09-21
Turn on auto add/delete for pen tool

volfied:
This is a tool setting I realized is default on in PS, but not in DS, which was causing #1681 

So we now turn this feature on, and using pen tool, when you switch to a closed path/shape and try using it again, pen tool automatically starts a new layer as it should.

Addresses #1681


mcilroyc:
such return

PR #2554, opened 2015-09-21
Suspend policies during modal tool states

iwehrman:
1. Convert `disablePolicies` and `reenablePolicies` to `suspendPolicies` and `restorePolicies`, allowing policy actions to be added during the suspended state. This currently does not allow you to remove a policy during the suspended state that was added before the suspended state. This also doesn't allow you to suspect policies twice without restoring them in the middle. The former feature could be added if necessary.
2. Suspend policies while in the type modal tool state.
3. Fix a bug in which clicking a never-before-selected layer with the type tool would fail as a result of the `updateTextProperties` event being handled before the selection change, and hence before initialization.

### DIFF
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -194,10 +194,11 @@ define(function (require, exports) {
     removePointerPolicies.modal = true;
 
     /**
-     * Set the propagation mode for the given policy kind.
+     * Set the propagation mode for the given policy kind. If no mode is supplied,
+     * a default mode will be used.
      *
      * @param {number} kind
-     * @param {number} mode
+     * @param {number=} mode
      * @return {Promise}
      */
     var setMode = function (kind, mode) {
@@ -205,9 +206,15 @@ define(function (require, exports) {
         switch (kind) {
         case PolicyStore.eventKind.KEYBOARD:
             setModeFn = adapterUI.setKeyboardPropagationMode;
+            if (mode === undefined) {
+                mode = adapterUI.keyboardPropagationMode.FOCUS_PROPAGATE;
+            }
             break;
         case PolicyStore.eventKind.POINTER:
             setModeFn = adapterUI.setPointerPropagationMode;
+            if (mode === undefined) {
+                mode = adapterUI.pointerPropagationMode.ALPHA_PROPAGATE;
+            }
             break;
         default:
             return Promise.reject(new Error("Unknown kind:" + kind));
@@ -240,8 +247,7 @@ define(function (require, exports) {
         
         policyStore.suspend(kind);
 
-        var mode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
-            setModePromise = this.transfer(setMode, kind, mode),
+        var setModePromise = this.transfer(setMode, kind),
             syncPoliciesPromise = this.transfer(syncPolicies, kind);
 
         return Promise.join(setModePromise, syncPoliciesPromise);
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -27,76 +27,56 @@ define(function (require, exports) {
     var Promise = require("bluebird");
 
     var adapterUI = require("adapter/ps/ui"),
-        adapterOS = require("adapter/os"),
-        events = require("js/events"),
+        adapterOS = require("adapter/os");
+
+    var events = require("js/events"),
         locks = require("js/locks"),
         PolicyStore = require("js/stores/policy"),
         EventPolicy = require("js/models/eventpolicy"),
         KeyboardEventPolicy = EventPolicy.KeyboardEventPolicy;
 
     /**
-     * The previous policies and default modes during temporary policies disabling
-     * {
-     *     keyboardPolicies: EventPolicySet,
-     *     pointerPolicies: EventPolicySet,
-     *     keyboardMode: number,
-     *     pointerMode: number
-     * }
-     * @type {?object}
-     */
-    var _cachedPolicyState = null;
-
-    /**
      * For a given policy kind, fetch the master policy list and set it in photoshop
-     * If a policyListID is provided, then attempt to remove it on error
      *
      * @param {string} kind
-     * @param {number=} policyListID
-     *
      * @return {Promise} Resolves when polices are set in photoshop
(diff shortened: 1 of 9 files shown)


iwehrman:
Happy to chat about this, @mcilroyc. Also, please give it a bit of testing because... things.

src/js/actions/policy.js
```diff
@@ -226,67 +219,84 @@ define(function (require, exports) {
...
+                _cachedPolicyState.pointerMode = pointerMode;
+
+                return adapterUI.setPointerPropagationMode(
+                        { defaultMode: adapterUI.pointerPropagationMode.ALPHA_PROPAGATE });
+            });
+
+        var syncKeyboardPoliciesPromise = _syncPolicies.call(this, PolicyStore.eventKind.KEYBOARD),
+            syncPointerPoliciesPromise = _syncPolicies.call(this, PolicyStore.eventKind.POINTER);
+
+        return Promise.join(
```
mcilroyc: I see what you did here.  My original assumption was that there might be some interaction between setting policies, and the mode ... but I guess that's not the case, so OK!

src/js/actions/policy.js
```diff
@@ -310,7 +320,7 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var onReset = function () {
-        reenablePolicies = null;
+        _cachedPolicyState = null;
```
mcilroyc: :neckbeard:

src/js/actions/tools.js
```diff
@@ -414,6 +415,42 @@ define(function (require, exports) {
...
+        if (event.kind._value === "mouse" && event.tool && event.tool.ID === "ArtT") {
+            cloakPromise = this.transfer(ui.cloak);
+        }
+
+        return Promise.join(changeStatePromise, policyPromise, cloakPromise);
+    };
+    handleToolModalStateChanged.reads = [];
+    handleToolModalStateChanged.writes = [];
+    handleToolModalStateChanged.transfers = [policy.suspendPolicies, policy.restorePolicies,
+        changeModalState, "ui.cloak"];
```
mcilroyc: For consistency, either change to `"policy."`  or `ui.cloak`

iwehrman:
@mcilroyc: I've pushed a second commit that significantly reworks policy suspension and restoration. Previously it was not possible for policy IDs installed before suspension to be removed during suspension, and this limitation was causing problems. For example, when exiting out of the superselect type sub-tool, we would attempt to swap policies before the toolModalStateChange event was processed, which lead to a `removePolicies` failure. In the new world, you can both add and remove policies while policies are suspended, and the right thing happens upon restoration: removed policies are not restored, but added policies are. To accomplish this, I moved all of the state that was previously in the policy action module into the action store. The store itself now has `suspend` and `restore` methods which handle moving aside the installed policies and creating the correct set upon restoration. While I was in there, I made a few other cleanups as well:

1. There is now a `policy.setMode` action which does what it says. It also emits a new `MODE_CHANGED` event which the policy store listens for. The policy store tracks the mode so that the `suspend` and `restore` functions can roll back and restore modes as well as policies without any module state.
2. I converted all the helper functions into full-blown actions, which I think is preferable. I also removed crufty error handling from `syncPolicies` because that doesn't really make sense in the modern world, in which stores maintain state and all state is reset whenever there's an error. (There's also no reason to think that a `syncPolicies` failure is something that should be simple to recover from.)
3. `suspendPolicies` and `resumePolicies` are now parametrized by the event kind, which makes the most sense for code re-use. I added new actions `suspendAllPolicies` and `restoreAllPolicies` which just call the eponymous actions for both keyboard and pointer policies.

By the way, you were previously wondering about the counter manipulation. Here's what that's all about: When we move aside the old event policy sets and install a new one, it's still possible to add new policy lists. We want to make sure that when new policy lists are added, their IDs won't conflict with the existing IDs from the suspended policies, otherwise when we try to merge them back together during the restore operation there will be ugly conflicts. The simple solution I came up with was just to have the fresh policy set installed during suspension inherit the counter value from the suspended policy set, so any new policy IDs will necessarily not conflict with any previous ones.

Back to @mcilroyc!

mcilroyc:
Looks great.  Nice to have all state ripped out of the action.  I'll do some light testing in the morning before merging.

mcilroyc:
I can repro the following behavior on your branch.  It works as expected on master.

- Export (which pops a folder chooser dialog)
- Click New Folder Button
- Attempt to type in the box to name the new folder
- Result: can not type

src/js/actions/policy.js
```diff
@@ -204,127 +186,171 @@ define(function (require, exports) {
...
-            return Promise.reject(new Error("Can not re-renable keyboard policies, no previously cached state"));
+    var suspendPolicies = function (kind) {
+        var policyStore = this.flux.store("policy");
+        if (policyStore.isSuspended(kind)) {
+            return Promise.reject(new Error("Policies are already suspended"));
         }
+        
+        policyStore.suspend(kind);
+
+        var mode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
```
mcilroyc: 1. It seems a little weird to set the keyboardPropMode inside this generic policy function, since it will get called for any "kind" of policy
2. The default value used to be FOCUS_PROPAGATE, right?  Is this change due to fitz's recent changes in core?  If so, do you know the minimum required build with that?
3. It is also suspicious that you are not somewhere in the `suspend` process setting the default pointer mode ... but maybe that's because its the default value anyway, as you noted in beforeStartup...but anyway I'm not convinced it's not an omission.

mcilroyc:
I can repro the previous problem on 674, so assigning back to @iwehrman for a look-see

src/js/actions/policy.js
iwehrman: Yeah, I think I missed a generalization step here. Back to the well...

iwehrman:
OK, I fixed the problem in `suspend` by 1) making the `mode` param in `setMode` optional and having an event-specific default (FOCUS for keyboard, ALPHA for pointer), and 2) omitting the mode in `suspend`. Back to @mcilroyc. Fingers crossed!

mcilroyc:
bombs away

Issue #2400, opened 2015-09-11
Export dialog remains open when new file dragged into PS 


DivyaPrabhakar:
![Exports dialog still open](https://www.dropbox.com/s/ebd1tfvvp4zd839/exports_still_open.gif?raw=1)

mcilroyc:
I refute this being high priority, but I guess I wasn't at the meeting, so...

mcilroyc:
@ktaki if you run in to issues when verifying this fix, please make sure it is not caused by this separate issue: #2563

ktaki:
Verified the fix.

PR #2561, opened 2015-09-21
Properly merge dismissal policy during dialog open of previously registered dialog

mcilroyc:
This addresses #2400.

Because the dialog state is a nested map, it must be merged deeply.  Otherwise, an undefined dismissalPolicy blows away the previous policy instead of the desired behavior of leaving it untouched.

Note that dragging files onto the PS icon (to open as a new document) seems to also suffer from some weird issues, especially if layers are selected?  Anyway, when it works, then this change addresses 2400.


volfied:
That explains it.

Issue #2548, opened 2015-09-21
Errors closing CC Lib asset after editing


chadrolfs:
drag an asset onto canvas
double-click to open/edit
change it and close (click save to prompt)

Result: `Action documents.close failed: TypeError: undefined is not a function`

shaoshing:
@chadrolfs once this is fixed, you will need latest Jenkin build (>= 676) in order the have the graphic's preview updated correctly to the changes via the close-and-save dialog.

iwehrman:
FBNC to @chadrolfs.

Issue #2565, opened 2015-09-22
Unhandled exception in libraries store when closing document


iwehrman:
1. Open a new artboard document.
2. Close the document:

```
Executing action documents.close after waiting 1ms; 1/0
fluxcontroller.js:563 Action documents.close failed: TypeError: Cannot read property 'get' of null
    at Fluxxor.createStore._handleClosedGraphicDocument (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/src/js/stores/library.js:344:58)
    at Store.__handleAction__ (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:618:16)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:167:41)
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:2057:20
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:3698:12
    at baseForOwn (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:2175:11)
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:2199:13
    at Dispatcher.doDispatchLoop (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:150:4)
    at Dispatcher._dispatch (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:139:11)
    at defaultDispatchInterceptor [as dispatchInterceptor] (file:///Users/wehrman/Desktop/photoshop-Release_x86_64%202/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:93:4)
[...]

PR #2576, opened 2015-09-22
CC Libraries: resolve issues related to graphic editing.

shaoshing:
This PR fixed the `deleteGraphicTempFiles` transfer due to a typo in my previous PR. It also fixed a possible circular dependency issue when resolving action dependencies, by checking the dependency stack. 

Resolved issues: #2548 #2565

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -221,7 +221,16 @@ define(function (require, exports, module) {
     FluxController.prototype._initActionLocks = function () {
         var actionDependencies = new Map();
 
-        var _resolveDependencies = function (action) {
+        var _resolveDependencies = function (action, stack) {
+            stack = stack || new Set();
+            
+            // Check the stack to prevent circular dependency between actions (e.g. A -> B -> C -> A).
+            if (stack.has(action)) {
+                return new Set();
+            }
+            
+            stack.add(action);
+            
             // Map from an action to the set of all unsynchonrized actions to
             // which it may transfer, including via sub-action transfers.
             var dependencies = actionDependencies.get(action);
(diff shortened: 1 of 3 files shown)


src/js/fluxcontroller.js
```diff
@@ -221,7 +221,16 @@ define(function (require, exports, module) {
     FluxController.prototype._initActionLocks = function () {
         var actionDependencies = new Map();
 
-        var _resolveDependencies = function (action) {
+        var _resolveDependencies = function (action, stack) {
+            stack = stack || [];
```
iwehrman: I like this change! But I think it would make more sense for this to be a `Set` also? Then you can do a `log n` `has` membership check instead of an `O(n)` `contains` check. Also, we should document the parameters and return type of this function while we're in here.

src/js/fluxcontroller.js
```diff
@@ -221,7 +221,16 @@ define(function (require, exports, module) {
...
+            
+            // Check the stack to prevent circular dependency between actions (e.g. A -> B -> C -> A).
+            if (stack.indexOf(action) !== -1) {
+                return new Set([]);
```
iwehrman: I think `new Set()` is equivalent to `new Set([])`.

iwehrman:
Back to @shaoshing for a slight simplification. Thanks for adding this feature to the controller!

shaoshing:
@iwehrman I'v addressed your comments, but I am working at another issues related to the close-doc dialog. Will update this PR later.

shaoshing:
Addressed your comments @iwehrman, please review again.

iwehrman:
Brilliant.

PR #2577, opened 2015-09-22
Implement the must have PS shortcuts

volfied:
Implements some of the shortcuts/actions described in the most used PS shortcuts card. 

Did not implement merge layers yet, because it's going to take longer than these and will be riskier. So please review this PR separately.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1258,7 +1258,7 @@ define(function (require, exports) {
 
         return Promise.all(showPromises);
     };
-    setVisibilitySelectedInCurrentDocument.reads = [locks.JS_DOC, locks.PS_DOC];
+    setVisibilitySelectedInCurrentDocument.reads = [locks.JS_DOC];
     setVisibilitySelectedInCurrentDocument.writes = [];
     setVisibilitySelectedInCurrentDocument.transfers = [setVisibility];
 
@@ -1451,12 +1451,13 @@ define(function (require, exports) {
      * @param {Document} document Document for which layers should be reordered
      * @param {number|Immutable.Iterable.<number>} layerSpec Either an ID of single layer that
      *  the selection is based on, or an array of such layer IDs
-     * @param {number} targetIndex Target index where to drop the layers
+     * @param {number|string} target If a number, the target index to move layers to
+     *                               If a string, the reference enum type ("front", "back", "previous", "next")
      *
      * @return {Promise} Resolves to the new ordered IDs of layers as well as what layers should be selected
      *, or rejects if targetIndex is invalid, as example when it is a child of one of the layers in layer spec
      **/
-    var reorderLayers = function (document, layerSpec, targetIndex) {
+    var reorderLayers = function (document, layerSpec, target) {
         if (!Immutable.Iterable.isIterable(layerSpec)) {
             layerSpec = Immutable.List.of(layerSpec);
         }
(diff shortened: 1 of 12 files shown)


src/js/actions/layers.js
```diff
@@ -1239,6 +1239,30 @@ define(function (require, exports) {
...
+            return Promise.resolve();
+        }
+
+        var showPromises = currentDocument.layers.selected.map(function (layer) {
+            return this.transfer(setVisibility, currentDocument, layer, visible);
+        }, this).toArray();
+
+        return Promise.all(showPromises);
+    };
+    setVisibilitySelectedInCurrentDocument.reads = [locks.JS_DOC, locks.PS_DOC];
```
mcilroyc: you're not actually reading the PS_DOC, right?   I know it doesn't actually matter here because of the transfer, but asking anyway.

src/js/actions/layers.js
```diff
@@ -1465,6 +1489,130 @@ define(function (require, exports) {
...
+        if (document && layers === undefined) {
+            layers = document.layers.selectedNormalized;
+        }
+
+        if (!document || !layers) {
+            return Promise.resolve();
+        }
+
+        var layerIDs = collection.pluck(layers, "id"),
+            targetIndex = 0;
```
mcilroyc: note to self: is this safe w/ & w/out background?

src/js/actions/layers.js
```diff
@@ -1770,12 +1918,26 @@ define(function (require, exports) {
...
+            
+        if (document === undefined) {
+            document = applicationStore.getCurrentDocument();
+        }
+
+        if (fromDocument === undefined) {
+            fromDocument = applicationStore.getCurrentDocument();
+        }
+
+        if (fromLayers === undefined) {
```
mcilroyc: need to guard against no-current-doc case?

src/nls/de/menu.json
```diff
@@ -91,6 +91,7 @@
         },
         "CREATE_CLIPPING_MASK": "Schnittmaske erstellen",
         "CREATE_ARTBOARD": "Neue Zeichenfläche erstellen",
+        "DUPLICATE": "Auswahl duplizieren",
```
mcilroyc: why is this in your commit?

mcilroyc:
review complete.  I'll merge unless any of my mild comments moves you to change anything.

src/nls/de/menu.json
volfied: We already had duplicate string in EDIT, so I copied it across to Arrange as well, to save l10n some time.

volfied:
@mcilroyc Thank you for review, I've switched from calculating the index to using PS wrappers, and catching errors (because PS "move" breaks on layer tree edges). 

Please review/merge https://github.com/adobe-photoshop/spaces-adapter/pull/150 before accepting this one.

src/js/actions/layers.js
volfied: The menu item itself has enable rule to not allow no document cases.

src/js/actions/layers.js
volfied: Switched to use references instead of indices.

src/js/actions/layers.js
volfied: True

src/nls/de/menu.json
mcilroyc: :star2:

PR #2578, opened 2015-09-22
Update panel sizes on display configuration change

iwehrman:
We currently call `ui.updatePanelSizes` when we receive a `displayConfigurationChanged` event from the adapter, but this may arrive before the browser has redrawn the UI with the new UI sizes. This changes the panels to call debounced helper function to update the panel sizes only after 500ms have passed without a new `displayConfigurationChanged` event.

With this change, it's possible to have drag the Design Space application window from one display to another and have the on-canvas UI (e.g., guides overlays) continue to work correctly.

### DIFF
--- a/src/js/jsx/Toolbar.jsx
+++ b/src/js/jsx/Toolbar.jsx
@@ -30,8 +30,11 @@ define(function (require, exports, module) {
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
         classnames = require("classnames");
 
+    var os = require("adapter/os");
+
     var ToolbarIcon = require("jsx!js/jsx/ToolbarIcon"),
-        Zoom = require("jsx!js/jsx/Zoom");
+        Zoom = require("jsx!js/jsx/Zoom"),
+        synchronization = require("js/util/synchronization");
 
     var Toolbar = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("tool", "application", "preferences")],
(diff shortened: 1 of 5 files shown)


src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -77,7 +78,7 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            return !Immutable.is(this.state.uiState, nextState.uiState) ||
+            return !_.isEqual(this.state.uiState, nextState.uiState) ||
```
iwehrman: This was an unrelated bug; `uiState` is not an immutable object. At some point we should convert it to be an immutable object, though!

src/js/jsx/Toolbar.jsx
```diff
@@ -95,9 +98,15 @@ define(function (require, exports, module) {
 
         // On startup, we want to make sure center offsets take pinned toolbar
         componentDidMount: function () {
+            this._updateToolbarWidthDebounced = synchronization.debounce(this._updateToolbarWidth, this, 500);
+            os.addListener("displayConfigurationChanged", this._updateToolbarWidthDebounced);
```
iwehrman: Whoopsie daisy: the toolbar was previously not updating it's dimensions at all on `displayConfigurationChange`.

src/js/jsx/Toolbar.jsx
```diff
@@ -95,9 +98,15 @@ define(function (require, exports, module) {
...
             this._updateToolbarWidth();
         },
 
+        componentWillUnmount: function () {
+            os.removeListener("displayConfigurationChanged", this._updatePanelSizesDebounced);
```
volfied: I think this is supposed to remove updateToolbarWidthDebounced. Bad c/p :)

iwehrman: Derp.

Issue #2572, opened 2015-09-22
Sampler trace errors 


placegraphichere:
3 Error messages (may be the same issue multiple times) when attempting to sample text-to-text and shift click sample:

```
Export: Found configured port: 59596
fluxcontroller.js:553 Finished action type.initFontList in 303ms with RTT 325ms; 2/3
fluxcontroller.js:543 Executing action type.initFontList after waiting 305ms; 2/2
export.js:862 Export: Generator plugin connection established
fluxcontroller.js:553 Finished action export.afterStartup in 316ms with RTT 316ms; 2/2
fluxcontroller.js:553 Finished action type.initFontList in 193ms with RTT 498ms; 1/2
fluxcontroller.js:543 Executing action ui.afterStartup after waiting 491ms; 1/1
fluxcontroller.js:553 Finished action ui.afterStartup in 8ms with RTT 499ms; 1/1
fluxcontroller.js:543 Executing action application.afterStartup after waiting 500ms; 1/0
fluxcontroller.js:522 Enqueuing action ui.updatePanelSizes; 1/0
fluxcontroller.js:553 Finished action application.afterStartup in 113ms with RTT 613ms; 1/1
fluxcontroller.js:543 Executing action ui.updatePanelSizes after waiting 36ms; 1/0
fluxcontroller.js:553 Finished action ui.updatePanelSizes in 146ms with RTT 182ms; 1/0
fluxcontroller.js:522 Enqueuing action sampler.showHUD; 0/0
fluxcontroller.js:543 Executing action sampler.showHUD after waiting 1ms; 1/0
fluxcontroller.js:563 Action sampler.showHUD failed: TypeError: Cannot read property 'characterStyle' of null
    at Object._getSourceColor (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/sampler.js:77:41)
    at Object._calculateSampleTypes (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/sampler.js:105:32)
    at Object.<anonymous> (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/sampler.js:403:46)
From previous event:
    at Object.showHUD (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/sampler.js:390:14)
[...]

PR #2582, opened 2015-09-22
Initialize uninited layers during sample operations

volfied:
Addresses #2572, @iwehrman please review since you've suggested the solution.


iwehrman:
Looks good!

Issue #2580, opened 2015-09-22
CMD+f not working on master


placegraphichere:
Dev build 664+master does not allow me to type anything in app search, exits search. Console shows no errors
 
![search-not-clickable](https://cloud.githubusercontent.com/assets/3229167/10033085/d2c0ca12-614c-11e5-9415-e4c1844d7049.gif)

 ```Enqueuing action dialog.openDialog; 0/0
fluxcontroller.js:543 Executing action dialog.openDialog after waiting 1ms; 1/0
fluxcontroller.js:522 Enqueuing action shortcuts.removeShortcut; 1/0
fluxcontroller.js:553 Finished action dialog.openDialog in 24ms with RTT 25ms; 1/1
fluxcontroller.js:543 Executing action shortcuts.removeShortcut after waiting 5ms; 1/0
fluxcontroller.js:553 Finished action shortcuts.removeShortcut in 18ms with RTT 23ms; 1/0```

chadrolfs:
de020a7f2f07b0907b5a021d87513f4b8783e7a2 is the first bad commit
commit de020a7f2f07b0907b5a021d87513f4b8783e7a2
Author: Cory McIlroy <mcilroy@adobe.com>
Date:   Mon Sep 21 18:25:43 2015 -0500

    Properly merge dismissal policy during dialog open of previously registered dialog

mcilroyc:
I'm going to close this, taking @placegraphichere's PR merge as an indication that he, the reporter, agree's it is fixed.

PR #2583, opened 2015-09-22
Turn off dismissOnDialogOpen for search dialog as it has a sub dialog

volfied:
Addresses #2580, while keeping @mcilroyc's change. Problem was, `dismissOnDialogOpen` is true by default, and search dialog has a sub dialog, which mcilroy's recent fix caused this buggy behavior to show.

@iwehrman please review since it's urgent.


iwehrman:
This _looks_ fine to me. Re-assigning to @placegraphichere to test and, if the fix solves his problem, to merge.

PR #2588, opened 2015-09-23
Listen to pathComponentSelectionChanged events in vector tool

volfied:
During vector edit tool, any time a path selection changes, we now get an event with the ids of paths being selected. If there is no path selected, we switch back to select tool.

Requires pdgv.666 or later.


mcilroyc:
:six: :six: :six:

Issue #2592, opened 2015-09-23
Zoom can be set to < 5% or > 3200%


ktaki:
Using down arrow and up arrow, the zoom level number can be set to out side of 5% and 3200%. Also, entering number outside of the range (including negative numbers) from keyboard is accepted if you type it more than once consecutively.

mcilroyc:
@ktaki note that in addition to fixing the bug, arrow keys are now disabled for this input.

ktaki:
Verified the fix. up and down arrow keys have been disabled. Number input has been adjusted.

PR #2595, opened 2015-09-23
Disable stepping on 0 step for number input, and limit zoom UI

volfied:
Addresses #2592 

On NumberInput, if step is provided as 0, we disable stepping. Default is 1, so this shouldn't affect other NumberInputs.

Also, set the min/max of Zoom UI to 5,3200 so it is within the acceptable realm of zoom values, and disabled stepping for it.


mcilroyc:
I'm not understanding why we are entirely disabling stepping for the zoom number input.  Was that a design decision outside of this referenced bug?

volfied:
That was entirely my decision. I don't think ability to increase 100% zoom by 1% at a time has any value, and user has much better known keyboard shortcuts to increase zoom at any point.

mcilroyc:
decision is now documented!

Issue #2121, opened 2015-08-19
No way to remove layer effects applied from CC Lib within DS


ktaki:
Examples: Color overlay, Gradient overlay.

chadrolfs:
The workarounds are:
1. switch to classic PS and remove them there
2. apply a different CC Library effect that doesn't have those (effectively over writing and removing them)

ktaki:
Good suggestion @chadrolfs !

placegraphichere:
I think we should show a placeholder item, without options that is labeled with the effect name. Bonus points if we allow users to delete them. This is similar to unsupported brushes in libraries. We should not change anything, instead simply acknowledge the item for the user

chadrolfs:
Feature work required ^^

iwehrman:
FBNC to @ktaki.

ktaki:
Is it difficult to add visibility icon next to trash can? People do not want to delete but just want to hide sometimes to see the changes in appearance. @placegraphichere How do you think?

iwehrman:
@ktaki There is now a visibility icon next to unsupported layer effects. FBNC 2 U!

ktaki:
Looks great!

PR #2602, opened 2015-09-23
Effects Panel: display unsupported layer effects

shaoshing:
**requires adapter branch [shao/layereffect] to work**

This PR displays unsupported layer effects and allows deleting them (#2121). This new feature is compatible with the copy/paste style buttons and the sampler tool.

layers-effects


src/js/jsx/sections/style/UnsupportedEffectList.jsx
```diff
@@ -0,0 +1,142 @@
...
+                            this, effect.type, effect.index);
+                    
+                    return (
+                        <div className="unsupported-effect-list__effect">
+                            <div className="unsupported-effect-list__effect__title">
+                                {effect.title}
+                                <span className="unsupported-effect-list__effect__index">{indexStr}</span>
+                            </div>
+                            <ToggleButton
+                                title="Delete"
```
iwehrman: This should be a localized string.

src/js/models/layer.js
```diff
@@ -310,7 +322,13 @@ define(function (require, exports, module) {
      * @param {object} layerEffect instance of a layer effect such as a Shadow
      */
     Layer.prototype.setLayerEffectByType = function (layerEffectType, layerEffectIndex, layerEffect) {
-        return this.setIn([layerEffectType + "s", layerEffectIndex], layerEffect);
+        if (!LayerEffect.TYPES.has(layerEffectType)) {
+            throw "Invalid layer effect type: " + layerEffectType;
```
iwehrman: We should only ever throw error objects so that they can be handled (e.g., with `.message` and `.stack`) uniformly at a higher level. So, these should be changed to something like `throw new Error("Invalid layer...`

src/js/models/layer.js
```diff
@@ -544,7 +567,19 @@ define(function (require, exports, module) {
...
     Layer.prototype.hasLayerEffect = function () {
-        return !this.innerShadows.isEmpty() || !this.dropShadows.isEmpty();
+        return !!this.effects.find(function (effects) { return !effects.isEmpty(); });
+    };
+    
+    /**
+     * True if the layer has unsupported layer effect.
+     * @return {boolean}  
+     */
+    Layer.prototype.hasUnsupportedLayerEffect = function () {
```
iwehrman: Since this has an `O(n)` lookup, it might be worthwhile to make this a cached property instead of a method. See, e.g., `unsupported` and `object.getCachedSpecs` above in this file. (I would be happy to explain how that works offline, btw.)

src/js/models/layer.js
```diff
@@ -544,7 +567,19 @@ define(function (require, exports, module) {
      * @return {boolean}  
      */
     Layer.prototype.hasLayerEffect = function () {
-        return !this.innerShadows.isEmpty() || !this.dropShadows.isEmpty();
+        return !!this.effects.find(function (effects) { return !effects.isEmpty(); });
```
iwehrman: Same here, re: caching, actually.

iwehrman:
Well, this seems great! I just have a few tiny suggestions but am eager to merge! :musical_note: :notes:

shaoshing:
Hi @iwehrman, thanks for reviewing; I've addressed your comments. now back to you.

iwehrman:
Fantastic!

Issue #2600, opened 2015-09-23
Zoom level can be less than 5% when app window is moved from high res screen to low res


ktaki:
Set zoom level to 5% in high res screen. Move the app window to low res screen. Pan the canvas (issue #2598) the zoom level display is updated to 2.5%. This should stay at the minimum value allowed (5%).

Note that if the zoom level is set to 3200% at low res screen and move the app window to high res, the zoom level stays at 3200%. This is fine.

tiburbage:
I think the minimum zoom factor is 0.27%. The zoom factor should never go below that, but will be doubled when dragging from a Retina display to a non-HiDPI display, and halved when going in the other direction. This keeps the apparent sizes of things uniform when going between the displays.

iwehrman:
FBNC to @ktaki.

ktaki:
The PR #2603 fixed #2598, not this issue. Please review.

iwehrman:
FBNC to @ktaki again!

ktaki:
Entering number less than 5% will result in 5%.
With Cmd+minus, the smallest value is 0.3%
If App window is moved from retina to normal display, 0.3% becomes 0.1%. 

In all above situations, standard Ps keeps 0.15% as the minimum. Can we do the same? At least, we should have a consistent minimum zoom value.

volfied:
@ktaki After playing with this, even if I allow 0.15, PS actually keeps the zoom value at 0.27%... Despite us passing it a zoom value of 0.15

I think at those zoom levels, document size comes into play, and it's not worth our time to make this as close to PS as possible. If you disagree, please let me know, and I will allow 0.15 to be a valid input. Window screen density allowing those zoom sizes is not a big deal for me, we should aspire to keep the user within reasonable zoom levels in our own UI.

ktaki:
So, Ps lies about the minimum zoom value? Fine, case closed.

Issue #2598, opened 2015-09-23
Zoom level display not updated automatically when app window was moved between high res and low res Whenmonitors


ktaki:
When app window is moved to high res screen to low res screen, or vice versa, the zoom level percentage is automatically adjusted. The zoom level display, however, is not updated automatically. After panning the canvas, the number is updated.

ktaki:
This has been fixed with PR #2603.

PR #2603, opened 2015-09-23
Update transform after updating panel sizes

volfied:
Addresses #2600.

Just emitting change in _handlePanelResized was not enough, as the zoom change wasn't being accounted for by the panel resizing. @iwehrman please review.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -241,6 +241,9 @@ define(function (require, exports) {
         return this.dispatchAsync(events.ui.PANELS_RESIZED, sizes)
             .bind(this)
             .then(function () {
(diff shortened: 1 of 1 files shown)


iwehrman:
My only concern is that, sadly, we call `updatePanelSizes` too often. When the UI changes, all the components that contribute panel sizes separately call `updatePanelSizes` one after another. This isn't harmful, it just isn't as efficient as it could be. This makes the problem somewhat worse. I'm only noting this so that someone else is aware of an opportunity for future optimization. In the meantime, this seems like the best solution.

ktaki:
Note this addressed #2598, but not #2600.

PR #2606, opened 2015-09-24
Move duplicateNewSheets handler into the move event handler

volfied:
JP moved duplicateNewSheets events into the "move" events, so they always come to us in right order.

Addresses #2510. Requries pgdv 668 or later.


mcilroyc:
This all looks in order.  Worked on 669.

Issue #2593, opened 2015-09-23
Overlay offsets not being correctly reset on error


iwehrman:
1. Help > Test Action Failures
2. Pan the canvas.

Result: the UI flashes because the overlay offsets are incorrect after the controller reset.

iwehrman:
Over to @chadrolfs to verify.

PR #2609, opened 2015-09-24
Preserve overlay offsets on reset

iwehrman:
Do not reset panel boundaries after a controller reset because:

1. The panel boundaries shouldn't have changed anyway; and
2. We don't yet have any good way to tell the individual panel components that they should reset their sizes when the controller resets.

Ideally we would have some kind of event that the components could easily listen to which would cause them to reset their sizes. And probably there should be a more general mixin for panels that contribute to the panel boundaries. But, for now, this super hack is sufficient to address #2593:


mcilroyc:
alright.

Issue #2594, opened 2015-09-23
Bounds highlight blinks when selecting shapes via marquee select


chadrolfs:
new doc, draw some shapes (rect, ellipse, pen), with Select tool marquee select shapes

Result: while marquee dragging, bounds highlight for shapes blink.

PR #2610, opened 2015-09-24
Don't update mouse over highlights while in marquee state

volfied:
It was causing highlights to blink as it removed all highlights if marquee was on. Addresses #2594


shaoshing:
:+1:

Issue #2505, opened 2015-09-16
CMD+f Errors sometimes when using this file


placegraphichere:
Error is hard to reproduce, but seems to happen when I mulit-select a bunch of text objects that have missing fonts (Lust script), then try to open the App Search using CMD+f.

http://cl.ly/2O3x100N2P1E

```
Opening an un-registered dialog without providing a dismissalPolicy
js/fluxcontroller.js:563 Action search.toggleSearchBar failed: TypeError: Cannot read property 'searchItems' of undefined
    at Fluxxor.createStore.getState (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/stores/search.js:147:65)
    at React.createClass.getStateFromFlux (file://./js/jsx/search/SearchBar.jsx:60:43)
    at getInitialState (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/fluxxor/build/fluxxor.js:482:20)
    at mergedResult [as getInitialState] (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6019:17)
    at new Constructor (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6258:54)
    at ReactCompositeComponentMixin.mountComponent (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:6791:16)
    at null.<anonymous> (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:10068:19)
    at wrapper [as mountComponent] (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13968:31)
    at Object.ReactReconciler.mountComponent (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:14781:35)
    at ReactDOMComponent.ReactMultiChild.Mixin._mountChildByNameAtIndex (file:///Applications/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:13622:40)
[...]

PR #2618, opened 2015-09-24
Fix search after reset

iwehrman:
1. Register search providers in `search.beforeStartup` instead of in `Search`'s `componentWillMount` method so that search providers are re-registered after a controller reset.
2. Ensure that the search command is enabled either if there is no open document or if there is a supported document open.

Addresses #2505 and #1888.

@shaoshing: 2) above tweaks your recent unsupported document change in 1e52903c3095d859496750784ecaa926fc8b42d6.

### DIFF
--- a/src/js/actions/search.js
+++ b/src/js/actions/search.js
@@ -24,7 +24,11 @@
 define(function (require, exports) {
     "use strict";
 
-    var dialog = require("./dialog");
+    var Promise = require("bluebird");
+
+    var dialog = require("./dialog"),
+        search = require("../stores/search"),
+        locks = require("../locks");
 
     /**
      * The search bar dialog ID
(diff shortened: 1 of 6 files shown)


src/js/actions/search.js
```diff
@@ -48,9 +52,22 @@ define(function (require, exports) {
         }
         return this.transfer(dialog.openDialog, ID);
     };
-    toggleSearchBar.reads = [];
+    toggleSearchBar.reads = [locks.JS_SEARCH];
```
volfied: do you mean JS_DIALOG?

volfied:
Just one question @iwehrman

iwehrman:
YES! Fixed now. Back to @volfied.

Issue #2611, opened 2015-09-24
Upper limit of stroke size not set


ktaki:
It should be set to 288 pix to follow standard Ps.

ktaki:
Verified the fix.

PR #2619, opened 2015-09-24
Enforce a maximum stroke size of 288px.

iwehrman:
What it says!

Addresses #2611.


volfied:
I have so many questions.

PR #2623, opened 2015-09-24
Render icon bar and empty column on startup

iwehrman:
This changes the initial render of `PanelSet` to draw the icon bar and a single empty column instead of nothing to improve the experience of the initial UI load. It's not perfect, but I think it's better.

Addresses: concerns.


volfied:
Good improvement.

PR #2625, opened 2015-09-24
fix spaces-design for latest adapter

pineapplespatula:
quick fix for latest adapter changes related to vector mask work

### DIFF
--- a/src/js/tools/ellipse.js
+++ b/src/js/tools/ellipse.js
@@ -43,12 +43,13 @@ define(function (require, exports, module) {
      * @constructor
      */
     var EllipseTool = function () {
-        var resetPromise = descriptor.playObject(toolLib.resetShapeTool()),
+        var resetObj = toolLib.setShapeToolMode(toolLib.toolModes.SHAPE),
             firstLaunch = true;
             
         /** @ignore */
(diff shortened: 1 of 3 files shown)


src/js/tools/ellipse.js
```diff
@@ -43,7 +43,7 @@ define(function (require, exports, module) {
      * @constructor
      */
     var EllipseTool = function () {
-        var resetPromise = descriptor.playObject(toolLib.resetShapeTool()),
```
volfied: This should be like the other two shape tools, and only the play object be defined outside the selectHandler.

If you call playObject outside the selectHandler, you're essentially calling this function only once in the life time of the app, and in all other instances of select handler being called, the promise will immediately return since it has been resolved.

volfied:
One question.

pineapplespatula:
back to you @volfied

volfied:
Can you rebase the commits and make the commit message describe what it's changing? 

Maybe something like: "Use updated tool reset APIs for shape tools"

volfied:
One more thing @pineapplespatula

pineapplespatula:
done, back to @volfied

Issue #2613, opened 2015-09-24
X and Y coordinates for a selected layer are w.r.t canvas, not the artboard


DivyaPrabhakar:
If the selected shape is on an artboard, the X and Y coordinates should be in relation to the artboard. 

Repro: 

1. Create two artboards, AB1 and AB2
2. Create a shape layer and place it on AB2 in the top left corner
3. X Y coordinates read X: (some non-zero value), Y: 0. 

Expected behavior is  X: 0     Y: 0

iwehrman:
Over to @chadrolfs for verification.

chadrolfs:
Hmmm, well I'm not totally sure it's fixed.  There are some confusing things.  Watch this:
1st: X/Y of AB is 850, 0.  While dragging, you can hold down Space to reposition and as you do, you get the coords according to what the AB is (850/0) but once you let go and stop drawing the rect, the coords change to -1/0 (which is what you're going for with this fix).  So I guess expected behavior is correct after you've finished drawing but not during.

https://www.dropbox.com/s/u6lgdikrv378mxr/XY-Coords.mov?dl=0

Once you are outside the AB, things get very interesting tho.
Trying to set a shape outside AB to 0/0 will always move it to the left and up and reset it to 1024.  Not sure I understand that.

Anyway, based on the video and what I've explained, if the fix doesn't include "while dragging" or outside AB, I think we can close this.

DivyaPrabhakar:
I saw the same issues @chadrolfs was referring and a couple of new ones so I filed bugs for each of them and will close this one because it was an issue filed only about the X Y coordinates in the input fields, which is now fixed.

PR #2626, opened 2015-09-24
Add relativeChildBounds, and use it for position inputs

volfied:
Will address #2613, and show layer position in relation to it's owner artboard.

 - Added `Bounds.prototype.relativeTo` method which returns a new bounds object with updated position in relation to the given bounds object.
 - Added `relativeChildBounds` cached property that calculates the bounds of a layer relative to it's owner artboard.
 - Added `selectedRelativeChildBounds` cached property that grabs all the selected layer's relativeChildBounds.
 - Add `relative` flag to `actions.transform.setPosition` to use relative child bounds during transform calculations, subsequently added it to `_getMoveLayerActions` as well.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -106,11 +106,13 @@ module.exports = function (grunt) {
         },
         lintspaces: {
             src: [
-                "*.js",
-                "*.json",
+                "*",
                 "src/**/*.json",
                 "src/**/*.jsx",
-                "src/**/*.js"
+                "src/**/*.js",
+                "src/**/*.svg",
+                "src/**/*.less",
+                "!src/**/*.gif"
             ],
             options: {
                 newline: true,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -106,12 +106,10 @@ module.exports = function (grunt) {
         },
         lintspaces: {
             src: [
-                "*",
-                "src/**/*",
-                "test/**/*",
-                "!**/*.ogg",
-                "!**/*.otf",
-                "!**/*.png"
+                "*.js",
+                "*.json",
+                "src/**/*.jsx",
+                "src/**/*.js"
             ],
             options: {
                 newline: true,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -108,6 +108,7 @@ module.exports = function (grunt) {
             src: [
                 "*.js",
                 "*.json",
+                "src/**/*.json",
                 "src/**/*.jsx",
                 "src/**/*.js"
             ],
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -67,11 +67,14 @@ define(function (require, exports) {
      * @param {Layer} targetLayer
      * @param {{x: number, y: number}} position
      * @param {Array.<{layer: Layer, x: number, y: number}>} moveResults payload for updating each layer's bounds
+     * @param {boolean} relative If true, will use teh artboard relative bounds of the layer
      *
      * @return {Immutable.List<{layer: Layer, playObject: PlayObject}>}
      */
-    var _getMoveLayerActions = function (document, targetLayer, position, moveResults) {
-        var overallBounds = document.layers.childBounds(targetLayer),
+    var _getMoveLayerActions = function (document, targetLayer, position, moveResults, relative) {
+        var overallBounds = relative ?
+                document.layers.relativeChildBounds(targetLayer) :
+                document.layers.childBounds(targetLayer),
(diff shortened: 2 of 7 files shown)


volfied:
@chadrolfs @ktaki This is a high risk change, can you guys test out swap and transform features a bit in this branch when you have time?

src/js/actions/transform.js
```diff
@@ -381,9 +385,9 @@ define(function (require, exports) {
                 positions: []
             },
             layerOneActions = _getMoveLayerActions
-                .call(this, document, layers.get(0), newPositions.get(0), payload.positions),
+                .call(this, document, layers.get(0), newPositions.get(0), payload.positions, false),
```
pineapplespatula: why are you specifying false, instead of handling relative === undefined in the function

volfied: Because I want that flag to be explicit across the codebase, just coding defensively.

pineapplespatula: it would be more defensive to add a line to your function 
    if (relative === undefined) relative = false;

chadrolfs:
on attempting to load this branch, I get 
`Uncaught TypeError: Cannot read property 'SHAPE' of undefined  rectangle.js:51`

volfied:
Chad's error was related to his bower link being out of date.

I added a new commit (6fb4e0c) which allows us to account for nesting changes when you setPosition of a layer into another artboard. This has a small cosmetic bug where if you drag a layer using mouse into another artboard, it's universal coordinates flash for a brief second before it being re-rendered. I'll be working on that, but I think it's good to go to master this way (barring all bugs you guys find)

src/js/actions/transform.js
iwehrman: Quasi-related: this is the sort of argument that might be better as a property of an `options` arguments.

src/js/actions/transform.js
```diff
@@ -334,7 +337,8 @@ define(function (require, exports) {
                 historyStateInfo: {
                     name: strings.ACTIONS.SET_LAYER_POSITION,
                     target: documentLib.referenceBy.id(document.id)
-                }
+                },
+                synchronous: false
```
iwehrman: Can you add a comment here explaining why this flag is used?

src/js/actions/transform.js
```diff
@@ -321,7 +324,7 @@ define(function (require, exports) {
      *
      * @return {Promise}
      */
-    var setPosition = function (document, layerSpec, position) {
+    var setPosition = function (document, layerSpec, position, relative) {
```
iwehrman: Missing JSDoc on the new param.

src/js/actions/transform.js
```diff
@@ -1044,10 +1050,18 @@ define(function (require, exports) {
             }
         }, this);
 
+        _moveToArtboardHandler = function () {
+            var appStore = this.flux.store("application"),
```
iwehrman: Since you're in here touching this, it would probably be better to make this an action that transfers to `resetIndex`.

iwehrman:
Looks great, seems to work well. Back to @volfied. A deeper testing pass would still be helpful though, @chadrolfs or @ktaki.

volfied:
Rebased, addressed reviews, and made changes to `_getMoveLayerActions` so it returns the action just for the group layer. Problem was, if we moved a group with older code, we would move it's children to another artboard, but the group itself would not move since those layers got piecemeal translated.

Also, debounce handler has a 200 ms delay for "moveToArtboard" because we get one for each layer that's being transferred, and resetting index mid action breaks things.

volfied:
@iwehrman Back to you.

src/js/actions/transform.js
iwehrman: Still want this comment :)

Gruntfile.js
```diff
@@ -106,12 +106,11 @@ module.exports = function (grunt) {
         },
         lintspaces: {
             src: [
-                "*",
-                "src/**/*",
-                "test/**/*",
-                "!**/*.ogg",
-                "!**/*.otf",
-                "!**/*.png"
+                "*.js",
```
iwehrman: I don't like the idea of not checking the svg files as well as the other text files in the root directory (the README, LICENSE, Gruntfile, etc.). How much time does this check actually save? More than 1s?

iwehrman:
Still missing a comment, plus complaints about the lint changes.

volfied:
@iwehrman added a comment, and also realized that (using unix `time` command), providing specific extensions to look for is much faster than giving it a whole directory. So lintspaces now checks the same amount of files but does it about 30 times faster.

iwehrman:
Zowie! I'm sold.

Issue #2621, opened 2015-09-24
Can't read text while renaming a library asset


davealonzo:
<img width="419" alt="screen shot 2015-09-24 at 2 37 27 pm" src="https://cloud.githubusercontent.com/assets/12259315/10087794/0dd38660-62cb-11e5-8c8a-d4ab6d142c69.png">

chadrolfs:
@designedbyscience should this be assigned to you?

volfied:
@chadrolfs Can you confirm the fix?

Issue #2615, opened 2015-09-24
Double clicking to edit missing library smart objects causes error


volfied:
1 - Create a graphic element
2 - Place graphic element in image
3 - Delete the graphic element from the library
4 - Double click the graphic element layer in the image to edit

Result:
```
Action superselect.doubleClick failed: TypeError: Cannot read property 'getPrimaryRepresentation' of undefined
    at Object.openGraphicForEdit (file:///Users/baaygun/Desktop/photoshop-Release_x86_64/Adobe%20Photoshop%20…p/Contents/Required/Plug-Ins/Spaces/www/src/js/actions/libraries.js:536:37)
```

volfied:
FBNC to @chadrolfs

PR #2627, opened 2015-09-24
Various improvements for the libraries panel

shaoshing:
1 can edit large document format (PSB, here is a sample asset for testing purpose http://adobe.ly/1LAQoVB)
2 handle error when double-click smart object while its associated element is deleted. (#2615)
3 automatically scroll to the last locally updated graphic asset in the libraries panel.
4 match asset rename input style with the layers panel. (#2621)

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -97,7 +97,9 @@ define(function (require, exports) {
         REP_LAYERSTYLE_TYPE = "application/vnd.adobe.layerstyle",
         REP_COLOR_TYPE = "application/vnd.adobe.color+json",
         REP_PNG_TYPE = "image/png",
-        REP_PHOTOSHOP_TYPE = "image/vnd.adobe.photoshop";
+        REP_PHOTOSHOP_TYPE = "image/vnd.adobe.photoshop",
+        // Large Document Format
+        REP_PSB_TYPE = "application/photoshop.large";
 
     /**
      * List of acceptable image representations that PS can place as
@@ -128,7 +130,8 @@ define(function (require, exports) {
      * @type {Set}
      */
     var EDITABLE_GRAPHIC_REPRESENTATION_TYPES = new Set([
-        REP_PHOTOSHOP_TYPE
+        REP_PHOTOSHOP_TYPE,
+        REP_PSB_TYPE
     ]);
     
     /**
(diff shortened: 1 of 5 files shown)


src/js/actions/libraries.js
```diff
@@ -128,7 +128,8 @@ define(function (require, exports) {
      * @type {Set}
      */
     var EDITABLE_GRAPHIC_REPRESENTATION_TYPES = new Set([
-        REP_PHOTOSHOP_TYPE
+        REP_PHOTOSHOP_TYPE,
+        "application/photoshop.large" // PSB
```
volfied: Where is REP_PHOTOSHOP_TYPE defined? can you define a constant for `application/photoshop.large` in the same place?

src/js/actions/libraries.js
```diff
@@ -533,10 +534,15 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var openGraphicForEdit = function (element) {
+        if (!element) {
+            log.debug("[CC Lib] openGraphicForEdit: element is missing. maybe it is already deleted.");
```
volfied: Please add a // FIXME comment in here explaining how CEP libraries panel does it for future reference.

src/js/jsx/sections/libraries/Library.jsx
```diff
@@ -76,6 +76,15 @@ define(function (require, exports, module) {
...
          */
         lastLocallyCreatedElement: null,
+        
+        /**
+         * Store the last locally updated graphic element and modified time. Used to determine whether the 
+         * library list should scroll to reveal the updated graphic.
+         * 
+         * @type {AdobeLibraryElement}
+         */
+        lastLocallyUpdatedGraphic: null,
```
volfied: These look like they should be in the state, instead of open variables in the component. Is there a reason you didn't keep them in the state?

volfied:
Got couple nits, and one concern about component private vars vs state.

src/js/jsx/sections/libraries/Library.jsx
shaoshing: actually, putting `lastLocallyUpdatedGraphic` in state will trigger unnecessary UI redraw, as the main purpose of these variables are used to decide whether the asset list should scroll after a redraw due to change in `this.props.library`.

shaoshing:
Hi @volfied , thanks for reviewing :+1: . Addressed your comments, so back to you

PR #2631, opened 2015-09-25
Debounce handleWindowResize in DocumentHeader

iwehrman:
This is the only instance of a window `resize` handler that both executes more than once and isn't debounced. This means it's executing extremely frequently during window resizes, which may or may not materially affect the frequency with which layout and painting happens. I'm arguing with @jsbache about this offline, but this seems like a good and safe change to make either way.

Addresses: concerns.


volfied:
Nice use of the setState callback.

PR #2632, opened 2015-09-25
Change Libraries panel to hide on collapse with CSS

designedbyscience:
Before, the panel content for Libraries was removed if the library was collapsed, now it is just hidden with CSS (same as the other panels).


iwehrman:
Looks great, works great, tastes great.

PR #2634, opened 2015-09-25
Add .25, .5 and 1 percent to available zoom_increments

volfied:
Helps address #2600, and work more like Photoshop.


iwehrman:
OK!

PR #2635, opened 2015-09-25
Command Click on Column Toggles has different behavior

designedbyscience:
This is more an experiment than anything else, using a PR so we can have a conversation about it.

For the proposed single column mode, there are a lot of ways we could get there...this is one of them. Instead of moving all of the panels into a single column, which has some issues around space (why we moved to two column in the first place), this simply allows the user to quickly switch between Properties and Layers/Libraries. 

Behavior is:
2 columns open: Command Click on either column toggle button closes both panels
2 columns closed: Command Click on either column toggle button open both panels 
1 column open: Command Click on either column toggle button switches the open panel for the closed panel

For your consideration, @placegraphichere

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -193,8 +193,19 @@ define(function (require, exports, module) {
 
         /** @ignore */
         _handleColumnVisibilityToggle: function (columnName) {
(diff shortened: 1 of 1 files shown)


davealonzo:
@designedbyscience Your proposal sounds great Eric!

Issue #2570, opened 2015-09-22
Toggling fill and stroke visibilities changes their colors when multiple layers are selected


ktaki:
Create two rectangles. Change the colors of fill and stroke for each rectangle. Use different colors. Select the both rectangle in layers panel. Click visibility icons of fill and strokes in the Appearance panel to turn off and then on the visibility.

RESULT: 
The fill colors for the both rectangles become black. 
The stroke colors for the both rectangles become one of the stroke colors.

mcilroyc:
FBNC to @ktaki

ktaki:
Verified the fix.

PR #2636, opened 2015-09-25
Re install existing fill/stroke colors on re-enable

volfied:
This will address #2570, but more importantly, fix toggling of strokes/fills so they behave correctly.

 - Change `setFillEnabled` and `setStrokeEnabled` to pass in a `null` value for color.
 - If color being passed to `setFillColor` and `setStrokeColor` is null, and color is being enabled, we will use the last known fill/stroke colors from the layer models for each layer itself, using default fill/stroke color as fallback. This allows us to disable/enable strokes/fills of multiple layers without overwriting/losing values.

@ktaki please test this on the branch, this is a big change and I want to make sure I'm not introducing new bugs.

@mcilroyc your expertise with layerActions and fill/stroke would be very appreciated in reviewing.

### DIFF
--- a/src/js/actions/shapes.js
+++ b/src/js/actions/shapes.js
@@ -225,15 +225,12 @@ define(function (require, exports) {
      * 
      * @param {Document} document
      * @param {Immutable.List.<Layer>} layers list of layers being updating
-     * @param {Color} color color of the strokes, since photoshop does not provide a way to simply enable a stroke
      * @param {object} options
      * @param {boolean=} options.enabled
      * @return {Promise}
      */
-    var setStrokeEnabled = function (document, layers, color, options) {
-        // TODO is it reasonable to not require a color, but instead to derive it here based on the selected layers?
-        // the only problem with that is having to define a default color here if none can be derived
-        return setStrokeColor.call(this, document, layers, color, options);
+    var setStrokeEnabled = function (document, layers, options) {
+        return setStrokeColor.call(this, document, layers, null, options);
     };
     setStrokeEnabled.reads = [];
     setStrokeEnabled.writes = [locks.PS_DOC, locks.JS_DOC];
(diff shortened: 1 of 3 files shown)


src/js/actions/shapes.js
```diff
@@ -292,17 +288,56 @@ define(function (require, exports) {
                 eventName,
                 options.coalesce);
 
-            var colorPromise = layerActionsUtil.playSimpleLayerActions(document, layers, strokeObj, true, options);
+            
+            if (!color && options.enabled) {
```
mcilroyc: I guess this is a long-standing issue, but the JSDoc should reflect `color`'s optional nature

src/js/jsx/sections/style/Fill.jsx
```diff
@@ -168,12 +167,10 @@ define(function (require, exports, module) {
          * @param {boolean} isChecked
          */
         _toggleFillEnabled: function (event, isChecked) {
-            var color = this.props.fill && collection.uniformValue(this.props.fill.colors) || Color.DEFAULT;
-
             this.getFlux().actions.shapes.setFillEnabled(
                 this.props.document,
                 this.props.layers,
-                color,
+                null,
```
mcilroyc: I think you can just remove this argument entirely from this method, and just change the shape action accordingly.  Same for `setStrokeEnabled`

src/js/actions/shapes.js
```diff
@@ -292,17 +288,56 @@ define(function (require, exports) {
...
+
+                colorPromise = layerActionsUtil.playLayerActions(document, actions, true, options);
+            } else {
+                // remove the alpha component based on ignoreAlpha param
+                psColor = color ? color.toJS() : null;
+                if (psColor && options.ignoreAlpha) {
+                    delete psColor.a;
+                }
 
+
```
mcilroyc: :space_invader:

ktaki:
Tested the branch. The original issue was fixed. No new issues found.

mcilroyc:
Thanks @ktaki.  @volfied I have a few nits.  I think my suggestion of removing that param will clean up old cruft ... but if you want to leave it for future flexibility, OK.  Back to you, if briefly.

volfied:
@mcilroyc Thank you for review, I cleaned up the functions.
@ktaki, thank you for testing, I'm glad there are no obvious issues.

Issue #2272, opened 2015-08-28
Design Space is partially unresponsive during startup when Generator is missing 


pineapplespatula:
Since i use a locally built version of PS without any plugins i dont have generator. When I launch i get a series of errors about the missing socket 
![screen shot 2015-08-28 at 4 53 58 pm](https://cloud.githubusercontent.com/assets/1008531/9559236/18addfc6-4da5-11e5-94bc-d71c6df68a8c.png)

until DS gives up on finding Generator some functionality is missing. Specifically keyboard shortcuts. 

Steps to reproduce :
1: remove Generator
2: refresh DS
3: Hit 'T' to try and switch to type tool 

Result
nothing

expected 
Type Tool

mcilroyc:
I can repro this.  BUT, because most things seem to be working *except* shortcuts ... I'm inclined to think this is not specific to export, but is caused by shortcuts not being properly registered between the before and after startup services.  @iwehrman or @volfied any immediate ideas?

This can be repro'd by adding the following to *any* afterStartup method, but export or help are easy choices  
```
        return Promise.delay(20000).then(function () {
            return Promise.resolve("FOO");
        });
```

iwehrman:
This is still an issue: 
![image](https://cloud.githubusercontent.com/assets/1075154/9915781/bfb454e0-5c6d-11e5-81db-84a8c27c5857.png)
Bumping to high priority and handing off to @mcilroyc.

volfied:
@chadrolfs can you confirm the fix?

PR #2637, opened 2015-09-25
Mark the controller active after beforeStartup instead of after afterStartup

iwehrman:
This effectively unblocks menu commands and shortcut commands from running before `afterStartup` is complete. It's possible there was some good reason for delaying this state change until after `afterStartup`, but I can't currently see what it might be.

The moral of the story is that, from now on, if the app requires some initialization to run without error, it should go in `beforeStartup`. If it's able to run without the initialization then it can go in `afterStartup`.

Addresses #2272.


volfied:
Nice.

Issue #1942, opened 2015-08-06
Undo after deleting a non-visible group doesn't restore selection or visibility state


chadrolfs:
This can get so confusing in large docs.

https://www.dropbox.com/s/upk3srcctz7pxhw/Undo-Blasphemy.mov?dl=0

chadrolfs:
I think we had a discussion somewhere else about issues/difficulties around restoring selections after undo but in this case, you're not even in the same artboard which is part of the confusion.

mcilroyc:
Hey @chadrolfs  - in trying to repro this issue, I noticed that after undo, the visibility of the group gets reset.  Is this expected?  Or a regression?  

Should the visibility change itself be undo-able?  Right now it seems that the visibility change is NOT tracked as a history state.

chadrolfs:
Regression.  After undo, the on canvas state and visibility of the group that gets restored after undo is out of sync.  This is not expected.  There's 2 bugs here: 1.) the selection isn't restored, the user ends up with some other layer selected that is way down the layers panel, and 2.) the state of visibility is out of sync with on-canvas (the more serious bug).

Visibility by itself should not be undo-able (same as it is in PS classic).

mcilroyc:
Hey @chadrolfs, since the more serious sub-bug #2 (layer visibility out of sync wrt canvas) has been addressed, do you think we can kick this out of M3?

It seems like the behavior now is to select the parent group of the un-deleted-and-previously-invisible layer/group, which doesn't seem unreasonable.

chadrolfs:
I think I got the 2nd issue confused.  As I originally reported it, it isn't actually addressed.  It being out of sync is indeed fixed but restoring the visibility state isn't.  Quick repro: drag out 3 rects, select one, turn visibility off, then delete, then undo.  Result: layer is returned but with visibility on.  

If you repeat this in standard PS, the visibility state is retained via undo and step backwards.  If it comes to it, I guess we can defer this but it has big potential to really miff users.

chadrolfs:
I'm totally fine deferring or never fixing the issue restoring the selection, but visibility state needs to be fixed.

mcilroyc:
[...]

PR #2638, opened 2015-09-25
During undo/redo, optimistically copy layer visibility from current state...

mcilroyc:
...and then also query from photoshop the visibility and update the layerstructure accordingly.

This should resolve the more serious of the two sub-bugs in #1942

@iwehrman could you please take a first pass at the performance implications of this?  The total time spent in the undo action seems to be unchanged on Vermilion.  I may be missing something.

Also please consider my design decision to allow a second re-draw of the layers panel in the rare-ish case that an otherwise hidden layer becomes visible during undo/redo because that layer was involved.  repro steps with two layers A and B (I use groups in Vermilion):

1. move layer A
1. move layer B
1. make layer A *invisible*
1. move layer B
1. undo twice: note that layer A is still invisible
1. undo again, and note that layer A become visible on canvas, and then shortly thereafter the panel reflects the visibility change

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -1113,6 +1113,17 @@ define(function (require, exports, module) {
     };
 
     /**
+     * Merges properties into the layers based on a map of simple property maps, keyed by layer ID
+     *
+     * @param {Immutable.Map.<number, Immutable.Map>} layerMap
+     */
+    LayerStructure.prototype.setLayerPropsMap = function (layerMap) {
+        return this.mergeDeep({
+            layers: layerMap
+        });
+    };
+
+    /**
      * Update the bounds of the given layers.
      *
      * @private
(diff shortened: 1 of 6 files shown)


mcilroyc:
UPDATE: I just added a [this change](https://github.com/adobe-photoshop/spaces-design/pull/2638/files#diff-56fdf0585052e728b2dbf61f6f5872d5R153) that I'm sure is fix all kinds of other bugs ... but also allows more correct timing comparison.  Now I believe that this change is increasing the undo time by ~200ms on vermilion.

src/js/models/layerstructure.js
```diff
@@ -1540,5 +1551,24 @@ define(function (require, exports, module) {
...
+     * Given a set of layers representing the current state, overlay the visibility status
+     * of any matching layers
+     *
+     * TODO is there a way to do this with mergeWith?
+     *
+     * @param {LayerStructure} layerTree
+     * @return {LayerStructure}
+     */
+    LayerStructure.prototype.overlayWithLayers = function (layerTree) {
+        var nextLayers = layerTree.layers.map(function (layer, key) {
```
volfied: Can't you use `mergeDeep` here if `layerTree only has `<id: {id, visible}>` mapping?

volfied: I guess not if you want to make sure you're not adding non existent layers (is that possible)

volfied:
Can't say I'm a fan of `overlayWithLayers` method being just `visible` specific. Can you change that to `overlayVisibilityWithLayers`?

mcilroyc:
@volfied I was trying to keep some generality about the process from the history's perspective so that we may apply future logic to the idea of "overlaying" stuff current>history.  But anyway, I agree with your suggestion - it makes sense to keep it clear and concise in this layerstructure change. will fix

mcilroyc:
I rebased and squashed with a new commit that renamed the offending methods to be visibility-specific, per @volfied

volfied:
What could go wrong?

Issue #2568, opened 2015-09-22
Invalid prop `color` supplied to ColorFields


iwehrman:
`Warning: Failed propType: Invalid prop 'color' supplied to 'ColorFields', expected instance of 'Record'. Check the render method of 'ColorPicker'.`

iwehrman:
Fixed by #2644.

PR #2644, opened 2015-09-25
Set ColorFields expected color proptype to HSVAColor

volfied:
ColorPicker keeps the state.color in HSVAColor type, so ColorFields should expect that.

Addresses #2568


iwehrman:
Derp.

Issue #2551, opened 2015-09-21
Can't set hue if saturation is 0


iwehrman:
In the color picker, changes to the hue don't stick when the saturation is set to 0. This is because colors with zero saturation and different hues are equivalent, but the value typed into the input should still remain.
![image](https://s3.amazonaws.com/f.cl.ly/items/2R3B0a3n070K0b041C2O/Screen%20Recording%202015-09-21%20at%2003.01%20PM.gif)

iwehrman:
Fixed by #2647.

PR #2647, opened 2015-09-25
Persist the hue value across 0 saturations in color picker

volfied:
Addresses #2551, problem was, we were parsing the Hsv value into rgb, which kept losing the hue when saturation was 0.


iwehrman:
LGTM.

Issue #2643, opened 2015-09-25
Ungroup: layers panel does not seemingly update correctly


placegraphichere:
It takes some time before the layers panel will update when a large group is ungrouped. The layers do not seem to respond until another click is made, or other UI is interacted with.

Example is using the vermillion PSD and ungrouping one of the large top level groups inside one of the artboards.

iwehrman:
I can confirm that the panel isn't updated until you click on something to change the selection state.

volfied:
@placegraphichere this should be fixed. can you confirm?

iwehrman:
Kicking over to @chadrolfs to verify.

PR #2653, opened 2015-09-26
Initialize layers after ungroup and after resetting the selection

iwehrman:
1. We should always be calling `initializeLayers` after `resetSelection` because that action causes teh selection to change. So, now we are.
2. `ungroupSelected` doesn't always call `resetSelection`. If it doesn't we need to call `initializeLayers` manually. We need to do so manually whenever an action changes the selection state of the layers without going through `select` or (now) `resetSelection`. If you see any other actions that might change the selection state without going through one of those actions, please let me know!

Addresses #2643.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -430,27 +430,6 @@ define(function (require, exports) {
     };
 
     /**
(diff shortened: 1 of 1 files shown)


volfied:
Initialize all the things.

Issue #2646, opened 2015-09-25
Reverting a file results in Uncaught TypeError


chadrolfs:
Open AdventureCoHomePage_Final.psd (in my drop) 
move a layer
File > Revert

Results: 
`Action history.revertCurrentDocument failed: TypeError: undefined is not a function`

PR #2655, opened 2015-09-26
Fix the 'revert' history store handler

mcilroyc:
This should address #2646.

Look you guys, I don't know why this is necessary.  I also do not know when it broke, other than that it was at least several sprints ago, before the major plugin version change.  

Prior to this commit, if I stick a breakpoint at stores/history.js line 444, I could execute that same code in the console.  If I tried to step in to it, it exploded.  I was stumped.  I moved the logic out of that low-value abstraction (as a method of the historystate model) and just did tha business directly in to the history store.  It worked.  Is that obvious to anyone why?

### DIFF
--- a/src/js/models/historystate.js
+++ b/src/js/models/historystate.js
@@ -71,17 +71,5 @@ define(function (require, exports, module) {
             (this.documentExports && this.documentExports !== documentExports);
     };
 
-    /**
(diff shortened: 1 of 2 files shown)


volfied:
Maybe some premature cleanup in immutable causing the breakage? Either way, if this fixes it. I'm merging it.

Issue #1609, opened 2015-06-18
Layer selection is slow


iwehrman:
In the mega-artboards document, selecting layers in the panel takes >300ms in debug mode on a retina MBP. On-canvas targeting takes > 450ms. Yes, some of this overhead is due to the post-condition verification, but I think we can do better. 

This is a tracking bug. At least one related issue may be #1464.

mcilroyc:
@chadrolfs in PR #2657 @iwehrman  mentioned three bugs.  I assigned one (this one) to you.  This was a fairly major change to layer selection (esp. w/ modifiers) and needs a careful eye.

chadrolfs:
Yeah, much better, I'm seeing between 100-200ms on retina.

Issue #1470, opened 2015-06-12
Select layers with modifier should use batchPlay


DesignSpaceBot:
See [here](https://github.com/adobe-photoshop/spaces-design/blob/4bd89afb900bb46f72dc10eb89e6fe5ab8b4e5f1/src/js/actions/layers.js#L492-L504). We should test for the modifier up front and then use `batchPlay` instead of waiting for the select action to finish before submitting the property getter.

-- @iwehrman at [2015-05-24T16:25:51Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1470)

iwehrman:
Fixed by #2657.

PR #2657, opened 2015-09-26
Fully optimistic layer selection

iwehrman:
The changes `layers.select` to be fully optimistic: the complete selection post-state is calculated up front in JavaScript, and the layer model is updated in parallel with the Photoshop play command instead of in sequence. Previously we only did this when no modifier was used. Even if Photoshop worked correctly this would be somewhat faster, but as is there is also a terrible bug in which the `addUpTo` modifier does not take the visibility (w.r.t. expanded/collapsed ancestors) of layers into account, such that selecting a range of collapsed groups would also select all of their descendants.

Note that the specification of `addUpTo` requires the notion of a "pivot" element, which represents the starting point of a range selection. This is tracked in `LayerStructure.prototype.pivotID` the lazy `pivot` property that refers to the pivot layer model (or null, if there is no pivot layer).

Addresses #2434, #1470 and (I do declare!) #1609.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -756,8 +756,9 @@ define(function (require, exports) {
      * @param {Document} document Owner document
      * @param {Layer|Immutable.Iterable.<Layer>} layerSpec Either a single layer that
      *  the selection is based on, or an array of such layers
-     * @param {string} modifier Way of modifying the selection. Possible values
-     *  are defined in `adapter/src/lib/layer.js` under `select.vals`
+     * @param {string=} modifier Way of modifying the selection. Possible values
+     *  are defined in `adapter/src/lib/layer.js` under `select.vals`.
+     *  Default is similar to "select", but the pivot layer will always be cleared.
      *
      * @returns {Promise}
      */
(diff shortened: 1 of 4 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -167,7 +167,14 @@ define(function (require, exports, module) {
                 }
             }
 
-            this.getFlux().actions.layers.select(this.props.document, this.props.layer, modifier);
+            // The clicked layer may an have out-of-date document models due to
+            // the aggressive SCU method in LayersPanel.
```
mcilroyc: ramble: seems like the most react-y way to do this would be for the method that calls the flux action be at the same component as the aggressive SCU, and would thus have ready access to current props/state.  That could be wrapped in a function and provided as a property, pushed in to the sub-component which may not get re-rendered very often.  Maybe that doesn't make sense here for some reason... maybe the aggressive SCU is *too high* up the chain.

src/js/actions/layers.js
```diff
@@ -768,55 +769,87 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
+        if (modifier !== undefined && layerSpec.size > 1) {
```
volfied: If default is "select", why not allow multiple layer specs?

volfied:
I got one code comment, and a wish that pivot's purpose was explained in the select action.

src/js/actions/layers.js
```diff
@@ -768,55 +769,87 @@ define(function (require, exports) {
...
+                nextSelected = document.layers.exposed.filter(function (layer) {
+                    var index = document.layers.indexOf(layer);
+
+                    return firstIndex <= index && index <= lastIndex;
+                });
+            }
+            break;
+        default:
+            nextSelected = layerSpec;
+            nextPivot = null;
```
mcilroyc: if "select" is the default modifier, then it seems weird they are treated slightly differently (nextPivot assignment).  Maybe there's a good reason, just checking.

mcilroyc:
back to @iwehrman for a few questions and doc requests

src/js/actions/layers.js
iwehrman: Why? This was very intentional.

src/js/jsx/sections/layers/LayerFace.jsx
iwehrman: I think you basically just mean that `LayerFace` should have an `onClick` handler and `LayersPanel` should execute the action. I guess that makes sense? I'm a little reticent to push more complexity up into `LayersPanel` though...

src/js/actions/layers.js
iwehrman: In particular, if you don't pass a modifier, no pivot is set. This is the case when we call select internally (i.e., not from the layer face's click handler.)

src/js/actions/layers.js
iwehrman: See here: https://github.com/adobe-photoshop/spaces-design/pull/2657#discussion_r40575589

src/js/actions/layers.js
volfied: Yeah, but if you pass "select" with a big layerSpec here, it's going to complain at us. And that should be allowed. You can later on do the same check to set the pivot only if it was actually passed. Or we need to document somewhere that if you want this action to select multiple layers, don't provide any modifiers (even though select is a perfectly valid one)

iwehrman:
Rebased and cleared up what the meaning of the default modifier is. Back to @mcilroyc.

PR #2658, opened 2015-09-26
Initialize the font list only it necessary

iwehrman:
We're currently initializing the font list twice for no reason. The `Type` initiates before each render if it hasn't already been initialized, but `Type` apparently renders twice in succession before the first call is complete. The fix is to just have the `type.initFontList` action itself check to see if the font list has already been initialized and to bail out early if so. This saves a few hundred milliseconds on startup time.

Addresses: concerns.


volfied:
:cake:

Issue #790, opened 2015-06-12
Initializing big documents is slow


DesignSpaceBot:
I think this is partially system dependent.  I'm seeing ~3500ms.  Zorana saw 10000-15000ms

-- @chadrolfs at [2015-03-12T21:09:30Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/790)

chadrolfs:
This is much better than it was previously.  I'm closing this.  If anyone disagrees, re-open and push to M4

PR #2660, opened 2015-09-27
Lazy document initialization

iwehrman:
This PR is in the same spirit as #2439, which delayed loading detailed information about layers until their first selection, but for documents. It changes `documents.initInactiveDocuments` to only load a small subset of essential document properties during startup, and `documents.selectDocument` to finish loading all of the properties when the document is first selected. It does _not_ change `initActiveDocument` or `updateDocument` or anything else that might initialize a new document after startup. An uninitialized document is detected simply by having a null `layers` property.

With 7 moderately sized documents open, this reduces overall startup time from ~6.5 seconds to ~4.5 seconds, which is a ~30% reduction.

Although #2439 was a fairly simple architectural change, this is even simpler. Still, these sorts of performance optimizations always introduce some risk, so I'd appreciate some help testing this.

Partially, indirectly addresses #790.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -156,6 +156,19 @@ define(function (require, exports) {
     };
 
     /**
+     * Get just a document descriptor for the given document reference. This is a light
+     * version of _getDocumentByRef, which skips properties and other information that are
+     * unnecessary for inactive documents.
+     *
+     * @private
+     * @param {object} reference
+     * @return {Promise.<object>}
+     */
+    var _getInactiveDocumentByRef = function (reference) {
+        return descriptor.multiGetProperties(reference, _documentProperties);
+    };
+
+    /**
      * Get an array of layer descriptors for the given document descriptor.
      *
      * @private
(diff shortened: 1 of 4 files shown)


volfied:
Changes look good, but the layer initialization PR burned my tongue one too many times. I'm going to let this one sit for a while and think about where, if any, we request another document's properties *before* selecting/initializing it.

src/js/actions/documents.js
```diff
@@ -647,11 +665,14 @@ define(function (require, exports) {
...
-                var payload = {
-                    selectedDocumentID: document.id
-                };
-
-                this.dispatch(events.document.SELECT_DOCUMENT, payload);
+                if (!document.layers) {
+                    // The now-active document has yet to be fully initialized.
+                    return this.transfer(updateDocument);
+                } else {
+                    this.dispatch(events.document.SELECT_DOCUMENT, {
```
mcilroyc: did you intend not to use `return this.dispatchAsync...`?  Also, in the case that you transfer to updateDocument, wouldn't we also need to emit SELECT_DOCUMENT for the application store?

mcilroyc:
Also looks good to me.  I am worried about places throughout where we assume layers is non-null ... but I guess mostly we rely on first fetching current document, so ok?

src/js/actions/documents.js
iwehrman: 1. No: `dispatch` is perfectly safe inside `then` handlers. 
2. No: `updateDocument`, when no ID is supplied, updates the _active_ document, which updates the application store accordingly.

iwehrman:
Well, I figured out one time I hadn't anticipated previously in which `layers` is expected to be non-null: the search bar. So, I'm gonna fix that up...

iwehrman:
Rebased and pushed another commit which fixes search. When the search bar is toggled open, we check for uninitialized documents and call `updateDocument` on them. This _can_ be a little slow, but... it seems worth it?

I also changed `updateDocument` to not fetch guide info for non-active documents, since a) this failed and b) it wasn't necessary anyway, because that was already happening correctly in `selectDocument`. Back to @volfied!

PR #2662, opened 2015-09-28
Enables a single column mode, where all panels appear in a single column

designedbyscience:
Another small-screen/single column option. Here, there is a menu option for Single Column Mode under the Window menu. 

With this option, all of the panels now appear in a single column, and the toggle panels toolbar disappears. I adjusted some of the sizing of both the panels and the headers in order to fit more on the screen.

Attached is the screen at approximately the resolution of a 13" Macbook Air

vermilionartboards_psd___6_25___dropdown__rgb_8____

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -56,6 +56,16 @@ define(function (require, exports, module) {
          * @type {?DOMElement}
          */
         _target: null,
+        
+        /**
+         * A unique key to append the Dialog ID. 
+         * This is a simple solution to prevent collision of dialog ID names when
+         * a component doesn't unmount before another instance mounts
+         *
+         * @private
+         * @type {Number}
+         */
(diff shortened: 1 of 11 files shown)


designedbyscience:
Now updated with less compressed styling for the layers panel.

src/js/jsx/shared/Dialog.jsx
```diff
@@ -101,12 +101,13 @@ define(function (require, exports, module) {
...
 
             return {
-                open: openDialogs.has(this.props.id)
+                open: openDialogs.has(this.props.id + this._uniqkey)
             };
         },
 
         componentWillMount: function () {
-            this.getFlux().store("dialog").registerDialog(this.props.id, this._getDismissalPolicy());
+            this._uniqkey = (new Date()).getTime();
```
volfied: Can you declare `_uniqKey` at the top of component with JSDoc, and reasons for it's existence?

Issue #2659, opened 2015-09-27
Active document and export button hidden with too many tabs


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/10123931/99cac14c-64fb-11e5-9680-40d053cd6a7a.png)

At least the header buttons and the active document should, I think, always be visible.

chadrolfs:
Related to #2620

Issue #2620, opened 2015-09-24
Doc titles no longer squish or take expanded focus when selected or moused over. 


chadrolfs:
![screen shot 2015-09-24 at 2 36 44 pm](https://cloud.githubusercontent.com/assets/8442631/10087628/e3215088-62c9-11e5-8007-8a0aba2796fb.png)

chadrolfs:
doc titles are hidden and extended under panel and beyond app frame.

PR #2663, opened 2015-09-28
Make sure document tab bar shrinks to fit export button

designedbyscience:
2 bugs, 1 stone

Addresses #2620 and #2659

### DIFF
--- a/src/style/document-header.less
+++ b/src/style/document-header.less
@@ -63,6 +63,7 @@
     position: relative;
     overflow: hidden;
     display: flex;
+    flex-shrink: 1;
 }
 
 .export-header {


mcilroyc:
hovering over the tab makes the tab full-sized, and it seems a bit spastic to me.

mcilroyc:
just to be more clear, here is an example vid of the effect I'm arguing against:
http://cl.ly/0x3n2m1P2C2B
I'm going to assign back to @designedbyscience to force him to answer for this, or tell me to shut up in public.

designedbyscience:
I am restoring previous intended behavior that was lost in some Exports work. There is a lot of utility to seeing the whole filename in the tab (there is much than could be done better in the title bar, but I _think_ that making this less spastic is a fair amount of work)

Shut up @mcilroyc :lemon:

mcilroyc:
#shuttinup

Issue #2557, opened 2015-09-21
Truncation of assets will happen after around 8 assets in Export dialog


ktaki:
![screen shot 2015-09-21 at 4 03 53 pm](https://cloud.githubusercontent.com/assets/2029597/10007506/b5ab1c66-607a-11e5-843c-76d289334145.png)

chadrolfs:
should have an ellipsis (...)

mcilroyc:
ellipsis added by PR.  FBNC to @ktaki

ktaki:
Verified the fix.

PR #2664, opened 2015-09-28
Assets in the export dialog end in ellipsis instead of truncation

DivyaPrabhakar:
Any overflow of assets in the export dialog is hidden and ends in an ellipsis instead of truncating. 
Minor CSS change, addresses #2557.


mcilroyc:
LGTM

PR #2665, opened 2015-09-28
transform panel visual cleanup

placegraphichere:
- spacing and margins
- distribute icons spacing
- disconnected/connected icons match visually


volfied:
Looks good.

Issue #2597, opened 2015-09-23
Unable to move more than one artboard at a time via Select tool click and drag on title


chadrolfs:
create a new doc with multiple artboards
select on canvas by Shift + click on titles with Select tool
click + drag on title

Result: artboards don't move (arrow keys work)

volfied:
if I don't get to it in time, superselect click/drag first attempts a selection, and only starts a drag by using OS.postEvent if selection is valid. The logic seems to be broken for this case.

volfied:
Good and bad news, our code is correct and we send a LMD event to Photoshop... so drag code in core is not starting.

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
verified fix but also verified regression of #1613

PR #2672, opened 2015-09-28
Multiple artboards in selection for dragging / policies

volfied:
This change requires CL 1044998, which will be available in pgdv.mac.688

Problem was that I was preventing both artboard tool switch AND drag operation from happening when multiple artboard selection border was clicked. So now I only prevent the tool switch in PS side.

But this presented a new problem: We would get two "move" events, where one would have "0,0", and because we were debouncing, we would handle the 0,0 event, and immediately return, eating up the other event. So now, if there are any artboards selected, "move" events all are considered legal.

I also noticed that we would draw one big policy frame all selected artboards... however, if oyu had two artboard selected that were not in a row/column, this would create a fake policy on the empty corners, and PS would still draw the handle bars around non policy areas, creating wrong behavior. So I fixed that while I was in there.

Addresses #2597

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1023,14 +1023,19 @@ define(function (require, exports) {
         _layerTransformHandler = synchronization.debounce(function (event) {
             this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
             
+            var appStore = this.flux.store("application"),
+                currentDoc = appStore.getCurrentDocument(),
+                artboardsSelected = currentDoc.layers.selected.some(function (layer) {
+                    return layer.isArtboard;
+                });
+                
             // If it was a simple click/didn't move anything, there is no need to update bounds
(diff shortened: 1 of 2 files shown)


src/js/actions/transform.js
```diff
@@ -1023,14 +1023,19 @@ define(function (require, exports) {
         _layerTransformHandler = synchronization.debounce(function (event) {
             this.dispatch(events.ui.TOGGLE_OVERLAYS, { enabled: true });
             
+            var appStore = this.flux.store("application"),
+                currentDoc = appStore.getCurrentDocument(),
+                artboardsSelected = currentDoc.layers.selected.some(function (layer) {
```
mcilroyc: I think `hasArtboardSelected` would be a good candidate for a layerstructure prop.  You've already used it twice in this PR!  (future suggestion)

mcilroyc:
LGTM.  what an unfortunate `diff` !

Issue #2196, opened 2015-08-24
On-canvas UI should be defined in rems instead of pixels


iwehrman:
So that our on-canvas UI scales along with the rest of the UI based on screen size, we should define our on-canvas UI in rems instead of pixels. (We can observe the font-size at runtime and use that for a rems-to-pixels conversion for compatibility with DOM APIs if necessary.)

Compare the images below, which show the same UI simulated with different screen sizes:
![image](https://cloud.githubusercontent.com/assets/1075154/9454607/262d02ce-4a78-11e5-9f02-5080068d604f.png)
![image](https://cloud.githubusercontent.com/assets/1075154/9454610/2d404ada-4a78-11e5-8477-ec66478f2dfb.png)

volfied:
@iwehrman please close once you merge the rem PR.

iwehrman:
Leaving this open because it isn't aware of `displayConfigurationChanged` events. Self-assigning.

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
Closing this, but #2861 is related.

PR #2688, opened 2015-09-29
Use REMs to draw sampler overlay and artboard adders

volfied:
Addresses https://github.com/adobe-photoshop/spaces-design/issues/2196

That's me in the sampler
That's me in the artboards

@iwehrman please review


designedbyscience:
:100:

iwehrman:
I like this, but can we push the multiplication into a helper function like `pxToRem`?

volfied:
OK @iwehrman, made the change.

iwehrman:
Radical.

Issue #2641, opened 2015-09-25
Select All shortcut results in Uncaught TypeError and bad state


chadrolfs:
New doc
create a shape 
cmd/ctrl + a to select all

Result: Uncaught TypeError: Cannot read property 'get' of undefined

Additional note: if you reset DS and then Select All, it works without error but you are unable to move the Artboard (separate issue?)

volfied:
This happens in DocumentExports.getUniformAssets method. Passing to @mcilroyc

PR #2689, opened 2015-09-29
DocumentExports.getLayerExports will always return an Immutable List

mcilroyc:
This was really the practical use of this method anyway, which allowed some cleanup.

Addresses #2641.  In some cases, new layers did not have layer-level exports defined properly in the `ExportDocuments` model.

This PR also tries harder to safely default artboards to "exportEnabled" and having one default asset.  When opening a document that has undefined export-related `extension` values, we assume this is a fresh document (never opened in DS, or at least the exports were not monkeyed with), so we default the values accordingly.  This is a request by @davealonzo.


volfied:
So that's why. Thank you!

Issue #2656, opened 2015-09-26
Datalist hairline alignment issues


iwehrman:
Here are three instances of expanded `Datalist` components:

1. Blend mode for groups:
![image](https://cloud.githubusercontent.com/assets/1075154/10120054/ff23d486-645e-11e5-8c2d-952a3d72a909.png)
Note that the hairline is not visible when the `Datalist` is expanded. Compared to the next two examples, this should be considered the "correct" implementation of the design.

2. Blend mode for type/vectors:
![image](https://cloud.githubusercontent.com/assets/1075154/10120056/14a44174-645f-11e5-8863-442df5faf913.png)
So close... note that the hairline is peeking out from behind the `Select` menu.

3. Libraries list:
![image](https://cloud.githubusercontent.com/assets/1075154/10120063/42dac25c-645f-11e5-9100-4c3a6232bca2.png)
Argh! The `Select` component should both be widened to cover the entire hairline, and the entry titles should be left aligned with the text in the input, as in the previous two examples.

iwehrman:
The Type datalists have problems too:

1. Type face has a peeking hairline: 
![image](https://cloud.githubusercontent.com/assets/1075154/10123990/9ee5314c-64fd-11e5-8650-9e289ec2734a.png)

2. Type weight has all that and a menu alignment problem:
![image](https://cloud.githubusercontent.com/assets/1075154/10123993/ac27f8da-64fd-11e5-9c6f-f2ca81d93ffe.png)

placegraphichere:
these are all set individually for some reason. So time just needs needs to be made to get them into shape. 

@davealonzo also mentioned that we could try to hide the underscore when open too. Assigning to myself

placegraphichere:
See #2690

iwehrman:
Fixed by #2690.

PR #2690, opened 2015-09-29
Dialog Dropdown alignment fixes

placegraphichere:
- fix for #2656 
- fixed blendmodes for group, shape, text
- fixed typeface, weight
- fixed effects blendmode
- fixed export scale, format
- fixed library dialog

- typeface dialog width

### DIFF
--- a/src/style/shared/dialog.less
+++ b/src/style/shared/dialog.less
@@ -85,16 +85,16 @@ dialog {
     font-size: 100%;
     display: flex;
     justify-content: flex-start;
-    border: none;
     width: @panel-column-width - (@dialog-margin*2) - (@dialog-padding*2);
     max-height: 25rem;
     height: auto;
     border-radius: @dialog-radius;
-    background:@bg-border;
-    border:@hairline solid @underlines;
+    background: @bg-border;
+    border: @hairline solid @underlines;
     padding: @dialog-padding;
     margin: @dialog-margin;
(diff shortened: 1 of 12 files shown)


src/style/shared/dialog.less
```diff
@@ -109,9 +109,27 @@ dialog::backdrop {
...
 }
 
+.column-28 + .dialog-blendmodes{
+    width: 29.2rem;
+}
+
+.column-14 + .dialog-blendmodes{
+    width: 15.6rem;
+    left: 5.1rem;
+    margin-top: -1px;
```
iwehrman: Do you think this will work at all resolutions?

iwehrman:
One more in Export scale:
image

iwehrman:
One bug, one question. Back to @placegraphichere.

iwehrman:
Also, a process suggestion: when you're actively working on fixing a bug, it's nice to set the "Fix In Progress" label on that bug. It helps a) make sure nobody else starts working on it without talking to you, and b) helps us remember to verify that the bug is closed when you're done. Also, whoever merges your changes should go back and swap the label to "Fixed But Not Closed" and re-assign it to either QE or the original filer to verify the fix.

placegraphichere:
back to you @iwehrman

iwehrman:
Very nice.

Issue #2608, opened 2015-09-24
Checkbox styling problems


iwehrman:
On a 5k iMac with `font-size: 62.5%`, checkboxes look like this:
![image](https://cloud.githubusercontent.com/assets/1075154/10063270/a84b01ec-6220-11e5-8620-4712b57fae5e.png)

placegraphichere:
take a look at #2691

chadrolfs:
PR is still open assigned to Seth

iwehrman:
FBNC to @designedbyscience.

designedbyscience:
I have another PR which addresses this even further #2908

PR #2691, opened 2015-09-29
Checkbox styling

placegraphichere:
- fix for #2608

### DIFF
--- a/src/style/shared/button.less
+++ b/src/style/shared/button.less
@@ -218,7 +218,7 @@ button {
         top: 0;
         left: 0;
         background: transparent;
-        border: .21rem solid @warm-gray;
+        border: .21rem solid @icon-active;
         border-radius: .21rem;
         &:after {
             content: '';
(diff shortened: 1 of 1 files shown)


src/style/shared/button.less
```diff
@@ -259,38 +262,8 @@ button {
             opacity: 1;
             transform: rotate(0);
         }
-    }
-}
-
-// Fixing for Retina
-@media (-webkit-min-device-pixel-ratio: 2) {
```
mcilroyc: Just double checking in case this was a merge issue or something ...  What changed such that this retina fix is no longer necessary?

designedbyscience: Things still look off on Retina, so I think not

iwehrman:
Does anyone know if this improves the state of things or makes it worse?

iwehrman:
On all my displays this looks better. @designedbyscience: I'm not sure what your comment above means. I am ready to merge this. Should I wait?

iwehrman:
Welp, nobody spoke up, so I'm merging this since AFAICT it makes things less bad!

Issue #2564, opened 2015-09-22
Color picker swatch has radius edge UI issue


chadrolfs:
![ice_screenshot_20150922-095211](https://cloud.githubusercontent.com/assets/8442631/10025132/8aefa088-610f-11e5-9ea9-3711353b25b0.png)

iwehrman:
Still happening:
![image](https://cloud.githubusercontent.com/assets/1075154/10085946/9b964060-62c0-11e5-88be-44ce1940d868.png)

designedbyscience:
Okay, I am attaching a little test file here, to demonstrate general findings:

1) At any color with alpha < 1, those little white corners appear, if the background is a light color
2) For opaque colors, we can prevent that by removing the z-index from the inner element.

Below file is actually HTML, but Github doesn't let me upload those. 
[border-radius-rendering-issue.txt](https://github.com/adobe-photoshop/spaces-design/files/4477/border-radius-rendering-issue.txt)

shaoshing:
@chadrolfs can you verify this?

chadrolfs:
Aaaalmost.  Looks great on non-retina.  On retina, top left corner still...
http://cl.ly/image/1r1f3z3E0i1P

designedbyscience:
Updates to this one:

For opaque colors, we shouldn't be seeing the white corners anywhere.
For colors with alpha < 1, there will be white corners still and I'm not sure what we should do as that is a Chrome rendering bug that appears in standard desktop Chrome still

chadrolfs:
This looks fixed to me now.  @designedbyscience?

chadrolfs:
ah, i see what you're seeing.  http://cl.ly/image/2M252d363B0m

designedbyscience:
Yea... I only have a mitigation I'm working on, and I'm not sure it is any better than we have now

designedbyscience:
It is only particularly noticeable in a small range of alpha

chadrolfs:
possibly deferrable.

iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/12539308/969371b6-c2a4-11e5-8807-060ccb71ca62.png)
Can we just give up on this particular border radius as a workaround for this bug?

chadrolfs:
hmmm, i dunno, now that it's showing up more obvious without alpha, we should probably fix if we can.

PR #2695, opened 2015-09-29
Updates to color swatches

designedbyscience:
1) Fix alignment of Effects color swatch, which should have lined up with the Appearance color swatches
2) Correct the stroke swatch, which I had done wrong and @davealonzo nicely pointed out to me (and happily made some of the code cleaner)
3) Attacked that corner radius issue #2564 some more: This issue is largely to blame on Chrome, and can be seen in normal desktop Chrome as well. It seems to have several compounding issues
* Flexbox + z-index create it for any color
* Any color with alpha < 1 also creates it

I had written a hack for the opacity issue previously, which was actually not necessary as I was able to remove the z-index instead. So for opaque colors, it looks good. Anything less than opaque still shows white corners, though it only looks really bad in the range 0.85 < alpha < 1

### DIFF
--- a/src/js/jsx/sections/style/Fill.jsx
+++ b/src/js/jsx/sections/style/Fill.jsx
@@ -115,8 +115,6 @@ define(function (require, exports, module) {
 
             var fillOverlay = function (colorTiny) {
                 var fillStyle = {
-                    height: "100%",
-                    width: "100%",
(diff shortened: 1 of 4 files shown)


shaoshing:
[LGTM](http://lgtm.in/i/691QZwaSp)

PR #2697, opened 2015-09-29
remove uneeded call to remove Shortcut

pineapplespatula:
According to Ian we already wipe the shortcuts onReset


src/js/actions/tools.js
```diff
@@ -742,7 +742,6 @@ define(function (require, exports) {
      */
     var onReset = function () {
         descriptor.removeListener("toolModalStateChanged", _toolModalStateChangedHandler);
-        this.transfer(shortcuts.removeShortcut, "vectorModeSwitch");
```
iwehrman: Also, this promise should have been blocking the promise from this function...

Issue #2601, opened 2015-09-23
Artboard does not show as selected in layers panel when active on canvas after layer delete


chadrolfs:
new doc with one shape inside the AB and one outside
select the layer outside and delete 

Result: AB is now selected on canvas but layers panel does not confirm that.

https://www.dropbox.com/s/idz9i8m3yxnhtpr/AB-selectionConfusion.mov?dl=0

volfied:
I think recent changes in SCU of LayersPanel fixed this. back to @chadrolfs

chadrolfs:
Nope, unfortunately not.

iwehrman:
FBNC to @chadrolfs.

PR #2704, opened 2015-09-29
Initialize newly selected layers after layer deletion

volfied:
Just another initialize. RemoveLayers dispatches event, event changes selection, selected layers aren't initialize, so document store doesn't emit a change event. So we have to initialize those layers to get a change event. Boy saves world.

Addresses #2601


iwehrman:
:hamburger: :fries:

Issue #2699, opened 2015-09-29
Sampling stroke fails across artboard multi-layered doc


chadrolfs:
https://photoshop.slack.com/files/zorana/F0BGMNP1R/sampling.mp4

you can use explore-mt-hood.psd to repro (in my transfer folder on sjshare/photoshop).

Open psd from my transfer
(Select tool) select Activities rect in INDEX AB (Rounded Rectangle 1)
apply a stroke (change size and color)
apply an inner shadow
select Activities rect in Mt. Hood AB (also named Rounded Rectangle 1)
hit i for Sampler
mouse over Rounded Rect 1 in INDEX AB
hit space and select stroke

chadrolfs:
![screen shot 2015-09-29 at 11 16 34 am](https://cloud.githubusercontent.com/assets/8442631/10173875/7ed7e134-669b-11e5-9f40-f9b412308e47.png)

volfied:
Please re-assign after steps are added.

volfied:
I don't know what's happened, but none of these shape layers have their `AGMStrokeStyleInfo` fields available to us.

This is not an initialization issue, PS is not initializing these shapes completely.

volfied:
https://watsonexp.corp.adobe.com/#bug=4066169

isonno:
This is as designed.

Layers created without strokes (or more specifically, without the stroke machinery activated) do not have `AGMStrokeStyleInfo` attributes.  The fill color is buried in the Layer's `adjustment` attribute, and the `fillEnabled` attribute is set to `null` (vs. set to `false` if `AGMStrokeStyleInfo` is present and there is no fill specified).

We could explore some way to force the creation of `AGMStrokeStyleInfo` attributes, but this could incur significant overhead and may have backwards compatibility issues.  It's more expedient to deal with this in JS code (CopyCSS also works around this issue).

volfied:
John is right, fix incoming.

tiburbage:
OK, accepting AsDesigned for Watson bug and closing.

PR #2717, opened 2015-09-29
Add stroke to layers without a stroke object during setStroke command

volfied:
Another artifact from multiple strokes age, where we checked to see if all layers had strokes and if not, we called _refreshStrokes. Problem was, with transactions, we wouldn't play the actual setStroke objects yet, so we'd be trying to grab a non existent stroke from PS, and things would burn.

I removed the check, and need to call _refreshStrokes, because we can recreate the stroke ourselves using available strokeProperties.

Addresses #2699

### DIFF
--- a/src/js/actions/shapes.js
+++ b/src/js/actions/shapes.js
@@ -187,34 +187,23 @@ define(function (require, exports) {
 
         options = _mergeOptions(options, document, strings.ACTIONS.SET_STROKE);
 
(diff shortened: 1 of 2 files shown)


volfied:
In my absence, I give @shaoshing full control of this PR and branch to address any review changes he or any one else may have.

shaoshing:
This looks good. And also fixed setting stroke alignment.

Issue #2669, opened 2015-09-28
Entering and exiting mask mode leaves layer icon indicating it has a mask 


chadrolfs:
select a layer, hit m to enter mask mode, hit m again to exit it (without adding a mask)
Result: layer is still red

Layer indicator should be cleared when the mask is deleted.

mcilroyc:
FBNC to @chadrolfs  - please note that this fix introduces a history bug for which we are blocked by core.  So THIS bug should be fixed, but please note that history store gets nuked by this "double m" workflow now.

PR #2718, opened 2015-09-29
Jecohen/2669

pineapplespatula:
fix for issue #2669 requires a very recent jenkins build of PS and accompanying adapter PR https://github.com/adobe-photoshop/spaces-adapter/pull/154

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -40,6 +40,7 @@ define(function (require, exports) {
 
     var events = require("../events"),
         guides = require("./guides"),
+        layers = require("./layers"),
         locks = require("js/locks"),
         policy = require("./policy"),
         ui = require("./ui"),
@@ -558,6 +559,7 @@ define(function (require, exports) {
 
         var initPromise,
             dispatchPromise,
+            removePromise,
             policyPromise;
 
         // if we are swtiching to vector mask mode, make sure the layer has a vector mask
@@ -601,7 +603,29 @@ define(function (require, exports) {
                 .then(function (policyID) {
                     this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
                 });
+            removePromise = Promise.resolve();
         } else {
+            // delete empty masks when leaving mask mode
+            removePromise = this.transfer(layers.resetLayers, currentDocument, currentLayer)
+                .then(function () {
+                    currentLayer = appStore.getCurrentDocument().layers.selected.first();
+
+                    if (currentLayer.vectorMaskEnabled && currentLayer.vectorMaskEmpty) {
+                        return descriptor.playObject(vectorMaskLib.deleteVectorMask());
+                    } else {
+                        return Promise.resolve();
+                    }
+                })
+                .bind(this)
+                .then(function () {
+                    var payload = {
+                            documentID: currentDocument.id,
+                            layerIDs: Immutable.List.of(currentLayer.id),
+                            vectorMaskEnabled: false
+                        };
+                    return this.dispatch(events.document.history.optimistic.REMOVE_VECTOR_MASK_FROM_LAYER, payload);
+                });
+
             var pointerPolicyID = toolStore.getVectorMaskPolicyID();
          
             if (pointerPolicyID) {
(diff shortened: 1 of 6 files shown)


pineapplespatula:
dont merge this yet, i'm fixing an issue where i overzealously delete things

src/js/actions/export.js
```diff
@@ -969,7 +969,7 @@ define(function (require, exports) {
     exports.exportLayerAssets = exportLayerAssets;
     exports.exportDocumentAssets = exportDocumentAssets;
     exports.setUseArtboardPrefix = setUseArtboardPrefix;
-    exports.afterStartup = afterStartup;
+    // exports.afterStartup = afterStartup;
```
iwehrman: This.

src/js/actions/tools.js
```diff
@@ -600,7 +602,28 @@ define(function (require, exports) {
                 .then(function (policyID) {
                     this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
                 });
+            removePromise = Promise.resolve();
         } else {
+            // delete empty masks when leaving mask mode
+            removePromise = this.transfer(layers.resetLayers, currentDocument, currentLayer)
+            .then(function () {
```
iwehrman: The tabbing is wrong here.

src/js/actions/tools.js
```diff
@@ -621,44 +644,53 @@ define(function (require, exports) {
...
-                    return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
+                    return Promise.join(initPromise, dispatchPromise, policyPromise, removePromise);
                 })
                 .then(function () {
-                    descriptor.batchPlayObjects([vectorMaskLib.enterFreeTransformPathMode()],
-                        { synchronous: false });
+                    if (!currentLayer.vectorMaskEmpty) {
+                        return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+                            .then(function () {
+                                descriptor.batchPlayObjects([vectorMaskLib.enterFreeTransformPathMode()],
```
iwehrman: Are you intentionally throwing this promise away? Or should you be doing something with it?

pineapplespatula: not intentionally, but i'm not clear on how  synchronous: false affects promises.

src/js/actions/export.js
pineapplespatula: I'll take it out before i think the PR is merge-able, but i prefer to leave it in while the PR is review only since my local build doesn't have generator .

pineapplespatula:
@iwehrman  back to you. I'm still seeing a concurrence thing, but i think this PR now fixes enough bugs that introducing that one is still worth it.

pineapplespatula:
ian is in meetings all day, @mcilroyc can you look at this one too?

mcilroyc:
@pineapplespatula how do you commit masks now?

pineapplespatula:
it should commit. When i was hitting m. it would commit 90% of the time and clear 10%, if your hitting it much more often then I'll take a closer look as to why photoshop is lying to us

pineapplespatula:
@mcilroyc that should fix the issue you're seeing.

src/js/actions/tools.js
```diff
@@ -600,7 +602,29 @@ define(function (require, exports) {
...
+                    }
+                })
+                .bind(this)
+                .then(function () {
+                    var payload = {
+                            documentID: currentDocument.id,
+                            layerIDs: Immutable.List.of(currentLayer.id),
+                            vectorMaskEnabled: false
+                        };
+                    return this.dispatch(events.document.history.optimistic.REMOVE_VECTOR_MASK_FROM_LAYER, payload);
```
mcilroyc: I think both this and the "ADD" variant should be non-history changing events.  ("events.document.REMOVE...").  As of right now, a "ghostly" unnecessary history state is being added each time "m" is pressed.  The real history changing event happens when the mask is drawn, n'est pas?

pineapplespatula: adding a reveal all vector mask is what happens when you enter vector mask mode. This is a history changing event. 
Deleting said vector mask is also a history event.
I guess i'm not clear on what you are suggesting.

mcilroyc:
If I click a layer that has no mask, and I click "m" twice, two history states are created in DS, but NONE are created in photoshop (or at least we don't get the historyState events from photoshop). This is very bad, from a history store perspective. 

ss

pineapplespatula:
Photoshop does have 2 history states from the above action. I had made my events optimistic(which may or may not have been correct) i just added a PR to change the type of event they are which seems to solve the error you are seeing, but i am very unclear as to what variety of event these should be.

pineapplespatula:
Removed the history change after talking to Cory, we will need a separate issue/core bug around not receiving history states for these events.

src/js/actions/tools.js
mcilroyc: It seems like this will dispatch *always* when exiting vector mode, which is not desirable for existing+unchanged vector modes.

mcilroyc:
I get the following when "press M twice" with an existing masked layer:

ss

mcilroyc:
Two new comments: one I think is a bug (history related, but independent of the core bug) and the other is an error I'm getting.  back to @pineapplespatula.  If this is fixed in other PRs... LMK which.

pineapplespatula:
When you press M twice you are hitting the endModalToolState command which should register as a commit, but instead registers as a cancel. and we're getting an error there. this is a known issue that I'm handling this in a different branch and shouldn't block this PR

src/js/actions/tools.js
```diff
@@ -601,7 +603,32 @@ define(function (require, exports) {
                 .then(function (policyID) {
                     this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
                 });
+            removePromise = Promise.resolve();
         } else {
+            // delete empty masks when leaving mask mode
+            removePromise = this.transfer(layers.resetLayers, currentDocument, currentLayer)
+                .bind(this)
+                .then(function () {
+                    currentLayer = appStore.getCurrentDocument().layers.selected.first();
```
mcilroyc: are you protected from multiple layers being selected?  Probably somewhere... but still, this seems a bit dangerous.  File it away for future reference.

src/js/actions/tools.js
```diff
@@ -622,44 +649,53 @@ define(function (require, exports) {
                     currentTool.nativeToolName());
 
                 firstLaunch = false;
-                return Promise.join(initPromise, defaultPromise, resetPromise, dispatchPromise, policyPromise);
+                return Promise.join(initPromise, defaultPromise, resetPromise, dispatchPromise, policyPromise,
```
mcilroyc: is it actually safe to call initPromise and removePromise in parallel?

src/js/actions/tools.js
```diff
@@ -622,44 +649,53 @@ define(function (require, exports) {
...
-                    return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
+                    return Promise.join(initPromise, dispatchPromise, policyPromise, removePromise);
                 })
                 .then(function () {
-                    descriptor.batchPlayObjects([vectorMaskLib.enterFreeTransformPathMode()],
-                        { synchronous: false });
+                    if (!currentLayer.vectorMaskEmpty) {
+                        return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+                            .then(function () {
+                                return descriptor.batchPlayObjects([vectorMaskLib.enterFreeTransformPathMode()],
```
mcilroyc: `playObject` would have sufficed?  just sayin

src/js/tools/superselect.js
```diff
@@ -70,19 +72,26 @@ define(function (require, exports, module) {
...
-                .then(function () {
-                    var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                        OS.eventKind.LEFT_MOUSE_DOWN);
-                    return this.transfer(policy.addPointerPolicies, pointerPolicy);
-                })
-                .catch(function () {
-                    // Silence the errors here
-                });
+            var currentLayers = currentDocument.layers.selected,
+                currentLayer = currentLayers.first();
```
mcilroyc: same basic question as before - do we need to test that selected.size ===1?

mcilroyc:
I have a few more comments inline that you may want to consider for future PRs.  Also, file this away... I now think that these two events may need to be "nonoptimistic" history events, but no sense in worrying about that until we get the core change in.

Issue #2713, opened 2015-09-29
Artboard coordinates are incorrect after deleting a layer


chadrolfs:
new doc
draw rect shapes; one shape inside the AB and one outside
with Select tool delete the second layer drawn outside the AB

Result: AB is now selected but coordinates are off

https://www.dropbox.com/s/160dh280hb9lh1n/AB-badCoord-afterDelete.mov?dl=0

chadrolfs:
![screen shot 2015-09-29 at 3 12 19 pm](https://cloud.githubusercontent.com/assets/8442631/10179678/6ac6ee8a-66bc-11e5-85cb-15c9d51dfb50.png)

chadrolfs:
found confirming fix of #2601

volfied:
In case I can't fix this before I leave, here is the problem:

With removeLayer transferring to `initializeLayers` for the selected layers. I get the updated bounds of the artboard. And then I get a autoCanvasResizeShift event, and I translate all the bounds by that amount, where the artboard's bounds were already up to date, and didn't need translating.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
With these steps, yes, this bug is not reproducible but #2754 still occurs

PR #2721, opened 2015-09-30
Enqueue initialize instead of transferring to in removeLayers

volfied:
Problem:

On deleting a layer, we would transfer to  initialize the newly selected layer so it would actually get doc store to emit a change event. However, if the deleted layer caused a canvas resize shift, we would handle this by translating all the layers in the document by the shift. And we would translate the newly selected layer's shifted boundaries. So things would go to shift.

So now, instead of transferring to initializeLayers, I enqueue that action. This way, handleCanvasShift gets played first, and then init layers gets played, and the bounds are correctly shifted / initialized.

@iwehrman I don't know whether you'll agree with this fix. But I can't figure out a cleaner way. So please review and argue with me tomorrow.

Addresses #2713

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1030,12 +1030,12 @@ define(function (require, exports) {
                 var nextDocument = this.flux.store("document").getDocument(document.id),
                     selected = nextDocument.layers.selected;
 
-                return this.transfer(initializeLayers, nextDocument, selected);
+                this.flux.actions.layers.initializeLayers(nextDocument, selected);
             });
     };
     removeLayers.reads = [locks.PS_DOC];
     removeLayers.writes = [locks.JS_DOC];
-    removeLayers.transfers = ["history.queryCurrentHistory", initializeLayers];
+    removeLayers.transfers = ["history.queryCurrentHistory"];
     removeLayers.post = [_verifyLayerIndex, _verifyLayerSelection];
 
     /**
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1030,12 +1030,12 @@ define(function (require, exports) {
                 var nextDocument = this.flux.store("document").getDocument(document.id),
                     selected = nextDocument.layers.selected;
 
-                this.flux.actions.layers.initializeLayers(nextDocument, selected);
+                return this.transfer(initializeLayers, nextDocument, selected);
             });
     };
     removeLayers.reads = [locks.PS_DOC];
     removeLayers.writes = [locks.JS_DOC];
-    removeLayers.transfers = ["history.queryCurrentHistory"];
+    removeLayers.transfers = ["history.queryCurrentHistory", initializeLayers];
     removeLayers.post = [_verifyLayerIndex, _verifyLayerSelection];
 
     /**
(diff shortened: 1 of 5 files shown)


iwehrman:
Sorry, I don't understand your explanation of the problem. In particular, I don't understand how calling initializeLayers after removeLayers is over solves the problem. In any case, it seems like a better fix would be to call initializeLayers a _second_ time somewhere else. I really don't want to break removeLayers like this.

volfied:
Transfer chain is this: `layers.deleteSelected` => `layers.removeLayers` => `initializeLayers`

When `deleteSelected` plays the `"delete"` descriptor, PS emits us a `autoCanvasResizeShift` event, telling us all layers have been shifted by x/y. As a reaction to this event, we enqueue `layers.handleCanvasShift` action.

Meanwhile, `removeLayers` transfers to `initializeLayers`, which calls `resetLayers` on the layer we just selected as result of the "delete" command. And because in PS world, bounds have already shifted, we get the latest and shifted bounds for this layer. And then `initializeLayers` finishes, and thus `layers.deleteSelected` finishes.

And we play the enqueued `layers.handleCanvasShift` action, which grabs the newest document model at the play time, and dispatches a `TRANSLATE` message for ALL the layers in the document by the canvas shift. This corrects all the layers fine, except the one we recently initialized. That layer gets shifted twice, and #2713 occurs.

I see couple solutions:

 - A complicated flag system, where we set a private boolean variable to true when we get `autoCanvasResizeShift`. In turn, during removeLayers, if this flag is true, we add IDs of all initialized layers into an array as another private variable. And during `handleCanvasShift`, we subtract the IDs in this array from arrays we're shifting, since they are more up to date. And we reset this array everytime we get the `autoCanvasResizeShift` event, so in cases we don't get it due to deletion, we don't accrue extra layers.

- On `autoCanvasResizeShift` handler (not the action, but the handler that enqueues the action), we grab the current document model (that has the non-initialized layer in it), and pass that to `handleCanvasShift`. And in `handleCanvasShift`, instead of just dispatching a translate message, we dispatch a setPosition, with all the updated positions of these outdated model's layers. This, in theory, would return the identical bounds as `initializeLayers` would get on the newly selected layer, while keeping everything else true. (This seems safest to me, no flags, no nonsense)

volfied:
@iwehrman I implemented the second solution I suggested. It seems to work.

Please review.

volfied:
While Ian is out, I'm also adding a postCondition verification for this solution so we can test it a bit better.

src/js/actions/layers.js
```diff
@@ -2130,24 +2130,33 @@ define(function (require, exports) {
...
-                    x: event.to.horizontal,
-                    y: event.to.vertical
-                }
+                documentID: document.id,
+                positions: positions
             };
-        
-        return this.dispatchAsync(events.document.TRANSLATE_LAYERS, payload);
+    
+        return this.dispatchAsync(events.document.history.optimistic.REPOSITION_LAYERS, payload);
```
iwehrman: Why the new event here? This would seem to cause a history state to be created?

volfied: TRANSLATE_LAYERS, in it's handler, would eventually call `document.layers.translateLayers` on the most recent document. 

Instead, I call it here on an outdated model, and then use REPOSITION_LAYERS to set the bounds directly on the updated model.

Not sure about the history state thing, @mcilroyc can you advise? Should i create an identical nonoptimistic dispatch?

mcilroyc: Using `events.document.history.optimistic.REPOSITION_LAYERS` definitely seems wrong, in that it would create a NEW history state.  It is not clear to me which of the two is correct in your situation:
1) Change to `events.document.history.amendment.REPOSITION_LAYERS` so that this change will be updated on the document model cached as part of the *current* state
2) Change to `events.document.REPOSITION_LAYERS` which would have no history consequences at all.

iwehrman:
@jsbache: this is an important bug to discuss, even if @volfied's workaround is sufficient for the 16.1 release.

@mcilroyc: can you take this PR over, making the required history state changes? Once you've done that, please re-assign it back to me for review.

mcilroyc:
@iwehrman back to you.  I pushed up another commit that uses a non-history version of REPOSITION_LAYERS and I got rid of the old TRANSLATE_LAYERS completely since it is now deprecated.

iwehrman:
Welp, let's give it a try!

Issue #2722, opened 2015-09-30
Transform handles flash at the beginning of drag


chadrolfs:
draw several rect shapes
click-drag one, then another

Result: transform handles flash at the beginning of the drag

Expect: transform handles to show on mouse up

iwehrman:
FBNC to @chadrolfs.

PR #2723, opened 2015-09-30
Hide transform controls at the beginning of drag

volfied:
Addresses #2722, transform controls are hidden in two conditions: We set tool options, or we go in a modal state. During click/drag we first select the layer, and then initiate drag. And during that brief instance, transform controls of the newly selected layer flash.

So now, we first disable transform controls, and then try to select. And after initiating drag, we play a modalSafe toolOptions call to re-set the transform controls, so they appear at the end of drag.

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -28,6 +28,7 @@ define(function (require, exports) {
         Immutable = require("immutable");
 
     var descriptor = require("adapter/ps/descriptor"),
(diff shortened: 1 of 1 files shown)


volfied:
In my absence I give @iwehrman full control of this PR and address/change the code as he or anyone else comments

iwehrman:
I'm definitely a little concerned that this will exacerbate the drag-start latency. But, I can't really notice the difference in brief testing. So, OK!

Issue #2719, opened 2015-09-29
Drag image from libraries freezes the app


zoranagee:
File New
Drag an image from libraries
Application freezes

NOte: I am using jenkins playgroundev build 697 with the JS bits that Chad sent (so what's labeled as sprint 32?)

chadrolfs:
Depending on what you drag, it may freeze.  See #2562

chadrolfs:
just repro'd this.

chadrolfs:
gets stuck on place

chadrolfs:
only way I've found out of it is to right-click on app icon in OS dock > Quit and then you get a prompt to Place, then you can click Cancel on the prompt to save to continue working.

shaoshing:
this will need Jenkin build (>= 703) to fully work, @chadrolfs

PR #2728, opened 2015-09-30
Suspend/Restore all policy around the placeEvent of graphic asset.

shaoshing:
This PR fix the issue that PS is not responding to mouse click in the transform mode after a graphic asset is dopped on the canvas. This is related to #2719 and requires latest Jenkin build (>= 703) to fully work.

Since this is related to the event policy, @mcilroyc can you reivew my PR? Thank you.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -160,6 +160,15 @@ define(function (require, exports) {
      * @type {LibrarySyncStatus}
      */
     var _librarySyncStatus;
+    
+    /**
+     * Event handlers initialized in beforeStartup.
+     *
+     * @private
+     * @type {function()}
+     */
+    var _placeEventHandler,
+        _toolModalStateChangedHandler;
 
     /**
      * Finds a usable representation for the image element that PS will accept
@@ -791,12 +800,18 @@ define(function (require, exports) {
             })
             .bind(this)
             .then(function (path) {
-                return this.transfer(policyActions.addKeydownPolicy, true, os.eventKeyCode.ENTER, null)
-                    .bind(this)
-                    .then(function (policyID) {
+                var hasAlt = this.flux.stores.modifier.getState().alt;
+                
+                return Promise.bind(this)
+                    .then(function () {
+                        // Suspend all policies to allow moving and transforming the new graphic layer.
+                        if (!hasAlt) {
+                            return this.transfer(policyActions.suspendAllPolicies);
+                        }
+                    })
+                    .then(function () {
                         var docRef = docAdapter.referenceBy.id(currentDocument.id),
-                            placeObj = libraryAdapter.placeElement(docRef, element, path, location),
-                            hasAlt = this.flux.stores.modifier.getState().alt;
+                            placeObj = libraryAdapter.placeElement(docRef, element, path, location);
 
                         // This command is played "asynchronously", which means that Photoshop will
                         // allow CEF to run while the command is being played (i.e., while in the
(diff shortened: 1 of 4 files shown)


src/js/actions/documents.js
```diff
@@ -826,28 +827,40 @@ define(function (require, exports) {
...
-        if (!document) {
-            var error = new Error("Place event received without a current document");
-            return Promise.reject(error);
+            replace = false,
+            promise = Promise.bind(this);
+            
+        if (policyStore.areAllSuspended()) {
+            // If the policies are all suspended for drag-n-drop graphic asset, we restore them after the 
+            // asset is placed.
+            promise = promise.then(function () {
```
mcilroyc: I think more commonly in our code we would see: `promise = this.transfer(...`   and then a few lines below: `} else { promise = Promise.resolve():`   

... just a suggestion.

src/js/actions/libraries.js
```diff
@@ -806,12 +812,9 @@ define(function (require, exports) {
                         // So, although the command does not resolve until the placement has finished,
                         // it does resolve before the layer model has been updated because the event
                         // has not yet been received. For more details see issue #2177.
-                        var placeElementPromise = descriptor.playObject(placeObj, { synchronous: false })
-                            .bind(this)
-                            .finally(function () {
-                                return this.transfer(policyActions.removeKeyboardPolicies, policyID, true);
-                            });
-                            
+                        return descriptor.playObject(placeObj, { synchronous: false });
```
mcilroyc: I think you should also catch errors here, and restorePolicies in that case.  

Related question: after the user places the object (and while policies are suspended) is it possible for them to "cancel" the operation (via ESC?)?  I'm trying to think of ways that we could initiate the placeObject flow, but then NOT end up getting an event back from photoshop.

mcilroyc:
first pass looks good.  A couple of things noted.

src/js/actions/libraries.js
shaoshing: Handled the possible exception below.

src/js/actions/libraries.js
```diff
@@ -1160,6 +1192,21 @@ define(function (require, exports) {
...
+        }.bind(this));
+            
+        descriptor.addListener("toolModalStateChanged", function (event) {
+            var isPlacingGraphic = this.flux.store("library").isPlacingGraphic(),
+                modalStateCancelled = event.reason && event.reason._value === "cancel";
+
+            if (isPlacingGraphic && modalStateCancelled) {
+                this.flux.actions.libraries.handleCompletePlacingGraphic();
+            }
+        }.bind(this));
```
shaoshing: Added this for the event of hitting ESC to cancel the placeElement event.

shaoshing:
Hi @mcilroyc just updated this PR to address your concerns.

src/js/actions/libraries.js
```diff
@@ -821,26 +829,50 @@ define(function (require, exports) {
...
+            .then(function () {
+                var policyStore = this.flux.store("policy");
+                
+                if (policyStore.areAllSuspended) {
+                    return this.transfer(policyActions.restoreAllPolicies);
+                }
+            });
+    };
+    handleCompletePlacingGraphic.reads = [locks.JS_APP];
+    handleCompletePlacingGraphic.writes = [];
```
mcilroyc: does PLACE_GRAPHIC_UPDATED not write to an of our stores?  just *seems* like there should be a write lock.  but maybe not

src/js/stores/library.js
```diff
@@ -504,6 +522,15 @@ define(function (require, exports, module) {
         },
         
         /**
+         * Ture if DS is in placing graphic mode.
+         * 
+         * @return {boolean}
+         */
+        isPlacingGraphic: function () {
```
mcilroyc: side note, for future, I think we should strive to use the semi-standard `getState()` to provide simple store properties instead of individual getter methods like this.

src/js/stores/library.js
```diff
@@ -114,6 +119,8 @@ define(function (require, exports, module) {
                 events.libraries.ASSET_RENAMED, this._handleElementRenamed,
                 events.libraries.ASSET_REMOVED, this._handleElementRemoved,
                 
+                events.libraries.PLACE_GRAPHIC_UPDATED, this._handlePlaceGraphicUpdate,
+
```
mcilroyc: what's with all this space?

src/js/actions/libraries.js
```diff
@@ -1160,6 +1192,21 @@ define(function (require, exports) {
                 ELEMENT_COLORTHEME_TYPE
             ]
         });
+        
+        descriptor.addListener("placeEvent", function () {
```
mcilroyc: I think you need to assign both these handlers to private variables, so that they can be removed upon reset.  See other handlers in on other actions as an example.

mcilroyc:
Ok @shaoshing  - I have one legit complaint about the listeners not being removed onReset, and then some minor nits.  Almost there.  back to you.

shaoshing:
Thanks for the thorough review, I addressed more commnets. Now back to you @mcilroyc 
For the getter functions, I will create separate PR to clean up the codes.

Issue #2735, opened 2015-09-30
Can't marquee select multiple layers on the canvas


tiburbage:
If I have multiple shapes defined, and drag out a marquee over them on canvas with the Select Tool, nothing happens. Based on "pre-selection edge highlighting", they appear as if they will be selected, but then no selection occurs. I can marquee select them individually. Testing with pgdev.1615 w/ DS master 8142f96 (Mac or Win)

1 - Create a document with at least 2 shapes (say non-overlapping)
2 - Switch to the Select Tool
3 - Drag out a marquee over a couple of shapes

Expect: an shapes the bounding box of which falls under the marquee should be selected

mcilroyc:
can you re-check with latest github master now @tiburbage ?

tiburbage:
@mcilroyc  Looks good on Mac and Win at 024d8ef. Should I just close and remove the "Fixed but not closed" label, or remove all the labels?

mcilroyc:
closing it is sufficient, labels can stay put

PR #2736, opened 2015-09-30
Don't pass a modifier to layers.select through marquee in normal case

volfied:
Layers.select doesn't like it when you pass it "select" as the modifier (even tho that's the default modifier) and multiple layers... Well guess who did that until now? Marquee select.

@mcilroyc please review as this is a FIXNOW bug. #2735


mcilroyc:
love it

Issue #2715, opened 2015-09-29
Close document throws error if the "next" document has not been initialized


mcilroyc:
one way to repro:
- create two documents (cmd+n, 2x)
- reset design space
- close the active document
- see exception

Assigning to myself, fix in progress.

iwehrman:
For posterity: caused by #2660.

shaoshing:
@chadrolfs can you verify?

chadrolfs:
Now if I create 3 docs, reset DS and then close them.  On the last one, console spews:
`Action documents.close failed: Error: Next selected document ID should be null iff there are no open documents`
 
![screen shot 2015-09-30 at 3 36 28 pm](https://cloud.githubusercontent.com/assets/8442631/10208563/077cd0f6-6789-11e5-8dc4-a42a18fde2ee.png)

shaoshing:
Cory just fixed this, @chadrolfs please verify again.

PR #2738, opened 2015-09-30
Bug fix in disposeDocument, better handling for final document close

mcilroyc:
Numero #2715 part dos.   Let `null` document ID pass through (if last doc was just closed), and do NOT query history.


shaoshing:
:+1:

PR #2739, opened 2015-10-01
Only add a unique key suffix to Datalists

designedbyscience:
As to allow reference by dialog id in other places


mcilroyc:
looks good, works good

PR #2747, opened 2015-10-01
Reference tool feature, dummy UI for now

DivyaPrabhakar:
This is the implementation for the reference tool feature. Currently, there is dummy UI set up to imitate the different reference points available to the user. 

The commit history is kind of messed up, if that is a problem let me know but I know all the correct changes are in this PR. Now it just says @iwehrman made them all.

### DIFF
--- a/src/img/ico-reference-none.svg
+++ b/src/img/ico-reference-none.svg
@@ -1,3 +1,28 @@
-<svg id="rotation-none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="100%" height="100%" viewBox="0 0 24 24">
+<?xml-stylesheet type="text/css" href="style.css" ?>
+<svg id="reference-none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="100%" height="100%" viewBox="0 0 24 24">
+<defs>
+	<style type="text/css"><![CDATA[
+		.reference_point rect{
+      	}
+    ]]></style>
+    <script type="text/ecmascript">
+      var a = document.getElementsByClass("reference_point"); 
+      for (i=0; i< a.length; a++) {
+        a[i].addEventListener("load",function(){
+            var id = a[i].id;
+            id.addEventListener("mousedown",function(){alert('hello world!')},false);
+        },false);
+      }
+  </script>
+</defs>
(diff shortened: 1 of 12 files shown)


src/img/ico-reference-none.svg
```diff
@@ -1,3 +1,28 @@
-<svg id="rotation-none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="100%" height="100%" viewBox="0 0 24 24">
+<?xml-stylesheet type="text/css" href="style.css" ?>
+<svg id="reference-none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="100%" height="100%" viewBox="0 0 24 24">
+<defs>
+	<style type="text/css"><![CDATA[
```
mcilroyc: should I care that this indentation is wrong?

src/js/actions/transform.js
```diff
@@ -231,7 +235,40 @@ define(function (require, exports) {
         // Used to calculate the new top/left positions of all layers in relation to top/left
         // of the group
         var overallWidthRatio = size.w ? size.w / overallBounds.width : 0,
-            overallHeightRatio = size.h ? size.h / overallBounds.height : 0;
+            overallHeightRatio = size.h ? size.h / overallBounds.height : 0,
+            overallLeftOffset = 0,
+            overallTopOffset = 0,
+            direction;
+
+        if (size.hasOwnProperty("w")) {
```
mcilroyc: The JSDoc for `size` indicates there will always be a `w` property.  Can you reconcile this by updating the JSDoc and explaining (here) the situation in which it is missing?

src/js/util/ui.js
```diff
@@ -114,8 +114,60 @@ define(function (require, exports) {
         }
     };
 
+    /**
+     * Parses the given refPoint string and returns the keys to be passed into
+     * a bounds object
+     *
+     * @param {string} refPoint Two character string denoting the active reference point
+     *                          [lmr][tcb]
```
mcilroyc: TCB

#elvis

src/js/actions/transform.js
```diff
@@ -231,7 +235,40 @@ define(function (require, exports) {
...
+            direction = refPoint.charAt(0);
+            switch (direction) {
+                case "l":
```
mcilroyc: I don't think you really need to specify this case, since there is not default case.  Unless you left it in for "self documentation" sake, then carry on

src/js/actions/transform.js
```diff
@@ -231,7 +235,40 @@ define(function (require, exports) {
         // Used to calculate the new top/left positions of all layers in relation to top/left
         // of the group
         var overallWidthRatio = size.w ? size.w / overallBounds.width : 0,
-            overallHeightRatio = size.h ? size.h / overallBounds.height : 0;
+            overallHeightRatio = size.h ? size.h / overallBounds.height : 0,
+            overallLeftOffset = 0,
```
mcilroyc: This may be on overreaction, but my gut wishes that all the character evaluation of the `refPoint.charAt()` should be stuffed away in a util file, along side its cousin `getPositionKeysByRefPoint()`.  TBH, I think this "decode weird photoshop data" stuff belongs in our adapter layer, but that's a post-oct discussion.

src/js/actions/transform.js
```diff
@@ -390,9 +431,9 @@ define(function (require, exports) {
                 positions: []
             },
             layerOneActions = _getMoveLayerActions
-                .call(this, document, layers.get(0), newPositions.get(0), payload.positions),
+                .call(this, document, layers.get(0), newPositions.get(0), "lt", payload.positions),
```
mcilroyc: Is "lt" a default value?  If so, can we just pass `undefined` here and default it down deeper?  This "hardcoded" value just seems out of place here.

src/js/stores/ui.js
```diff
@@ -146,6 +146,8 @@ define(function (require, exports, module) {
          */
         components: UI_COMPONENTS,
 
+        _referencePoint: null,
```
mcilroyc: needs JSDoc

src/js/stores/preferences.js
```diff
@@ -99,6 +99,10 @@ define(function (require, exports, module) {
             return this._preferences;
         },
 
+        get: function (key, unsetValue) {
```
mcilroyc: This isn't even a nit, it's borderline pedantry, but why not just use the existing pattern of:
```
 var preferences = this.flux.store("preferences").getState();
        if (preferences.get("showFirstLaunch", true))
...
```

mcilroyc:
@DivyaPrabhakar I'm going to assign back to you for a few of the nits I had. (There are also some non-actionable comments in there too).  Can you please update the PR description with just a little more detail on what this change introduces.   Also is there a min build version?

src/js/actions/transform.js
mcilroyc: I see that maybe you were just following a pattern that exists below... so if you don't know the answer to this, ignore it.  It may be a hard-won lesson that just wasn't documented correctly.

mcilroyc:
Also @DivyaPrabhakar I'm getting an error.  Maybe because I don't have the preference set yet?  Can you make sure it works if you nuke your preference and reload?

`Action documents.open failed: Error: Invalid reference point provided: a
    at Object.getPositionKeysByRefPoint (file:///Users/mcilroy/PlaygroundBuilds/711/Adobe%20Photoshop%20CC%202015.app/Contents/Required/Plug-Ins/Spaces/www/src/js/util/ui.js:129:19)`

src/img/ico-reference-none.svg
DivyaPrabhakar: In this case, it is correct.

src/js/actions/transform.js
DivyaPrabhakar: I am pretty sure the existing pattern is wrong and the inputs are optional so I will change it in both places.

src/js/actions/transform.js
DivyaPrabhakar: Yeah, I kept that in there for self-documentation, just to further iterate in addition to the JSDoc, the format the reference point takes.

src/js/stores/preferences.js
DivyaPrabhakar: @iwehrman can you comment on this since you wrote it? I can change it but I wanted to make sure there wasn't a specific reason for your method first.

DivyaPrabhakar:
FYI: I have made the changes @mcilroyc suggested and responded to his other comments. I created a new PR (#2764), with the edits @mcilroyc made to this PR, and it is currently being reviewed by @shaoshing.

Issue #2678, opened 2015-09-28
If layer is locked, do not enter vector mask editing mode


ktaki:
Select locked layer and press 'm' to enter vector mask editing mode.
Layer icon turns red. 
Click on rectangle tool and draw a rectangle.

![screen shot 2015-09-28 at 2 56 00 pm](https://cloud.githubusercontent.com/assets/2029597/10149881/400d44c6-65f1-11e5-9b86-7a8621cac6a9.png)

mcilroyc:
FBNC to @ktaki

ktaki:
Verified the fix.

PR #2750, opened 2015-10-01
do not enter mask mode if layer is locked

pineapplespatula:
fix for #2678 
I'm not sure if the reciprocal is also true, that if you lock a layer we should exit mask mode. maybe @placegraphichere has some thoughts. but in the mean time this is a nice little low risk fix


mcilroyc:
seems reasonable

PR #2760, opened 2015-10-02
add vector mask menu item

pineapplespatula:
Adding menu item so anyone can ever find this feature.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -2161,17 +2161,13 @@ define(function (require, exports) {
      */
     var editVectorMask = function () {
         var toolStore = this.flux.store("tool"),
-            superselectVector = toolStore.getToolByID("superselectVector");
+            currentVectorMaskMode = toolStore.getVectorMode();
 
-        return this.transfer(tools.select, superselectVector)
-            .then(function () {
-                // select and activate knots on Current Vector Mask
-                return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
-            });
+        return this.transfer(tools.changeVectorMaskMode, !currentVectorMaskMode);
     };
     editVectorMask.reads = [locks.JS_TOOL];
     editVectorMask.writes = [locks.PS_DOC];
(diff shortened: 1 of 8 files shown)


src/js/actions/layers.js
```diff
@@ -2161,17 +2161,13 @@ define(function (require, exports) {
      */
     var editVectorMask = function () {
         var toolStore = this.flux.store("tool"),
-            superselectVector = toolStore.getToolByID("superselectVector");
+            curentVectorMaskMode = toolStore.getVectorMode();
```
mcilroyc: `curent` :musical_keyboard:

src/js/models/menubar.js
```diff
@@ -235,7 +235,17 @@ define(function (require, exports, module) {
...
                 (document !== null) && hasPreviousHistoryState,
             "later-history":
-                (document !== null) && hasNextHistoryState
+                (document !== null) && hasNextHistoryState,
+            "one-layer-selected-can-have-vector-mask":
+                (document !== null) &&
+                !document.unsupported &&
+                (document.layers !== null) &&
+                (document.layers.selectedNormalized.size === 1) &&
+                (document.layers.selected.every(function (layer) {
```
mcilroyc: consider moving this logic into a layerstructure property at some point in the future?

mcilroyc:
this doesn't work for me?  Anyway, back to you to fix the misspelling at least.

iwehrman:
I'm not actually sure if we want to add this to the menu item. @placegraphichere and I are thinking that we want a toolbar icon for this (see #2766), and it's confusing to have it in the menu given that it isn't actually a menu command.

pineapplespatula:
I marked this as UI and review until @placegraphichere chimes in

iwehrman:
I'm closing this out because the only reason I think we otherwise would like this menu item is to make it available in search. But, I think the better thing to do is to just add all (most?) global shortcuts to the search. This will get us things like mask mode as well tool activation, layers panel navigation, etc.

Issue #2729, opened 2015-09-30
Limit maximum number of effect instances to be 10 per effect


ktaki:
Otherwise, Ps may crash. The menu item in '+' flyout should be dimmed when the max number has been reached.

chadrolfs:
Core needs to know.  Please file a watson bug saying don't let people do with an action descriptor and assign to Jesper or JP.

shaoshing:
created watson bug and assigned to Taki https://watsonexp.corp.adobe.com/#bug=4068367

mcilroyc:
prevented on the DS side by @shaoshing 's PR

shaoshing:
Taki, this layer effect might save your from clicking your mouse :P http://adobe.ly/1jDyVku

ktaki:
Thanks @shaoshing. Looks good.

PR #2761, opened 2015-10-02
Restricts the maximum number of layer effects user can create

shaoshing:
This PR restricts the maximum number (10) of layer effects user can create. If reach maximum number, the coresponding menu item will be greyed out. for #2729

For you convenience, you can import this layer effect (http://adobe.ly/1jDyVku) to your cc library to quickly populate these effects.

### DIFF
--- a/src/js/jsx/sections/style/EffectsPanel.jsx
+++ b/src/js/jsx/sections/style/EffectsPanel.jsx
@@ -32,7 +32,10 @@ define(function (require, exports, module) {
         classnames = require("classnames");
 
     var os = require("adapter/os"),
-        LayerEffect = require("js/models/effects/layereffect");
+        LayerEffect = require("js/models/effects/layereffect"),
+        strings = require("i18n!nls/strings"),
+        collection = require("js/util/collection"),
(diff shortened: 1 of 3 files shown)


src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -176,7 +177,21 @@ define(function (require, exports, module) {
...
+         *
+         * @private
+         * @param {string} effectName - accept: strokeEffects, colorOverlays, innerShadows, dropShadows
+         * @return {boolean}
+         */
+        _canAddEffect: function (effectName) {
+            var effects = collection.pluck(this.props.document.layers.selected, effectName);
+            
+            // Return false if any of the selected layers reach maximum effect count.
+            return collection.zip(effects).size < MAX_EFFECT_COUNT;
```
mcilroyc: performance nit: `zip` is kind of a heavy function just for getting the max size.  It would be one thing if we were actually caching the value for use elsewhere, but here it is just wasted effort.  It may not really be significant in this case though.

mcilroyc:
LGTM

Issue #2743, opened 2015-10-01
CC Lib: dragging in empty or missing asset doesn't place and leaves app in bad state


chadrolfs:
create a new doc
load library that contains an empty or missing asset (e.g. Test - OK to accidentally nuke)
drag missing or empty graphic onto canvas (e.g. I am 3D)

Icon indicates it's empty or missing:  <img width="51" alt="screen shot 2015-09-30 at 5 13 35 pm" src="https://cloud.githubusercontent.com/assets/8442631/10227136/818071f4-6820-11e5-82be-7bc95fdc3078.png">

Result: it doesn't place and you are stuck in whatever tool you started with

shaoshing:
Fixes are:

* disabled drag-n-drop for empty graphic asset
* hovering over empty asset will show "Graphic has no content"

chadrolfs:
after you attempt to drag and drop an empty asset, it fails as expected but subsequent attempts to drag valid assets also fail.

shaoshing:
Hmm, I am unable to reproduce this; drang-n-drop on empty and valid asset are both working. Do you mind record a video for me?

chadrolfs:
have you tried on Windows?  I can't seem to repro any more on Mac, but it looks like there are some serious performance issues on Windows.

chadrolfs:
Blocked by #2805

chadrolfs:
Able to test current master a16a13d1e39f01cd6f4e302ce729ccfb9ca6db1d with FC staging build (1600) to verify this fix

Issue #2547, opened 2015-09-21
Unable to open/edit some CC Lib assets


chadrolfs:
Load shared library (Test - OK to accidentally nuke) 
drag long file name asset (ac69f7a1d57a28...) onto canvas and double-click to edit

Result: it won't open 

AI and SVG don't open either (assuming as designed because that requires a different application.  classic results in a dialog that is probably suppressed in DS).

shaoshing:
@chadrolfs Currently we only support editing PS files in DS. In classic, you can edit images (png, jpeg, with limitation) and PS files; editing unsupported files will show a modal dialog saying "file cannot directly edited in PS". 

Adding @davealonzo to this conversation for advise on handling unsupported files.

davealonzo:
I would suggest to ignore double-click to edit for now until we can get better support for messaging in the UI and for editing such file types.

Also, I asked Bradee if Ps plans on supporting editing SVGs from libraries. She said they will be in November. I am inquiring more about what they're up to.

chadrolfs:
@davealonzo are you saying we should ignore 2x-click on canvas or in layers panel or in CC Lib panel or all of the above (I assume this is for Library assets only, correct)?  What about the Enter key?  Are you saying we should not allow Library asset editing at all?  

Disabling this functionality probably affects editing objects/layers in general (shape, text, pixel, embedded smart objects, linked smart objects).   

Need clarity on how to proceed on each path.

davealonzo:
OK... did some heavy soul searching on this one with the help of Chad, Shaoshing and Ian. With continuing to support editable library assets in mind (e.g. psd, psb), here's the proposal:

**Hover title modifications for Libraries panel**
* For editable asset types, add "Double click graphic to edit" to second line
* For non-editable asset types, add "Not editable in this app" to second line

**Double click interaction to edit**
* Implement double click interaction to thumbnail graphic (if editable)
* Implement double click interaction to placed canvas asset (if editable)
* Continue to support double click interaction in Layers panel (if editable) 
* Ignore double click interactions for all non editable instances (informed by the hover title)

Yes?

chadrolfs:
Yes!  Sounds great.

PR #2762, opened 2015-10-02
CC Libraries - issues and improvement

shaoshing:
Added the following bug fixes and improvement:

* Better handling on empty graphic asset: disabled drag-n-drop for empty graphic asset, hovering over empty asset will show "Graphic has no content". Below is what the thumbnail looks like for empty graphic asset - #2743 

screen shot 2015-10-02 at 10 49 12 am

* Mouse over the graphic thumbnails will tell you whether the graphic is editable in DS. - #2547

screen shot 2015-10-02 at 10 46 00 am

* Ease log count when syncing large library collection. We differentiate sync event and download event and supress those related to download.

* Refactor: added libraryActions.getState and removed unnecessary getter functions.


src/js/actions/libraries.js
```diff
@@ -821,7 +827,13 @@ define(function (require, exports) {
                         // So, although the command does not resolve until the placement has finished,
                         // it does resolve before the layer model has been updated because the event
                         // has not yet been received. For more details see issue #2177.
-                        return descriptor.playObject(placeObj, { synchronous: false });
+                        return descriptor.playObject(placeObj, {
+                            synchronous: false,
+                            historyStateInfo: {
```
mcilroyc: I'm not convinced that this is doing what you are expecting.  When I drag a graphic asset onto canvas, I see two history states.  Same behavior on master as this PR branch.
~[ss](http://cl.ly/image/2K1G2z333f0k)

I think there is a separate issue here related to that first history state for the "wrong" document, btw.

Am I missing something?  Here is a vid: http://cl.ly/1o0O3g0g3Z0E

mcilroyc:
This all looks good on first pass.  I have some history questions.

shaoshing:
@mcilroyc thanks for reviewing and the vid. I confirmed that undo "drop asset" works without the history state option, so I reverted the changes. Not sure why it was not working a few days back; might related to the keyboard policy I worked on. For the first history state (named "open" with different document ID), I suspect it has to do with the place event; I observed same history state for placing linked / embedded SO.

shaoshing:
BTW: the "place linked / embedded SO" is not working in DS now. I will file an issue.

mcilroyc:
nice clean up overall.

PR #2764, opened 2015-10-02
Reference tool feature, with correct UI

DivyaPrabhakar:
This PR includes the UI and logic for the reference tool feature. 

A couple things to note:

1. In TransformPanel.jsx, we needed to embed the SVG directly within the .jsx file otherwise we would need 9 different icons to simulate the 9 different reference points. With this approach, though it is lengthy, it allows us to have one icon that can support all 9 options. 

2. Currently, the reference point will not resize correctly if the "Lock Width and Height Proportional" is set. I found a bug and I am working on this currently but I don't know if this is considered Feature Complete + bug or Feature Complete blocker. @chadrolfs, comments on this matter?


src/js/actions/transform.js
```diff
@@ -231,7 +235,42 @@ define(function (require, exports) {
...
+                case "c":
+                    overallTopOffset = (overallBounds.height - size.h) / 2;
+                    break;
+                case "b":
+                    overallTopOffset = (overallBounds.height - size.h);
+                    break;
+            }
+        }
+
+        // var resizeProportionalNewSize = _calculateNewSize(overallBounds, size, targetLayer.proportionalScaling);
```
iwehrman: Stale comment.

src/js/jsx/sections/transform/Position.jsx
```diff
@@ -52,35 +53,37 @@ define(function (require, exports, module) {
                 return collection.pluckAll(layers, ["kind", "locked", "isBackground"]);
             };
 
-            return !Immutable.is(getSelectedChildBounds(this.props), getSelectedChildBounds(nextProps)) ||
+            return this.props.referencePoint !== nextProps.referencePoint ||
```
iwehrman: Can you define `getDefaultProps` and `propTypes` (grep the source for examples) to both provide a default reference point and document its type (a string, required)?

src/js/jsx/sections/transform/Position.jsx
```diff
@@ -127,9 +131,11 @@ define(function (require, exports, module) {
                 layers = document.layers.selected,
                 bounds = document.layers.selectedRelativeChildBounds;
 
-            var disabled = this._disabled(document, layers),
-                tops = collection.pluck(bounds, "top"),
-                lefts = collection.pluck(bounds, "left");
+            var disabled = this._disabled(document, layers);
```
iwehrman: For stylistic consistency, all these vars should be collapsed into one declaration.

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -25,10 +25,13 @@ define(function (require, exports, module) {
...
 
     var React = require("react"),
+        Fluxxor = require("fluxxor"),
+        FluxMixin = Fluxxor.FluxMixin(React),
+        StoreWatchMixin = Fluxxor.StoreWatchMixin,
         classnames = require("classnames");
 
-    var SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
-        AlignDistribute = require("jsx!./AlignDistribute"),
+    // var Button = require("jsx!js/jsx/shared/Button"),
```
iwehrman: Stale comment.

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -43,6 +57,16 @@ define(function (require, exports, module) {
             
             return true;
         },
+
+        /**
+         * Set the reference point on click.
+         *
+         * @private
+         * @param {string} referencePoint One of "a", "b", "c"
```
iwehrman: Stale comment.

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -66,15 +98,67 @@ define(function (require, exports, module) {
                             <div className="control-group">
                                 <Size document={this.props.document} />
                             </div>
-                            <div className="control-group">
-                                <div className="control-group reference-mark">
-                                    <SVGIcon CSSID="reference-cm" />
-                                </div>
+                            <div title="Reference point">
```
iwehrman: This should be a localized string.

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -66,15 +98,67 @@ define(function (require, exports, module) {
...
+                                    L7.000,14.000 L5.000,14.000 L5.000,16.000 L7.000,16.000 L7.000,18.000
+                                    L9.000,18.000 L9.000,16.000 L14.000,16.000 L14.000,18.000 L16.000,18.000
+                                    L16.000,16.000 L18.000,16.000 L18.000,14.000 L16.000,14.000 L16.000,9.000
+                                    L18.000,9.000 L18.000,7.000 L16.000,7.000 L16.000,5.000 ZM20.000,6.000
+                                    L20.000,3.000 L17.000,3.000 L17.000,6.000 L20.000,6.000 ZM20.000,13.000
+                                    L20.000,10.000 L17.000,10.000 L17.000,13.000 L20.000,13.000 ZM17.000,17.000
+                                    L17.000,20.000 L20.000,20.000 L20.000,17.000 L17.000,17.000 ZM9.000,9.000
+                                    L14.000,9.000 L14.000,14.000 L9.000,14.000 L9.000,9.000 ZM10.000,13.000
+                                    L13.000,13.000 L13.000,10.000 L10.000,10.000 L10.000,13.000 Z" />
+                                    <rect className={getReferenceClasses("rb")} id="rb"
```
iwehrman: I think it would be better to format these `<rect>` elements the way we format the rest of our JSX elements with one attribute on each line. (That would probably be going overboard for the `<path>` above, though.)

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -66,15 +98,67 @@ define(function (require, exports, module) {
...
                                 <Size document={this.props.document} />
                             </div>
-                            <div className="control-group">
-                                <div className="control-group reference-mark">
-                                    <SVGIcon CSSID="reference-cm" />
-                                </div>
+                            <div title="Reference point">
+                                <svg id="reference-none" preserveAspectRatio="xMidYMid" width="100%"
+                                    height="100%" viewBox="0 0 24 24">
+                                <path d="M21.000,16.000 L21.000,21.000 L16.000,21.000 L16.000,19.000
```
iwehrman: I think this should be indented.

src/js/stores/preferences.js
```diff
@@ -99,6 +99,10 @@ define(function (require, exports, module) {
             return this._preferences;
         },
 
+        get: function (key, unsetValue) {
```
iwehrman: JSDoc needed. Since I wrote this, here it is:
```
* Directly get the value for the given preference key, or an unsetValue if the key does not exist.
* 
* @param {string} key
* @param {?string} unsetValue
* @return {?string} 
```

src/js/stores/ui.js
```diff
@@ -146,6 +146,15 @@ define(function (require, exports, module) {
          */
         components: UI_COMPONENTS,
 
+        /**
+         * Reference point for a layer when adjusting W and H
+         * Default is "lt" for left, top
```
iwehrman: Elsewhere in the code you're using `"tl"`...

src/js/stores/ui.js
```diff
@@ -182,6 +192,7 @@ define(function (require, exports, module) {
             this._zoom = null;
             this._transformMatrix = null;
             this._inverseTransformMatrix = null;
+            this._referencePoint = "lt";
```
iwehrman: Ditto.

src/js/util/ui.js
```diff
@@ -83,7 +83,7 @@ define(function (require, exports) {
      * @param {number} y Vertical coordinate
      * @param {number=} opacity Opacity of the owner layer at the pixel
      * @return {Promise.<Color>}
-     */
+ */
```
iwehrman: Broken whitespace.

src/js/util/ui.js
```diff
@@ -114,8 +114,60 @@ define(function (require, exports) {
         }
     };
 
+    /**
+     * Parses the given refPoint string and returns the keys to be passed into
+     * a bounds object
+     *
+     * @param {string} refPoint Two character string denoting the active reference point
+     *                          [lmr][tcb]
+     *
```
iwehrman: No newline is necessary here. Also, you don't need this much hanging indentation.

src/js/util/ui.js
```diff
@@ -114,8 +114,60 @@ define(function (require, exports) {
...
+     * @return {{x: string, y: string}} Keys referring to the value in Bounds object to get the
+     * desired reference point
```
iwehrman: ... but more than this :)

src/style/panel.less
```diff
@@ -455,3 +455,9 @@
         min-height: 3rem;
     }
 }
+
+#reference-none {
```
iwehrman: I could be wrong, but _generally_ we do not use CSS ids for styling. Perhaps @designedbyscience and @placegraphichere have carved out some exceptions for SVG. But, I would have expected a class selector instead.

src/style/sections/transform/align-distribute.less
```diff
@@ -35,3 +35,11 @@
     width: 2rem;
     height: 2rem;
 }
+
+.reference-point {
+    opacity: 0;
+}
+
+.reference-point__active {
+  opacity: 1.0;
```
iwehrman: Incorrect indentation. Four-space tabs everywhere.

src/js/actions/transform.js
```diff
@@ -231,7 +235,42 @@ define(function (require, exports) {
         // Used to calculate the new top/left positions of all layers in relation to top/left
         // of the group
         var overallWidthRatio = size.w ? size.w / overallBounds.width : 0,
-            overallHeightRatio = size.h ? size.h / overallBounds.height : 0;
+            overallHeightRatio = size.h ? size.h / overallBounds.height : 0,
+            overallLeftOffset = 0,
+            overallTopOffset = 0,
+            direction;
+
+        // This is the case if the width input was changed
```
iwehrman: Here's the most important request: please write a paragraph-length explanation of how this works using a `// comment`. The algorithm for this was complicated enough that we all screwed it up at least a few times, but you've now cracked the code. Explain this to your future self! (And also the reviewers.)

iwehrman:
The icon is now in the wrong place. it seems to have moved to the right, but I think it should be left-aligned with the rotation icon:
image

iwehrman:
This works well _except_ in the case of proportional scaling. If I use the top-right reference point with a 400x400 group and scale the height proportionally, I think I expect a 200x200 group, but that's not what happens:
http://cloud.wehrman.org/1w1t1X0x0b22

iwehrman:
Great work so far! You're getting close. Back to you, @DivyaPrabhakar.

src/js/actions/transform.js
```diff
@@ -209,12 +212,13 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer} targetLayer
-     * @param {{w: number, h: number}} size
+     * @param {{w=: number, h=: number}} size
```
shaoshing: Just a tip that you can use properties param in the future to better describe object param:

```
@param {object} size
@param {number=} size.w
@param {number=} size.h
```

src/js/actions/transform.js
```diff
@@ -231,7 +235,42 @@ define(function (require, exports) {
...
+            overallLeftOffset = 0,
+            overallTopOffset = 0,
+            direction;
+
+        // This is the case if the width input was changed
+        if (size.hasOwnProperty("w")) {
+            direction = refPoint.charAt(0);
+            switch (direction) {
+                case "l":
+                    break;
```
shaoshing: Is this intended for showing the complete cases? can you leave a comment to explain the reason?

src/js/actions/transform.js
```diff
@@ -231,7 +235,42 @@ define(function (require, exports) {
...
+            // if the height input was changed
+            direction = refPoint.charAt(1);
+            switch (direction) {
+                case "t":
+                    break;
+                case "c":
+                    overallTopOffset = (overallBounds.height - size.h) / 2;
+                    break;
+                case "b":
+                    overallTopOffset = (overallBounds.height - size.h);
```
shaoshing: nit: You are not using brackets for `case "r"`, so maybe remove brackets here to keep consistency?

src/js/actions/transform.js
shaoshing: nit: can you remove these comments?

src/js/actions/transform.js
```diff
@@ -390,9 +433,9 @@ define(function (require, exports) {
                 positions: []
             },
             layerOneActions = _getMoveLayerActions
-                .call(this, document, layers.get(0), newPositions.get(0), payload.positions),
+                .call(this, document, layers.get(0), newPositions.get(0), undefined, payload.positions),
```
shaoshing: `this` is not in use in the helper function, so using `call` is not necessary? Please search for `_getMoveLayerActions`

src/js/actions/transform.js
```diff
@@ -495,7 +539,7 @@ define(function (require, exports) {
             sizePromise = descriptor.playObject(resizeObj);
         } else {
             var resizeLayerActions = layerSpec.reduce(function (actions, layer) {
-                var layerActions = _getResizeLayerActions.call(this, document, layer, size, payload.sizes);
+                var layerActions = _getResizeLayerActions.call(this, document, layer, size, payload.sizes, refPoint);
```
shaoshing: same problem as `_getMoveLayerActions ` described above.

src/js/actions/ui.js
```diff
@@ -509,6 +518,27 @@ define(function (require, exports) {
     zoomInOut.transfers = [zoom];
 
     /**
+     * Set the global resize reference point.
+     *
+     * @param {string} referencePoint One of "tl", "tm", "tr", "cl", "cm", "cr", "bl", "bm", "br".
```
shaoshing: Instead of copy-paste the possible value of referencePoint across the project, you can make use of `@typedef`, for example:

```
@typedef {string} ReferencePointString - One of "tl", "tm", "tr", "cl", "cm", "cr", "bl", "bm", "br".
```

and then replace all param docs of referencePoint to 
```
@param {ReferencePointString} referencePoint
```

src/js/actions/ui.js
```diff
@@ -570,12 +600,18 @@ define(function (require, exports) {
                 return this.flux.actions.ui.zoomInOut({ "zoomIn": true, "preserveFocus": true });
             }.bind(this));
 
-        return Promise.join(osPromise, owlPromise, pathPromise, zoomInShortcutPromise)
+        // Initialize the reference point from preferences
+        var preferences = this.flux.store("preferences"),
+            referencePoint = preferences.get(REFERENCE_POINT_PREFS_KEY, "lt"),
```
shaoshing: mystic string. maybe consider using a const instead. for example `DEFAULT_REFERENCE_POINT`

src/js/stores/preferences.js
```diff
@@ -99,6 +99,10 @@ define(function (require, exports, module) {
...
+            return this._preferences.get(key, unsetValue);
+        },
+
```
shaoshing: syntactic sugar alert: this function is a wrapper that is almost doing nothing, but the cost is that we could confused this with other data types like Immutable.Map. Besides, this is not consistent with other places that are using `getState`.

src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -66,15 +98,67 @@ define(function (require, exports, module) {
...
+                                        onClick={this._handleReferenceClick.bind(this, "lc")}/>
+                                    <rect className={getReferenceClasses("rt")} id="rt"
+                                        x="16" y="2" width="5" height="5"
+                                        onClick={this._handleReferenceClick.bind(this, "rt")}/>
+                                    <rect className={getReferenceClasses("mt")} id="mt"
+                                        x="9" y="2" width="5" height="5"
+                                        onClick={this._handleReferenceClick.bind(this, "mt")}/>
+                                    <rect className={getReferenceClasses("lt")} id="lt"
+                                        x="2" y="2" width="5" height="5"
+                                        onClick={this._handleReferenceClick.bind(this, "lt")}/>
```
shaoshing: Not sure if I like the implementation by pasting the entire SVG, but we should encapsulate this as ReferencePoint component, just like Position, Size components.

src/js/util/ui.js
shaoshing: nil :P

src/js/actions/ui.js
shaoshing: Also use the more comment way

src/js/jsx/sections/transform/TransformPanel.jsx
shaoshing: nit: please remove the commented code.

src/js/jsx/sections/transform/TransformPanel.jsx
shaoshing: this is the copy-paste problem I talked above :)

shaoshing:
Ha, Ian and I both reviewd your PR. I think there might be some duplicates so feel free to ignore if there are. As Ian said, this is an excellent job!

src/js/actions/transform.js
shaoshing: maybe you have to set the `referencePoint` for `_getMoveLayerActions` here; I tried the swap position and it raise an error.

src/js/jsx/sections/transform/TransformPanel.jsx
DivyaPrabhakar: Grunt complains when I keep the <rect> elements in one line because they are too long. Obvioulsy it is the same for the Path. :/

src/js/stores/ui.js
DivyaPrabhakar: Thanks for this catch. For the record, "lt" is correct and I changed all the "tl"'s.

chadrolfs:
double peer review is super great, thanks @iwehrman and @shaoshing

src/js/stores/preferences.js
iwehrman: I'm the guilty party here. I thought it was really useful though, because otherwise the only state we get from the preferences store is just the map, which we only use to call `get` on. It seemed like a fair way to save a step... But, if you don't like it, @DivyaPrabhakar can rip it out!

src/js/actions/transform.js
iwehrman: Drive-by comment: I think both formats are fine. I prefer the more compact, direct representation for _small_ records, but the expanded version with many `@param` instances for bigger ones. But, either is cool with me!

DivyaPrabhakar: I think I will separate them for this case to further iterate that both do not have to be set, and in a lot of cases they in fact do not occur simultaneously. I know the "=" means the same but further clarification in transform.js doesn't hurt anyone in my opinion.

src/js/stores/preferences.js
DivyaPrabhakar: Imma keep it

src/js/jsx/sections/transform/Position.jsx
DivyaPrabhakar: Done, just double check I did it right?

DivyaPrabhakar:
@iwehrman @shaoshing - I made the changes to the PR that you guys gave me. Two things: 

1.) Regarding @shaoshing's _getMoveLayerActions, I don't think it should be hardcoded (as mentioned in #2747) and undefined breaks the swap tool like @shaoshing mentioned. I will fix this to not be hardcoded but I think it requires more time and care to make sure all the instances of swap are working after the change. I propose after this PR  gets merged, I will do a code cleanup for this specifically, as a bug.

2.) Regarding the raw SVG placed in TransformPanel.jsx, I agree it is ugly and getting it into its own component would be great. However, I think this is a code cleanup again, which I can be responsible for after this PR gets merged. 

What do you guys think of these propositions?

chadrolfs:
I've run into 2 issues intermittently in my brief testing (note: not easy to repro)

1. After setting ref point and changing W/H, performing an on canvas transform may result in
`Action transform.setSize failed: Error: Host is in a modal state and the request cannot be executed.`

2. I have gotten into a state where trying to change W/H values, X and Y will change instead
https://www.dropbox.com/s/gwfmydw2q7feez8/XY-RefPointIssue.mov?dl=0

chadrolfs:
screen shot 2015-10-05 at 3 47 33 pm

chadrolfs:
One more, 3. It's sometimes takes several clicks to set the desired ref point

```Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 0ms; 1/0
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 6ms with RTT 6ms; 1/0
fluxcontroller.js:533 Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 0ms; 1/0
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 5ms with RTT 5ms; 1/0
fluxcontroller.js:533 Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 0ms; 1/0
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 6ms with RTT 6ms; 1/0
fluxcontroller.js:533 Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 0ms; 1/0
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 11ms with RTT 11ms; 1/0
fluxcontroller.js:533 Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 0ms; 1/0
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 6ms with RTT 6ms; 1/0
fluxcontroller.js:533 Enqueuing action superselect.marqueeSelect; 0/0
fluxcontroller.js:533 Enqueuing action ui.setReferencePoint; 1/0
fluxcontroller.js:554 Executing action superselect.marqueeSelect after waiting 1ms; 1/1
fluxcontroller.js:564 Finished action superselect.marqueeSelect in 5ms with RTT 6ms; 1/1
fluxcontroller.js:554 Executing action ui.setReferencePoint after waiting 5ms; 1/0
fluxcontroller.js:564 Finished action ui.setReferencePoint in 13ms with RTT 18ms; 1/0```

chadrolfs:
@iwehrman @shaoshing either of you run into any of those issues?

chadrolfs:
2nd issue happens on master so not specific to this branch

shaoshing:
@chadrolfs , regarding your issues:

1. I cannot reproduce this one
2. I can reproduce in master as well, so this should be an existing bug. Here's how to reproduce https://www.dropbox.com/s/1fe0jpo59xu0idj/rp-2.mov?dl=0
3. confirmed. cc @DivyaPrabhakar . reproduce: https://www.dropbox.com/s/ph6su27rcmfotm4/rp-3.mov?dl=0

shaoshing:
@DivyaPrabhakar can you also rebase master?

DivyaPrabhakar:
@shaoshing @chadrolfs I know the reason for 3 and I can confirm it is because the svg is very small. I can fix this when @placegraphichere gets back or I can make the svg bigger and therefore the click targets will also become bigger.

shaoshing:
@DivyaPrabhakar your suggestion sounds good. since we have a deadline for this feature and the current issues / nits are not deal breakers, I am gonna merge it and we can do the improvement and code cleanup in separate PR. Good work!

Issue #2122, opened 2015-08-20
Need missing font and layout change warning indicators in layers panel


ktaki:
In standard Ps, yellow and white triangles like:

<img width="241" alt="screen shot 2015-08-19 at 5 17 32 pm" src="https://cloud.githubusercontent.com/assets/2029597/9372696/71acaf28-4696-11e5-854e-688282a49cfb.png">

No indicators in DS, but it will result in the following modal dialogs when trying to edit the layer.

<img width="795" alt="screen shot 2015-08-19 at 5 15 37 pm" src="https://cloud.githubusercontent.com/assets/2029597/9372738/e67c5c22-4696-11e5-9ec2-91e434790988.png">

<img width="806" alt="screen shot 2015-08-19 at 5 15 19 pm" src="https://cloud.githubusercontent.com/assets/2029597/9372739/ec07b880-4696-11e5-8217-f0964c22f169.png">

iwehrman:
Can you confirm my intuition about what the yellow and white warning signs mean?

ktaki:
Yellow triangle indicates the layer has missing font(s).
White triangle indicates it may cause layout change (or reflow of text.) This is a warning that it might be different version of fonts were used when the layer was originally created.

placegraphichere:
Discussed this with @davealonzo and I think we'd like to just maintain the popups for now. They're a necessary evil at the moment. I've got the icon we need for this in #2765. Just a single icon for now. Not two for both cases. Since the difference in the color/no-color icon is probably not a understandable difference for most people, I certainly would not be able to tell you the "alert" difference looking at them.

chadrolfs:
I'm not seeing the missing font icon show up in the layers panel when opening a document with missing fonts.  Are the icons hooked up correctly?

designedbyscience:
They aren't hook up in that PR, just added to our asset set

pineapplespatula:
We only seem to have an icon for missing(yellow triangle in classic) and not for reflow( white triangle in classic) 

I will have both warnings use the missing icon for now.

chadrolfs:
works for me, thanks!

iwehrman:
This is essentially FBNC after #2845, but I'm re-assigning it to @placegraphichere so that we can get the extra warning icon. @ktaki please feel free to test with the caveat that both conditions show the same icon at the moment.

iwehrman:
Also demoting this to medium priority.

ktaki:
[...]

PR #2765, opened 2015-10-02
Missing Font Icons

placegraphichere:
- does not fix, but for use with issue #2122 
- text layer
- text: font missing
- text: font mask
- text: font missing mask


iwehrman:
I got this one you guys!

Issue #1519, opened 2015-06-12
Min and max values for type properties


DesignSpaceBot:
Here are the min and max values for type properties. Should they match?

Ps standard:
Font size (Size):    0.01 pt / 1296 pt
Leading (Line) :     0.01 pt / 5000 pt
Tracking (Letter):   -1000 / 10000

DS:
Font size (Size):    0 pt / 5400 pt
Leading (Line) :    No limits, any number can be entered.
Tracking (Letter):   No limits, any number can be entered.

-- @ktaki at [2015-06-04T00:01:48Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1519)

mcilroyc:
FBNC to @ktaki

ktaki:
Font size (Size): 0.04 pt / 1296 pt  ----> The min value 0.04 is probably OK. (Standard Ps has 0.01)
Leading (Line) : 0.01 pt / 5000 pt  ---> These match with standard Ps
Tracking (Letter): -1000 / 1000  ---> Max value is too small. Standard Ps has 10,000 as max.

iwehrman:
Argh, I misread your bug description above. I pushed commit 63daa6c38399c4f1240912383d1e67472b48024c directly to master to fix the max tracking value. The min font size value is a standard PS rounding error. I think we should just let that one go... 

FBNC to @ktaki again.

ktaki:
Verified the fix. Need to make the field a bit wider. Opened an issue #2821.

PR #2769, opened 2015-10-02
Enforce minimum and maximum font, leading and tracking sizes

iwehrman:
What it says! See #1519 for details on the enforced sizes.


src/js/jsx/sections/style/Type.jsx
```diff
@@ -536,6 +551,14 @@ define(function (require, exports, module) {
                 }),
                 fillBlendFormline;
 
+            // Convert the given point value to pixels
+            var toPixels = function (value) {
```
mcilroyc: bleh, this doesn't seem like it belongs in a JSX file

mcilroyc:
ok.

PR #2772, opened 2015-10-02
Color Stop Finesse

placegraphichere:
- getting ready for more colors stops
- clean up for layers panel and range slider background color
- more libraries tweaks for hover/selection
- color stops ONLY need to be changed in `colors.less`


iwehrman:
:no_good:

PR #2774, opened 2015-10-02
color variable for library sample backgrounds

placegraphichere:
one more of these color changes before we get to themes


iwehrman:
:open_hands:

Issue #2781, opened 2015-10-03
Opening the search bar is slow


iwehrman:
There are two reasons for this:

1. Documents that are inactive and uninitialized must be initialized before the search bar dialog can be opened.
2. Formatting layer names is a major bottleneck when there are many layers in the currently open documents. The culprit is the `O(n^2)` algorithm used by `getShortestUniquePaths` in `util/path`.

Issue #2759, opened 2015-10-02
Search: returning to app after defocus result in unresponsive field


chadrolfs:
with or without doc open
cmd/ctrl + f
arrow down/up to expose datalist
cmd/ctrl + tab to defocus app (repeat to refocus)
attempt to click/type in field

Result: unresponsive
workaround: click outside search to dismiss and re-enter.

Issue #2057, opened 2015-08-14
Search bar loses focus when you switch to another app


hamburge:
It should either not lose focus or immediately regain it instead.

ktaki:
Verified the fix.

PR #2782, opened 2015-10-03
Misc search improvements

iwehrman:
1. Display a read-only "initializing.." message while initializing background documents instead of showing nothing for a long period of time. This improves the user experience. Partially addresses #2781.
2. Do a better job preserving search focus when tabbing away and back. The search-in-progress is no longer committed. Focus is correctly preserved when tabbing away. For some reason it still loses focus  though when you click on another application window!? Partially addresses #2759 and #2057.
3. Speed up formatting of layer name results by not reducing the layer ancestry to use the "shortest unique paths". I think this makes more sense from a UX perspective, and it's also happens to be terribly slow. Also partially addresses #2781.
4. Rename "Current Documents" to "Open Documents".
5. When formatting layer search results, do not include the layer name itself in the secondary ancestry string.
6. Bonus: fixed a couple stray instances of "..." (three periods) which should be "…" (ellipses).

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -28,12 +28,14 @@ define(function (require, exports, module) {
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
-        _ = require("lodash"),
         Immutable = require("immutable"),
-        collection = require("js/util/collection"),
-        strings = require("i18n!nls/strings");
+        _ = require("lodash");
 
-    var Datalist = require("jsx!js/jsx/shared/Datalist"),
+    var os = require("adapter/os");
+
+    var collection = require("js/util/collection"),
+        strings = require("i18n!nls/strings"),
+        Datalist = require("jsx!js/jsx/shared/Datalist"),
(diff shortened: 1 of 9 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -102,12 +108,60 @@ define(function (require, exports, module) {
...
+            os.removeListener("activationChanged", this._handleActivationChanged);
+        },
+
         componentDidUpdate: function (prevProps, prevState) {
+            if (!prevState.ready && this.state.ready) {
+                // In case some documents were not initialized, but now are
+                this._initSearchProviders();
+            } else if (prevState.ready && !this.state.ready) {
+                // In case there are newly uninitiazlied documents
+                this.getFlux().actions.documents.initializeDocumentsThrottled();
```
pineapplespatula: we do not seem to be doing anything with the promise returned from here. should we be waiting for it to resolve?

iwehrman: We can't block anything in React using this promise, unfortunately. Worst case scenario, you could end up calling `initializeDocumentsThrottled` a second time before the first one is finished. But, that's why I'm using the throttled version of the action; the second call should be a no-op.

pineapplespatula:
LGTM

Issue #2731, opened 2015-09-30
Guide drop zones are incorrect with Single Column Mode


chadrolfs:

chadrolfs:
<img width="1440" alt="screen shot 2015-09-30 at 2 16 28 pm" src="https://cloud.githubusercontent.com/assets/8442631/10206859/022020d2-677e-11e5-9cc5-97919f909918.png">

chadrolfs:
re-assign to @volfied if needed. You have less than half a day.  :stuck_out_tongue_winking_eye:

designedbyscience:
This is a cloaking issue, I believe

iwehrman:
I re-did the overlay offsets stuff last, so I'll take this one.

PR #2784, opened 2015-10-04
Reset UI borders when toggling single-column mode

iwehrman:
This is a very narrow fix for #2731, in which toggling single-column mode fails to update the UI borders (i.e., the overlay offsets and cloaking rectangle).

After the next release, the border/panel/column management really needs some cleanup and additional structure... (CC @shaoshing, who may care to think about this during his quiet moments.)


pineapplespatula:
LGTM

Issue #2785, opened 2015-10-04
Border policies not updated reliably


iwehrman:
Two bugs, easily spotted with `Help > Show Policy Frames`:

1. When dragging a text layer on canvas, the border polices are not moved along with the layer.
2. When closing documents in rapid succession, border policies frequently remain once all the documents are closed, never to be changed again.

These sorts of issues are invisible to the user until she or he tries to target some on-canvas UI that happens to be coincident on one of these ghostly border policies.

iwehrman:
Hypothesis: this is the root cause of #2224. If you get in this state, you won't be able to start creating a rectangle from within one of these ghostly border polices.

iwehrman:
Fixed by #2790 and #2803.

Issue #2224, opened 2015-08-26
Create Rect doesn't always work


zoranagee:
Open up the explore mt. hood - edit file (on my transfer in sjshare)
select the BG layer in the Top Nav folder
create a rectangle

RESULT: nothing happens; if I randomly continue to try to draw out it occasionally will draw out a rect

video for this on my transfer as well ("CreateRect.mp4")

chadrolfs:
I CNR with current resources so maybe this is fixed. (using dev 523 w/master @ ffa24f9)

iwehrman:
I've seen this bug. I think it might be related to multiple monitors. I also had trouble reproducing it reliably, but at one point I couldn't make rectangles in the bottom or right side of the canvas only.

chadrolfs:
@tiburbage can you try to repro this.

ktaki:
eMailed Tom B if he could repro this.

tiburbage:
I tried without success to have the tool fail to create a Rectangle shape using the steps and content Zorana cited. I did notice that the first Rectangle I drew just above the BG layer had a BLACK fill, so it was "invisible" over the black BG layer but was clearly created in the Layers hierarchy. Changed the fill color and there it was...

zoranagee:
Yes I've had that happen as well and I did check for this. I'll try to reproduce as well.

Apologize for errors and brevity. Tapping tiny keys.

On Aug 31, 2015, at 18:55, tiburbage <notifications@github.com<mailto:notifications@github.com>> wrote:


I tried without success to have the tool fail to create a Rectangle shape using the steps and content Zorana cited. I did notice that the first Rectangle I drew just above the BG layer had a BLACK fill, so it was "invisible" over the black BG layer but was clearly created in the Layers hierarchy. Changed the fill color and there it was...

-

ktaki:
Nobody seems to be able to reproduce this. @zoranagee Can you try it again and let me know.

chadrolfs:
I think this may be the same bug as #2192

ktaki:
Here are the steps to reproduce.

Create AB doc with Web (1920x1080) template. Save and close the file.
In Finder, select the file and enter Cmd+d six times to duplicate the file. 7 files in total.
Reset DS. Make sure the superselect tool is the current tool.
Select the seven files and drag them to the app window to open the files.
Select rectangle tool and try draw a rectangle on canvas
[...]

PR #2790, opened 2015-10-04
Misc border policy fixes

iwehrman:
1. Call `resetBorderPolicies` after `resetLayers` because bounds may have changed.
2. Call `resetBorderPolicies` at the _end_ of `disposeDocument` to avoid an atomicity violation due to lack of concurrency control among non-sequential invocations of `this.transfer`. This is a SuperHack :tm:.

The latter bug is quite nasty. I realized that it existed in theory a few weeks ago, but I verified that it's causing the second problem above. In particular: `disposeDocument` transfers `resetBorderPolicies` and another action which itself transfers to `resetBorderPolicies`, but does not serialize the two transfers. The executions of the two instances of `resetBorderPolicies` interleave in a _very bad way_. This is exactly the sort of concurrency controller that the controller should be providing for us, but in this case is not. I'll file a separate issue...

Addresses #2785. Also addresses, I hypothesize, #2224.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -445,15 +445,17 @@ define(function (require, exports) {
                     recentFilesPromise = this.transfer(application.updateRecentFiles),
                     updateTransformPromise = this.transfer(ui.updateTransform),
                     deleteTempFilesPromise = this.transfer(libraryActions.deleteGraphicTempFiles,
-                        documentID, isDocumentSaved),
-                    policyPromise = this.transfer(toolActions.resetBorderPolicies);
+                        documentID, isDocumentSaved);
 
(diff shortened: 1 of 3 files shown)


src/js/actions/documents.js
```diff
@@ -637,6 +639,7 @@ define(function (require, exports) {
 
         if (!document) {
             return Promise.resolve();
+            // return this.transfer(toolActions.resetBorderPolicies);
```
pineapplespatula: This should be deleted

pineapplespatula:
one nit, otherwise looks good.

iwehrman:
Whoops! Back to you, @pineapplespatula.

pineapplespatula:
woo.

Issue #2791, opened 2015-10-04
Controller transfer is not currency-safe


iwehrman:
First observed in the wild in #2790, the problem is that `this.transfer` does not provide the same concurrency-safety guarantees that top-level actions provide. For example, if a transaction transfers to two actions concurrently, their execution may be interleaved _even if they read and write the same state_. The controller protects against this among top-level actions, but not among intra-action transfers. 

A possible solution may be to rewrite `transfer` to make use of per-action `AsyncDependencyQueue` instances, which sounds like fun.

shaoshing:
Assign back to @iwehrman in case he has other plan for this issues.

iwehrman:
Nope! The only other action item here would to reconsider our use of many `transfer` calls in place of coarser actions that manage concurrency manually in order to improve performance. An example is `tools.beforeStartup` which was previously transferring to `addShortcut` once for each tool shortcut. After #2790, these are all serialized because they all write to the policy store. I added a new action `addShortcuts` which bulk-adds shortcuts using a single call to update the policies, and that is unsurprisingly much faster! There may be other cases where we can do the same thing. But I'm still going to close this bug. I'm just writing this here for @shaoshing so we're on the same page :sunflower:

PR #2794, opened 2015-10-05
Concurrency-safe transfer

iwehrman:
This a review-only PR which relies upon #2790. Besides reverting the second part of #2790 (pushing the transfer to `resetBorderPolicies` to the end of `disposeDocument`), it also makes a deeper change to the `FluxController` to use per-action queues to satisfy `transfer` requests. This ensures that the execution of transfers from within a single action are not interleaved in an unsafe way. 

There is some performance penalty to this. In particular, I note that `tools.beforeStartup` takes significantly longer to complete. This is not entirely unexpected because this is quite intentionally more conservative with respect to concurrency. Generally we'll pay a price in places where we currently transfer to a number of actions concurrently which are in conflict. This may well be unsafe, and so sorting out the concurrency control slows it down. I think `tools.beforeStartup` is more expensive because of the N concurrent `addShortcut` transfers. That also shows the solution to these slowdowns: just write a better action which is `addShortcuts`.

I'll discuss this more in the meeting this morning. I'd like to talk in particular about whether or not we think this should be merged before the next release.

Addresses #2791.


iwehrman:
One weird bug: panning the canvas does not seem to cloak the non-UI bounds. The artboard plus buttons, for example, remain when you pan.

iwehrman:
I fixed some existing bugs and found a few more, which are also now fixed in this branch. The updated commit has a record of the changes. Once #2810 is merged, I'll rebase this against that and see how startup performance is. If it's OK and if some testing doesn't turn up other issues, I think we should merge this for the upcoming release. I think I'll sleep better...

iwehrman:
I'm seeing really terrible bugs with stale border policies being left around only by virtue of selecting different layers. As soon as #2810 is merged, and this is rebased, I'm going to remove the review-only label.

iwehrman:
This is ready for testing. It seems to add ~600ms to startup time. I don't notice any other seriously ill effects. Personally, I think it's worth it. 

@chadrolfs and @ktaki: can you please test this branch and let me know if you notice anything weird? Especially let me know if you notice any performance regressions.

iwehrman:
This is rebased again and ready for testing.

chadrolfs:
going in...

chadrolfs:
This doesn't look good:

screen shot 2015-10-06 at 1 36 17 pm

chadrolfs:
not undo/redo safe

chadrolfs:
never mind, that bug is on master, not specific to this branch

iwehrman:
Sorry, is there a bug here that I should look at, @chadrolfs?

chadrolfs:
@iwehrman nope, issue I call out above is not specific to this branch, is on master and now logged.

ktaki:
I think Chad is pointing the issue #2831.
Of the three cases reported in the issue so far, the following two cases can be reproduced in the transfer-queue branch. 

case 1:

create a new doc
move the AB twice
cmd/ctrl + z
Result: coordinates are off

case 2:

Add AB to the left or top

iwehrman:
I believe these bugs are reproducible on master and have nothing to do with the race conditions currently present in the controller. Unfortunately this branch does not fix _all_ the bugs :)

chadrolfs:
@iwehrman you are correct.

src/js/actions/documents.js
```diff
@@ -465,17 +465,16 @@ define(function (require, exports) {
...
                         resetHistoryPromise,
                         updateTransformPromise,
                         recentFilesPromise,
-                        deleteTempFilesPromise);
-            })
-            .then(function () {
-                // This must run after the above transfers. See #2785.
-                return this.transfer(toolActions.resetBorderPolicies);
+                        deleteTempFilesPromise,
+                        policyPromise);
```
shaoshing: I saw the `policyPromise` has the comment `This must run after the above transfers`, so is it safe to put it within the other promises to run concurrently?

shaoshing:
Hey @iwehrman I read all the code and also learnt how the whole job queue works for actions and transfers. This is really great stuff. I have only one question and I cannot wait to merge it.

src/js/actions/documents.js
iwehrman: That should be OK. The transfer queue will ensure that the policy transfer doesn't run at the same time with anything else that changes the policies. I moved it previously (and added that comment) to try to clarify that running it concurrently with the other reset actions would cause an atomicity violation. But, with the transfer queue that should no longer be necessary.

shaoshing:
:+1: :+1:

Issue #2683, opened 2015-09-28
Opening a file while should ends vector mask mode


ktaki:
Open a image file. Enter vector mask mode and draw a rectangle. Open another image file. (or close the opened file without saving it and reopen it.) It is still in vector mask mode. Doc title is in blue. Drawing a shape, however, only draw a path. Vector mask is not created.

mcilroyc:
FBNC to @ktaki

ktaki:
The scenario above (Open a image file. Enter vector mask mode and draw a rectangle. Open another image file) will cause the following error.

![screen shot 2015-10-02 at 3 33 05 pm](https://cloud.githubusercontent.com/assets/2029597/10259262/1b501a28-691b-11e5-85cf-c0ba8c036e09.png)

mcilroyc:
FBNC to @ktaki

ktaki:
Verified the fix.

PR #2798, opened 2015-10-05
All open commands should close mask mode

pineapplespatula:
fix for issue #2683

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -42,6 +42,7 @@ define(function (require, exports) {
         guides = require("./guides"),
         layers = require("./layers"),
         locks = require("js/locks"),
+        log = require("js/util/log"),
         policy = require("./policy"),
         ui = require("./ui"),
         shortcuts = require("./shortcuts"),
@@ -550,9 +551,13 @@ define(function (require, exports) {
         var currentLayers = currentDocument.layers.selected,
             currentLayer = currentLayers.first();
 
-        // vector mask mode requires an avtive layer
+        // vector mask mode requires an active layer
         if (!currentLayer) {
-            return Promise.resolve();
+            if (vectorMaskMode) {
+                return Promise.resolve();
+            } else {
+                return this.dispatchAsync(events.tool.VECTOR_MASK_MODE_CHANGE, vectorMaskMode);
+            }
         }
 
         var nonVectorModeLayer = (currentLayer.kind === currentLayer.layerKinds.BACKGROUND ||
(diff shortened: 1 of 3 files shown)


mcilroyc:
I don't know if this is a separate bug or not, but I get the following error.  repro: open two docs, switch to first, go in to "m" mode, switch to second doc:

ss

pineapplespatula:
This is an issue of PS not resolving the modal promise until after the end of the modal tool state, This may end with a cancel event which we read as a failure (despite that it is a valid end condition for us)

pineapplespatula:
Ok this should be fixed now. Give it a try

src/js/actions/export.js
```diff
@@ -969,7 +969,7 @@ define(function (require, exports) {
     exports.exportLayerAssets = exportLayerAssets;
     exports.exportDocumentAssets = exportDocumentAssets;
     exports.setUseArtboardPrefix = setUseArtboardPrefix;
-    exports.afterStartup = afterStartup;
+    // exports.afterStartup = afterStartup;
```
mcilroyc: derp

src/js/actions/tools.js
```diff
@@ -704,6 +709,25 @@ define(function (require, exports) {
...
+     * @return {Promise}
+     */
+    var enterPathModalState = function () {
+        descriptor.batchPlayObjects([vectorMaskLib.enterFreeTransformPathMode()],
+                { synchronous: false })
+            .catch(function () {
+                // Silence the errors here since we cannot guarantee that PS will not throw an error
+                // when we end the modal tool state in a valid case
+            });
+        return Promise.resolve();
```
mcilroyc: I'm skeptical that we don't want to return this playObject promise.  I get that we don't want it to bubble out an exception, but seems like it could at least wait for it to resolve.

Also, it seems like we'd at least want to *log* the errors.  It seems like over time our goal is to try our hardest not to make ps calls that will error, so we'll want to use the messages as diagnostics.  eh?

mcilroyc:
Seems to work.  A few comments inline though.  back to @pineapplespatula

pineapplespatula:
@mcilroyc back to you

src/js/actions/tools.js
```diff
@@ -552,7 +553,11 @@ define(function (require, exports) {
 
         // vector mask mode requires an avtive layer
```
mcilroyc: sp `avtive`

src/js/actions/tools.js
```diff
@@ -552,7 +553,11 @@ define(function (require, exports) {
...
         if (!currentLayer) {
-            return Promise.resolve();
+            if (vectorMaskMode) {
+                return Promise.resolve();
+            } else {
+                return this.dispatch(events.tool.VECTOR_MASK_MODE_CHANGE, vectorMaskMode);
```
mcilroyc: I think you meant dispatchAsync.

pineapplespatula:
@mcilroyc once more, this time with feeling

PR #2810, opened 2015-10-05
Add shortcuts.addShortcuts to add shortcuts

iwehrman:
1. Add a new `shortcuts.addShortcuts` action to add shortcuts in bulk.
2. Use one call to `shortcuts.addShortcuts` in `tools.beforeStartup`. Previously, we were doing separate transfers for each call to `shortcuts.addShortcut` (singular) which was both bad for performance (N redundant policy calls in the adapter instead of one) and a concurrency-safety problem waiting to happen. 

Addresses: concerns.

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -678,11 +678,11 @@ define(function (require, exports) {
         descriptor.addListener("toolModalStateChanged", _toolModalStateChangedHandler);
 
         // Setup tool activation keyboard shortcuts
-        var shortcutPromises = tools.reduce(function (promises, tool) {
+        var shortcutSpecs = tools.reduce(function (specs, tool) {
             var activationKey = tool.activationKey;
 
             if (!activationKey) {
-                return promises;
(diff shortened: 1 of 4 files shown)


src/js/actions/tools.js
```diff
@@ -695,37 +695,43 @@ define(function (require, exports) {
...
-            return promises;
+            return specs;
         }.bind(this), []);
     
         _vectorMaskHandler = function () {
             var toolStore = this.flux.store("tool"),
                 vectorMode = toolStore.getVectorMode() || false;
             
-            return this.transfer(changeVectorMaskMode, !vectorMode);
+            this.flux.actions.tools.changeVectorMaskMode(!vectorMode);
```
iwehrman: This is an unrelated fix. It's not correct to transfer from an event handler like this because the locks are no longer being held. Instead, a new action should be enqueued.

shaoshing:
LGTM! :+1:

PR #2814, opened 2015-10-05
CC Libraries - addresses rename input focus issue

shaoshing:
This PR addresses #2197, also added SCU for LibraryList component to improve performance.

### DIFF
--- a/src/js/jsx/sections/libraries/LibraryList.jsx
+++ b/src/js/jsx/sections/libraries/LibraryList.jsx
@@ -25,6 +25,7 @@ define(function (require, exports, module) {
     "use strict";
 
     var React = require("react"),
+        _ = require("lodash"),
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React);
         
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -291,7 +304,7 @@ define(function (require, exports, module) {
             return (<div className="libraries__bar__top__content">
                 <Datalist
                     ref="list"
-                    list={listID}
+                    list="libraryes"
```
iwehrman: Typo :tractor:

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -71,10 +72,23 @@ define(function (require, exports, module) {
...
         componentDidUpdate: function () {
-            if (this.refs.input) {
-                this.refs.input.getDOMNode().focus();
+            var renameInput = this.refs.input;
+            
+            if (renameInput) {
+                window.setTimeout(function () {
+                    renameInput.acquireFocus();
+                    renameInput.getDOMNode().focus();
+                }, 100);
```
iwehrman: This is a bummer, but I understand that there are weird focus issues going on here. We might want to increase this timeout just a little more though to be on the safe side?

shaoshing: For some reason, using `element.getDOMNode().focus()` alone does not focus the rename input property. It only select the text but not in a focused state (without showing the blue bottom border). I am not confident that using `setTimeout` and `acquireFocus` is a good solution. So anyone who worked on the Focusable component before, please let me know your thoughts.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -71,10 +72,23 @@ define(function (require, exports, module) {
...
         componentDidUpdate: function () {
-            if (this.refs.input) {
-                this.refs.input.getDOMNode().focus();
+            var renameInput = this.refs.input;
+            
+            if (renameInput) {
+                window.setTimeout(function () {
+                    renameInput.acquireFocus();
+                    renameInput.getDOMNode().focus();
+                }, 250);
```
shaoshing: Increased the wait time from 100 to 250 for a safer auto focus.

shaoshing:
@iwehrman assign back to you as your are the first reviewer. Let me know if you have other thoughts about the setTimeout solution, I'd love to switch to a more robust way; otherwise, it is ready for merge. Thank you

iwehrman:
I encountered this problem when working around a similar issue elsewhere in the libraries panel. We have some serious blur/focus issues with our core components. I think, in the very short term, that this is probably the best we can do given how many other high-priority issues we have to address for the upcoming release. So, let's check in this hack now and re-think how we manage these issues in a few weeks.

Issue #2642, opened 2015-09-25
Attempting to edit/open an CC Library asset results in Open dialog


chadrolfs:
create a new doc
drag an asset onto canvas (Test - OK to accidentally nuke: Climber)
hit Enter

Result: Open dialog

shaoshing:
The exact steps to reproduce this:

1 switch to DS, double click to open any graphic for edit.
2 switch to classic and close the document 
3 switch to DS, open the same graphic again and then close it
4 open the same graphic -- open dialog appears.

iwehrman:
FBNC to @chadrolfs.

PR #2815, opened 2015-10-05
CC Libraries - fix open-dialog issue when double click graphic for edit.

shaoshing:
This PR addresses issue #2642 by removing stale edit status in the initial load (when you open the graphic in DS and close the document in classic).

### DIFF
--- a/src/js/stores/library.js
+++ b/src/js/stores/library.js
@@ -30,7 +30,8 @@ define(function (require, exports, module) {
         CCLibraries = require("file://shared/libs/cc-libraries-api.min.js");
 
     var events = require("../events"),
-        log = require("js/util/log");
+        log = require("js/util/log"),
+        math = require("js/util/math");
     
     /**
      * Regular expression for removing invalid characters in graphic asset's temp filename.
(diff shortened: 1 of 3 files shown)


src/js/stores/library.js
```diff
@@ -196,9 +198,19 @@ define(function (require, exports, module) {
...
-                return [parseInt(documentID), status];
-            });
+            var editStatusArray = _.reduce(payload.editStatus, function (result, status, documentID) {
+                // Remove orphan edit status that are not associate with any open documents.
+                
+                var document = this.flux.stores.document.getDocument(documentID),
+                    documentName = document ? document.name : "";
+                
+                if (documentName.indexOf(status.elementID) !== -1) {
+                    result.push([parseInt(documentID), status]);
```
iwehrman: I'm not sure why documentID is a string here? Can you just use `document.id` instead? In any case, you shouldn't use `parseInt` (and _especially_ not without a radix) because it returns `NaN` if it fails to parse. Instead, if you have to parse the string, use the helper in `js/util/math` and test for null.

src/js/stores/library.js
```diff
@@ -196,9 +199,25 @@ define(function (require, exports, module) {
             this._libraryCollection = payload.collection;
             this._updateLibraries(payload.lastSelectedLibraryID);
             
-            var editStatusArray = _.map(payload.editStatus, function (status, documentID) {
-                return [parseInt(documentID), status];
-            });
+            var editStatusArray = _.reduce(payload.editStatus, function (result, status, documentIDStr) {
+                // Remove orphan edit status that are not associate with any open documents.
+
+                var documentID = math.parseNumber(documentIDStr);
```
shaoshing: @iwehrman `payload.editStatus ` is loaded from preference and it is stored as string using `JSON.stringify`, its keys are stringified to match JSON spec, that's why the documentID is string here. I gave it a better name, and use `math.parseNumber` as suggested to deal with edge cases.

iwehrman:
OK!

Issue #2424, opened 2015-09-11
Export dialog flickers while panning the canvas


DivyaPrabhakar:
When the export dialog is open, the user can still move the canvas behind the dialog and this causes the dialog to flicker and only rest when the panning stops. 

![Export dialog flicker](https://www.dropbox.com/s/ohymkie6uw5npjp/export-flickers.gif?raw=1)

iwehrman:
Nice catch!

mcilroyc:
i hope you're both happy!  FBNC to @DivyaPrabhakar

DivyaPrabhakar:
Verified on my end.

PR #2822, opened 2015-10-06
Dismiss the export dialog on wheel events

iwehrman:
Addresses #2424. 

Ideally I think these would just be suppressed, but I'm not sure how to do that and I _am_ sure how to do this! All dialogs with a `dismissOnCanvasClick` policy will have the same behavior. 

PS: actually, I do know how to suppress these events: by using a policy over the canvas region. But, that seems fairly risky compared to this simple change.

Also note: this does _not_ address a similar problem with pinch-to-zoom. To handle that, we would need to do something at a very different level because there are no adapter os events for the pinch gesture. For example, the cloaking mechanism would have to know whether or not there is a dialog open that is on top of the canvas. So, this is definitely still an issue, but one that I think will have to be addressed separately!

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -169,6 +169,23 @@ define(function (require, exports, module) {
         },
 
         /**
+         * Dismiss the dialog on mouse wheel events.
+         *
+         * @private
+         */
+        _handleMouseWheel: function () {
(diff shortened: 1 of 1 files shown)


mcilroyc:
well, my hats off to you sir.  I kind of can't believe how well this works.  This solution had not occurred to me, dude.

occurred to me

PR #2823, opened 2015-10-06
expand hit rigion on svg elements

pineapplespatula:
in the reference tool I created transparent hit rectangles that are larger than the drawn rectangles.


src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -128,31 +128,49 @@ define(function (require, exports, module) {
                                     L14.000,9.000 L14.000,14.000 L9.000,14.000 L9.000,9.000 ZM10.000,13.000
                                     L13.000,13.000 L13.000,10.000 L10.000,10.000 L10.000,13.000 Z" />
                                     <rect className={getReferenceClasses("rb")} id="rb"
-                                        x="16" y="16" width="5" height="5"
+                                        x="16" y="16" width="5" height="5"/>
+                                    <rect className={getReferenceClasses("rb_hit")} id="rb_hit"
```
iwehrman: It seems like you really just want _no_ `className` on these? Isn't that what will happen anyway here?

pineapplespatula:
back to you

iwehrman:
OK!

Issue #2652, opened 2015-09-25
Ungroup does not create a history state


iwehrman:
1. Open Vermilion, select the last artboard.
2. Ungroup.
3. Undo. 

Result: nothing, because the Undo menu item is disabled.

iwehrman:
Fixed by #2827.

PR #2827, opened 2015-10-06
Menu stores now listens for UNGROUP_SELECTED

mcilroyc:
This addresses #2652 
When ungrouping layers, the menu was not recognizing this change, and so in some cases the "undo" menu item was left inactive.


iwehrman:
BOOM!

PR #2830, opened 2015-10-06
 refactor out code to determine if vector modes are legal on a given layer

pineapplespatula:


src/js/models/layerstructure.js
```diff
@@ -826,6 +826,19 @@ define(function (require, exports, module) {
     }));
 
     /**
+     * Determine whether a vector mask can be created on a layer 
+     *
+     * @param {Layer} layer
+     * @return {boolean}
+     */
+    Object.defineProperty(LayerStructure.prototype, "canHaveVectorMask", objUtil.cachedLookupSpec(function (layer) {
```
iwehrman: I thought that we only supported this for a given layer selection? Can this just be a property and not a lookup? I.e., shouldn't this always operate on the given layer selection? For example: if no layers are selected, then the answer is no. If many layers are selected, then the answer hinges on the first selected layer, right?

iwehrman:
Once you make the above change, can you _also_ make use of it in `DocumentHeader` here: 
https://github.com/adobe-photoshop/spaces-design/blob/34908b0e35a1cbf660c90d502fe7e891816c5ee3/src/js/jsx/DocumentHeader.jsx#L229-L235

I'm out until Monday, so please bounce this over to @mcilroyc when it's ready for another look.

iwehrman:
What's the deal here, @pineapplespatula? This has been waiting for over a month.

pineapplespatula:
it got ignored during the end game, @iwehrman :repeat_one:

iwehrman:
Can this be used to control enablement of the vector mask icon in the document header too?

iwehrman:
Great, thanks!

PR #2833, opened 2015-10-06
Fixed size of reference tool

designedbyscience:
Okay, chatted with @placegraphichere and decided on keeping the icon at the 24px sizing that it natively is...which ended up being achieved by deleting some CSS.


iwehrman:
Crispy.

Issue #2799, opened 2015-10-05
Slow Generator failure blocks File > New


iwehrman:
When Generator fails to start, `exports.beforeStartup` holds the `JS_EXPORT` write lock for a very long time. The `documents.createNew` action also requires the `JS_EXPORT` lock because. This prevents File > New from working while Generator is slowly failing. 

In some sense, this is how it should be working. On the other hand, it would be nice to break up at least one of these actions so that system can remain responsive while Generator is starting up. Even when Generator is installed correctly, this will cause the system to be unresponsive while Generator is first starting up, which can take a while even in the best case.

CC @mcilroyc

iwehrman:
@pineapplespatula: do you mind verifying the fix?

pineapplespatula:
Nicely done this looks good

PR #2834, opened 2015-10-06
 Improve export service startup to not lock other actions, better error handling on post-export asset updates

mcilroyc:
There are three components to this PR, the first two of which address #2799, and the other fixes an unreported bug that I noticed while testing.

1. When export service completes its initialization, it updates the store's "available" flag via a separate flux action (instead of transfer) so that it only gets the JS_EXPORT lock when it needs it, thus not preventing other actions in the meantime.
1. Remove the GENERATOR lock entirely.  Because the plugin's websocket interface will allow concurrent calls, there is no practical need for this lock. *This* lock (held by the export's afterStartup action) was preventing the `document.close` action which uses generator to clean up old temp files.
1. Allow (only) the post-export asset update to fail silently - these updates are not strictly necessary and should not cause a reset.  See below for more info.

The unreported bug:
Changing documents during a batch export will throw a bunch of errors when trying to update the assets in the "extension data" of the now-inactive document.  So now we just swallow those errors.  This could mean that the assets don't get updated correctly with the "STABLE" status, nor the file location, but that seems like an acceptable risk.  This might be fixable on the core side in the future by allowing extension data updates of inactive documents, if it becomes important.

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -303,16 +303,6 @@ define(function (require, exports) {
     };
 
     /**
-     * Update the export store with the new service availability flag;
-     *
-     * @param {boolean} available
-     * @return {Promise}
-     */
-    var _setServiceAvailable = function (available) {
-        return this.dispatchAsync(events.export.SERVICE_STATUS_CHANGED, { serviceAvailable: !!available });
-    };
-
-    /**
      * Update the export store with the new service busy flag;
      *
      * @param {boolean} busy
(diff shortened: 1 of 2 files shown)


src/js/actions/export.js
```diff
@@ -180,23 +180,32 @@ define(function (require, exports) {
...
+                function (e) {
+                    log.error("Export Failed for asset %d of layerID %d, documentID %d, with error: ",
+                        assetIndex, layer && layer.id, document.id, e);
+
+                    var assetProps = {
+                        filePath: "",
+                        status: ExportAsset.STATUS.ERROR
+                    };
+
+                    return this.flux.actions.export.updateExportAsset(document, _layers, assetIndex, assetProps,
```
iwehrman: Are you sure this is what you want to do here? I agree with the "enqueue an action" instead of "transfer to the action", but I don't think you want to block this action on the result of the "enqueue an action" promise. Almost always the two correct choices are 1) transfer and use the promise to block the parent action; or 2) enqueue a new action but ignore the promise.

src/js/actions/export.js
```diff
@@ -856,7 +878,7 @@ define(function (require, exports) {
                 .bind(this)
                 .then(function () {
                     log.debug("Export: Generator plugin connection established");
-                    return _setServiceAvailable.call(this, true);
+                    return this.flux.actions.export.setServiceAvailable(true);
```
iwehrman: Same here.

iwehrman: But, in this case, I'm actually not sure why a transfer wouldn't do the trick.

iwehrman:
Other than those action Q's, looks fine.

src/js/actions/export.js
mcilroyc: I suppose I get on board with (on both counts) not returning this "enqueue an action" promise.  However, the reason I'm doing this here is item No. 1 on the PR description.

mcilroyc:
un-returned.  back to @iwehrman

iwehrman:
Lovin' it :hamburger: :fries:

Issue #2770, opened 2015-10-02
Place linked / embedded SO freezes the app


shaoshing:
1 menu: file > Place Linked / Embedded ...
2 select any file and hit return
3 app frozen

This is probably related to keyboard and mouse policy as well.

iwehrman:
FBNC. Over to @chadrolfs to verify.

shaoshing:
@chadrolfs This is fixed but it has the side effect described in #2849 which I am working on

Issue #2836, opened 2015-10-06
Artboard add UI shows both standard and DS add UI; gets stuck in standard PS tools


chadrolfs:
new doc,
drag a CC Library graphic asset onto canvas 
click on AB title

Result: 
![ab-doubleui](https://cloud.githubusercontent.com/assets/8442631/10324692/72f48980-6c40-11e5-9223-bdfd2d0161ee.png)

iwehrman:
@chadrolfs Please file a high-priority core bug for this and assign to CJB.

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4069723

chadrolfs:
Worse than I thought: 
new doc,
drag a CC Library graphic asset onto canvas, Enter to place
hit r key for Rect tool

Result: you get standard PS rotation tool.  if you hit Q, you go into quick mask mode.

shaoshing:
I've updated how we restore policies for placing graphic asset in PR 2837. However this issue is not entirely fiexd becaue we are not receiving the "placeEvent" from Photoshop (729), so the new layer is not added to DS.

iwehrman:
@shaoshing is a new Watson bug needed for the missing `placeEvent`?

shaoshing:
@iwehrman Jesper updated the watson bug saying the `placeEvent` needs to be disabled for graphic asset for now due to 

```
	// Until This code knows how to record and playback a place that references
	// a library elementRef, just disable recording for library places. Otherwise
	// they would record with references to the file in the libs cache folder
	// which is no good, especially in the case of a linked place (you'd end up
	// with a local link to the cache folder file on playback).
```

I think I will hook up the addLayer to the `toolModalStateChange` event and that should fix the problem. working on a fix now.

shaoshing:
I am pausing my work on this until Jesper and Tai figure out whether or not Photoshop should emit the `placeEvent` for placing graphic from libraries.

shaoshing:
Jesper's work in dev.739 fixes the issue perfectly. @chadrolfs can you verify?

chadrolfs:
Getting access or stuck with standard PS tools looks fixed but I'm still seeing the standard PS AB UI on windows with dev.721 and master @ 3bc3356

chadrolfs:
hmmmm, maybe it is fixed, having a hard time repro'ing now.

chadrolfs:
[...]

PR #2837, opened 2015-10-06
Suspend/Restore event policies around menu's place commands.

shaoshing:
This is similar to libraries drag-n-drop, the event policies have to be suspended before executing place commands via the menu item, and then restored afterward; otherwise, DS will be freezed. Addresses issue #2770

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -28,7 +28,8 @@ define(function (require, exports) {
 
     var adapter = require("adapter"),
         ps = require("adapter/ps"),
-        ui = require("adapter/ps/ui");
+        ui = require("adapter/ps/ui"),
+        descriptor = require("adapter/ps/descriptor");
 
     var events = require("js/events"),
         locks = require("js/locks"),
@@ -36,6 +37,7 @@ define(function (require, exports) {
         log = require("js/util/log"),
         global = require("js/util/global"),
         headlights = require("js/util/headlights"),
+        policyActions = require("./policy"),
         preferencesActions = require("./preferences"),
         searchActions = require("./search/menucommands");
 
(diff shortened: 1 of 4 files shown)


src/js/actions/menu.js
```diff
@@ -63,18 +74,38 @@ define(function (require, exports) {
...
+                return ps.performMenuCommand(payload);
+            })
+            .then(function (success) {
+                // FIXME: After the M2 release, remove the debug conjunct
+                if (global.debug && !success) {
+                    throw new Error("Menu command not available: " + payload.commandID);
+                }
+            })
+            .catch(function () {
+                if (!isPlaceCommand) {
```
iwehrman: I'm confused as to why this happens when it's _not_ a place command that fails?

src/js/actions/menu.js
```diff
@@ -63,18 +74,38 @@ define(function (require, exports) {
...
-                throw new Error("Menu command not available: " + payload.commandID);
-            }
-        });
+        var isPlaceCommand = payload.commandID === _PLACE_LINKED_MENU_ID ||
+                payload.commandID === _PLACE_EMBEDDED_MENU_ID;
+        
+        return Promise.bind(this)
+            .then(function () {
+                if (isPlaceCommand) {
+                    this.dispatchAsync(events.menus.PLACE_COMMAND, { executing: true });
```
iwehrman: You probably want to just use `this.dispatch` here? Otherwise, the dispatch may not be complete by the time your transfer command executes.

src/js/actions/menu.js
iwehrman: Even if this is what you want, can you leave a comment here explaining why there is a failure and why it's OK to catch it and what the transfer accomplishes?

shaoshing: We suspend the policies before running the place command, so here we want to make sure they are restored if any exceptions raised. I also added comment to explain to reason.

shaoshing:
Hi @iwehrman , thanks for reviewing; I've addressed your comments. I also made the change to ease issue #2836 so that placing graphic from cc libraries won't steal all the policies. Right now we are waiting the core to fix the other part of the issue. Back to you.

src/js/actions/menu.js
iwehrman: Yes, but this says `if (!placeCommand)`. I understand that we want to clean up after suspending policies as a result of  place command. But it seems like this will do the cleanup only after executing something _other than_ a place command. Sorry if I'm just not getting it :)

shaoshing:
@iwehrman I've fixed the the obvious bug you pointed out; I appreciate for double checking it and did not let it slip thought :D

iwehrman:
OK, let's give it a whirl!

Issue #2826, opened 2015-10-06
Sampler: Shift + click to sample Typeface fails


chadrolfs:
create two type layers with different properties (including Typeface)
with one type layer selected, hit i for the Sampler tool
mouse over the other type layer, hold down Shift and click

Result: `Action sampler.click failed: TypeError: undefined is not a function`

chadrolfs:
![screen shot 2015-10-06 at 8 44 17 am](https://cloud.githubusercontent.com/assets/8442631/10313804/52b7e0e8-6c06-11e5-9677-782223a2ad22.png)

iwehrman:
FBNC to @chadrolfs.

Issue #2777, opened 2015-10-03
Sampling Text via HUD sometimes fails on first click, takes 2nd or double-clicks


chadrolfs:
create two different text layers (different size, color, typeface)
select one, switch to Sampler
mouse over second, press Space to bring up HUD
click on Typeface icon

Result: nothing is applied
click again or in last step above double-click

iwehrman:
FBNC to @chadrolfs.

PR #2838, opened 2015-10-06
Sampler - addresses 2777 & 2826

shaoshing:
This PR addresses the following sampler issues:

#2777 Sampling Text via HUD sometimes fails on first click, takes 2nd or double-clicks

#2826 Sampler: Shift + click to sample Typeface fails. This is fixed by increasing the clickable area of the type button in the HUB.


iwehrman:
LGTM!

Issue #2824, opened 2015-10-06
Clicking 'x' in search box does not dismiss the search dialog


ktaki:
Reported by John F. I wonder if this 'x' ever worked. If we follow the Export dialog's case, we might want to simply remove it. (Or, if this simple 'x' is useful for reminding users clicking it to dismiss the dialog, we might want to add it to Export dialog also.)

iwehrman:
FBNC to @ktaki. The "X" is designed to clear the input. It is now only visible when the search field is non-empty.

ktaki:
A Ha! Opened #2848 as UI issue.

PR #2839, opened 2015-10-06
Search - show clear input button only when the search input has value.

shaoshing:
Related to issue #2824 (Clicking 'x' in search box does not dismiss the search dialog) I was also confused at the beginning but the X button in the search box is for clearing the input. I made some change to display the X button only when the input has value, to ease the confusion.

### DIFF
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -48,6 +48,8 @@ define(function (require, exports, module) {
             value: React.PropTypes.string,
             // Callback to handle change of selection. Return false will cancel the selection.
             onChange: React.PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/Datalist.jsx
```diff
@@ -48,6 +48,8 @@ define(function (require, exports, module) {
             value: React.PropTypes.string,
             // Callback to handle change of selection. Return false will cancel the selection.
             onChange: React.PropTypes.func,
+            // Callback to handle change of input
+            onInputChange: React.PropTypes.func,
```
iwehrman: One more prop!

iwehrman:
Nice work. Thanks!

PR #2841, opened 2015-10-06
Some extra new lines escaped from our test

shaoshing:
These empty lines contain only spaces and tabs were not considered as extra lines in our test. Let's destroy them and make sure our test can capture them in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "grunt-jsonlint": "^1.0.4",
     "grunt-jsxhint": "^0.6.0",
     "grunt-lintspaces": "^0.7.0",
+    "node-lintspaces": "shaoshing/node-lintspaces",
     "jscs-jsdoc": "shaoshing/jscs-jsdoc#v1.1.0.x",
     "jscs-trailing-whitespace-in-source": "0.0.1",
     "lodash": "^3.10.1",
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -997,7 +997,6 @@ define(function (require, exports) {
                 documentID: documentID,
                 path: event.in._path
             });
-            
 
             if (!saveAs) {
                 return;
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "grunt-jsonlint": "^1.0.4",
     "grunt-jsxhint": "^0.6.0",
     "grunt-lintspaces": "^0.7.0",
+    "node-lintspaces": "shaoshing/node-lintspaces",
```
iwehrman: Is this extra package actually being used? Does `grunt-lintspaces`somehow pick it up?

shaoshing: yes the new package is in use. `grunt-lintspaces` relies on `node-lintspaces` and the later one does not check lines with spaces or tabs. I made a tiny fix to its regexp to address these lines.

iwehrman:
So, is node-lintspaces a "peer dependency" here? I'm not sure I knew this was possible... maybe I did, but I didn't realize it was this easy?

iwehrman:
Also, you should submit your regexp fix back to the main node-lintspaces project, too :smile:

shaoshing:
I've never use `peerDependencies` before, but I looked it up and it is more likely to be used by a plugin to resolve possible dependencies conflicts with other plugins or the host. DS here is more like the host and may not getting any benefit from it. Please let me know if I am wrong :D  BTW, submiting a PR to the project sounds great!

package.json
iwehrman: Well, I have no idea why this works (@mcilroy suggests `grunt` magic?) but it does seem to work. Note that I had to `rm -rf node_modules` and then `npm install` again for the change to take effect. But, hey, better linting!

iwehrman:
@shaoshing: let us know if/when `grunt-lintspaces` accepts your PR so we can remove the extra dependency.

PR #2845, opened 2015-10-06
 add WarningLevels to layers for missing fonts

pineapplespatula:
This PR requires Jenkins build 725. It fixes issue #2122

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -132,15 +132,24 @@ define(function (require, exports) {
             setFacePromise = locking.playWithLockOverride(document, layers, setFacePlayObject, typeOptions),
             updatePromise = this.transfer(updatePostScript, document, layers, postscript, family, style);
 
-        return Promise.join(updatePromise, setFacePromise).bind(this).then(function () {
-            if (!modal) {
-                return this.transfer(layerActions.resetBounds, document, layers);
-            }
-        });
+        return Promise.join(updatePromise, setFacePromise)
+            .bind(this)
+            .then(function () {
+                if (!modal) {
+                    var anylayerTextWarning = layers.some(function (layer) {
+                        return layer.textWarningLevel > 0;
+                    });
+                    if (anylayerTextWarning) {
+                        return this.transfer(layerActions.resetLayers, document, layers, true);
+                    } else {
+                        return this.transfer(layerActions.resetBounds, document, layers);
+                    }
+                }
+            });
     };
     setPostScript.reads = [locks.JS_DOC];
     setPostScript.writes = [locks.PS_DOC, locks.JS_UI];
-    setPostScript.transfers = [updatePostScript, layerActions.resetBounds];
+    setPostScript.transfers = [updatePostScript, layerActions.resetBounds, layerActions.resetLayers];
     setPostScript.modal = true;
 
     /**
@@ -188,15 +197,24 @@ define(function (require, exports) {
             setFacePromise = locking.playWithLockOverride(document, layers, setFacePlayObject, typeOptions),
             updatePromise = this.transfer(updateFace, document, layers, family, style);
 
-        return Promise.join(updatePromise, setFacePromise).bind(this).then(function () {
-            if (!modal) {
-                return this.transfer(layerActions.resetBounds, document, layers);
-            }
-        });
+        return Promise.join(updatePromise, setFacePromise)
+            .bind(this)
+            .then(function () {
+                if (!modal) {
+                    var anylayerTextWarning = layers.some(function (layer) {
+                        return layer.textWarningLevel > 0;
+                    });
+                    if (anylayerTextWarning) {
+                        return this.transfer(layerActions.resetLayers, document, layers, true);
+                    } else {
+                        return this.transfer(layerActions.resetBounds, document, layers);
+                    }
+                }
+            });
     };
     setFace.reads = [locks.JS_DOC];
     setFace.writes = [locks.JS_UI, locks.PS_DOC];
-    setFace.transfers = [updateFace, layerActions.resetBounds];
+    setFace.transfers = [updateFace, layerActions.resetBounds, layerActions.resetLayers];
     setFace.modal = true;
 
     /**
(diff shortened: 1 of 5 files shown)


iwehrman:
What's the state of the "layout will change" dialog? I'm worried that we're suppressing it and that we're actually re-laying out the text, but the warning icon doesn't go away ever. Is that what's happening? If it's automagically re-laying out the text, should we be resetting that property after first selecting the layer?

iwehrman:
Actually, it seems like if I edit the text layer, the icon is fixed. But if I change the font the icon stays the same. (To answer my own question, the dialog is suppressed and I guess it does the re-layout.) From what I can tell, both cases should cause a re-layout, but changing the font doesn't change the icon. So, that's a bug!

iwehrman:
Other than that, looks good so far :grimacing:

pineapplespatula:
i dont think we want to suppress these dialogs. Do we do that in JS or does the plugin supress all dialogs?

pineapplespatula:
I added an update to our document model @iwehrman  back to you

src/js/stores/document.js
```diff
@@ -907,7 +907,8 @@ define(function (require, exports, module) {
             var documentID = payload.documentID,
                 layerIDs = payload.layerIDs,
                 document = this._openDocuments[documentID],
-                nextLayers = document.layers.setCharacterStyleProperties(layerIDs, { postScriptName: postScriptName }),
+                tmpLayers = document.layers.setCharacterStyleProperties(layerIDs, { postScriptName: postScriptName }),
+                nextLayers = tmpLayers.setTextWarningLevel(layerIDs, 0),
```
iwehrman: Are we sure the warning level is now zero?

iwehrman: Also, are you sure that this is the _only_ property change that will cause a re-layout? What about if you change the size, tracking, leading or alignment?

pineapplespatula:
@iwehrman  back to you

src/js/models/layerstructure.js
```diff
@@ -1604,7 +1604,8 @@ define(function (require, exports, module) {
 
             // .set is used here instead of merge to eliminate the other styles
             var nextText = layer.text.set(styleProperty, nextStyle),
-                nextLayer = layer.set("text", nextText);
+                tmpLayer = layer.set("text", nextText),
+                nextLayer = tmpLayer.set("textWarningLevel", 0);
```
iwehrman: It seems like this this will remove every warning after any change. But, surely we still want to keep the warning level around for missing fonts?

pineapplespatula: modifying layers with missing fonts, ( with a suppressed dialog) should substitute out the missing font. 
at least that is the behavior in classic

pineapplespatula:
@iwehrman back to you

ktaki:
@pineapplespatula Clicking on a text layer does not insert an I beam cursor. Is this caused by your change?

iwehrman:
Is this specific to @pineapplespatula's branch? I can't repro the issue either on his branch or on master.

ktaki:
It was only happening with Jeffrey's branch. Master was fine. Tested with Jeffrey's branch again. CNR. Possibly it was something else.

src/js/actions/type.js
```diff
@@ -132,15 +132,24 @@ define(function (require, exports) {
...
-            if (!modal) {
-                return this.transfer(layerActions.resetBounds, document, layers);
-            }
-        });
+        return Promise.join(updatePromise, setFacePromise)
+            .bind(this)
+            .then(function () {
+                if (!modal) {
+                    var anylayerTextWarning = layers.some(function (layer) {
+                                return layer.textWarningLevel > 0;
```
iwehrman: Move this and the next line back one tab stop.

src/js/actions/type.js
```diff
@@ -132,15 +132,24 @@ define(function (require, exports) {
...
+                    } else {
+                        return this.transfer(layerActions.resetBounds, document, layers);
+                    }
+                }
+            });
     };
     setPostScript.reads = [locks.JS_DOC];
     setPostScript.writes = [locks.PS_DOC, locks.JS_UI];
-    setPostScript.transfers = [updatePostScript, layerActions.resetBounds];
+    setPostScript.transfers = [updatePostScript, layerActions.resetBounds,layerActions.resetLayers];
```
iwehrman: Missing space after the comma. (I wonder why JSCS doesn't catch this? CC @shaoshing, our other resident whitespace nitpicker.)

src/js/actions/type.js
```diff
@@ -188,15 +197,24 @@ define(function (require, exports) {
...
-            if (!modal) {
-                return this.transfer(layerActions.resetBounds, document, layers);
-            }
-        });
+        return Promise.join(updatePromise, setFacePromise)
+            .bind(this)
+            .then(function () {
+                if (!modal) {
+                    var anylayerTextWarning = layers.some(function (layer) {
+                            return layer.textWarningLevel > 0;
```
iwehrman: Move this and the next line back one tab stop.

src/js/actions/type.js
```diff
@@ -278,11 +296,22 @@ define(function (require, exports) {
         var updatePromise = this.transfer(updateColor, document, layers, color, modal, options),
             setColorPromise = layerActionsUtil.playSimpleLayerActions(document, layers, playObject, true, typeOptions);
 
-        return Promise.join(updatePromise, setColorPromise);
+        return Promise.join(updatePromise, setColorPromise)
+            .bind(this)
+            .then(function () {
+                if (!modal) {
+                    var anylayerTextWarning = layers.some(function (layer) {
+                            return layer.textWarningLevel > 0;
```
iwehrman: Ditto.

iwehrman: (Repeat until EOF.)

iwehrman: Also please add a newline before the `if` statement.

iwehrman:
Just some spacing nits remain. @ktaki: I can't see any reason why this would change the behavior of text selection, so I'm not going to block this PR for that unless you can repro.

pineapplespatula:
fixed. let me know if you would prefer i rebase down to one commit before merging?

src/js/actions/type.js
shaoshing: Because we don't have `requireSpaceAfterComma`, which is a new feature to JSCS 6 days agao.
https://github.com/jscs-dev/node-jscs/blob/94f175eec822f62528e6e5ca5aab0eb1de037243/lib/rules/require-space-after-comma.js

iwehrman:
OK!

Issue #2843, opened 2015-10-06
Document header resets when mask mode is active 


DivyaPrabhakar:
Repro: 

1. Create a lot of new documents so there is overflow in the document header
2. Press "m"
3. Header should clear and then open all the documents again. Exaggerated flicker.

iwehrman:
The problem specifically here is the exaggerated flicker. It should be possible to repro this also with Help > Test Action Failure. The fix is to ensure that the DocumentHeader sCU does not allow rendering when the controller is resetting.

mcilroyc:
To @DivyaPrabhakar for validation of the fix

DivyaPrabhakar:
Looks good @mcilroyc!

PR #2847, opened 2015-10-07
Delay rendering document header until inactive documents are loaded

iwehrman:
This prevents the document tabs from mounting one at a time, which is distracting. Yes, it delays the render somewhat, but since we initialize documents lazily now, it shouldn't slow things down too much. This also prevents a re-render from happening when the controller resets.

Addresses #2843.

### DIFF
--- a/src/js/jsx/DocumentHeader.jsx
+++ b/src/js/jsx/DocumentHeader.jsx
@@ -63,7 +63,8 @@ define(function (require, exports, module) {
                 applicationState = applicationStore.getState(),
                 documentIDs = applicationState.documentIDs,
                 document = applicationStore.getCurrentDocument(),
-                count = applicationStore.getDocumentCount();
+                count = applicationStore.getDocumentCount(),
+                inactiveDocumentsInitialized = applicationState.inactiveDocumentsInitialized;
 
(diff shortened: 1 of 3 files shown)


src/js/jsx/DocumentHeader.jsx
```diff
@@ -107,7 +109,18 @@ define(function (require, exports, module) {
...
+            // the UI doesn't blink unnecessarily.
+            if (this.props.active && !nextProps.active) {
+                return false;
+            }
+
+            if (!nextState.inactiveDocumentsInitialized) {
+                return false;
+            }
+
+            return this.state.inactiveDocumentInitialized !== nextState.inactiveDocumentsInitialized ||
```
mcilroyc: Document vs. Documents ??

mcilroyc:
one type, I think.  Also needs rebase.  back to @iwehrman

src/js/jsx/DocumentHeader.jsx
iwehrman: Derp.

iwehrman:
Typo fixed and rebased. Back to @mcilroyc.

PR #2851, opened 2015-10-07
Redraw on-canvas UI when display configuration changes

iwehrman:
This causes the on-canvas UI (i.e., the `SamplerOverlay` and `SuperselectOverlay`) to be redrawn upon receiving a `displayConfigurationChanged` event using pixel values that are current for the new browser root-size (i.e., with the same rem value but with updated pixel value). These events are emitted when the screen resolution changes or the when the window is moved between displays. Note for testing: Chromium doesn't seem to update it's CSS breakpoints when you just change the screen resolution on a single display. But it should do so, and hence is testable, when dragging the window between displays.

This also fixes the erroneously named and defined `pxToRem` function. In fact, we're converting a rems to pixels. The function was also off by a factor of 1/16, which forced the callers to use rem values inflated by a factor of 16 in order to get reasonable-looking pixel values out. This fixes the function by renaming it, eliminating the 1/16 error factor, and factoring out 1/16 from all the call sites. _dusts hands off_

Addresses #2196.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -526,6 +526,9 @@ define(function (require, exports) {
     zoomInOut.writes = [];
     zoomInOut.transfers = [zoom];
 
+    /**
+     * Emit a DISPLAY_CHANGED event.
+     */
     var handleDisplayConfigurationChanged = function () {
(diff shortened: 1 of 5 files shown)


mcilroyc:
REMarkable work

PR #2852, opened 2015-10-07
Narrow fix for another resetBorderPolicies race

iwehrman:
This is another very narrow fix for another `resetBorderPolicies` race that results in ghostly border policies. In this case, it's between the `layers.select` action and its transfer to `layers.initializeLayers`. Their calls to `resetBorderPolicies` interleave in a bad way, and the border polices are never cleaned up, which causes issues like in #2224. See for yourself:

Nightmare fuel

Fixing this one was easy (`initializeLayers` always calls `resetBorderPolicies` anyway, so I just removed the other, redundant call), but these sure are fun to find! Recall that #2794 intends to make this entire class of bugs go away forever.


mcilroyc:
sho nuf

PR #2856, opened 2015-10-07
Cycle through color formats with up/down in the color picker input.

iwehrman:
What it says! It skips "name" if the current color doesn't have a name. 

Addresses: UX concerns.


designedbyscience:
Useful!

chadrolfs:
Fantastic, love it!

Issue #2855, opened 2015-10-07
Doc titles no longer showing


chadrolfs:
Open several documents

Result: nothing in title bar.  they are still listed under the Window menu

iwehrman:
I probably broke this somehow with #2847.

DivyaPrabhakar:
@chadrolfs happened to me too. I checked and yes, @iwehrman it was your PR

chadrolfs:
looks fixed now with last git pull

iwehrman:
I can't repro this. Are there different steps?

DivyaPrabhakar:
@iwehrman close all documents in Ds and Ds completely. Restart it and then try to create new document.

mcilroyc:
FBNC to @chadrolfs

PR #2857, opened 2015-10-07
Always call initInactiveDocuments after startup

iwehrman:
Always call `initInactiveDocuments` after startup to ensure that the `inactiveDocumentsInitialized` store state is set.

Addresses #2855.


mcilroyc:
I can't actually repro the underlying issue, but this looks ok!

Issue #2859, opened 2015-10-07
Re-entering Mask mode to edit results in free transform error


chadrolfs:
Open an image (target a pixel layer)
hit m to enter Mask Mode
hit r for Rect tool and drag out a shape
hit a to edit vector anchor points and drag an anchor point
hit m to exit Mask mode
hit m to re-enter Mask mode
hit a to edit and click on an anchor point

Result: `Photoshop returned from free transform path modal state with an error`

https://www.dropbox.com/s/51w812lij670ury/Mask-edit-Err.mov?dl=0

pineapplespatula:
exiting the modal state without changing anything results in a PS error. I think it is too late in the cycle to change this in PS, 
i could change the log.error to just a warning, if you would prefer

PR #2862, opened 2015-10-07
this should not be an error

pineapplespatula:
fix for issue #2859


iwehrman:
We should not log conditions that are neither exceptional nor useful to developers. Is this message useful?

pineapplespatula:
it would only be useful as a reminder to fix the Core issue. i created a watson bug 4070280 for that purpose and removed the logged error

iwehrman:
I agree that the log message should be removed, but the branch is now failing the CI tests because the `log` variable is unused.

Issue #2650, opened 2015-09-25
missing dialogs cause policy swap issues


chadrolfs:
![screen shot 2015-09-25 at 4 41 56 pm](https://cloud.githubusercontent.com/assets/8442631/10114510/6a423708-63a6-11e5-9e2a-e9d899593326.png)
![screen shot 2015-09-25 at 4 50 18 pm](https://cloud.githubusercontent.com/assets/8442631/10114511/6a429702-63a6-11e5-80e9-a663cc738f68.png)

mcilroyc:
I think this may have been addressed by #2662.  In any case, I can no repro in 9070a9a43c5d458f416d8cae19f6376c80d9f6d4.  @chadrolfs back to you, or please provide detailed repro steps if you can.

chadrolfs:
re-assigning to you until #2865 is merged

Issue #2776, opened 2015-10-03
Focus: arrow keys move layer on canvas when in Type non-edit mode


chadrolfs:
Open or create a document with a couple of type layers
select them with the Select tool, click in size field and up/down arrow

Result: layer moved on canvas

Works as expect in Type edit mode.

PR #2865, opened 2015-10-07
Don't add/remove arrow keyboard policies on focus

iwehrman:
This removes a fertile source of errors with keyboard policies. Many moons ago @pineapplespatula had to add these so that arrow keys worked correctly in CEF text inputs while in the type modal tool state. This creates a bit of a concurrency nightmare, however, which was never fully addressed. See, e.g., #2650. However, @jfitz-adobe has recently relieved us of the need to install these arrow keyboard policies. This removes the `Focusable` keyboard policies entirely, and thus reduces the number of places where we can hurt ourselves with policy manipulation.

Should fully address #2650. Requires >= Mac pgdev.734, or whatever the latest Windows build was when this PR was filed.

(Also, it would be great to verify this on Windows too.)

### DIFF
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -320,7 +320,7 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var beforeStartup = function () {
-        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
+        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.FOCUS_PROPAGATE,
             keyboardModePromise = this.transfer(setMode, PolicyStore.eventKind.KEYBOARD,
                 defaultKeyboardMode),
             defaultPointerMode = adapterUI.pointerPropagationMode.ALPHA_PROPAGATE,
--- a/src/js/jsx/mixin/Focusable.js
+++ b/src/js/jsx/mixin/Focusable.js
@@ -26,47 +26,14 @@ define(function (require, exports, module) {
 
     var React = require("react");
 
-    var OS = require("adapter/os"),
-        UI = require("adapter/ps/ui"),
-        EventPolicy = require("js/models/eventpolicy"),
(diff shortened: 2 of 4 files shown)


mcilroyc:
I'm on 734, and this branch seems to prevent arrow keys from working in our Datalist (can't navigate up/down) and our NumberInput (can't scale up/down on things like stroke width).  @iwehrman

iwehrman:
Hmm, ok. I definitely tested it with `NumberInput` at least, but I'll test it again and then send it back to @jfitz-adobe.

iwehrman:
I concur: arrow keys now _only_ work when we're in the type modal tool state. Review-only for now.

chadrolfs:
isn't that issue #2776?

iwehrman:
I've pushed another commit which eliminates all instances of arrow key `ALWAYS_PROPAGATE` policies, and instead just changes the default _mode_ to `FOCUS_PROPAGATE`. I'm a little concerned about making such a deep change, but this seems to be the recommendation of @jfitz-adobe. I'm not aware of any problems, but I can't claim to have thoroughly tested the change either. Please test, @chadrolfs!

chadrolfs:
testing...

chadrolfs:
if i'm in text edit mode and enter Appearance Opacity, focus is dropped after the first arrow up or down

chadrolfs:
^^ that happens on master as well

iwehrman:
@chadrolfs re: opacity, DS is almost working "correctly" in this case because PS currently doesn't support changing layer opacity while in a modal text state. When you bump the opacity the first time, the action ends the modal tool state. This seems to return focus to Photoshop, which blurs the input. I'm not actually sure why focus is returned to Photoshop. In the short term, we could make this field read-only while in a modal tool state. I'm not sure if that's preferable though?

iwehrman:
(In any case, I would not call that a high priority issue.)

mcilroyc:
Are we all on board with merging this?

chadrolfs:
I think so.

Issue #2821, opened 2015-10-06
Need to make tracking field wider 


ktaki:
The maximum number allowed is 10,000.

![screen shot 2015-10-05 at 5 26 09 pm copy](https://cloud.githubusercontent.com/assets/2029597/10297381/f11f9954-6b86-11e5-9403-87c2cf477c40.png)
![screen shot 2015-10-05 at 5 26 19 pm copy](https://cloud.githubusercontent.com/assets/2029597/10297382/f3978be2-6b86-11e5-8a60-65148826150b.png)

designedbyscience:
Back for verification

ktaki:
Verified the fix.

Issue #2571, opened 2015-09-22
Overflowing text in non-English languages


iwehrman:
DE:
![image](https://cloud.githubusercontent.com/assets/1075154/10028514/bfa7d6da-6121-11e5-9e6b-05306302701d.png)

FR:
![image](https://cloud.githubusercontent.com/assets/1075154/10028518/c56b566e-6121-11e5-8186-98c637ba71c4.png)

JA:
![image](https://cloud.githubusercontent.com/assets/1075154/10028522/ccb61cc4-6121-11e5-852a-9c8583ca62c9.png)

iwehrman:
@DivyaPrabhakar: can you just try to expand the widths of the inputs _a little bit_ to accommodate the wider strings as shown in the pictures above? It might not be possible for JA. I don't want to make any major changes right now; just small tweaks.

iwehrman:
[Watson 4058099](https://watsonexp.corp.adobe.com/#bug=4058099)

DivyaPrabhakar:
FBNC @iwehrman

iwehrman:
Fixed by #2871 and #2918.

fzhu1978:
![jpn_20151016 m 262](https://cloud.githubusercontent.com/assets/11431959/10615627/c29b07a4-7715-11e5-8a93-13bc6dae27dc.png)
Still have issue on Japanese.

DivyaPrabhakar:
@iwehrman I'll take this again

DivyaPrabhakar:
@fzhu1978 This has be fixed but it is not closed. Passing it back to you so you can verify the fix.

chadrolfs:
Been sitting in FBNC for a while, @ktaki can you verify and close?  Thanks.

jsbache:
Not a core bug. I am removing the ps core tag

ktaki:
Only issue I found is in Japanese Effect panel. With Stroke, Size label has wrapped.

![screen shot 2015-10-26 at 3 08 07 pm](https://cloud.githubusercontent.com/assets/2029597/10744310/39f8bc08-7bf6-11e5-9632-30c9d8fb1ba0.png)

This can be deferred to M4.

ktaki:
Looks good!

PR #2871, opened 2015-10-08
Number input for text values made larger

DivyaPrabhakar:
Regarding issue #2821 

Before: 

After:

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -683,6 +683,7 @@ define(function (require, exports, module) {
                                 disabled={locked}
                                 min={MIN_TRACKING}
                                 max={MAX_TRACKING}
+                                size="column-5"
                                 onChange={this._handleTrackingChange}
                                 valueType="size" />
                         </div>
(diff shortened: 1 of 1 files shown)


iwehrman:
Dumb question: should we change `column-4` to `column-5` on the leading input too? It seems to fit, and it might help with the "aut..." clipping?

iwehrman:
img

DivyaPrabhakar:
@iwehrman I thought of that too in order to help with #2571 but I didn't know if we would prefer for that text-input field to be tight  rather than have the little space at the end, since we are going to do language specific changes any way for other parts of the UI. @designedbyscience or @placegraphichere if you think that little extra space doesn't matter, we can change that to column-5 and also and then have one less language specific change to make.

iwehrman:
As long as we agree that it doesn't look scrunched (I think it's OK), changing both seems better for visual consistency and because it makes the leading translation somewhat less bad.

DivyaPrabhakar:
@iwehrman Sure, I'll make the changes and the solves one of the translation grievances simultaneously so that is good.

designedbyscience:
Great, doesn't change any other alignments

Issue #2849, opened 2015-10-07
Placing CC Lib asset breaks scrim layer highlighting


chadrolfs:
https://www.dropbox.com/s/kty0gmxwvhhwypo/CCLibAsset-breaksLayerHighlight.mov?dl=0

iwehrman:
Weird. Looks like a policy problem. I wonder if @shaoshing's current branch in his open PR #2837 solves this?

shaoshing:
@iwehrman this is not related to the PR. I will look into it.

iwehrman:
FBNC to @chadrolfs.

PR #2874, opened 2015-10-08
Addresses issue that restoring policy will set to incorrect mode

shaoshing:
In some tool handlers, we are using `UI.setPointerPropagationMode` to set pointer mode directly and did not dispatch a `mode changed` event to the policy store, and thus the actual pointer mode is out of sync with the store. This will cause problem if we suspend/restore policy actions as the pointer mode might be restored to a wrong one. This PR replaces `UI.setPointerPropagationMode` calls with `policyActions.setMode` to make sure the policy store is in sync. This will address issue #2849

### DIFF
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -320,29 +320,20 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var beforeStartup = function () {
-        var keyboardMode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
-            keyboardDescriptor = {
-                defaultMode: keyboardMode
-            },
-            keyboardModePromise = adapterUI.setKeyboardPropagationMode(keyboardDescriptor)
-                .bind(this)
-                .then(function () {
-                    this.dispatch(events.policies.MODE_CHANGED, {
-                        kind: PolicyStore.eventKind.KEYBOARD,
-                        mode: keyboardMode
-                    });
-                }),
-            pointerMode = adapterUI.pointerPropagationMode.ALPHA_PROPAGATE,
-            // Alpha is the default pointer mode, so we don't need to change it here
-            pointerModePromise = this.dispatchAsync(events.policies.MODE_CHANGED, {
-                kind: PolicyStore.eventKind.POINTER,
-                mode: pointerMode
-            });
+        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
+            keyboardModePromise = this.transfer(setMode, PolicyStore.eventKind.KEYBOARD,
+                defaultKeyboardMode),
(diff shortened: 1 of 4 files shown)


src/js/actions/tools.js
```diff
@@ -418,7 +418,7 @@ define(function (require, exports) {
     selectTool.reads = [];
     selectTool.writes = [locks.JS_TOOL, locks.PS_TOOL];
     selectTool.transfers = [resetBorderPolicies, policy.removePointerPolicies, installShapeDefaults,
-        policy.removeKeyboardPolicies, policy.addPointerPolicies, policy.addKeyboardPolicies,
+        policy.removeKeyboardPolicies, policy.addPointerPolicies, policy.addKeyboardPolicies, policy.setMode,
```
iwehrman: It doesn't look like this action is actually transferring to `policy.setMode`? Should it be?

src/js/actions/policy.js
```diff
@@ -320,29 +320,21 @@ define(function (require, exports) {
...
-                        kind: PolicyStore.eventKind.KEYBOARD,
-                        mode: keyboardMode
-                    });
-                }),
-            pointerMode = adapterUI.pointerPropagationMode.ALPHA_PROPAGATE,
+        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
+            keyboardModePromise = this.transfer(setMode, PolicyStore.eventKind.KEYBOARD,
+                defaultKeyboardMode),
+            defaultPointerMode = adapterUI.pointerPropagationMode.ALPHA_PROPAGATE,
             // Alpha is the default pointer mode, so we don't need to change it here
```
iwehrman: I think when I wrote this I didn't realize we were actually changing the mode dynamically. I wonder if we should set it here even though it is the "default" in case, e.g., there is a controller error and we reset while in a non-default mode. What do you think?

iwehrman:
Great catch here, @shaoshing. One question remaining about a dangling transfer declaration, and one question about a possible improvement for additional robustness. Back to you!

src/js/actions/tools.js
shaoshing: yes, it is required; I triggered an error of not doing so and then realized that `selectTool.transfers` is actually a superset of actions, including those used in each tool's `selectHandler` and `deselectHandler`. For example, `policy.setMode`, `policy.addKeyboardPolicies`, `shortcuts.addShortcut`

shaoshing:
Hi @iwehrman I've made the change to set the default pointer mode in the case we need to reset the actions, and anwersered your question about the `setMode` transfer, so back to you.

src/js/actions/tools.js
iwehrman: Ah, yes. The problem is that the tool select handlers aren't actually actions. They should be, and we should transfer to them. But, that's a problem for a different day. Thanks for the explanation.

Issue #2861, opened 2015-10-07
CC Libraries asset names do not scale with UI


iwehrman:
Here's the UI with a root font-size of 43%:
![image](https://cloud.githubusercontent.com/assets/1075154/10351418/4be74210-6cfe-11e5-8b72-d120d7b85f6a.png)

Here's the UI with a root font-size of 100%:
![image](https://cloud.githubusercontent.com/assets/1075154/10351407/46f3c10c-6cfe-11e5-9bf2-0a6a644c07d7.png)

iwehrman:
Marking this as FIX NOW because it will block any future work to correct the UI size on SP3. See: #2235.

chadrolfs:
:+1:

chadrolfs:
Eric can you take this?  If not, please pass along to Shaoshing.  Thanks!

iwehrman:
Fixed by #2875.

PR #2875, opened 2015-10-08
Added specific font size in rems to library title inputs

designedbyscience:
So now it will change based on the `<html>` font size

Addresses #2861


iwehrman:
Looks great!

Issue #2809, opened 2015-10-05
Vector Mask mode adds weird policy area that is never removed


iwehrman:
I always get the following strange green policy area when entering vector mask mode:
![image](https://cloud.githubusercontent.com/assets/1075154/10295358/13180274-6b75-11e5-8148-c9bf48ae4b6e.png)

iwehrman:
@chadrolfs: can you please verify that there are no longer any odd policy rectangles when bouncing in to and out of mask mode?

PR #2879, opened 2015-10-08
fix bounds of pointer policy for vector mask mode

pineapplespatula:
This PR modifies the pointer policies associated with vector mask mode to refresh on uiStore change events. and to use the correct bounds.

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -45,6 +45,7 @@ define(function (require, exports) {
         policy = require("./policy"),
         ui = require("./ui"),
         shortcuts = require("./shortcuts"),
+        synchronization = require("js/util/synchronization"),
         system = require("js/util/system"),
         utilShortcuts = require("js/util/shortcuts"),
         EventPolicy = require("js/models/eventpolicy"),
@@ -516,8 +517,32 @@ define(function (require, exports) {
      * @type {function()}
      */
     var _toolModalStateChangedHandler,
-        _vectorMaskHandler;
+        _vectorMaskHandler,
+        _vectorMaskPolicyHandler;
     
+    /**
+     * helper function to determine cloaking bounds. 
+     *
+     * @private
+     * @return {{
+     *             x: number, 
+     *             y: number, 
+     *             width: number,
+     *             height: number
+     *         }}
+     */
+    var _getVectorMaskPolicyBounds = function () {
+        var uiStore = this.flux.store("ui"),
+            cloakRect = uiStore.getCloakRect();
+        
+        return {
+            x: cloakRect.left,
+            y: cloakRect.top,
+            width: cloakRect.right - cloakRect.left,
+            height: cloakRect.bottom - cloakRect.top
+        };
+    };
+
     /**
      * Change the tool's vector mask mode
      * This function will both dispatch a vector_mask_mode_changed event for the store
@@ -529,7 +554,6 @@ define(function (require, exports) {
     var changeVectorMaskMode = function (vectorMaskMode) {
         var flux = this.flux,
             toolStore = flux.store("tool"),
-            uiStore = flux.store("ui"),
             prevVectorMode = toolStore.getVectorMode(),
             currentTool = toolStore.getCurrentTool(),
             firstLaunch = true;
@@ -595,13 +619,7 @@ define(function (require, exports) {
         }
          
         if (vectorMaskMode) {
-            var centerOffsets = uiStore.getCenterOffsets(),
-                area = {
-                    x: centerOffsets.left,
-                    y: centerOffsets.top,
-                    width: centerOffsets.right - centerOffsets.left,
-                    height: centerOffsets.top - centerOffsets.bottom
-                },
+            var area = _getVectorMaskPolicyBounds.call(this),
                 pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
                     OS.eventKind.LEFT_MOUSE_DOWN,
                     {},
@@ -735,6 +753,7 @@ define(function (require, exports) {
     var beforeStartup = function () {
         var flux = this.flux,
             toolStore = this.flux.store("tool"),
+            uiStore = this.flux.store("ui"),
             tools = toolStore.getAllTools();
 
         // Listen for modal tool state entry/exit events
@@ -779,8 +798,7 @@ define(function (require, exports) {
         }.bind(this), []);
     
         _vectorMaskHandler = function () {
-            var toolStore = this.flux.store("tool"),
-                vectorMode = toolStore.getVectorMode() || false;
+            var vectorMode = toolStore.getVectorMode();
             
             this.flux.actions.tools.changeVectorMaskMode(!vectorMode);
         }.bind(this);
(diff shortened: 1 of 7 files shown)


src/js/actions/tools.js
```diff
@@ -791,6 +792,39 @@ define(function (require, exports) {
             fn: _vectorMaskHandler
         });
 
+        _vectorMaskPolicyHandler = function () {
```
pineapplespatula: I dont know what to debounce here to make this get executed so frequently

src/js/actions/tools.js
```diff
@@ -791,6 +792,39 @@ define(function (require, exports) {
...
+            var vectorMaskMode = toolStore.getVectorMode() || false;
+
+            if (vectorMaskMode) {
+                var pointerPolicyID = toolStore.getVectorMaskPolicyID();
+         
+                if (pointerPolicyID) {
+                    return this.transfer(policy.removePointerPolicies, pointerPolicyID)
+                        .bind(this)
+                        .then(function () {
+                            var cloakRect = uiStore.getCloakRect(),
```
iwehrman: Fix the tabbing here. Also, can you factor this out into a private helper function that is usable both here and above? Something like `_getVectorMaskPolicyBounds`?

src/js/actions/tools.js
iwehrman: The function itself is what you might debounce, to get a new function which is then made the UIStore change event handler.

src/js/actions/tools.js
```diff
@@ -779,8 +781,7 @@ define(function (require, exports) {
         }.bind(this), []);
     
         _vectorMaskHandler = function () {
-            var toolStore = this.flux.store("tool"),
-                vectorMode = toolStore.getVectorMode() || false;
+            var vectorMode = toolStore.getVectorMode() || false;
```
iwehrman: Instead of doing `|| false` here it would be better to change the tool store so that the default value is `false` instead of `null`.

src/js/actions/tools.js
```diff
@@ -791,6 +792,39 @@ define(function (require, exports) {
             fn: _vectorMaskHandler
         });
 
+        _vectorMaskPolicyHandler = function () {
+            var vectorMaskMode = toolStore.getVectorMode() || false;
```
iwehrman: And here w.r.t. `|| false`.

iwehrman:
I'm out until Monday, so please bounce this over to @mcilroyc when it's ready for another look.

pineapplespatula:
@mcilroyc  can you take a look at these pointer issues?

src/js/actions/tools.js
```diff
@@ -790,6 +808,35 @@ define(function (require, exports) {
...
+                            return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
+                        })
+                        .then(function (policyID) {
+                            this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
+                        });
+                }
+            } else {
+                return Promise.resolve();
+            }
+        }.bind(this), this, DEBOUNCE_DELAY, false);
```
mcilroyc: FYI I don't think you need this `.bind(this)` since the second argument to `debounce` does it for you.

src/js/actions/tools.js
```diff
@@ -790,6 +808,35 @@ define(function (require, exports) {
             modifiers: {},
             fn: _vectorMaskHandler
         });
+        var DEBOUNCE_DELAY = 200;
```
mcilroyc: missing newline.  And also this is missing JS Doc.

mcilroyc:
can you please fill out the description of the PR with some info, and in JSDoc about what the heck _vectorMaskPolicyHandler does, etc.

pineapplespatula:
@mcilroyc back to you

src/js/actions/tools.js
```diff
@@ -790,6 +808,37 @@ define(function (require, exports) {
...
+
+        // recreate pointer policies if we're in Vector Mask Mode on UI Store change events
+        _vectorMaskPolicyHandler = synchronization.debounce(function () {
+            var vectorMaskMode = toolStore.getVectorMode();
+
+            if (vectorMaskMode) {
+                var pointerPolicyID = toolStore.getVectorMaskPolicyID();
```
mcilroyc: I believe this handler should call a flux action instead of transferring.  I would suggest one `refreshVectorPolicy` action that would handle the transfers to the the remove/add policy actions.  Part of this is just that "it feels gross" to transfer from `beforeStartup` on a listener.  Paging Dr @iwehrman for a second opinion.

src/js/actions/tools.js
mcilroyc: this still looks weird being pinned to shortcutSpecs's tuches

mcilroyc:
assigning this back to @pineapplespatula because it has the review only label on it.  I feel pretty sure that I don't like the way the handler is structured (see prev comment), but I also can't make a strong argument so it doesn't necessarily need to block the PR.

src/js/actions/tools.js
iwehrman: I think it would be preferable for the entire function to be an action, which then performs these transfers. What should be debounced is the call to the `handleVectorMaskModeChange` action.

pineapplespatula:
@iwehrman  back to you

src/js/actions/tools.js
```diff
@@ -727,6 +745,43 @@ define(function (require, exports) {
     enterPathModalState.modal = true;
 
     /**
+     * recreate pointer policies if we're in Vector Mask Mode on UI Store change events
+     *
+     * return {Promise}
+     */
+    var handleVectorMaskUIChange = function () {
+        var toolStore = this.flux.store("tool"),
```
iwehrman: You're reading state from the tool store here, so you need at least a `JS_TOOL` read lock.

src/js/actions/tools.js
```diff
@@ -727,6 +745,43 @@ define(function (require, exports) {
...
+                        var area = _getVectorMaskPolicyBounds.call(this),
+                            pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
+                                OS.eventKind.LEFT_MOUSE_DOWN,
+                                {},
+                                area);
+                    
+                        return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
+                    })
+                    .then(function (policyID) {
+                        this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
```
iwehrman: I think this is writing tool store state, so actually you need the `JS_TOOL` write lock.

iwehrman: (And, given that you're adding the write lock, there's no need to also add the read lock.)

iwehrman:
Missing lock. Back to @pineapplespatula. Also, can you describe how I can verify this fix? Does it address a particular issue?

pineapplespatula:
It addresses the green box issue with policy frames showing

iwehrman:
Please reference the issue number in your PR summary so that whoever merges it knows how to verify the fix, and so that he or she can go back and mark the original issue as FBNC. 

FFR: addresses #2809.

PR #2881, opened 2015-10-09
"move" event bug fixes

mcilroyc:
The primary bug this addresses is #1613, but there are a few other possibly-unreported bugs that are fixed.

## 1
When we option-dragging multiple artboards, we get two move events.  The debouncing of the event handler was causing us to NOT process the fancy event that notifies us of the duplication.  So, we weren't adding the layers to the model...and then things blew up downstream.  I rearranged the handler to only debounce the "regular" move events.  I'll be honest, I don't know why this needs to be debounced at all, but I wanted to make the most narrow change I could.

## 2
I essentially undid commit 529eb5e34a667ad3aa043a18bfc5ebae02ba309f from barkin (via PR #2672) which I can only assume means that the underlying core (events) behavior has changed.  "move" events were being ignored if it includes a truthy `trackerEndedWithoutBreakingHysteresis` AND if there are no artboards selected.  This resulted in "double counting" events when artboards were selected... which was causing a phantom history state (bad!).  I removed the latter requirement, and I feel like it is doing the right thing now.

## 3a
Fixed a bug that was causing improper model caching in our history store when moving text layers.  I fixed that, AND made it safe for moving mixture of txt and non-txt layers.  After both the resetLayers and resetBounds action complete, we dispatch a single history "finalization" event.  I think this is a step in the right direction towards a future of clearly specified history transactions (ie, not burying them inside resetLayers or resetBounds).

## 3b
Added a call to resetBorderPolicies at the end of addLayers action.  This fixed  a bug with option-drag not updating bounds around the newly created duplicate layer(s)

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -408,11 +408,15 @@ define(function (require, exports) {
                     naiveLayerCount = document.layers.all.size + layerSpec.length;
 
                 return nextLayerCount !== naiveLayerCount;
+            })
+            .tap(function () {
+                return this.transfer(tools.resetBorderPolicies);
             });
     };
     addLayers.modal = true;
     addLayers.reads = [locks.PS_DOC];
     addLayers.writes = [locks.JS_DOC];
+    addLayers.transfers = [tools.resetBorderPolicies];
     addLayers.post = [_verifyLayerIndex, _verifyLayerSelection];
 
     /**
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1103,9 +1103,22 @@ define(function (require, exports) {
                         otherLayers = currentDoc.layers.allSelected.filterNot(function (layer) {
                             return layer.kind === layer.layerKinds.TEXT;
                         }),
+
+                        // note that in this case, the debouncing is critical even for just one "move" event
+                        // because the historyState event must be processed first for the following
+                        // "amend history" workflow to function correctly
                         textLayersPromise = this.flux.actions.layers.resetLayers(currentDoc, textLayers),
-                        otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers);
-                    return Promise.join(textLayersPromise, otherLayersPromise);
+                        otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers, true);
+
+                    // When moving a mixture of both text and non-text layers, the individual actions do not
+                    // affect history.  Instead, a single, separate event is dispatched to provide a unified
+                    // finalization of this history transaction with the updated model.
+                    return Promise.join(textLayersPromise, otherLayersPromise)
+                        .bind(this)
+                        .then(function () {
+                            return this.dispatchAsync(events.history.FINALIZE_HISTORY_STATE,
+                                { documentID: currentDoc.id });
+                        });
                 }
             }, this, 200);
 
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1092,14 +1092,8 @@ define(function (require, exports) {
 
             // Handle the normal move events with a debounced function
             var debouncedMoveHandler = synchronization.debounce(function () {
-                var artboardsSelected = currentDoc.layers.selected.some(function (layer) {
-                    return layer.isArtboard;
-                });
-
-                // If it was a simple click/didn't move anything, there is no need to update bounds
-                // But, if we're moving multiple artboards, we get one event with 0,0, which means
-                // we hit this and we have to update bounds
-                if (event.trackerEndedWithoutBreakingHysteresis && !artboardsSelected) {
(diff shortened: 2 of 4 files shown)


mcilroyc:
Assigning to @iwehrman since we've discussed this a bit.  I know that @iwehrman may not get to this immediately.  Additionally, @shaoshing you may want to look at the scope of this PR in case there is overlap with anything you're working on.

src/js/actions/transform.js
```diff
@@ -1084,22 +1084,45 @@ define(function (require, exports) {
...
-            // we hit this and we have to update bounds
-            if (event.trackerEndedWithoutBreakingHysteresis && !artboardsSelected) {
-                return Promise.resolve();
-            }
+                currentDoc = appStore.getCurrentDocument();
 
+            // Handle the normal move events with a debounced function
+            var debouncedMoveHandler = synchronization.debounce(function () {
+                // short circuit based on this trackerEndedWithoutBreakingHysteresis event flag
+                if (event.trackerEndedWithoutBreakingHysteresis) {
```
iwehrman: Can you explain the semantics of this property here? For posterity, if nothing else?

src/js/actions/transform.js
```diff
@@ -1112,19 +1135,9 @@ define(function (require, exports) {
...
-                    textLayersPromise = this.flux.actions.layers.resetLayers(currentDoc, textLayers),
-                    otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers);
-
-                return Promise.join(textLayersPromise, otherLayersPromise);
+                return debouncedMoveHandler();
             }
-        }, this);
+        }.bind(this);
 
         _moveToArtboardHandler = synchronization.debounce(function () {
```
iwehrman: We seemed to agree yesterday that `debounce` with 0 timeout was weird at best. Does this seem OK to you?

src/js/actions/transform.js
```diff
@@ -1084,22 +1084,45 @@ define(function (require, exports) {
...
+                        // note that in this case, the debouncing is critical even for just one "move" event
+                        // because the historyState event must be processed first for the following
+                        // "amend history" workflow to function correctly
+                        textLayersPromise = this.flux.actions.layers.resetLayers(currentDoc, textLayers),
+                        otherLayersPromise = this.flux.actions.layers.resetBounds(currentDoc, otherLayers, true);
+
+                    // When moving a mixture of both text and non-text layers, the individual actions do not
+                    // affect history.  Instead, a single, separate event is dispatched to provide a unified
+                    // finalization of this history transaction with the updated model.
+                    return Promise.join(textLayersPromise, otherLayersPromise)
```
iwehrman: In the future we may want a transaction-like infrastructure for history. In the meantime, this seems great.

src/js/actions/layers.js
```diff
@@ -408,11 +408,15 @@ define(function (require, exports) {
                     naiveLayerCount = document.layers.all.size + layerSpec.length;
 
                 return nextLayerCount !== naiveLayerCount;
+            })
+            .tap(function () {
+                return this.transfer(tools.resetBorderPolicies);
```
iwehrman: This seems correct to me.

src/js/actions/transform.js
iwehrman: Actually, can you go even further and leave a comment that describes what events you anticipate, in what order, and what they all mean?

mcilroyc: no, I have absolutely no idea what it is.  I'm hoping to find out more from JP, and will elaborate the comment if/when I do.

src/js/actions/transform.js
mcilroyc: yes, that's why I used 200 as the delay for this one.  That seemed to be the number used elsewhere.  I considered making it a global constant (like, in global.js) but I'll at least make it a const in this module.

mcilroyc: OH ignore above.  I see you were commenting on the MTA handler.  I do not think it makes sense as-is, but I also haven't noticed any problems during testing.  I think the result is possibly "under debouncing" which results in "over resetIndexing".

iwehrman: OK, seems fine to leave it for now then. Don't wake the beast.

iwehrman:
When I select two artboards (each with two rectangles inside) and option-drag copy them, I get:
`Verification of postcondition 1 failed for action layers.addLayers`, which is `_verifyLayerSelection`. This might be spurious however (since we're processing the events out of sync with the action?) because the selection looks OK to me.

Other than that, I can't break it. So, besides some requests for additional in-line documentation and a question about a different debounced handler, I'm basically satisfied for the time being. Also, to reiterate our discussion off-line: what need for a longer-term solution is to articulate a core request for how all these notifications ought to work, or at least how they could be changed to be significantly less broken.

iwehrman:
@shaoshing: I think this looks OK, but you might want to take a once-over too.

iwehrman:
Let's give it a whirl!

Issue #2864, opened 2015-10-07
CC Lib: force Create/Cancel when creating a new library


chadrolfs:
new doc
drag out some rects on canvas
click on Libraries drop down > Create New Library
give it a name and hit Enter twice, then hit Delete or drag an asset onto the canvas

Result: you're now in a state where you can continue working on your document, switch tools, etc, but you still haven't clicked Cancel or Create yet and you can continue to use assets for a library that is still active in the background.

mcilroyc:
@chadrolfs , @shaoshing claims this is fixed by his PR 2882

PR #2882, opened 2015-10-09
CC Libraries - tweaks for library creation and rename

shaoshing:
This PR introduces the following changes:

* During library creation, the current library's content will be hidden, to reduce confusion and give user clear instruction on where to input the library name. - addresses #2197

* Hitting return / esc can confirm / cancel the current library command (creation or rename). - addresses #2864 

* Handle edge case that asset may have empty representation and will cause exception.

### DIFF
--- a/src/js/jsx/sections/libraries/LibraryList.jsx
+++ b/src/js/jsx/sections/libraries/LibraryList.jsx
@@ -26,6 +26,7 @@ define(function (require, exports, module) {
 
     var React = require("react"),
         _ = require("lodash"),
+        Promise = require("bluebird"),
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React);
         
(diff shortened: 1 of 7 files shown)


src/js/jsx/shared/TextInput.jsx
```diff
@@ -138,6 +138,15 @@ define(function (require, exports, module) {
                 });
             }
         },
+        
+        /**
+         * Return the text input's current value
+         * 
+         * @return {string}
+         */
+        getValue: function () {
```
shaoshing: Added getter function `getValue` so that the parent component does not need to listen to its change event or keep a copy of the current value.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -82,39 +83,40 @@ define(function (require, exports, module) {
...
                 window.setTimeout(function () {
-                    renameInput.acquireFocus();
-                    renameInput.getDOMNode().focus();
-                }, 250);
+                    var libraryNameInput = this.refs.libraryNameInput;
+                    
+                    // Check the existence of the input again, in case it is unmounted after the timeout.
+                    if (libraryNameInput) {
+                        libraryNameInput.acquireFocus();
+                        libraryNameInput.getDOMNode().focus();
```
iwehrman: Crazy idea: what if you set up an `onBlur` handler here to handle the case where someone starts a rename/create operation but then wanders off to do something else in another part of the UI?

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -182,52 +184,61 @@ define(function (require, exports, module) {
...
-                this.setState({ command: null });
+        _handleLibraryNameInputKeydown: function (event) {
+            if (event.key === "Return" || event.key === "Enter") {
+                this._handleConfirmCommand();
+            } else if (event.key === "Escape") {
+                this._handleCancelCommand();
             }
         },
-
+
```
mcilroyc: et tu, shaoshing?

src/js/util/libraries.js
```diff
@@ -139,14 +139,14 @@ define(function (require, exports) {
...
-        return libraryActions.EDITABLE_GRAPHIC_REPRESENTATION_TYPES.has(representation.type);
+        return !!representation && libraryActions.EDITABLE_GRAPHIC_REPRESENTATION_TYPES.has(representation.type);
     };
     
     /**
      * Return the file extension of the element's primary representation.
      * 
      * @param {AdobeLibraryElement} element
-     * @return {?string}
+     * @return {string|undefined}
```
mcilroyc: I think the preferred syntax is `string=`

mcilroyc:
Looks good to me (only had some nits).  If you'd like to consider Ian's crazy idea (onBlur), then do so elsewhere.  This PR's headed out the door.

Issue #2558, opened 2015-09-21
New Export button on the doc title bar can be overwritten by doc names


ktaki:
Any other better place to be?

mcilroyc:
assigned to @DivyaPrabhakar , cc: @designedbyscience

iwehrman:
I don't see this problem any longer. FBNC to @ktaki.

ktaki:
This may be a minor issue, bug I see some unnatural behavior of showing document titles on the document bar. Try Cmd+n several times. The font size becomes smaller suddenly when reached to the middle of the bar with a half of the space still left on the bar. Can that be wait till the doc titles reaches near the end of the bar?

ktaki:
See the picture below. 8 document open and the last doc is selected, but it does not show full document name, in this case, "Untitled 8".

![screen shot 2015-10-12 at 1 27 58 pm](https://cloud.githubusercontent.com/assets/2029597/10437928/5506911c-70e5-11e5-819f-f2a1d1fea430.png)

chadrolfs:
Removing FIX NOW since this is a positional issue and issue as it was originally written no longer occurs (Export icon is no longer being overwritten by doc name).  The issue here now is that the last doc name is under the header icons, but clicking on it can correct.

ktaki:
No, clicking on it won't correct it, but I agree this is to be fixed by M3, not for UI F.

DivyaPrabhakar:
@ktaki @chadrolfs Isn't is supposed to go behind the icons because we agreed that we would have users scroll the document header if there are too many titles? What is the ideal case -- have all the document tabs resize so that they are all visible at once in the documents header and have no scroll?

chadrolfs:
I think the main issue right now is that they don't scroll well or fluidly and can get stuck, at least via mouse move or clicking on the 1/2 exposed title, especially on windows.  The full name should be exposed for the document that has focus IMO.

ktaki:
Chad is right. With exceptions of unusual long names, the doc bar should display full name when the doc is selected.

designedbyscience:
Back to @ktaki for verification

ktaki:
Looks good!!

PR #2883, opened 2015-10-09
Fine tune the document header resizing behavior

shaoshing:
This is related to issue #2558. Taki is not happy that the document header got shrinked in the middle. He is wondering if it is possible to trigger the shrinking only when the container is out of space. I couldn't come up with a easy solution for this (it requires some extra work to detect if whether or not the header will be out of space), so I adjusted its behavior to use smaller text when reach maximum header count, where the count is dynamically calculated based on the current width of the container. This looks good to me, at least the user won't be surprised that the text is shrinked in the middle.

Here is the before and after comparision: 

#### Before

tabs-resizing

#### After

new-tabs-resizing

### DIFF
--- a/src/js/jsx/DocumentHeader.jsx
+++ b/src/js/jsx/DocumentHeader.jsx
@@ -42,7 +42,7 @@ define(function (require, exports, module) {
         exportStore = require("js/stores/export");
 
     var DocumentHeader = React.createClass({
-        mixins: [FluxMixin, StoreWatchMixin("application", "document", "tool", "dialog")],
+        mixins: [FluxMixin, StoreWatchMixin("application", "document", "tool", "dialog", "preferences")],
 
         getInitialState: function () {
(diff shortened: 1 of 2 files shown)


shaoshing:
Added discussion label so we can compare different shrink strategy.

shaoshing:
Eric pointed out that this is not working well in non-retina screen, so I tried a different way to address both. I render two tab containers. one of them is invisible to the users and always using regualr tab size. Then I detect whether it has enough room to show all title completely. If not, then I re-render the second tab container in small tab size. This seems nicely fix the problem.

#### OSX

new-tabs2

#### WIN

new-tabs-win

shaoshing:
This is ready for merging if Erice approve the implementation using a invisible tab container and is happy about the result. :D

designedbyscience:
So Much Better.

Issue #2887, opened 2015-10-09
Opening a file without fixing broken link to Linked Smart Object errors and app unresponsive.


chadrolfs:
Open file with a broken Linked Smart Object (possibly a file saved from a different machine)
note: should force dialog - Cannot locate linked assets:
click Cancel

Result: UI blinks and app becomes unresponsive.

`Transfer from documents.allocateDocument to documents.updateDocument failed: TypeError: Expected [K, V] tuple: M`

chadrolfs:
<img width="917" alt="screen shot 2015-10-09 at 11 27 18 am" src="https://cloud.githubusercontent.com/assets/8442631/10402307/78d4af0c-6e79-11e5-9c3d-777fc9d45d68.png">
![screen shot 2015-10-09 at 11 30 26 am](https://cloud.githubusercontent.com/assets/8442631/10402306/78d3d7ee-6e79-11e5-988a-611df16def24.png)

chadrolfs:
using latest dev build 743 and master @ 20f8742f7188333db70ddac6688838d118723917

mcilroyc:
CNR.  back to @chadrolfs for extra guidance on repro.  Please also try branch mcilroy/issue-2887 to see if it fixes the issue.

chadrolfs:
what i get on your branch now:

![screen shot 2015-10-09 at 1 02 02 pm](https://cloud.githubusercontent.com/assets/8442631/10404332/c487fe60-6e86-11e5-9947-cbf70d0e2e80.png)

chadrolfs:
@ktaki or @shaoshing can you repro this?  sending you a test file...

shaoshing:
@chadrolfs @mcilroyc  I can reproduce with the test file. Running on dev.743 and latest master.

shaoshing:
Except for `initActiveDocument`, I am also getting two `Verification of postcondition 0 failed for action documents.open` errors.

<img width="680" alt="screen shot 2015-10-09 at 1 24 33 pm" src="https://cloud.githubusercontent.com/assets/118264/10404688/1cbb7fd8-6e89-11e5-931b-79858a0619cf.png">


<img width="682" alt="screen shot 2015-10-09 at 1 23 17 pm" src="https://cloud.githubusercontent.com/assets/118264/10404656/f22ae7ae-6e88-11e5-9a6f-b435790a4333.png">

chadrolfs:
Thanks @shaoshing!

[...]

PR #2888, opened 2015-10-09
Deal with smart objects with broken links

mcilroyc:
It seems that if a linked smart object has a broken link, the descriptor's `linked` property is still *true* and its `link` contains a *string* instead of the fully formed *object*.

This PR contains a very narrow fix for this scenario, and I'm sure it could probably use some further clean-up.  It feels like there's a better way to model the smart object properties.  But, that would take some digging, and this is a :red_circle: Fix Now bug.  Addresses #2887


chadrolfs:
verified fix works on this branch.  _merge me_  :grinning:

pineapplespatula:
LGTM

Issue #2639, opened 2015-09-25
Toolbar UI bounds are off by 1px on retina


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/10110649/6cfd80ce-6383-11e5-9702-429392ccdc08.png)

chadrolfs:
grabbing to verify

chadrolfs:
If #2264 is addressed, that could impact this.  As of now, this is fixed.

PR #2891, opened 2015-10-09
Small change to the toolbar width

DivyaPrabhakar:
Addresses issue #2639.

### DIFF
--- a/src/style/toolbar.less
+++ b/src/style/toolbar.less
@@ -35,7 +35,7 @@
 .toolbar {
     box-sizing: border-box;
     z-index: 1;
-    width: @toolbar-button-size;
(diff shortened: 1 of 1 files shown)


src/style/toolbar.less
```diff
@@ -35,7 +35,7 @@
 .toolbar {
     box-sizing: border-box;
     z-index: 1;
-    width: @toolbar-button-size;
+    width: @toolbar-button-size + .2rem;
```
mcilroyc: `.2` here and then `0.2` the line below.  I guess we can let it slide this time ; )

DivyaPrabhakar: @mcilroyc Never! Change coming up

Issue #2804, opened 2015-10-05
Typeface datalist field is empty for missing fonts


chadrolfs:
Open a file with missing fonts and select one of those layers in the layers panel

Result: Typeface field is empty
Expected: `missing`

![screen shot 2015-10-05 at 1 55 14 pm](https://cloud.githubusercontent.com/assets/8442631/10293344/db2921a6-6b68-11e5-8775-b16ac035115f.png)

ktaki:
We should show the font name with an indication of the font is missing. In standard Ps, the font name is surrounded by [ ].

designedbyscience:
FBNC over to @chadrolfs

PR #2894, opened 2015-10-10
Missing font now says [missing] instead of blank

DivyaPrabhakar:
References issue #2804 

Now looks like this:

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -463,11 +463,15 @@ define(function (require, exports, module) {
 
             // The typeface family name and style for display in the UI
             var familyName,
-                styleTitle;
+                styleTitle,
+                placeholderText;
 
             if (this.props.uniformLayerKind && !postScriptNames.isEmpty() && this.state.initialized) {
                 if (postScriptFamilyName) {
                     familyName = this._getPostScriptFontFamily(postScriptFamilyName);
+                    if (!familyName) {
+                        placeholderText = "[" + postScriptFamilyName + "]";
+                    }
                     
                     if (postScriptStyleName) {
                         styleTitle = this._getPostScriptFontStyle(postScriptStyleName);
@@ -602,7 +606,7 @@ define(function (require, exports, module) {
             return (
                 <div ref="type">
                     {fillBlendFormline}
-                    <div className="formline" >
+                    <div className="formline">
                         <Datalist
                             className="dialog-type-typefaces"
                             sorted={true}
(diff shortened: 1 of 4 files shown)


src/js/jsx/sections/style/Type.jsx
```diff
@@ -625,17 +625,17 @@ define(function (require, exports, module) {
...
+                                className="dialog-type-weights"
+                                sorted={true}
+                                title={styleTitle}
+                                list={weightListID}
+                                disabled={locked || !styleTitle}
+                                value={styleTitle}
+                                defaultSelected={postScriptFamilyName}
+                                options={familyFontOptions}
+                                onChange={this._handleTypefaceChange}
+                                size="weights column-22" />
```
designedbyscience: Could this be added via className rather than size (which theoretically is just for grid sizing)?

src/style/panel.less
```diff
@@ -332,6 +332,14 @@
     width: 28.8rem;
 }
 
+.column-full::-webkit-input-placeholder {
+    font-style: italic;
+}
+
+.weights::-webkit-input-placeholder {
```
designedbyscience: This also needs a more descriptive class name

src/js/jsx/shared/Datalist.jsx
```diff
@@ -90,7 +90,7 @@ define(function (require, exports, module) {
                 defaultSelected: null,
                 live: true,
                 startFocused: false,
-                placeholderText: "",
+                placeholderText: "[missing font]",
```
designedbyscience: This needs to be localized string

iwehrman: Also, this is in a shared component, so it needs a new prop to generalize it otherwise all datalists will have "missing font" placeholder text.

iwehrman:
I'm swiping this since @DivyaPrabhakar is out.

src/nls/root/strings.json
```diff
@@ -423,7 +423,7 @@
             "LINE": "Line",
             "ALIGN": "Align",
             "AUTO_LEADING": "auto",
-            "MISSING": "missing",
+            "MISSING": "[Missing Font]",
```
iwehrman: This is not actually used.

iwehrman:
OK, I rebased and pushed another commit. Now the typeface datalist shows `[MissingPostscriptName]` as the placeholder when the font is missing. The placeholder text is not italicised. The type weight datalist is disabled and empty in this case. 

Note that in some PSDs, I'm having trouble changing the type face of layers with missing fonts. But, it's happening to me on master too, so I think it's an unrelated bug. I'll file a separate issue for that.

Back to @designedbyscience and/or @DivyaPrabhakar.

designedbyscience:
Looks good

PR #2895, opened 2015-10-10
Adjusted toolbar elements to account for retina display fix

DivyaPrabhakar:
After the fix for issue #2639 got merged with PR #2891, @designedbyscience noticed the icons in the toolbar were off. This is the fix for that -- made a new less variable with the +0.2rem change and changed all the widths in toolbar.less to be of that width.


designedbyscience:
Looks good

PR #2896, opened 2015-10-11
KBSC Dialog updates

placegraphichere:
- added sampler modifiers, search, mask mode and export
- sampler shift click/spacebar
- icons for tools, search, mask mode and export
- small changes to spacing for titles


iwehrman:
Looks great.

PR #2897, opened 2015-10-11
Colors for B3 color stop

placegraphichere:
This is the second darkest of the 4 stops. It is also the “default”
standard Ps color stop.

- bg app color


iwehrman:
LGTM. Other fixes can happen later in the day.

PR #2898, opened 2015-10-11
Document Bar Tool Icons resized

placegraphichere:
- needs to be 16x16
- app search and maskmode icons


designedbyscience:
Look good

Issue #2749, opened 2015-10-01
Intro: Next page and prev page navigation


ktaki:
Currently, clicking right side of the window goes to the next page and clicking left side goes back to the previous page. This creates a problem if user clicks slightly left side of Continue button located at the center of the window and clicking again without moving a mouse on the next page. It goes back to previous screen.

Proposal: Since most of the time user wants to go forward, it might be make more sense if clicking any part of the window above the horizontal line goes forward, and clicking left arrow icon at the upper-left corner goes back to previous page. The icon glows when mouseover. It is obvious when you click the icon, you can go back.

ktaki:
IMHO, the left arrow is for going back to the previous screen. If we prefer to go back to the beginning, the icon should look like |< . Anyway, I think this is good enough to go with 16.1. The improvements are due for 17.0. Closing.

PR #2899, opened 2015-10-11
Intro Dialog Fixes

### LINKED ISSUE #2795: Modal dialogs are not centered correctly
image

image

placegraphichere:
- Next/prev buttons removed, just goes to the next. Back button returns
to the zero slide: #2749
- [edit] now fixes  #2795

### DIFF
--- a/src/style/help/first-launch.less
+++ b/src/style/help/first-launch.less
@@ -32,6 +32,58 @@
     height:66rem;
     border-radius: 0.8rem;
     overflow: hidden;
+    margin-left: 0;
+}
+.main__no-panel .first-launch__dialog, .main__no-panel .keyboard-shortcut__dialog {
+    margin-left: 0;
+}
+
+.main__one-panel .first-launch__dialog, .main__one-panel .keyboard-shortcut__dialog {
+    margin-left:  -(@panel-column-width / 2);
+}
+
+.main__both-panel .first-launch__dialog, .main__both-panel .keyboard-shortcut__dialog {
+    margin-left:  -(@panel-column-width);
+}
+
+// Intro and KBSC Dialogs at 1x
+
+@media (-webkit-min-device-pixel-ratio: 1) and (max-width: 1620px) {
+    .main__both-panel .first-launch__dialog, .main__both-panel .keyboard-shortcut__dialog {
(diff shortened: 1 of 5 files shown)


designedbyscience:
Are we no longer centering these dialogs on the canvas?

placegraphichere:
@designedbyscience see #2795, I describe the issue with centering there. Basically we need additional JS to center, not just CSS due to the possible column expand/collapse. I've probably got the code we need in the branch about positioning the export dialog from a few weeks ago

designedbyscience:
There is already a CSS class on div.main with one, two or no panels, to adjust the centering

placegraphichere:
neato, I'll check it out and update this

placegraphichere:
Now should fully address #2795. 

When width of app is too small based on 2 columns:
screen shot 2015-10-12 at 4 37 02 pm
screen shot 2015-10-12 at 4 30 22 pm

---

When different columns are open:
screen shot 2015-10-12 at 3 51 02 pm
screen shot 2015-10-12 at 3 50 50 pm
screen shot 2015-10-12 at 3 50 37 pm
screen shot 2015-10-12 at 3 50 16 pm
screen shot 2015-10-12 at 3 49 28 pm
screen shot 2015-10-12 at 3 49 16 pm

src/style/help/first-launch.less
```diff
@@ -32,6 +32,36 @@
     height:66rem;
     border-radius: 0.8rem;
     overflow: hidden;
+    margin-left: 0;
+}
+.main__no-panel .first-launch__dialog, .main__no-panel .keyboard-shortcut__dialog{
```
designedbyscience: Spaces before braces

iwehrman:
This does not fully address #2795. It seems like single-column mode is not taken into account, at least not consistently:
image

I think we should just get rid of the back arrow. The asymmetry feels weird, and the new behavior doesn't seem any more intuitive me than the old behavior. Could we just have an icon for the intro slide instead of the "home" button?

placegraphichere:
this is in a better state now. assigning to @iwehrman since he's still here today

iwehrman:
I agree this is in a better state. Merging away!

PR #2905, opened 2015-10-12
- fix for #2903

placegraphichere:


src/js/jsx/help/KeyboardShortcuts.jsx
```diff
@@ -153,7 +153,7 @@ define(function (require, exports, module) {
                                     <SVGIcon CSSID="tool-maskmode" />
                                 </Button>
                                 <span className="keyboard-shortcut__name">{shortcuts.TOOLS.MASKMODE}</span>
-                                <span className="keyboard-shortcut__instr">m</span>
+                                <span className="keyboard-shortcut__instr">M</span>
```
iwehrman: Actually, this almost certainly needs to be a localized string. I missed this in your last PR...

placegraphichere:
what portion needs to be localized? the single kbsc (ie "V") were there from the previous release

iwehrman:
Well, ok. In theory, our shortcuts are all localizable, which means in theory all of our text that refers to the shortcuts should be localizable. But, currently, I do not believe that the loc team has changed any shortcuts. So if none of the other shortcuts in here are localized then I guess this is OK for now.

Issue #2596, opened 2015-09-23
Add artboard canvas UI should not show when layer visibility is off


chadrolfs:
cmd/ctrl + n
select the AB in the layers panel and either click the visibility icon or use the shortcut ( cmd/ctrl + alt + , ) to turn visibility off.

Result: 
![screen shot 2015-09-23 at 1 36 24 pm](https://cloud.githubusercontent.com/assets/8442631/10058105/98f69634-61f8-11e5-9596-6e7adbd1ee9e.png)

volfied:
I think @DivyaPrabhakar should look into this, and learn a little about our overlay code. Please come bug me if you get stuck. Artboard adders are drawn in superselectOverlay.jsx

chadrolfs:
Marking this M3, but also happens in standard PS.  Logging watson issue

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4069676

chadrolfs:
Verified fix in standard PS, but not fixed on JS side.

iwehrman:
FBNC to @chadrolfs.

PR #2906, opened 2015-10-12
Don't draw artboard 'adders' if the artboard is not visible

mcilroyc:
Prevent the plus icons from showing up around the artboard if it is invisible.  Addresses #2596


iwehrman:
Brilliant.

PR #2907, opened 2015-10-12
Updates to single panel behavior

designedbyscience:
Changes to the sizes of the section in order to stop the layers panel from jumping around

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -254,7 +254,9 @@ define(function (require, exports, module) {
                             key: documentID,
                             disabled: disabled,
                             document: document,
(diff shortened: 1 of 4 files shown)


iwehrman:
Can you give me a little more of a hint about what's supposed to happen here? Indeed, the layers panel does not seem to bounce around, except when there are no selected layers. So, I guess it works?

designedbyscience:
When Effects and/or Exports are open, they only take up 15% of the height, and Layers & Libraries fills the rest
When Layers and Libraries are closed, Effects and Exports will fill the space depending on their contents (just uses flex box)

iwehrman:
I see, therefore I merge.

PR #2908, opened 2015-10-12
Using SVGs for checkboxes

designedbyscience:
All of the other checkbox solutions have some issues, so I just replaced all of the states with SVGs (courtesy of @placegraphichere)

It looks better, and consistent across standard and Retina displays


iwehrman:
LGTM.

Issue #2902, opened 2015-10-12
Inconsistent borders along window edge


iwehrman:
The toolbar does not have a panel border:
![image](https://cloud.githubusercontent.com/assets/1075154/10434988/03689ad4-70d2-11e5-8c38-6414548c3471.png)

The icon bar does have a panel border:
![image](https://cloud.githubusercontent.com/assets/1075154/10435001/1669bcd0-70d2-11e5-81fa-bdfc9c3ca7cd.png)

And so does single-column mode:
![image](https://cloud.githubusercontent.com/assets/1075154/10435005/29d5ea14-70d2-11e5-9f5c-55776f224491.png)

iwehrman:
CC  @placegraphichere or @designedbyscience

designedbyscience:
@placegraphichere says the right border is incorrect. I'm on it

iwehrman:
Fixed by #2909.

Issue #2878, opened 2015-10-08
Intra-panel borders do not scale with UI


iwehrman:
100%:
![image](https://cloud.githubusercontent.com/assets/1075154/10381932/9f5c1fd0-6dd3-11e5-9b14-fdaf7a8d15c4.png)

50%:
![image](https://cloud.githubusercontent.com/assets/1075154/10381940/b0c06de4-6dd3-11e5-8c82-5a34fc5da17c.png)

25%:
![image](https://cloud.githubusercontent.com/assets/1075154/10381947/bf921ade-6dd3-11e5-9669-d4863f07733b.png)

iwehrman:
Fixed by #2909!!!

PR #2909, opened 2015-10-12
Make borders responsive, make outer side borders equal

designedbyscience:
Addresses #2902 and #2878


iwehrman:
Lovin' it :hamburger: :fries:

PR #2910, opened 2015-10-12
Feedback from Z on intro text

placegraphichere:
- first slide
- swap slide
- switching slide
- small change to css for padding on long text lines

updates for intro GIFs

- switching
- toolset


iwehrman:
OK.

Issue #2914, opened 2015-10-12
Intro Screen: character not recognized on Windows machine


chadrolfs:
![ice_screenshot_20151012-163849](https://cloud.githubusercontent.com/assets/8442631/10441575/6c50c472-70ff-11e5-8591-9cadd8d662a8.png)

PR #2916, opened 2015-10-13
CMDmd-key-unicode

placegraphichere:
fix for #2914


iwehrman:
@chadrolfs can you verify this and, if it works, merge away?

chadrolfs:
still not fixed
ice_screenshot_20151012-171715

PR #2918, opened 2015-10-13
Localization fixes for Japanese and German

DivyaPrabhakar:
Fixed the input labels for overflow. 

Now we can use the :lang() pseudo selector in our CSS for localization purposes. 

references: #2571 

1.) @ktaki Please check all the labels and titles in this screenshot are correct. 

2.) @placegraphichere Please approve the spacings between the labels. 

3.) @placegraphichere Please approve the spacing between the opacity label and its input 

4.) @placegraphichere Let me know if the spacing between labels and inputs is ok here

### DIFF
--- a/src/js/jsx/Main.jsx
+++ b/src/js/jsx/Main.jsx
@@ -37,7 +37,8 @@ define(function (require, exports, module) {
         Help = require("jsx!js/jsx/Help"),
         Search = require("jsx!js/jsx/Search"),
         ExportModal = require("jsx!js/jsx/sections/export/ExportModal"),
-        Guard = require("jsx!js/jsx/Guard");
+        Guard = require("jsx!js/jsx/Guard"),
+        system = require("js/util/system");
         
     /**
      * Chrome does not play animated svg loaded from external file via svg:use. 
@@ -143,9 +144,13 @@ define(function (require, exports, module) {
             this.props.controller.off("unlock", this._handleControllerUnlock);
         },
 
+        componentDidMount: function () {
+            window.document.documentElement.lang = system.language;
+        },
+
         render: function () {
             var hasDocument = this.getFlux().store("application").getCurrentDocument();
-            
+
             var className = classnames({
                 main: true,
                 "main__ready": this.state.ready,
--- a/src/js/jsx/sections/style/ShadowList.jsx
+++ b/src/js/jsx/sections/style/ShadowList.jsx
@@ -30,7 +30,7 @@ define(function (require, exports) {
         Immutable = require("immutable"),
         classnames = require("classnames"),
         _ = require("lodash");
-        
+
     var LayerEffect = require("js/models/effects/layereffect"),
         collection = require("js/util/collection"),
         strings = require("i18n!nls/strings");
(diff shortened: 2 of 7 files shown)


src/js/jsx/Main.jsx
```diff
@@ -145,8 +145,8 @@ define(function (require, exports, module) {
 
         render: function () {
             var hasDocument = this.getFlux().store("application").getCurrentDocument();
-            
-            var className = classnames({
+
+            var className = classnames(window.navigator.language, {
```
iwehrman: You need to chop this string down so that it works in all German-language or French-language or Japanese-language locales. Quoting the docs [here](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language):

    Examples of valid language codes include "en", "en-US", "fr", "es-ES", etc.

As discussed previously, I suggest copying [this sniipet](https://github.com/adobe-photoshop/spaces-design/blob/4a0bf29d75535a49becf21a2b3c6ca2ab4cb8f23/src/index-build.html#L31-L43) into a new property called "language" in `js/util/system`.

src/style/panel.less
```diff
@@ -329,6 +329,28 @@
     .generate-columns(@n, (@i + 1));
 }
 
+/* Japanese Specific Label Sizings */
+.ja .label__medium.column-2.blur-label {
```
iwehrman: Also, these class names should have a prefix. Something like `.language__ja` and `.language__de`.

iwehrman:
Looks good otherwise! Back to @DivyaPrabhakar.

src/js/jsx/Main.jsx
```diff
@@ -145,7 +146,8 @@ define(function (require, exports, module) {
 
         render: function () {
             var hasDocument = this.getFlux().store("application").getCurrentDocument();
-            
+            window.document.documentElement.lang = system.language;
```
iwehrman: I think this should only happen once, probably in `componentDidMount`.

src/js/jsx/sections/style/ShadowList.jsx
```diff
@@ -40,7 +40,7 @@ define(function (require, exports) {
         BlendMode = require("jsx!./BlendMode"),
         ColorInput = require("jsx!js/jsx/shared/ColorInput"),
         ToggleButton = require("jsx!js/jsx/shared/ToggleButton");
-
+
```
iwehrman: :space_invader:

iwehrman:
Also, please bounce this to @designedbyscience for a final review because he understands the `lang` property of the `documentElement` better than I do.

DivyaPrabhakar:
@designedbyscience Passing this to you to look over once and make sure the psuedo-class :lang() is used correctly.

src/index-build.html
```diff
@@ -30,7 +30,8 @@
     <script>
         var locale = window.navigator.language,
             parts = locale.match(/\w+/),
-            language = parts.length > 0 && parts[0];
+            language = parts.length > 0 && parts[0],
```
designedbyscience: I think this line should end in a `;` not a `,`

src/index-build.html
```diff
@@ -30,7 +30,8 @@
...
+            document.documentElement.lang = language;
```
designedbyscience: And you may not need this at all, since it is handled in Main.jsx#componentDidMount

designedbyscience:
Small comments on index-build, otherwise, looks good

designedbyscience:
Great!

PR #2919, opened 2015-10-13
Fixed SVGIcon require in CheckBox

mcilroyc:
uses !jsx prefix now.  This was causing the minified sources to bork.


mcilroyc:
LGTM if I don't say so myself.

Issue #2920, opened 2015-10-13
Broken Smart Object layer icon contrast against new color stop is too low


chadrolfs:
<img width="264" alt="screen shot 2015-10-12 at 9 45 30 pm" src="https://cloud.githubusercontent.com/assets/8442631/10445743/5d1c501c-712b-11e5-97a8-1af0daf8fcbf.png">

chadrolfs:
<img width="258" alt="screen shot 2015-10-12 at 10 16 35 pm" src="https://cloud.githubusercontent.com/assets/8442631/10446012/f3190a12-712e-11e5-92d0-355a34063010.png">

iwehrman:
FBNC to @chadrolfs.

PR #2921, opened 2015-10-13
Color for error iconz

placegraphichere:
- error icon color
- warning icon color
- fix for #2920


iwehrman:
Review-only until we get a UIF exception.

iwehrman:
What it looks like now:
image

chadrolfs:
here's the other one updated: 
screen shot 2015-10-13 at 2 15 09 pm

iwehrman:
Apparently we don't need a UIF exception for this because we're only changing the color of an icon. Hence, bombs away!

Issue #2808, opened 2015-10-05
Tabbing out a Export Datalist (no change) will create unnecessary history state


mcilroyc:
This worked at one point, but presumably got regressed by recent datalist work.  This is not a critical error because the "extra" history states are at least consistent across PS/DS, but it is not the desired behavior.

mcilroyc:
fixed by #2922

Issue #2807, opened 2015-10-05
ESCing out of export dialog throws error


mcilroyc:
1. Add Asset
1. Click "PNG" drop down
1. Hit ESC key

![ss](http://cl.ly/image/2s382E3l0R1k/Image%202015-10-05%20at%204.56.29%20PM.png)

mcilroyc:
Asking @ktaki to validate this.  Note that I closed related bug #2808 myself, but wanted you to at least be aware of it.  I believe there is an annoyance remaining: after making a selection by hitting the enter key, the input loses focus, but I wanted to keep this bug fix narrow.

ktaki:
Verified the fix.

PR #2922, opened 2015-10-13
ExportList now uses autoSelect=false and changeOnBlur=false datalist params

mcilroyc:
This addresses #2807, #2808, and an unreported issue: extra history state when committing a change to the "format" datalist of export asset configurations.

The default value of the `DataList` parameter `autoSelect` was `true`, and this was triggering a bug when hitting ESC.  Interestingly, even with that parameter defaulting to true, the datalist did not behave autoSelecty... so by setting this false, we aren't losing any functionality.  YES, I do believe there may still be a bug lurking deep inside DataList, but I wanted to keep this PR out of it, if at all possible.

FWIW, there is still one annoying issue with these export inputs right now: after making a new selection in the datalist, the input losing focus.  This means that hitting tab will change layer selection, not tab to the next export input.

Update: This PR also changes the way ESC key is handled in the case when the TextInput is focussed, but the Select List is closed (like when you first tab into the datalist, or after you've selected with a click).  There WAS a call to `onChange(null)` which seems to have been necessary for the SearchBar, which used that to trigger a close of the search dialog, but was otherwise dangerous.  I modified that specific codepath to just pass the key event to SearchBar (via the onKeyDown prop of DataList), which allows for NOT calling the aforementioned `onChange(null)`.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -445,6 +445,9 @@ define(function (require, exports, module) {
                     }
                     break;
                 }
+                case "Escape":
+                    this.props.dismissDialog();
+                    return;
                 case "Backspace": {
                     if (this.refs.datalist.cursorAtBeginning() && this.state.filter.length > 0) {
                         // Clear filter and icon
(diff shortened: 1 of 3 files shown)


pineapplespatula:
I'm still hitting #2807 with this branch I get the error ` Transfer from export.updateLayerAssetFormat to export.updateExportAsset failed: Error: Failed to update asset: TypeError: Cannot read property 'toUpperCase' of null `

mcilroyc:
Thanks for catching that issue @pineapplespatula.  I've squashed it, I believe, so back to you.

pineapplespatula:
LGTM

Issue #2545, opened 2015-09-21
Unsupported formats need handling when opening via CC Lib asset edit original


chadrolfs:
Open a library containing unsupported formats (i.e. 3D)
drag asset onto canvas
double-click asset on canvas or layers panel 

Result: Opens and can edit by attempting to drag out a guide

closing results in: `Action documents.close failed: TypeError: Cannot read property 'call' of undefined`

Additional bug: clicking on title bar blinks toolbar

chadrolfs:
https://www.dropbox.com/s/u3so3c9us9gmraw/CCLib-Open3D.mov?dl=0

shaoshing:
The panels and guides are not completely disabled for unsupported documents. Things I found and will be fixed:

* guide creation when guide is enabled in classic
* copy/paste appearance/style buttons
* add layer effect button
* vector operation buttons
* should greyout libraries panel

chadrolfs:
Waiting on #2562 to be fixed in order to verify this.

shaoshing:
#2562 is replaced by #2743

chadrolfs:
This is a case specific to 3D so let me know if you want me to open as a separate issue and close this one.

1. you can still edit/change a 3D image
https://www.dropbox.com/s/ib374gz8g6sko0y/3D-edit-pt1.mov?dl=0

2. bad things ensue after you do if you try to edit again
https://www.dropbox.com/s/2m7goi49fhx9pnm/3D-edit-pt2.mov?dl=0

shaoshing:
@chadrolfs , 3D or other unsupported layers are living inside a PSD file, so we don't know if the PSD is supported or not until we open it and check its layer attributes, so the only option we have is to not update the changes to the graphic when you try to save a unsupported document. I've created a PR for this. Regarding this video https://www.dropbox.com/s/ib374gz8g6sko0y/3D-edit-pt1.mov?dl=0, I have difficulty on reproduce it; I did reproduce once then it seems to be disappeared. Do you have a clue of the exact steps to repdouce it?

chadrolfs:
You need to click outside the 3D shape on the grid.  Sometime single click on the top of the 3D object then quickly back on the grid outside it and drag

chadrolfs:
after a move/rotate or two, you can get to a point where you can't move again:
https://www.dropbox.com/s/4suxl3u0norm3jm/Move3D.mov?dl=0

mcilroyc:
[...]

PR #2923, opened 2015-10-13
CC Libraries - don't update graphic when saving unsupported document.

shaoshing:
As the title suggested, when you open a 3D graphic for edit, and try to save the document, DS won't update it to the graphic as we don't support editing 3D, also saving the changes will trigger errors afterward. This will address #2545

This PR also change the restriction on adding SO to CC Libiraries. You can now add any locally-linked SO, after that, the layer will turn into cloud-linked SO and adding it  to the libraries again is disabled.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -301,7 +301,7 @@ define(function (require, exports) {
             // cloud-linked SO
             // 
             // FIXME: Instead of waiting, Photoshop should emit an event to tell DS that a layer's smart object 
-            // type has changed.
+            // type has changed. Watson https://watsonexp.corp.adobe.com/#bug=4073805
             .delay(1000)
             .then(function () {
                 return this.transfer(layerActions.resetLayers, currentDocument, currentDocument.layers.selected);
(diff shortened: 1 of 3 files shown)


src/js/actions/libraries.js
```diff
@@ -668,36 +672,45 @@ define(function (require, exports) {
...
         
-        // If the element we're editing was deleted, we recreate it as a new element, so the changes aren't lost.
-        // Otherwise, we just remove the existing representations so we can add the new ones
-        if (element.deletedLocally) {
-            element = library.createElement(element.name, element.type);
-        } else {
-            // TODO should keep representation's stock data
-            element.removeAllRepresentations();
+        if (!document.unsupported) {
+            this.dispatch(events.libraries.UPDATING_GRAPHIC_CONTENT, { documentID: documentID, element: element });
```
mcilroyc: I think this violates our rule that `dispatchAsync` should be used instead of `dispatch` if it is not already wrapped immediately in a promise chain.  If this throws (maybe low risk), the result would not be pretty.

src/js/actions/libraries.js
```diff
@@ -297,12 +297,16 @@ define(function (require, exports) {
                     
                 return descriptor.playObject(createObj);
             })
+            // Wait for 1 second before Photoshop changes layer's smart object type from linked SO to 
+            // cloud-linked SO
+            // 
+            // FIXME: Instead of waiting, Photoshop should emit an event to tell DS that a layer's smart object
```
mcilroyc: yikes - is there a watson/trello card to track this request?

src/js/models/layer.js
```diff
@@ -288,7 +288,8 @@ define(function (require, exports, module) {
                 isLinked: this.isLinked,
                 vectorMaskEnabled: this.vectorMaskEnabled,
                 vectorMaskEmpty: this.vectorMaskEmpty,
-                textWarningLevel: this.textWarningLevel
+                textWarningLevel: this.textWarningLevel,
+                smartObject: this.smartObject
```
mcilroyc: having a hard time figuring out how this change is required for this PR.  I assume there's a reason.

mcilroyc:
back to @shaoshing for a couple of nits/questions.  @iwehrman this is on the edge of my understanding of libraries - so feel free to make a pass through this.

src/js/actions/libraries.js
shaoshing: Can you expand on the reason of the rule? I don't see the difference of invoking `dispatch` alone or follow with a promise chain, as it is running synchronously. Even if `dispatch` throws an error, it will be caught by the outer promise chain wrapped by the fluxcontroller, isn't it?

src/js/actions/libraries.js
shaoshing: no, but I will create a watson bug to request adding the event.

src/js/models/layer.js
shaoshing: Yes. `LayerFace` component relies on the `Layer#face` attribute in the SCU (https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/jsx/sections/layers/LayerFace.jsx#L62). I am adding the smart object so it will aware of the change in smart object type and update its icon accordingly.

shaoshing:
Thanks for reviewing @mcilroyc , replied your comments and have one question before I made the change for `this.dispatch`

src/js/actions/libraries.js
shaoshing: I created a watson bug https://watsonexp.corp.adobe.com/#bug=4073805 and included in the comment below.

src/js/actions/libraries.js
mcilroyc: I can't convince myself that you're wrong, so forget it!  We can discuss casually sometime at an engineering roundtable ...

Issue #1326, opened 2015-06-12
Refresh font menu after Typekit fonts are synced


DesignSpaceBot:
When a doc with missing fonts are opened, missing font dialog is shown, and matching Typekit fonts are presented to be synced. After the fonts are synced from Typekit, font menu must be refreshed to include the newly synced fonts. Typeface and Weight fields must be filled with right name and weight. Currently, Typeface says "missing", Weight is blank, and font menu does not include the synced fonts.

-- @ktaki at [2015-04-28T15:49:14Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1326)

volfied:
@pineapplespatula If there is an event when new fonts are installed, please update our local font list. If there isn't, please talk to Vinod to get such an event be notified.

placegraphichere:
came across this today. Is there a core bug filed?

ktaki:
I thought we had a watson bug since we talked about this in our scrum several times. I could not find it, so here it is.
https://watsonexp.corp.adobe.com/#bug=4072425

mcilroyc:
FBNC to @ktaki 
note from Ian in the PR: 
"the easiest way to trigger these changes is with Font Book."

ktaki:
Possible side effect of this fix. Layer names are changed after fonts were synced.

Steps:
Open 10tkfonts.psd. @iwehrman You have this file, right?
If no missing fonts, delete fonts in the Typekit site, and reopen the file.
In Resolve Missing Fonts dialog, click Resolve Fonts.

RESULTS:

![screen shot 2015-10-14 at 2 25 10 pm](https://cloud.githubusercontent.com/assets/2029597/10498291/86eceae6-727f-11e5-87f9-c88e522b3f91.png)

If this is unrelated to this issue, I will open new issue. Let me know.

mcilroyc:
Ian believes this was fixed by 2957 @ktaki

ktaki:
Verified the fix. When fonts are synced from Resolve Missing Fonts dialog, the yellow icons will be cleared automatically. When fonts are removed or added behind the seen, clicking on the layers panel will update the text layer icons.

iwehrman:
@ktaki Also, switching tools seems to cause photoshop to send us the event about the font list changing.

PR #2926, opened 2015-10-13
Sync font list changes

iwehrman:
With this change, we now listen for the new `fontListChanged` event, which happens when the system font list changes, handles it by updating the font list (if it is already initialized) when the font list changes. The selected text layers are also reset when this happens so that the layers panel reflects the change immediately. Note that sometimes it is necessary for the tool to be changed before Photoshop will send this event the first time.

Assigning this to @mcilroyc because I could use a second opinion on the `resetLayers` call. Which is to say, I don't think this should create a new history state, but also I'm updating the document model, so... ?

Addresses #1326.

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -661,24 +661,86 @@ define(function (require, exports) {
      * Initialize the list of installed fonts from Photoshop.
      *
      * @private
+     * @param {boolean=} force If true, re-initialize if necessary.
      * @return {Promise}
      */
-    var initFontList = function () {
+    var initFontList = function (force) {
         var fontStore = this.flux.store("font"),
             fontState = fontStore.getState(),
             initialized = fontState.initialized;
 
-        if (initialized) {
+        if (initialized && !force) {
             return Promise.resolve();
         }
 
         return descriptor.getProperty("application", "fontList")
             .bind(this)
-            .then(this.dispatch.bind(this, events.font.INIT_FONTS));
+            .then(this.dispatch.bind(this, events.font.INIT_FONTS))
+            .then(function () {
+                var document = this.flux.store("application").getCurrentDocument();
+                if (!document) {
+                    return;
+                }
+
+                var selected = document.layers.selected,
+                    typeLayers = selected.filter(function (layer) {
+                        return layer.isTextLayer();
+                    });
+
+                return this.transfer(layerActions.resetLayers, document, typeLayers, true);
+            });
     };
     initFontList.reads = [locks.PS_APP];
     initFontList.writes = [locks.JS_TYPE];
+    initFontList.transfers = [layerActions.resetLayers];
     initFontList.modal = true;
+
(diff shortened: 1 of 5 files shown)


iwehrman:
Also note, as suggested by @jsbache, the easiest way to trigger these changes is with Font Book.

src/js/actions/type.js
```diff
@@ -661,25 +661,87 @@ define(function (require, exports) {
...
+                if (!document) {
+                    return;
+                }
+
+                var selected = document.layers.selected,
+                    typeLayers = selected.filter(function (layer) {
+                        return layer.isTextLayer();
+                    });
+
+                return this.transfer(layerActions.resetLayers, document, typeLayers);
```
mcilroyc: Yes, I would think you'd want to pass a third `true` argument to amend history.  As written, if you undo *back to* this state, the document won't reflect whatever changes were found by this resetLayers.

iwehrman:
On @mcilroyc's advice, the `resetLayers` call now amends the history state so that undo will work correctly.

src/js/actions/type.js
```diff
@@ -661,25 +662,91 @@ define(function (require, exports) {
...
             .bind(this)
-            .then(this.dispatch.bind(this, events.font.INIT_FONTS));
+            .then(this.dispatch.bind(this, events.font.INIT_FONTS))
+            .then(function () {
+                var document = this.flux.store("application").getCurrentDocument();
+                if (!document) {
+                    return;
+                }
+
+                // FIXME: For performance reasons this ONLY resets non-exposed layers,
```
mcilroyc: resets *exposed* layers :tractor:

mcilroyc:
What's troubling about this is that for the re-exposed-text-layers never get the changed icon, even when the get selected.  It seems like it would be OK to set all textLayers to uninitialized so that at least they 'get fixed' once they are eventually selected.  A background job would be ideal!  anyway, I'm OK to merge this if you think this is the best compromise we can come up with safely now.  @iwehrman

iwehrman:
OK! This now solves all the problems I can reasonably think of. In particular, we now use a modified `resetLayers` call, which fully updates ONLY the selected layers and lazy-updates (i.e., only updates the non-lazy properties, thus returning the layer to an uninitialized state), to reset ALL type layers in ALL (initialized) documents, including inactive ones. The hope is that this modified `resetLayers` call is generally fast enough that we can get away with this. In practice, with a few large and a few small documents open, it seems to take < 1s to do this. 

Back to @mcilroyc to chop it up some more.

mcilroyc:
Seems good now. I think this new resetLayers functionality will be beneficial elsewhere in the future.  @tomwmcrae can you please add this to my permanent record under the folder "times I made ian's life hard for the good of humanity"

tomwmcrae:
@mcilroyc: Your permanent record has been updated.

Issue #2893, opened 2015-10-09
Transform controls extended over toolbar and document title bar in vector mask transform mode


ktaki:
In vector mask transform mode, rotation control sometimes extended over to toolbar and/or doc title bar making not possible to change tool or doc by clicking on them. Once this happens, exiting vector mask mode does not fix the problem. The cursor won't change to arrow over the toolbar or doc bar. Could this be another policy frame issue?

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

Issue #2892, opened 2015-10-09
superselect tool in vector mask mode is broken when you eneter vector mask mode through a different tool


pineapplespatula:
steps:

on a document with a pixel layer 
1: press 'r' to go to rect tool
2: press 'm' to go to vector mask mode
3: make a vector mask
4: press 'v' to got super select. 
5: click and drag outside of the mask to rotate

result: modal state ends
expected: same behavior as if we entered vector mask mode in super select tool

pineapplespatula:
@ktaki  do you see this behavior

ktaki:
Yes, I did on Friday, and I do today.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #2927, opened 2015-10-13
use alpha propogation while in super select tool

pineapplespatula:
fix for issues #2893 and  #2892

### DIFF
--- a/src/js/tools/superselect.js
+++ b/src/js/tools/superselect.js
@@ -75,11 +75,16 @@ define(function (require, exports, module) {
                         UI.pointerPropagationMode.ALPHA_PROPAGATE_WITH_NOTIFY);
                 });
         } else {
-            var currentLayers = currentDocument.layers.selected,
-                currentLayer = currentLayers.first();
+            var currentLayers = currentDocument.layers.selected;
+
+            if (currentLayers.isEmpty()) {
+                return Promise.resolve();
+            }
+
+            var currentLayer = currentLayers.first();
 
             if (currentLayer.vectorMaskEnabled) {
-                this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
+                return this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
                     .bind(this)
                     .then(function () {
                         currentLayer = applicationStore.getCurrentDocument().layers.selected.first();
@@ -87,7 +92,8 @@ define(function (require, exports, module) {
                             return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
                                 .bind(this)
                                 .then(function () {
-                                    // We are not transfering here, because we activly want to end the use of our locks
+                                    // We are not transfering here
+                                    // because we activly want to end the use of our locks
                                     this.flux.actions.tools.enterPathModalState();
                                 });
                         }
--- a/src/js/tools/superselect.js
+++ b/src/js/tools/superselect.js
@@ -38,6 +38,7 @@ define(function (require, exports, module) {
         VectorTool = require("./superselect/vector"),
         TypeTool = require("./superselect/type"),
         system = require("js/util/system"),
+        layerActions = require("js/actions/layers"),
         policy = require("js/actions/policy"),
         shortcuts = require("js/util/shortcuts"),
         SuperselectOverlay = require("jsx!js/jsx/tools/SuperselectOverlay"),
@@ -76,23 +77,21 @@ define(function (require, exports, module) {
         } else {
             var currentLayers = currentDocument.layers.selected,
                 currentLayer = currentLayers.first();
-            
-            if (currentLayer.vectorMaskEnabled && !currentLayer.vectorMaskEmpty) {
-                return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+
+            if (currentLayer.vectorMaskEnabled) {
+                this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
                     .bind(this)
                     .then(function () {
-                        var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                            OS.eventKind.LEFT_MOUSE_DOWN);
-                        return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
-                    })
-                    .then(function () {
-                        // We are not transfering here, because we activly want to end the use of our locks
-                        this.flux.actions.tools.enterPathModalState();
+                        currentLayer = applicationStore.getCurrentDocument().layers.selected.first();
+                        if (!currentLayer.vectorMaskEmpty) {
+                            return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+                                .bind(this)
+                                .then(function () {
+                                    // We are not transfering here, because we activly want to end the use of our locks
+                                    this.flux.actions.tools.enterPathModalState();
+                                });
+                        }
                     });
-            } else {
-                var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                    OS.eventKind.LEFT_MOUSE_DOWN);
-                return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
             }
         }
     };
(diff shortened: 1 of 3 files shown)


iwehrman:
First, fix all the whitespace!

iwehrman:
Second, explain why this change is necessary and why it is safe!

pineapplespatula:
this change is needed because the pointer policies on vector mask mode allowed users to put DS into an unusable state.
it is safe for a few reasons, it removes Always Propagate from the vector mask workflow
it simplifies the number of policies we utilize and does not add multiple policies on tool select actions 
i get to remove a bunch the infrastructure that propped up the always propagate modes which is no longer needed.

iwehrman:
From what I can tell, this breaks the select border overlays on hover. Do you see that too?

pineapplespatula:
yes, that should be fixed now

iwehrman:
This doesn't seem to address #2892, at least with the repro steps described. Can you confirm that it does for you?

pineapplespatula:
It addresses #2892 for me. are you still seeing it drop out of the modal state?

iwehrman:
Here's what I'm seeing:
http://cloud.wehrman.org/2Y0v2R0T1K38

pineapplespatula:
I was using a layer that previously had a mask. I've added the approriate resetLayers call @iwehrman give it another try

iwehrman:
Needs rebase.

pineapplespatula:
@iwehrman rebased

src/js/actions/layers.js
```diff
@@ -451,9 +451,10 @@ define(function (require, exports) {
      *  non-lazy properties updated, and will be returned to an uninitialized
      *  state. This is a performance optimization. NOTE: the selection state
      *  of the layer models passed in must be accurate for this to work correctly!
+     * @param {boolean} skipResetBorderPolicy If truthy do not reset Border Policies
      * @return {Promise}
      */
-    var resetLayers = function (document, layers, amendHistory, lazy) {
+    var resetLayers = function (document, layers, amendHistory, lazy, skipResetBorderPolicy) {
```
iwehrman: As mentioned offline, it seems like it might be better to remove this flag and just have `tools.resetBorderPolicies` be a no-op in mask mode. That seems preferable to this, but if that doesn't work then this is OK.

src/js/tools/superselect.js
```diff
@@ -76,23 +76,21 @@ define(function (require, exports, module) {
         } else {
             var currentLayers = currentDocument.layers.selected,
                 currentLayer = currentLayers.first();
-            
-            if (currentLayer.vectorMaskEnabled && !currentLayer.vectorMaskEmpty) {
-                return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+
+            if (currentLayer.vectorMaskEnabled) {
+                this.flux.actions.layers.resetLayers(currentDocument, currentLayer, false, false, true)
```
iwehrman: I think this should be a transfer effectively transferring from `tools.select`? Or is there some reason that this just enqueues an action instead of transferring? Do you want `tools.select` to finish before this finishes? (I realize that you explicitly do want that for the result of `enterPathModalState` below, but it's not clear if similar reasoning applies here.)

src/js/tools/superselect.js
```diff
@@ -120,6 +118,24 @@ define(function (require, exports, module) {
     };
 
     /**
+     * @private
+     */
```
iwehrman: Need a little explanation here, and also documentation of the return type.

src/js/tools/superselect.js
```diff
@@ -120,6 +118,24 @@ define(function (require, exports, module) {
...
+    var _pointerPolicyList = function () {
+        var toolStore = this.flux.store("tool"),
+            vectorMode = toolStore.getVectorMode(),
+            pointerPolicy = new PointerEventPolicy(UI.policyAction.ALPHA_PROPAGATE,
```
iwehrman: This should just be initialized in the first branch below, I think.

iwehrman:
Back 2 :pineapple: :fork_and_knife: (that's the best I can do until there's a :spatula:)

pineapplespatula:
back to :weary: : :man:  (so close...)

iwehrman:
Needs moar rebase.

src/js/tools/superselect.js
```diff
@@ -120,6 +119,27 @@ define(function (require, exports, module) {
...
 
     /**
+     * Vector Mode requires that do not swallow pointer commands like we do in normal superselect
+     * @private
+     * @return {Array.<PointerEventPolicy>}
+     */
+    var _pointerPolicyList = function () {
+        var toolStore = this.flux.store("tool"),
+            vectorMode = toolStore.getVectorMode(),
+            pointerPolicy = null;
```
iwehrman: No need to initialize this to null since you're going to overwrite immediately in either branch. Just a declaration will do fine here.

src/js/tools/superselect.js
```diff
@@ -120,6 +119,27 @@ define(function (require, exports, module) {
...
+
+        if (vectorMode) {
+            pointerPolicy = new PointerEventPolicy(UI.policyAction.ALPHA_PROPAGATE,
+                OS.eventKind.LEFT_MOUSE_DOWN);
+            return [pointerPolicy];
+        } else {
+            pointerPolicy = new PointerEventPolicy(UI.policyAction.NEVER_PROPAGATE,
+                OS.eventKind.LEFT_MOUSE_DOWN);
+            return [pointerPolicy];
```
iwehrman: Also, you can hoist these return calls out until after the if-then-else.

src/js/tools/superselect.js
```diff
@@ -120,6 +119,27 @@ define(function (require, exports, module) {
     };
 
     /**
+     * Vector Mode requires that do not swallow pointer commands like we do in normal superselect
```
iwehrman: Grammar nit.

iwehrman:
Plus a few more nits since you gotta go back in anyway.

pineapplespatula:
one more time back to @iwehrman

src/js/tools/superselect.js
```diff
@@ -76,23 +77,21 @@ define(function (require, exports, module) {
         } else {
             var currentLayers = currentDocument.layers.selected,
                 currentLayer = currentLayers.first();
```
iwehrman: If there is no selected layer, this will be null...

src/js/tools/superselect.js
```diff
@@ -76,23 +77,21 @@ define(function (require, exports, module) {
...
-            
-            if (currentLayer.vectorMaskEnabled && !currentLayer.vectorMaskEmpty) {
-                return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing())
+
+            if (currentLayer.vectorMaskEnabled) {
```
iwehrman: ... which will cause a null pointer exception here.

src/js/tools/superselect.js
```diff
@@ -76,23 +77,21 @@ define(function (require, exports, module) {
...
                     .bind(this)
                     .then(function () {
-                        var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                            OS.eventKind.LEFT_MOUSE_DOWN);
-                        return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
-                    })
-                    .then(function () {
-                        // We are not transfering here, because we activly want to end the use of our locks
-                        this.flux.actions.tools.enterPathModalState();
+                        currentLayer = applicationStore.getCurrentDocument().layers.selected.first();
```
iwehrman: Ditto here, and the line below.

iwehrman:
I think I found a few control paths that will cause NPEs. Back to @pineapplespatula to check it out.

pineapplespatula:
good catch, sending back to @iwehrman

src/js/tools/superselect.js
```diff
@@ -74,25 +75,27 @@ define(function (require, exports, module) {
...
-                var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                    OS.eventKind.LEFT_MOUSE_DOWN);
-                return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
+            var currentLayers = currentDocument.layers.selected;
+
+            if (currentLayers) {
+                var currentLayer = currentLayers.first();
+
+                if (currentLayer && currentLayer.vectorMaskEnabled) {
+                    this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
```
iwehrman: Sorry I missed this earlier, but this needs to be `return`ed.

src/js/tools/superselect.js
```diff
@@ -74,25 +75,27 @@ define(function (require, exports, module) {
...
-                        // We are not transfering here, because we activly want to end the use of our locks
-                        this.flux.actions.tools.enterPathModalState();
-                    });
-            } else {
-                var pointerPolicy = new PointerEventPolicy(UI.policyAction.ALWAYS_PROPAGATE,
-                    OS.eventKind.LEFT_MOUSE_DOWN);
-                return this.transfer(policy.addPointerPolicies, [pointerPolicy]);
+            var currentLayers = currentDocument.layers.selected;
+
+            if (currentLayers) {
```
iwehrman: `currentLayers` will always be truthy. You need to explicitly test it's size. How about:
```
if (currentLayers.isEmpty()) {
    return;
}
```

src/js/tools/superselect.js
```diff
@@ -74,25 +75,27 @@ define(function (require, exports, module) {
...
+                var currentLayer = currentLayers.first();
+
+                if (currentLayer && currentLayer.vectorMaskEnabled) {
```
iwehrman: If you do the above emptiness check, you can eliminate the first conjunct above because there will always be a first selected layer given that the selection is non-empty.

iwehrman:
Let's keep this party going! :dancer:  :hatching_chick: :tada:

pineapplespatula:
@iwehrman :tada: :tanabata_tree: :tangerine:

iwehrman:
Welp, darned if I can find anything else to complain about.

Issue #2929, opened 2015-10-13
Incorrect tooltip on inner shadow X/Y inputs


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/10465678/f603e050-71a4-11e5-8605-100f23b858c2.png)

https://watsonexp.corp.adobe.com/#bug=4071268

iwehrman:
Fixed by #2930.

PR #2930, opened 2015-10-13
Correct the inner shadow x/y input tooltip

iwehrman:
What it says!

image

Addresses #2929.

**NOTE** We might need a UIF exception for this (https://watsonexp.corp.adobe.com/#bug=4071268). So, please don't merge until I get permission from @jsbache.


iwehrman:
No UIF exception needed because we're just changing existing strings. Ready for review.

pineapplespatula:
Looks good.

Issue #1903, opened 2015-08-04
Clicking on an artboard dirties the document


shaoshing:
1) open any psd file
2) click on the art board
3) file tab shows changed state

![unsaved-status](https://cloud.githubusercontent.com/assets/118264/9066720/0fd60d1c-3a8e-11e5-9d84-4ecfeddf9c12.gif)

iwehrman:
@DivyaPrabhakar this might be another good one to investigate. The fix may or may not be obvious.

iwehrman:
This seems to be a core bug. [Watson 4071097](https://watsonexp.corp.adobe.com/#bug=4071097).

mcilroyc:
FBNC.  asking @chadrolfs to take a look.  The PR was non-trivial, so we'd be interested to find cases where an actual "dirtying" user action was not reported as such (but showing the doc as dirty).

chadrolfs:
related watson bug is closed/fixed

chadrolfs:
related to #2959 but this is fixed, closing.

PR #2931, opened 2015-10-13
Do not dirty the document when changing the selection.

iwehrman:
What it says! Requires Mac pgdev.761 or later. Over to @mcilroyc again because of a possibly controversial history state change: when `resetLayers` is called with `amendHistory`, I now suppress setting the dirty bit. This seems reasonable to me because in theory there was some previous change which _also_ set the dirty bit (hence the amendment), but I could use a second opinion.

Addresses #1903.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -442,7 +442,8 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer|Immutable.Iterable.<Layer>} layers
-     * @param {boolean=} amendHistory If truthy, update the current history state with latest document
+     * @param {boolean=} amendHistory If truthy, update the current history state
+     *  with latest document, and also do NOT dirty the document.
      * @return {Promise}
      */
     var resetLayers = function (document, layers, amendHistory) {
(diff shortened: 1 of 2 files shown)


src/js/actions/layers.js
```diff
@@ -474,6 +475,7 @@ define(function (require, exports) {
                     };
                 });
                 if (amendHistory) {
+                    payload.suppressDirty = true;
```
mcilroyc: This makes sense.  In fact, in theory, since the other code path here doesn't even *touch* history, I would expect that you could get away with ALL resetLayers suppressing dirty.  That said, I do think this change makes me a bit nervous.

src/js/stores/document.js
```diff
@@ -306,16 +306,20 @@ define(function (require, exports, module) {
...
             var documentID = payload.documentID,
                 layerObjs = payload.layers,
                 document = this._openDocuments[documentID],
                 nextLayers = document.layers.resetLayers(layerObjs, document),
-                nextDocument = document.set("layers", nextLayers);
+                nextDocument = document.set("layers", nextLayers),
+                suppressDirty = payload.suppressDirty;
 
-            this.setDocument(nextDocument, true);
+            this.setDocument(nextDocument, !suppressDirty);
```
mcilroyc: thanks for fixing this.  I'm writing it down to document that mad review skillz.

Issue #2860, opened 2015-10-07
Hitting Delete upon entering Mask mode should remove existing mask


chadrolfs:
Deleting a mask should be easy and straight forward.

Open an image (target a pixel layer)
hit m to enter Mask Mode
hit r for Rect tool and drag out a shape
hit m to exit Mask mode
hit m to re-enter Mask mode
hit Delete

Result: nothing
Expect: Mask is removed

Workaround: enter mask mode, hit a, then hit Delete.

chadrolfs:
If there is no mask, hitting Delete should do nothing if you're in Mask edit mode.

mcilroyc:
FBNC to @chadrolfs.   I'll note for the record that when you delete the existing mask, it seems a little unfortunate to me that you get kicked out of mask mode ... but I could be wrong, or maybe that's hard, dunno.

ktaki:
Hitting delete key deletes all paths. If multiple paths exit, I expect only selected path is deleted.

pineapplespatula:
I'm trying for an improved workflow

chadrolfs:
ah ok, does that cover #2977 case?

pineapplespatula:
yes

chadrolfs:
Closing this.  As the issue is written, this is fixed.  #2978 can cover #2977

PR #2935, opened 2015-10-13
pressing delete while in vector mask mode should sometimes delete the…

pineapplespatula:
… vector mask

fix for issue #2860

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -2225,6 +2225,53 @@ define(function (require, exports) {
     addVectorMask.reads = [locks.JS_TOOL];
     addVectorMask.writes = [locks.PS_DOC];
     addVectorMask.modal = true;
+
+    /**
+     * removes a vector mask on the selected layer
+     *
+     * @return {Promise}
+     */
+    var deleteVectorMask = function () {
+        var applicationStore = this.flux.store("application"),
+            currentDocument = applicationStore.getCurrentDocument();
+
+        if (currentDocument === null) {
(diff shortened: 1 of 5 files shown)


src/js/actions/layers.js
```diff
@@ -2225,6 +2225,54 @@ define(function (require, exports) {
...
+                        event = events.document.history.optimistic.REMOVE_VECTOR_MASK_FROM_LAYER;
+
+                    return this.dispatchAsync(event, payload);
+                });
+        } else {
+            return Promise.resolve();
+        }
+    };
+    deleteVectorMask.reads = [locks.JS_TOOL];
+    deleteVectorMask.writes = [locks.PS_DOC];
```
mcilroyc: I think this writes JS_DOC as well

mcilroyc:
Aside from the lock issue, I have two functional issues for you:
1) I can't delete the mask when in SS tool?
2) When I delete the mask in elip tool, it deletes the whole layer!

http://cl.ly/132R0L0N3U2N
part way through when I jump back in to mask mode while in SS tool, you can't see on the video but I'm pressing the delete key to no avail.  Then later you'll see I switch to E tool and press delete.

pineapplespatula:
good catch. @mcilroyc  back to you

src/js/tools/superselect/vector.js
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
```
mcilroyc: wut

Issue #2491, opened 2015-09-16
Drop shadow menu style nits


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9914387/9018f5c6-5c66-11e5-867f-e3f3d2c5c957.png)

1. The triangle is not fully connected to the top of the menu.
2. Datalists additionally indicate selection with a blue bar on the left of the selected option, but this is missing from the drop shadow menu.

placegraphichere:
I'll pick this up in the spectrum PR #2489

chadrolfs:
I think we can close this.

chadrolfs:
nvm, blue selection indicator - yes.  also arrow keys don't work like they do in other datatlists.

chadrolfs:
IMO, we either fix this for UIF this week or kick it out of M3 in favor of frying bigger fish.  @iwehrman what's your opinion?

iwehrman:
Only the keyboard navigation remains for this bug, so I've updated the title accordingly.

iwehrman:
This is a fairly big change to this component, so I'm kicking this bug out of M3.

PR #2936, opened 2015-10-13
dialog blue

placegraphichere:
- add blue border to dialogs
- first and last cause rounded corner issues
- final visual fix for #2491


iwehrman:
I don't think a UIF exception is required for this.

Issue #2780, opened 2015-10-03
Direct select tool shortcut is active when not in mask mode


iwehrman:
When not in mask mode, the `a` shortcut currently activates the direct select tool. It shouldn't do that.

mcilroyc:
FBNC to ... @chadrolfs ?  since he marked it high pri

PR #2937, opened 2015-10-13
move shortcut for vector superselect tool into vector mode aware handler

pineapplespatula:
fix for issue #2780

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -519,6 +519,7 @@ define(function (require, exports) {
      */
     var _toolModalStateChangedHandler,
         _vectorMaskHandler,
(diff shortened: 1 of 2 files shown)


src/js/actions/tools.js
```diff
@@ -519,6 +519,7 @@ define(function (require, exports) {
      */
     var _toolModalStateChangedHandler,
         _vectorMaskHandler,
+        _vectorSelectMaskHandler,
```
mcilroyc: this is an old problem, obviously, but I wonder if these shouldn't get nulled out by onReset?

mcilroyc:
LGTM

PR #2940, opened 2015-10-14
Center dialogs with JS calculation rather than CSS

designedbyscience:
Since the number of complicating factors for centering the dialogs keeps increasing (toolbar pinned/not pinned, 1 or 2 columns open, single column mode), we are switching to using JS to center the dialogs.

If there is enough space over the canvas for the dialog, we center it over the canvas (based on the cloaking bounds), otherwise, we center it on the screen


shaoshing:
LGTM

Issue #2819, opened 2015-10-06
Incorrect layer index after placing file


iwehrman:
1. Create a new artboard document.
2. Drag an image from the desktop onto the document.
3. Place the image.

Result: the empty layer remains next to the new layer that corresponds to the placed layer. But, the empty layer has been removed by Photoshop.
Expected: The artboard has only one child layer, which is the placed layer.

iwehrman:
CC @mcilroyc @shaoshing. Is this a dupe? I know we've gone back and forth on this with `addLayers`...

mcilroyc:
I'm not aware of a dupe.  It is semi-related to a bug on my plate, but not a dupe.  1613.  I can take this if neither of you has an affinity for it.

shaoshing:
For future reference:

The expected behaviors after placing file 

1. place file via menu item (place linked or embedded): empty layer will be replaced.
2. drag-n-drop local file: empty layer will be replaced.
3. drag-n-drop CC libraries graphic: empty layer will not be replaced.

https://www.dropbox.com/s/j2zso98s9qzd927/replace-layer.mov?dl=0

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
caveat: empty layer needs to be targeted (selected in layers panel) for it to be replaced.

mcilroyc:
yes, just confirming/agreeing, that is part of PS's heuristic, as far as we can tell

Issue #2651, opened 2015-09-25
React warnings about duplicate key props


iwehrman:
`Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of DropShadowList. See https://fb.me/react-warning-keys for more information.`

To repro, ungroup the last artboard in Vermilion.

iwehrman:
I'm also seeing this: 
`Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of UnsupportedEffectList. See https://fb.me/react-warning-keys for more information.`

iwehrman:
It's back! I'm seeing this:
` Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of DropShadowList. See https://fb.me/react-warning-keys for more information.` 

To repro, open Vermilion and select the following layer range:
![image](https://cloud.githubusercontent.com/assets/1075154/10297514/e9b9f4be-6b88-11e5-83d1-35ed3cb03e8c.png)

mcilroyc:
@iwehrman mind validating this?

iwehrman:
No warnings in sight. Fixed by #2942.

PR #2942, opened 2015-10-14
Fix incorrect layer index after placing file and calm React warning

shaoshing:
This PR fixes:

1 selected empty layer is not replaced after placing file. addresses issue #2819. I checked the three difference way of placing file and here is the summary:

1. place file via menu item (place linked or embedded): empty layer will be replaced.
2. drag-n-drop local file: empty layer will be replaced.
3. drag-n-drop CC libraries graphic: empty layer will not be replaced.

**Require dev > 743 to receive the replaced layer id in `placeEvent` attributes**

screen shot 2015-10-14 at 11 27 42 am

2 React warn about missing unique key in children. #2651

### DIFF
--- a/src/js/jsx/sections/style/ShadowList.jsx
+++ b/src/js/jsx/sections/style/ShadowList.jsx
@@ -440,6 +440,10 @@ define(function (require, exports) {
                 // Group into arrays of dropShadows, by position in each layer
                 var shadowGroups = collection.zip(collection.pluck(layers, "dropShadows"));
                 
+                if (shadowGroups.isEmpty()) {
+                    return null;
+                }
+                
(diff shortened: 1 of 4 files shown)


src/js/jsx/sections/style/ShadowList.jsx
```diff
@@ -452,15 +456,11 @@ define(function (require, exports) {
                     );
                 }, this).toList();
             } else {
-                shadowsContent = Immutable.List.of((
+                shadowsContent = (
```
shaoshing: The reason of changing from list to single react component is because React require all children to have a unique key prop, if you pass in a list/array as child component.

mcilroyc:
LGTM

Issue #2840, opened 2015-10-06
Reference point: Y moves as H changes


ktaki:
If a reference point is set to the center row or bottom row, changing H will change Y. This should not happen.

chadrolfs:
Yes, some wacky stuff is going on here

iwehrman:
FBNC to @ktaki and/or @chadrolfs.

ktaki:
Looks good!

PR #2948, opened 2015-10-14
Reference Point: Changing W and H does not change X and Y coordinates

DivyaPrabhakar:
References issue #2840 

Before: 

After: 
Correct behavior -- X and Y only changes when the reference point is changed. When one changes the Width or Height, the layer should move in the opposite direction as the reference point. For example, if the reference point is set to Bottom Right, changing the width should move the left position of the layer and changing the height should change the top position of the layer. The reference point serves as an anchor, therefore does not move. Thus, the X and Y coordinates should not move either when changing the height and width inputs.


iwehrman:
LGTM!

PR #2950, opened 2015-10-14
Do not show hidden options in datalist

### LINKED ISSUE #2465: Blend modes datalist doesn't filter the full list
drag out a rect shape
click in appearance blend mode datalist and type "Dis"
result: filters to Dissolve
Now type "Lum"
result: nothing is filtered even though Luminosity is in the list.

designedbyscience:
While looking into #2465, it became apparent that we were not longer filtering hidden options from datalists. This should fix #2465 in that we shouldn't have been able to see luminosity as an option anyways


iwehrman:
OK, sure!

Issue #2945, opened 2015-10-14
Dragging SVG in DS fails - empty drag (file not added)


zoranagee:
Select an SVG file in CC Libraries and drag to document

RESULT: nothing
EXPECTED: SVG file to be placed (works in standard)

chadrolfs:
I CNR, this works for me with current staging build 1676

zoranagee:
There are so many flavors of SVG - I'll add you to my library and you can try that file there.

chadrolfs:
Repro: (quick) - attempt to drag "Trash" from  library: Test - OK to accidentally nuke" onto canvas

chadrolfs:
Repro: (real-world) - from CC (icon on OS menu bar) click on Assets, search for SVG, add asset to a library, drag it from library to canvas

shaoshing:
@chadrolfs Thank you for the repro and file, I am looking at it now.

mcilroyc:
@shaoshing can you push this over to @chadrolfs if you think it is addressed now with his latest repro file?

PR #2952, opened 2015-10-14
CC Libraries: fix for BMP and SVG graphic types

shaoshing:
This PR fix BMP and SVG graphic types so that you can add them to libraries, or drag from libraries. It also added checker to prevent from creating graphic with wrong representation type. Addresses issue #2945

This is a testing library I created with all graphic types: http://adobe.ly/1QtMpK3


shaoshing:
BTW, the add graphic button is broken for locally-linked smart object and is fixed in #2923

mcilroyc:
LGTM

Issue #2949, opened 2015-10-14
Can't undo/revert after moving layer


iwehrman:
1. Open a document with a single layer. The document is clean.
2. Move the layer. The document is dirty. 

Result: Undo and Revert are disabled.

iwehrman:
Fixed by #2953.

PR #2953, opened 2015-10-14
menu store now listens for the new finalize_history event

mcilroyc:
This addresses #2949.  I recently added a new event that allows history state to be wrapped up and finalized after a complex transaction.  It is currently used only by the move handler.  The menu wasn't listening for this new event, so the menu was not showing "undo" as valid option after the first layer move with a clean doc.


iwehrman:
:+1::+1::+1:

Issue #2943, opened 2015-10-14
The icons on the Welcome screenshots of Design Space (Preview) work space are missing.


fzhu1978:
Watson bug 4072721
https://watsonexp.corp.adobe.com//#bug=4072721

Problem: The icons on the Welcome screenshots of Design Space (Preview) work space are missing.

Method:
1. Launch PS 20151012.M.251 build and switch into Design Space (Preview) work space on JPN OS.
2. Help->Design Space Introduction,check the welcome screen.

Result: The icons on the Welcome screenshots of Design Space (Preview) work space are missing.

Expected: No missing.

Notes:
1. This issue occur on both WIn and Mac OS on FRA/DEU/JPN/

fzhu1978:
Verified on 10/13 build, it still occurs.  Please help to fix @iwehrman.

DivyaPrabhakar:
FBNC @iwehrman @fzhu1978

iwehrman:
Let's just have @ktaki regress this one.

ktaki:
This still happens. It looks like svg file names have been translated.

![screen shot 2015-10-14 at 4 57 05 pm](https://cloud.githubusercontent.com/assets/2029597/10501420/30956206-7296-11e5-8a7a-a11de4f46d37.png)

ktaki:
Retested it. The icons are displayed this time.

PR #2954, opened 2015-10-14
Broken icons on Welcome Screen fixed

DivyaPrabhakar:
References #2943.

Before: 

After:

### DIFF
--- a/src/js/jsx/shared/Carousel.jsx
+++ b/src/js/jsx/shared/Carousel.jsx
@@ -35,6 +35,15 @@ define(function (require, exports, module) {
         headlights = require("js/util/headlights"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
         strings = require("i18n!nls/strings");
+
+    /**
+    * Storing the English file names manually otherwise in different languages, the SVGs cannot be found
+    * 
+    * @constant
+    * @type {array.<string>}
+    **/
+    var ICON_NAMES = ["tool-newSelect", "tool-newSelect", "layer-search-app", "tool-eyedropper",
+                            "layer-artboard", "swap", "workspace", "bird"];
         
     var Carousel = React.createClass({
         mixins: [FluxMixin],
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/Carousel.jsx
```diff
@@ -138,7 +142,7 @@ define(function (require, exports, module) {
                                 className={classSet}
                                 onClick={this._gotoItem.bind(this, idx, this.state.index)}>
                                 <SVGIcon
-                                        CSSID={strings.FIRST_LAUNCH.SLIDES[idx].ICON} />
```
shaoshing: Maybe we should remove these icon ids from the translation as well

src/js/jsx/shared/Carousel.jsx
```diff
@@ -125,6 +125,10 @@ define(function (require, exports, module) {
          * @return {Array.<ReactComponent>}
          */
         _buildNav: function () {
+            // The values in this array should match the icons in strings.FIRST_LAUNCH.SLIDES[idx].ICON
+            // Storing the English file names manually otherwise in different languages, the SVGs cannot be found
+            var iconNames = ["tool-newSelect", "tool-newSelect", "layer-search-app", "tool-eyedropper",
+                            "layer-artboard", "swap", "workspace", "bird"];
```
shaoshing: Can you define the array as constant at the top of the component?

shaoshing:
Looks good, just two nits related to translation cleanup and coding style. Back to you @DivyaPrabhakar

src/style/help/first-launch.less
```diff
@@ -364,6 +364,10 @@
     }
 }
 
+:lang(ja) .carousel__slide-button__continue {
+    bottom: 14rem;
+}
```
shaoshing: This feels like not a robust fix for the problem. In your screenshot, the first line of text is wrapped to the second line because it is added a special header for translation purpose, and it is possible that the translated text will take only one line, and will break your fix again. Same thing could happen for FR and DE. Ideally, the "continue" button should flow with the text, instead of having a fixed position, which is the root of all the problem.

src/js/jsx/shared/Carousel.jsx
DivyaPrabhakar: @shaoshing Do you mean remove them from nls/root/strings.json?

shaoshing: Yes @DivyaPrabhakar.

DivyaPrabhakar:
@shaoshing Regarding the "Continue" button, I agree with you but for it to be flexible for all languages, there is more work to be done that shouldn't be included in this PR since it is large enough to be its own issue. I will file a bug separately for that.

shaoshing:
Sounds great! Thank you @DivyaPrabhakar .

Issue #1781, opened 2015-07-21
Mousedown events are duplicated: clicks get preceded by Tooltips actions (2 clicks needed)


ktaki:
ui:diableTooltips and ui:enableTooltips are executed with the first click on a group icon in the layers panel. Then, the second click does layers:setGroupExpansion. This is Windows only issue.

iwehrman:
I can't seem to reproduce this. Note that currently groups are only expanded/collapsed when the group _icon_ is clicked. Or maybe I'm doing it wrong? Back to @ktaki for more info.

ktaki:
With a group already selected in the layers panel, one click on the group icon executes:

> Enqueuing action ui.disableTooltips; 0/0
> fluxcontroller.js:394 
> Executing action ui.disableTooltips after waiting 0ms; 1/0
> fluxcontroller.js:404 
> Finished action ui.disableTooltips in 80ms with RTT 80ms; 1/0
> fluxcontroller.js:373 
> Enqueuing action ui.enableTooltips; 0/0
> fluxcontroller.js:394 
> Executing action ui.enableTooltips after waiting 0ms; 1/0
> fluxcontroller.js:404 
> Finished action ui.enableTooltips in 70ms with RTT 70ms; 1/0
> fluxcontroller.js:373 

Another click executes:

> Enqueuing action layers.setGroupExpansion; 0/0
> fluxcontroller.js:394 
> Executing action layers.setGroupExpansion after waiting 0ms; 1/0
> fluxcontroller.js:404 
> Finished action layers.setGroupExpansion in 115ms with RTT 115ms; 1/0

The above two sections will be repeated for each expand/collapse actions.

iwehrman:
OK, I think I've seen it. I think the problem is that mousedown events are somehow being duplicated. It doesn't happen every time. It also doesn't seem specific the layers panel, but that's one place where two mousedowns (without a mouseup in between) causing the drag to begin but not end. Also, I keep trying to drag the entire PS window around and having that turn into a double-click-to-maximize-window operation. I think there might be something wrong in the adapter on windows with click events. @volfied or @pineapplespatula: can you please try to repro this? CC: @jfitz-adobe

iwehrman:
May be related to #1677 and https://watsonexp.corp.adobe.com//#bug=4019697. CC @chadrolfs.

iwehrman:
[...]

PR #2955, opened 2015-10-14
Delay drag start until a distance threshold is crossed

iwehrman:
On Windows (and very occasionally on Mac) mouse clicks can cause a mousedown as well as a mousemove event, even when the mouse is not actually moved during the click. This causes our drag logic to kick in, which also suppresses the following click event to prevent interference. In the case of false drags, this click suppression causes problems, e.g. failing to select layers in the layers panel, and probably other cases we haven't noticed.

@jsbache suggested that the drag operations should not start until the distance between the mousemove events and the original mousedown position exceeds some modest threshold. That's exactly what this PR does, and @chadrolfs' preliminary testing suggests that this addresses #1781. Concretely, the `Draggable` component waits to set the `dragging` bit until the distance between the mousemove and the original mousedown exceeds 5 pixels. 

Addresses #1781. :tada:

### DIFF
--- a/src/js/jsx/shared/Draggable.jsx
+++ b/src/js/jsx/shared/Draggable.jsx
@@ -54,6 +54,17 @@ define(function (require, exports, module) {
         FluxMixin = Fluxxor.FluxMixin(React),
         Immutable = require("immutable");
 
+    var math = require("js/util/math");
+
+    /**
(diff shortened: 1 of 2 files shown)


mcilroyc:
looks legit.  No obvious regression on mac at least.

PR #2957, opened 2015-10-14
Fix lazy resetLayers result collation 

iwehrman:
This fixes a dumb programming error from yours truly in the new lazy path through `resetLayers` which caused it to only return the very first layer's descriptor repeatedly. Whoops!

Should address the follow-up bug reported by @ktaki in #1326.


mcilroyc:
laps  :100: :runner:

Issue #2941, opened 2015-10-14
Layer re-ordering broken


chadrolfs:
Open a test multi-layered test file
attempt to re-order or drag layers to bottom of the stack

Result: layers reorder on canvas but not in layers panel and console errors result

iwehrman:
Can you post the stack trace?

chadrolfs:
master @ 7b4ccbcd26673653da9d19882fc3ad1fad0e903d with jenkins playground build 1676.

https://www.dropbox.com/s/7ldwcw15e27gol6/LayerReOrder-broke.mov?dl=0

chadrolfs:
`Action layers.reorder failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Move” is not currently available.`

chadrolfs:
`Transfer from documents.open to application.updateRecentFiles failed: SyntaxError: Invalid number`

chadrolfs:
![screen shot 2015-10-14 at 9 17 31 am](https://cloud.githubusercontent.com/assets/8442631/10490447/e8f86ef6-7255-11e5-8ddb-fe1487e9a7c6.png)
![screen shot 2015-10-14 at 9 29 02 am](https://cloud.githubusercontent.com/assets/8442631/10490446/e8f834a4-7255-11e5-9b0c-ad6b471bb09b.png)

chadrolfs:
moving something on canvas will force layers panel to update

shaoshing:
Steps to reproduce:

1 deselect all layers
2 reset DS
3 reorder any layer without selecting it
4 [BOOM]

iwehrman:
FBNC to @chadrolfs .

PR #2961, opened 2015-10-15
Fix for layer reordering

shaoshing:
We have lazy loading for layer attributes on startup (`layer.initialized === false`) until they are selected. When a layer is uninitialized and you reorder it without selecting it, DS will play the descriptor command but skip emmiting a change event because it thinks the layer is not initlaized and trying to avoid UI redraw (https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/stores/document.js#L180). Now, what you have in the layer panel, are layers remain unchanged but inside Photoshop the reorder command is completed. Then, if you try to reorder again, it will fail and throw error. 

I made a simple fix by selecting the dragged layer (if it haven't) before reordering; this will:

1. make sure the layer is selected and initialized
2. the dragged layer is shown as selected after reordering (before it is unselected). This matches behavior in classic PS.

Addresses issue #2941

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1650,7 +1650,7 @@ define(function (require, exports) {
                 return Promise.resolve();
             }
         }
-        
+
         return _getLayerIDsForDocumentID.call(this, document.id)
             .then(function (payload) {
                 return _getSelectedLayerIndices(document).then(function (selectedIndices) {
@@ -1668,10 +1668,17 @@ define(function (require, exports) {
                 } else {
                     return this.dispatchAsync(events.document.history.optimistic.REORDER_LAYERS, payload);
                 }
+            })
+            .then(function () {
+                // get the document with latest selected layers, after layer reordering.
+                var document = this.flux.store("application").getCurrentDocument();
+
+                return this.transfer(initializeLayers, document, document.layers.selected);
             });
     };
     resetIndex.reads = [locks.PS_DOC];
     resetIndex.writes = [locks.JS_DOC];
+    resetIndex.transfers = [initializeLayers];
     resetIndex.post = [_verifyLayerIndex, _verifyLayerSelection];
 
     /**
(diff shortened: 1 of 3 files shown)


shaoshing:
Steps to reproduce:

1 deselect all layers
2 reset DS
3 reorder any layer without selecting it
4 [BOOM]

iwehrman:
I'm a little wary of this particular fix. I don't like it when JSX has to call multiple actions in a row. I think the correct fix is just to have `layers.resetIndex` (which is transferred to by `layers.reorder`) just transfer to `initializeLayers` at the end of its operation. (This has been on my todo list for a while, but it never percolated to the top because I wasn't aware of a bug that it caused.) What do you think?

shaoshing:
That sounds like a better fix! I will dig into the code and try to figure it out. Thanks for the hint.

shaoshing:
@iwehrman  I've made the changes as suggested. However, I didn't transfer the `initializeLayers` in `resetIndex`, as this is a edge case when drag and drop an unintialized layer, and it seems unnecessary to always initialze the layers in `resetIndex`; Instead, I think `reorderLayers` is a better place for `initializeLayers`. 

Another change I made is adding `_handleReset` to the `draganddrop` store, so that we can make sure to recover to usable UI when bad thing happens (previously, if reorder failed, the ordered layer will stick at last mouse positon).

iwehrman:
I like the handleReset change, but I still think that `resetIndex` should finish by transferring to `initializeLayers`. The reason is that the layer selection can change during `resetIndex` and the rule is that newly selected layers must always be initialized. And the whole point of `resetIndex` is to fix things up if we've lost track of the layer order and selection. So although the bug in  `reorderLayers` is the only _known_ bug, it seems as though similar bugs could crop up with other uses of `resetIndex`. Also note that if there are no newly selected layers then the transfer to `initializeLayers` will resolve immediately. We should be liberal, not conservative, about calling `initializeLayers`!

If you disagree, I'm happy to keep arguing :)

shaoshing:
Hi @iwehrman , you comment sounds convincing, especially you spent more time then I do on the layer reset, so I decided to take your advice and moved the `initializeLayers` to `resetIndex`. Let me know if you have other concerns.

iwehrman:
Lovin' it :hamburger: :fries:

Issue #2958, opened 2015-10-14
export: cancel / ESC choose-folder dialog throws error


shaoshing:
Click cancel or hit ESC to close the choose-folder dialog during export and it will throw an error.

<img width="814" alt="screen shot 2015-10-14 at 4 05 14 pm" src="https://cloud.githubusercontent.com/assets/118264/10500490/68606be4-728d-11e5-9716-c06a52afb94d.png">

shaoshing:
Verified the fix.

PR #2962, opened 2015-10-15
Fixed handling when a user cancels out of the folder chooser dialog

mcilroyc:
This rotted in two remarkable ways.  So many feels.

Addresses #2958 

1. It would seem that the error returned by generator/JSX when a user cancels the dialog has changed from an object with a `message` property, to a simple string.  This is troubling, but I'm not sure how hard I want to try and figure out when/why it changed.
2. The previous approach was to throw a custom error in this case, and handle it in the export action that transferred to promptForFolder.  But, the transfer system is puking on that error now.  I'm fairly confident this used to work ... but anyway now I'm going with a simpler approach of simply resolving null in this case.  No fuss no muss.

The diff looks messier than it actually is.  The meat of the export logic got bumped in a couple of tab stops.

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -729,34 +729,34 @@ define(function (require, exports) {
             .then(function (baseDir) {
                 if (!baseDir) {
                     return Promise.resolve();
-                } else {
-                    _lastFolderPath = baseDir;
-
-                    var layerIdList = collection.pluck(layersList, "id");
-
-                    return _setAssetsRequested.call(this, document.id, layerIdList)
-                        .bind(this)
-                        .then(_setServiceBusy.bind(this, true))
-                        .return(quickAddPromise)
-                        .then(function () {
-                            // fetch documentExports anew, in case quick-add added any assets
-                            var documentExports = this.flux.stores.export.getDocumentExports(documentID, true);
-
-                            // Iterate over the layers, find the associated export assets, and export them
-                            var exportList = layersList.flatMap(function (layer) {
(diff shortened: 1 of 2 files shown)


src/js/actions/export.js
```diff
@@ -727,32 +727,36 @@ define(function (require, exports) {
...
-                }, this);
-
-                _batchExports.call(this, exportList);
-                return Promise.resolve();
-            })
-            .catch(ExportService.CancelPromptError, function () {
-                return Promise.resolve();
+                if (!baseDir) {
+                    return Promise.resolve();
+                } else {
```
shaoshing: suggestion: halting a promise chain in the middle could be painful; an easy way might be throw an exception (instead of returning a resolved promise) and add a `catch` at the end of the chain; in this way,  you can keep most of your code unchanged, and handle the edge cases nicely.

mcilroyc: Yes, I did consider that .. but I also felt like using exception handling for normal code paths is... an anti-pattern?  I suppose you could argue that "canceling a dialog" is a grey area.

shaoshing: Faire enough. Have you thought about removing `else { }` I am in favor of using singular `if` statement to deal with edge cases, which result in less indents and more readable.

shaoshing:
Back to Cory for my last question, then it is good to merge.

mcilroyc:
OK @shaoshing  - the Ol' Missouri Compromise.  I got rid of the `else` and the subsequent indent.  back to you.  thanks for the review.

shaoshing:
:+1:

Issue #2538, opened 2015-09-18
Collapsing or expanding AB or group in layers panel should also select it (like PS) does


chadrolfs:
This will help so you don’t end up with a layer selected that’s in a collapsed group/AB
related to #2192

iwehrman:
Clicking the disclosure arrow does not change the selection state in standard PS, and that's how it works in Design Space too.

iwehrman:
But, I agree that we should ensure that descendants of collapsed groups should not be selected.

chadrolfs:
Hmmm, I'm not sure what you're seeing but in standard PS, if you click on the disclosure arrow for a group or AB in the layers panel it collapses and changes the selection to the group/AB.

chadrolfs:
M3 if time.

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
I was going to bounce this because with the test file from Zorana, i can still get into a state where this doesn't work but that's only because it was saved in that state from a previous build where this was broken.  You won't be able to get into this state with files created/saved from here forward (optimistically).

Brilliant!

PR #2963, opened 2015-10-15
Update selection when collapsing groups

iwehrman:
This changes selection when collapsing groups to match Photoshop. It ensures that hidden children remain selected when a containing group is collapsed. In particular, this moves selection from selected descendants to the parent, but only if necessary, when collapsing groups. Otherwise, just clicking the group icon will (still) not change the selection. This is the same behavior as the standard Photoshop layer group disclosure triangles AFAICT.

Addresses #2538.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -783,47 +783,49 @@ define(function (require, exports) {
         // When collapsing layers, if a visible descendent is selected then the selection
         // is removed and moved up to the collapsing group.
         var layersToSelect = [],
-            layersToDeselect = [];
+            layersToDeselect = [],
+            playObjects = [];
 
-        layers.forEach(function (parent) {
-            var selectedDescendants = document.layers.strictDescendants(parent)
-                .filter(function (child) {
-                    return child.selected && !document.layers.hasCollapsedAncestor(child);
-                });
+        if (!expand) {
+            layers.forEach(function (parent) {
+                var selectedDescendants = document.layers.strictDescendants(parent)
+                    .filter(function (child) {
+                        return child.selected && !document.layers.hasCollapsedAncestor(child);
+                    });
 
-            // If there are any selected hidden descendants, deselect the
-            // children and select the parent if necessary.
-            if (!selectedDescendants.isEmpty()) {
-                layersToDeselect = layersToDeselect.concat(selectedDescendants.toArray());
-                if (!parent.selected) {
-                    layersToSelect.push(parent);
+                // If there are any selected hidden descendants, deselect the
+                // children and select the parent if necessary.
+                if (!selectedDescendants.isEmpty()) {
+                    layersToDeselect = layersToDeselect.concat(selectedDescendants.toArray());
+                    if (!parent.selected) {
+                        layersToSelect.push(parent);
+                    }
                 }
+            });
+
+            if (layersToSelect.length > 0) {
+                var selectRef = layersToSelect
+                    .map(function (layer) {
+                        return layerLib.referenceBy.id(layer.id);
+                    });
+
+                selectRef.unshift(documentLib.referenceBy.id(document.id));
+
+                var selectObj = layerLib.select(selectRef, false);
+                playObjects.push(selectObj);
             }
-        });
 
-        var playObjects = [];
-        if (layersToSelect.length > 0) {
-            var selectRef = layersToSelect
-                .map(function (layer) {
-                    return layerLib.referenceBy.id(layer.id);
-                });
+            if (layersToDeselect.length > 0) {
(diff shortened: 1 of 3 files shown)


src/js/actions/layers.js
```diff
@@ -732,6 +732,28 @@ define(function (require, exports) {
     resetLayerVisibility.modal = true;
 
     /**
+     * Initialize the uninitialized subset of the given layers by loading their lazy properties.
+     *
+     * @param {Document} document
+     * @param {Layer|Immutable.Iterable.<Layer>} layerSpec
+     * @return {Promise}
+     */
+    var initializeLayers = function (document, layerSpec) {
```
iwehrman: I just moved this up in the file so that `setGroupExpansion` could transfer to it. No changes here.

src/js/actions/layers.js
```diff
@@ -745,10 +767,6 @@ define(function (require, exports) {
             layers = Immutable.List.of(layers);
         }
 
-        if (descendants) {
```
iwehrman: Just moved this block down so that it came after the filtering step, which could cause a short circuit.

src/js/actions/layers.js
```diff
@@ -766,18 +834,31 @@ define(function (require, exports) {
...
                 documentID: document.id,
                 layerIDs: collection.pluck(layers, "id"),
+                selected: collection.pluck(layersToSelect, "id"),
+                deselected: collection.pluck(layersToDeselect, "id"),
                 expanded: expand
             });
 
-        return Promise.join(expansionPromise, dispatchPromise);
+        return Promise.join(expansionPromise, dispatchPromise, function () {
+            if (layersToSelect.length > 0) {
```
iwehrman: We could have selected new layers which are uninitialized, so we transfer at the end of the action to `initializeLayers`.

src/js/actions/layers.js
```diff
@@ -811,28 +892,6 @@ define(function (require, exports) {
     revealLayers.transfers = [setGroupExpansion];
 
     /**
-     * Initialize the uninitialized subset of the given layers by loading their lazy properties.
```
iwehrman: Again, this was just moved up. No changes.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -121,8 +121,15 @@ define(function (require, exports, module) {
...
-            this.getFlux().actions.layers.setGroupExpansion(this.props.document, layer,
-                !layer.expanded, descendants);
+            // The clicked layer may an have out-of-date document models due to
+            // the aggressive SCU method in LayersPanel.
+            var documentID = this.props.document.id,
+                documentStore = this.getFlux().store("document"),
+                currentDocument = documentStore.getDocument(documentID),
+                currentLayer = currentDocument.layers.byID(this.props.layer.id);
+
+            this.getFlux().actions.layers.setGroupExpansion(currentDocument, currentLayer,
```
iwehrman: This is the same workaround as in `_handleLayerClick` for the problem of stale documents as a result of the parent component's aggressive `shouldComponentUpdate` method. This is not great, but solving it in a nicer way is a task I'd like to delay.

src/js/actions/layers.js
```diff
@@ -758,6 +776,56 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
+        if (descendants) {
+            layers = layers.flatMap(document.layers.descendants, document.layers);
+        }
+
+        // When collapsing layers, if a visible descendent is selected then the selection
```
mcilroyc: would it save a few cycles to only do all this selection logic ONLY if `!expand`?

src/js/jsx/sections/layers/LayerFace.jsx
mcilroyc: slide

src/js/stores/document.js
```diff
@@ -428,14 +429,52 @@ define(function (require, exports, module) {
...
+
+            // If there will be a selection change, suppress change event for
+            // initial group expansion change.
+            this._updateLayerProperties(documentID, layerIDs, { expanded: expanded },
+                suppressChange);
+
+            if (!suppressChange) {
+                // A change event has already been emitted, and there are no
+                // further selection changes.
+                return;
```
mcilroyc: I think I need to lie down

iwehrman: We really need a store-level mechanism to suppress events while multiple changes are happening. But, not for 16.1!

iwehrman:
I rebased and pushed another commit with your `if (!expand)` suggested change. Back 2 @mcilroyc.

Issue #2800, opened 2015-10-05
Error dragging multiple files onto the canvas at once


iwehrman:
When I drag multiple PSDs onto a canvas at once (to place many smart objects), I get these errors:

```
Transfer from superselect.editLayer to tools.select failed: Error: Host is in a modal state and the request cannot be executed.
    at eval (eval at undefined, <anonymous>:1:5)
    at eval (native)
From previous event:
    at Descriptor._batchPlayImmediate (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:386:21)
    at Descriptor.batchPlay (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:519:25)
    at Descriptor.batchPlayObjects (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:544:21)
    at Descriptor.playObject (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:367:21)
    at Object.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/tools.js:387:35)
From previous event:
    at Object.selectTool (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/tools.js:382:14)
    at FluxController._applyAction (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:518:36)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:441:37)
    at AsyncDependencyQueue._processNext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:291:27)
[...]

PR #2965, opened 2015-10-15
Place object - policy and history changes

mcilroyc:
This addresses #2800, but makes some fairly significant changes to the way policies are suspended in/around the place object workflows.

The main change is that when entering a tool state, we suspend policies for ALL tools (except direct select, for mask mode compatibility), not just `txBx` as before.  Is this safe?  Seems OK to me, but there may be situations I'm not aware of.  Prior to this change, the enter key was being handled by DS and causing a SS "diveIn".  This seemed to work OK for single placements, but borks for multi-object placements.

When dragging an object from the library, the above change means that we no longer need to suspend policies directly from the libraries action, so I removed that.

The place-via-native-menu-commands flow is largely unchanged, but we now validate the need to suspend/restore polices herein, because of the tool modal events which may be interspersed.

When placing objects, we get historyState events for that object itself (diff doc ID) so now we ignore those.  This change was a long time coming.

Note: When TWO or more objects are placed (by dragging from OS) we get a succession of modal tool states, and then finally a series of placeEvent notifications.  The first N-1 events will not have a layer ID.  This is an open core bug that is deferred past this release.  To deal with this, we debounce if possible, and then call updateDocument as few times as possible for these events.  It does seem that sometimes for > 2 object placements we are not able to do this perfectly.

requires pgdev mac 806 (or win equiv)

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -53,7 +53,6 @@ define(function (require, exports) {
         shapeActions = require("./shapes"),
         typeActions = require("./type"),
         preferencesActions = require("./preferences"),
-        policyActions = require("./policy"),
         LibrarySyncStatus = require("js/models/library_sync_status");
 
     /**
@@ -821,12 +820,6 @@ define(function (require, exports) {
                 
                 return Promise.bind(this)
                     .then(function () {
-                        // Suspend all policies to allow moving and transforming the new graphic layer.
-                        if (!hasAlt) {
-                            return this.transfer(policyActions.suspendAllPolicies);
-                        }
-                    })
-                    .then(function () {
                         var docRef = docAdapter.referenceBy.id(currentDocument.id),
                             placeObj = libraryAdapter.placeElement(docRef, element, path, location);
 
@@ -877,7 +870,7 @@ define(function (require, exports) {
     createLayerFromElement.reads = [locks.CC_LIBRARIES, locks.JS_DOC, locks.JS_UI, locks.JS_APP];
     createLayerFromElement.writes = [locks.JS_LIBRARIES, locks.PS_DOC];
     createLayerFromElement.transfers = [layerActions._getLayerIDsForDocumentID, layerActions.addLayers,
-        policyActions.suspendAllPolicies, "libraries.handleCompletePlacingGraphic"];
+        "libraries.handleCompletePlacingGraphic"];
         
     /**
      * This event will be triggered when the user confirm or cancel the new layer 
@@ -886,19 +879,10 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var handleCompletePlacingGraphic = function () {
-        return this.dispatchAsync(events.libraries.PLACE_GRAPHIC_UPDATED, { isPlacing: false })
-            .bind(this)
-            .then(function () {
-                var policyStore = this.flux.store("policy");
-                
-                if (policyStore.areAllSuspended) {
-                    return this.transfer(policyActions.restoreAllPolicies);
-                }
-            });
+        return this.dispatchAsync(events.libraries.PLACE_GRAPHIC_UPDATED, { isPlacing: false });
     };
     handleCompletePlacingGraphic.reads = [locks.JS_APP];
     handleCompletePlacingGraphic.writes = [locks.JS_LIBRARIES];
-    handleCompletePlacingGraphic.transfers = [policyActions.restoreAllPolicies];
 
     /**
      * Applies the given layer style element to the active layers
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -79,10 +79,14 @@ define(function (require, exports) {
         
         return Promise.bind(this)
             .then(function () {
+                // This is a hack for the place linked/embedded menu commands which do not
+                // seem to promptly emit a toolModalStateChanged:enter event
                 if (isPlaceCommand) {
                     this.dispatch(events.menus.PLACE_COMMAND, { executing: true });
                     
-                    return this.transfer(policyActions.suspendAllPolicies);
+                    if (!this.flux.store("policy").areAllSuspended()) {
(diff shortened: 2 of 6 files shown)


src/js/actions/menu.js
```diff
@@ -75,14 +75,18 @@ define(function (require, exports) {
         }
 
         var isPlaceCommand = payload.commandID === _PLACE_LINKED_MENU_ID ||
-                payload.commandID === _PLACE_EMBEDDED_MENU_ID;
+            payload.commandID === _PLACE_EMBEDDED_MENU_ID;
```
mcilroyc: I'll fix this indent.  derp

src/js/actions/menu.js
```diff
@@ -75,14 +75,18 @@ define(function (require, exports) {
...
         
         return Promise.bind(this)
             .then(function () {
+                // This is a hack for the place linked/embedded menu commands which do not
+                // seem to promptly emit a toolModalStateChanged:enter event
```
shaoshing: Shall we create a watson bug for it?

src/js/actions/tools.js
```diff
@@ -484,14 +484,18 @@ define(function (require, exports) {
...
             changeStatePromise = this.transfer(changeModalState, modalState),
+            policyStore = this.flux.stores.policy,
+            toolsIdsThatSuspendPolicies = ["txBx", "arwT", "Rect"],
             policyPromise;
 
         // Suspend policies during type tool modal states
-        if (event.kind._value === "tool" && event.tool.ID === "txBx") {
-            if (modalState) {
+        if (event.kind._value === "tool" && toolsIdsThatSuspendPolicies.indexOf(event.tool.ID) > -1) {
+            if (modalState && !policyStore.areAllSuspended()) {
```
shaoshing: I am just curious about what will cause this situation `modalState && !policyStore.areAllSuspended()`

src/js/actions/libraries.js
```diff
@@ -886,19 +879,10 @@ define(function (require, exports) {
...
-        return this.dispatchAsync(events.libraries.PLACE_GRAPHIC_UPDATED, { isPlacing: false })
-            .bind(this)
-            .then(function () {
-                var policyStore = this.flux.store("policy");
-                
-                if (policyStore.areAllSuspended) {
-                    return this.transfer(policyActions.restoreAllPolicies);
-                }
-            });
+        return this.dispatchAsync(events.libraries.PLACE_GRAPHIC_UPDATED, { isPlacing: false });
```
shaoshing: We should probably clear out these events as their sole purpose was to suspend/restore the policies. I will make the change once this is merged.

src/js/actions/menu.js
shaoshing: any reason of cancelling the indent? this looks like we are defining two variables

mcilroyc: See my comment above (which you probably just missed due to page refresh)

src/js/actions/tools.js
```diff
@@ -484,14 +484,18 @@ define(function (require, exports) {
     var handleToolModalStateChanged = function (event) {
         var modalState = (event.state._value === "enter"),
             changeStatePromise = this.transfer(changeModalState, modalState),
+            policyStore = this.flux.stores.policy,
+            toolsIdsThatSuspendPolicies = ["txBx", "arwT", "Rect"],
```
shaoshing: before @iwehrman has the chance to point it out; Set may fit better here than Array.

iwehrman: Because `log 3 < 3`!!!

src/js/actions/tools.js
mcilroyc: note that this is actually the clean path through here.  In this case, `modalState` is maybe a bad var name because it means "entering a modal state".

If you are asking about why we'd be going in to modal state if policies might already be suspended ... that is a fair question.  It shouldn't happen, but given that PS sometimes emits those "late" modal tool states, this helps prevent overlap between this and the 'native menu' hack.

src/js/actions/tools.js
mcilroyc: fair.  Although it is moot now - see new commit coming soon.

mcilroyc:
@shaoshing @pineapplespatula I've udpated this PR to suspend for all modal tool states except "ptha".  This seems to maintain compatibility with mask mode.  @chadrolfs if you could take a pass through this, I would appreciate any early bug detection.

chadrolfs:
Layers panel doesn't update.  Moving artboard to force update results in index error
ice_screenshot_20151015-143949

mcilroyc:
just confirming that the issue chad mentioned above is a real bummer.  When we place multiple objects, we only get place event for the last one.  I noted in a code comment that this was a known/lame issue... but I didn't realize until now that it's a show-stopper.  Going to dig a bit, and then maybe open a watson :tada:

shaoshing:
Thanks for the update @mcilroyc

mcilroyc:
https://watsonexp.corp.adobe.com/#bug=4074678

src/js/actions/menu.js
mcilroyc: Created a non-M3 watson bug for this issue (not getting toolModalStateChanged).  We'll stick with this workaround for now.   https://watsonexp.corp.adobe.com/#bug=4074688

iwehrman:
I believe that the commit in branch `iwehrman/issue-3020`, when added to this PR, will address #3020.

mcilroyc:
@chadrolfs I've updated this PR based on the change Jesper made to send more than one placeEvent when placing multiple objects.  Unfortunately the fix is not perfect (see notes inline), but at least we now can recognize the situation and fall back on updateDocument.  I would appreciate another pass through this.  In addition to testing the associated bug, please take care to look at other places where we get in to PS modal tool state, and make sure we haven't lost any functionality.  Mask mode seems like a good candidate for regression, but my tests seem to indicate it works as expected.

iwehrman:
Here we go!

Issue #1980, opened 2015-08-11
Change the default pixel size for stroke to 1 px


ktaki:
Current default is 2 px. 1 px stroke is now default in standard Ps. Don't we want to follow it?

chadrolfs:
Should be a quick easy low risk fix if we have time to throw it into M3, if not M4.

ktaki:
Verified the fix.

PR #2968, opened 2015-10-15
change default pixel witdth of strokes to 1 px to match standard

pineapplespatula:
this should fix issue #1980


iwehrman:
I have some thoughts about how you could do this in a more complicated way... j/k lololo

Issue #2282, opened 2015-08-31
Choose your own adventure: pasting paths copied from Illustrator


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9587522/d44b07e8-4fd7-11e5-8231-3628986bc673.png)
Four four four four bugs in one!

placegraphichere:
does this need UX input? besides that blue type looking weird?

iwehrman:
Well... It's not completely clear what this needs. We almost never want to show a dialog from standard in Design Space, but it seems like we might not have a choice here in the short term. So _if_ we're going to use this dialog, then this is primarily an engineering issue of how to make it work correctly. In particular, pasting paths seems to automagically create a vector mask on the selected layer? Alternatively, we might be able to suppress this dialog and use some default option from among the four possibilities. In that case, do you have a favorite?

chadrolfs:
I think we should always paste as Shape Layer and suppress the dialog.

placegraphichere:
I don't that there is a perfect default. It largely depends on what the user is doing. If I'm just copying a shape from Ai, then I probably want a shape layer. If it's more complicated, or includes imagery maybe I want a smart object. Or if I'm trying to keep my assets in from non-destructing, a SO would do. 

If I had to simplify the options, I'd want to choose between shape layer and smart object. But that doesn't really help remove the dialog completely.

chadrolfs:
once you bring it in as a shape, you could create it as a library smart object.

placegraphichere:
Problem with an only shape approach is that a shape layer will simply create a shape box where any images once were. So I don't think it's a good idea to default to something that yields quite different results. Here's a screen grab of before and after copying from Ai to Ps using just shape:
![shape-layer-paste](https://cloud.githubusercontent.com/assets/3229167/9595681/0e591cc0-502f-11e5-9f94-8539f77faa9f.png)

chadrolfs:
Passing this to @shaoshing since he's been working with the place code.  What happens now is that the dialog is now suppressed and we paste as smart object but we can't commit the place.  User is essentially stuck.

Workaround: right-click on App is OS dock > Quit and then you get a Place dialog prompt.

mcilroyc:
As of 2969, you can commit the placement with ENTER.  FBNC to @chadrolfs

PR #2969, opened 2015-10-15
Fix pasting smart object copied from other sources (Illustrator) freezes Photoshop

shaoshing:
Yet another fix related to `place` command. Unfortunately, @mcilroyc's fix in #2965 using `toolModalStateChanged` does not cover this issue, due to the reason similar to place-object menu commands (https://github.com/adobe-photoshop/spaces-design/pull/2965/files#diff-0ade3704b4a758bac0c90e3fb8835095R82). 

This will address issue #2282

### DIFF
--- a/src/js/actions/edit.js
+++ b/src/js/actions/edit.js
@@ -35,7 +35,8 @@ define(function (require, exports) {
         menu = require("./menu"),
         collection = require("js/util/collection"),
         headlights = require("js/util/headlights"),
-        history = require("js/actions/history");
+        history = require("js/actions/history"),
+        policyActions = require("js/actions/policy");
 
     /**
      * Native menu command IDs for Photoshop edit commands.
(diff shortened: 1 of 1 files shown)


src/js/actions/edit.js
```diff
@@ -127,17 +128,26 @@ define(function (require, exports) {
...
+        // until the object is committed.
+        return this.transfer(policyActions.suspendAllPolicies)
+            .bind(this)
+            .then(function () {
+                return this.transfer(menu.nativeModal, {
+                    commandID: PASTE_NATIVE_MENU_COMMMAND_ID
+                });
+            })
+            .finally(function () {
+                if (this.flux.store("policy").areAllSuspended) {
```
mcilroyc: isn't this a function `areAllSuspended()`

shaoshing: :see_no_evil: You are truly right!

shaoshing:
@mcilroyc I've fixed the wrong function call. I saw another one but is fixed in your PR, so I will just leave it there to avoid conflict.

mcilroyc:
alright, looks good!

Issue #2974, opened 2015-10-15
Moving an overlapping layer with Select tool and modifiers results in marquee select error.


chadrolfs:
create two overlapping text layers:
<img width="182" alt="screen shot 2015-10-15 at 1 51 33 pm" src="https://cloud.githubusercontent.com/assets/8442631/10527210/172e82f8-7344-11e5-855c-d10b25555c0c.png">

Hold down Shift + click and drag

Result:
`Transfer from superselect.marqueeSelect to layers.select failed: Error: Select with modifier requires a single layer`

chadrolfs:
<img width="1031" alt="screen shot 2015-10-15 at 1 48 37 pm" src="https://cloud.githubusercontent.com/assets/8442631/10527230/36aa09f4-7344-11e5-93ee-e3691d5757b4.png">

iwehrman:
I think you can repro with just shift + marquee select.

chadrolfs:
Yep, amended steps above

PR #2975, opened 2015-10-15
Fix shift+marquee select

iwehrman:
Currently `layers.select` does not allow modifiers to be used when multiple layers are supplied. This is because the "pivot" calculation, which is used by many of the modifiers, doesn't make sense in the case of multiple layers. However, `marqueeSelect` uses the "add" modifier and provides multiple layers, which causes an error. This relaxes the check so that the only illegal case is when multiple layers are passed with the `addUpTo` modifier. Handling that operation really does require that only a single layer be passed in; otherwise, we just nullify the pivot and move on. (The pivot is used to calculate the final selection when shift-clicking-to-select-a-range in the layers panel.)

Addresses #2974.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -926,7 +926,7 @@ define(function (require, exports) {
      *
      * @param {Document} document Owner document
      * @param {Layer|Immutable.Iterable.<Layer>} layerSpec Either a single layer that
-     *  the selection is based on, or an array of such layers
+     *  the selection is based on, or a list of such layers.
      * @param {string=} modifier Way of modifying the selection. Possible values
      *  are defined in `adapter/src/lib/layer.js` under `select.vals`.
      *  Default is similar to "select", but the pivot layer will always be cleared.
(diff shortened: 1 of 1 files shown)


src/js/actions/layers.js
```diff
@@ -881,11 +881,8 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
-        if (modifier !== undefined && layerSpec.size > 1) {
-            return Promise.reject(new Error("Select with modifier requires a single layer"));
-        }
-
-        var selected = document.layers.selected,
+        var setPivot = modifier === undefined || layerSpec.size === 1,
```
pineapplespatula: I'm not clear on why `modifier === undefined` is a valid case for setPivot to be true

pineapplespatula:
@iwehrman  one nit/question

src/js/actions/layers.js
iwehrman: If you either don't set a modifier OR don't supply more than one layer, we deal with pivots correctly using the current logic. This is what `setPivot` means. With this PR, we also handle some modifiers (but not `addUpTo`) even when `setPivot === false`.

iwehrman: No modifier just means "make the current selection equal to the set of layers passed in". If you provide a modifier, we have to do some calculation that may take a pivot into account. That's why no modifier means all is well.

pineapplespatula: Sorry, I meant that this line should just be `var setPivot = layerSpec.size === 1` since the undefined case does not rely on setPivot

pineapplespatula:
one more round  :fist::fist:  @iwehrman

src/js/actions/layers.js
iwehrman: Oh, I see. It's functionally equivalent, but I guess a little simpler. Commit amended.

Issue #2977, opened 2015-10-15
Hitting delete key deletes all paths in edit mode


chadrolfs:
create a mask with multiple paths
hit a to enter anchor point edit mode
select an anchor point and hit delete

Result: entire mask is deleted
Expect: user should be able to delete specific paths or points

mcilroyc:
This looks to me like it works as expected now.  back to @chadrolfs

ktaki:
Looks good to me also.

chadrolfs:
almost.  Here are two cases where it fails
https://www.dropbox.com/s/m786bwdsmshqt95/Mask-marqueeAnchors.mov?dl=0
https://www.dropbox.com/s/pmvxb8egzzyinqq/Mask-inverted.mov?dl=0

pineapplespatula:
The inverted case we may not be able to do anything about, the path tools draw xor paths without modifier keys after the first path. and when you remove the first path drawn it can appear like a subtraction. we may be able to make them default to add, but i'm not sure. ( a workaround is holding down shift before drawing a path)

pineapplespatula:
also i cannot reproduce the first video can you provide more steps?

chadrolfs:
Yeah, CNR first case now.  Once you click outside the masks, it now selects all instead of deselecting.

chadrolfs:
Closing this, created #3044 for that specific case.

PR #2978, opened 2015-10-15
Change the delete mask workflow to be more useful

pineapplespatula:
fix for issue #2860 and #2977

### DIFF
--- a/src/js/actions/mask.js
+++ b/src/js/actions/mask.js
@@ -59,19 +59,18 @@ define(function (require, exports) {
                     return this.transfer(menuActions.native, payload)
                         .bind(this)
                         .then(function () {
-                            return this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
-                                .bind(this)
-                                .then(function () {
-                                    currentDocument = this.flux.store("application").getCurrentDocument();
-                                    currentLayer = currentDocument.layers.selected.first();
-                                    if (currentLayer && !currentLayer.vectorMaskEnabled) {
-                                        return this.transfer(layerActions.deleteVectorMask)
-                                            .bind(this)
-                                            .then(function () {
-                                                return this.transfer(toolActions.changeVectorMaskMode, false);
-                                            });
-                                    }
-                                });
+                            return this.transfer(layerActions.resetLayers, currentDocument, currentLayer);
+                        })
+                        .then(function () {
+                            currentDocument = this.flux.store("application").getCurrentDocument();
+                            currentLayer = currentDocument.layers.selected.first();
+                            if (currentLayer && !currentLayer.vectorMaskEnabled) {
+                                return this.transfer(layerActions.deleteVectorMask)
+                                    .bind(this)
+                                    .then(function () {
+                                        return this.transfer(toolActions.changeVectorMaskMode, false);
+                                    });
+                            }
                         });
                 } else {
                     return this.transfer(layerActions.deleteVectorMask)
--- a/src/js/actions/mask.js
+++ b/src/js/actions/mask.js
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2015 Adobe Systems Incorporated. All rights reserved.
+ *  
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"), 
+ * to deal in the Software without restriction, including without limitation 
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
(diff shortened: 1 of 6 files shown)


src/js/actions/mask.js
```diff
@@ -0,0 +1,92 @@
...
+                currentLayer = currentDocument.layers.selected.first();
+                
+                if (currentLayer && !currentLayer.vectorMaskEmpty) {
+                    var payload = { commandID: _CLEAR_PATH, waitForCompletion: true };
+                    return this.transfer(menuActions.native, payload)
+                        .bind(this)
+                        .then(function () {
+                            return this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
+                                .bind(this)
+                                .then(function () {
```
mcilroyc: I think this `then()` can be bumped out to the outer promise chain

src/js/actions/mask.js
```diff
@@ -0,0 +1,91 @@
...
+
+    var layerActions = require("./layers"),
+        menuActions = require("./menu"),
+        toolActions = require("./tools"),
+        locks = require("../locks");
+
+    var _CLEAR_PATH = 106;
+
+    /**
+     * Handle the deleteText for vector mask mode tools
```
mcilroyc: deleteText?  Also this description could provide a little more info since this is kind of a beefy method.  If it were just an action called "deleteMask" it would be obvious, but since it is a more generic "handleDelete" it needs some clarification.

src/js/actions/mask.js
```diff
@@ -0,0 +1,91 @@
...
+                
+                if (currentLayer && !currentLayer.vectorMaskEmpty) {
+                    var payload = { commandID: _CLEAR_PATH, waitForCompletion: true };
+                    return this.transfer(menuActions.native, payload)
+                        .bind(this)
+                        .then(function () {
+                            return this.transfer(layerActions.resetLayers, currentDocument, currentLayer);
+                        })
+                        .then(function () {
+                            currentDocument = this.flux.store("application").getCurrentDocument();
```
mcilroyc: could use `appStore` defined above

src/js/actions/mask.js
```diff
@@ -0,0 +1,91 @@
...
+        var appStore = this.flux.store("application"),
+            currentDocument = appStore.getCurrentDocument();
+
+        if (!currentDocument) {
+            return Promise.resolve();
+        }
+
+        var currentLayer = currentDocument.layers.selected.first();
+
+        return this.transfer(layerActions.resetLayers, currentDocument, currentLayer)
```
mcilroyc: can you explain why you have to start by resetting the layer?

pineapplespatula: Sure, 
We dont have a model for the Paths that photoshop is currently interacting with. If we had a notificaiton system in place that could tell us if the current mask was empty etc we wouldn't need to do this ( and should probably be logged on a backlog the implement), but since we don't have this we need to differentiate between the various cases of delete. namely selected component, no selected component, and empty mask. 

the first reset differentiates between the empty mask and non empty mask. 
the second reset differentiates between the non empty mask with and without selected components.

pineapplespatula:
@mcilroyc  back to you

src/js/actions/tools.js
```diff
@@ -230,6 +230,11 @@ define(function (require, exports) {
                 this.transfer(policy.removePointerPolicies, currentPolicy, true) : Promise.resolve(),
             guidePromise; // We want to set these after border policies
 
+        // if we are in vector mask mode we should not change pointer policies
```
mcilroyc: ugh, I think this is an indication of larger problem.  post-16.1 problems!

chadrolfs:
if this is going to cause more problems, I suggest we leave current behavior as is and address improvements in M4.  We can now easily delete a mask in mask mode.  Not ideal in multiple path cases, but it works and there are other more serious issues that need address'n.

pineapplespatula:
I think that Cory's comment is that resetBorderPolicies needs some rework post 16.1. Ian Cory and I have discussed a few different patterns that should work for handling the border policies more intelligently. (namely using a UIStore handler instead of calling resetBorderPolicies explicitly from a number of disparate places) 
Cory is this what you meant?

mcilroyc:
Yes, that's what I meant.  I think this PR is close to ready.  @pineapplespatula  I think you missed two of my comments (JSDoc, and appStore).

pineapplespatula:
@mcilroyc i changed the funtion name instead of making a more elaborate description.

Issue #2964, opened 2015-10-15
After drag, a click is lost


iwehrman:
1. Drag a layer.
2. Drop it. It doesn't matter where.
3. Click an unselected layer to change the selection.

Result: the click is not registered and the selection is not changed.

PR #2980, opened 2015-10-15
Fix lost click after drag

iwehrman:
We currently add a click handler when a drag operation starts which stops propagation of the click event. This is because in some cases the drag-ending mouseup is followed by a click event that causes weird things to happen (e.g., entering layer rename mode at the end of a drag). But, we don't _always_ get the click event, so this just removes the post-drag click handler after a short timeout in case the spurious click never arrives.

Addresses #2964.


pineapplespatula:
LGTM

PR #2981, opened 2015-10-15
add escape handlers for vector mask tools

pineapplespatula:
fix for issue # 2670 

there is still an issue with escaping from superselect/vector some of the time, but i think it should not block this PR, (and it have a feeling that it may be resolved by some of the other open vector mask PRs)


shaoshing:
LGTM. (enountered a bug while testing the new shortcut, but not introduced from this PR)

Issue #2946, opened 2015-10-14
Switching apps from Ds and returning: column tab buttons work incorrectly


placegraphichere:
- Open a doc, and open all columns
- swtich to another application (try CMD+tab)
- return to Ds (try CMD+tab, or just clicking back into Ds)
- click the button for either properties or layer column
- all columns close

- expected only the column clicked should change

This is how I've been able to repro, but it seems to come and go sometimes.

iwehrman:
May have something to do with the modifiers being set incorrectly? I bet it only happens when you cmd-tab?

placegraphichere:
that could be the case... I will edit with that note

tiburbage:
The Watson which referenced this issue has been closed, Jesper asserting he's not convinced action on the Spaces/C++ side is the right remedy.

https://watsonexp.corp.adobe.com//#bug=4074645

PR #2982, opened 2015-10-15
Clear stored modifier state when the application becomes inactive

iwehrman:
We currently store modifier state from the `adapterFlagsChanged` os event so that click handlers can determine which modifiers were active during the click. However, we only get `adapterFlagsChanged` events when the application is active. So, e.g., if you cmd-tab away from Design Space, we store modifier state with the command modifier enabled, but if you then click the title bar to activate the application again, no new `adapterFlagsChanged` events are emitted, and so the stored modifier state is now stale and the next click event will turn into a command click.

This works around the problem by clearing the modifier state whenever the application becomes inactive. As noted in the comment below, the adapter would ideally also emit an `adapterFlagsChanged` event whenever the application becomes active so that our state can be immediately brought up to date. It doesn't currently, however, but the remaining bug (i.e., if you hold command down, activate the application and click, expecting a command click but getting only a modifier-free click) seems much less severe. (Watson bug for the core request coming soon.)

Addresses #2946.

### DIFF
--- a/src/js/actions/keyevents.js
+++ b/src/js/actions/keyevents.js
@@ -99,12 +99,34 @@ define(function (require, exports) {
     var _handleFlagsChanged;
 
     /**
+     * Clears modifier state when the application becomes inactive.
+     *
+     * @private
+     * @type {function({{becameActive: boolean}})}
+     */
+    var _handleActivationChanged;
+
+    /**
      * Registers a key event handler to reflect adapter events back to the DOM.
      * 
      * @return {Promise}
      */
     var beforeStartup = function () {
         os.addListener(os.notifierKind.EXTERNAL_KEYEVENT, _externalKeyEventHandler);
+
+        _handleActivationChanged = function (event) {
+            if (!event.becameActive) {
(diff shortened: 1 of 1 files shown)


iwehrman:
Associated core bug: https://watsonexp.corp.adobe.com/#bug=4074645

pineapplespatula:
LGTM

jsbache:
Edge case. Let's defer

jsbache:
Why does DS cache the modifier flags. Modifiers should be available on the mouse-down commands. This is the bast (only) way to get event synchronized information

iwehrman:
@jsbache: the problem is that our event handlers are associated with particular DOM nodes, and so we rely on browser events to trigger operations like this instead of adapter events, which are disconnected from any particular DOM node.

iwehrman:
(Also, I think the last time I checked this information was not making it into the browser events, but I should probably verify that again.)

jsbache:
I am closing the Watson. Please create a jira issue if you don't get modifier keys on mouse events

Issue #2917, opened 2015-10-13
Typeface datalist field is not updated if text layers with missing fonts are selected consecutively


ktaki:
Text layer A has a font missing, the name is font_a.
Text layer B has a font missing, the name is font_b.

Select the text layer A in layers panel. The typeface datalist now shows [font_a]. This is correct.
Then, select the text layer B in layers panel. The typeface datalist still shows [font_a]. This should be updated to [font_b].

Tested the fix for #2804.

ktaki:
@DivyaPrabhakar When multiple text layer is selected (I remember we had a conversation on this topic on Slack yesterday), show the following.

1.) what should the datalist show if multiple missing font text layers are selected at the same time?

If the missing fonts selected are the same fonts, it should show the font name in []. 
If the missing fonts are different, show "[Mixed Fonts]"

2.) what should the datalist show if there is a combination of missing and not missing font text layers selected at the same time?

If you have one or more missing fonts in your selection, show "[Mixed Fonts]"

Since missing font is indicated by square brackets, you do not have to spell it out.

iwehrman:
FBNC to @ktaki again to text @DivyaPrabhakar's improvements.

ktaki:
@DivyaPrabhakar Can you double check the PR #3026 has all the fixes? Here are the test results.

Case | Placeholder Text | Result
------------ | ------------- | -------------
1 missing font | [name of the font] | Pass
2+ missing fonts  | [Mixed Fonts] | [Missing Font]
1 present font 	| name of font | Pass
2+ present fonts  | Mixed Fonts | Pass
1+ missing and present fonts  | [Mixed Fonts] | Mixed Fonts
1+ missing & all the same font  | [name of the font] | [Missing Font]
1+ present & all the same font  | name of the font | Pass

DivyaPrabhakar:
@ktaki Can you verify these? 

__2+ missing fonts__ -- I am seeing the correct result where it reads as [Mixed Fonts] (I am using the most recent master and latest build) 
![](https://www.dropbox.com/s/vfbh69u9bk7g80t/2%2Bmissing%20fonts.gif?raw=1)

__1+ missing and present fonts__ -- I can verify this one is incorrect. I will change it now. 

__1+ missing & all the same font__  -- Again I see the correct result. I took one missing font layer, copy pasted it a couple times, and then selected all the layers of the same missing font. I get the result of [name of font]
[...]

PR #2983, opened 2015-10-16
Typeface datalist updates with missing font name 

DivyaPrabhakar:
References #2917 

When type layer is clicked on, if the font is missing, the typeface datalist will fill with the [*name of missing font*]. If multiple missing font layers are selected, the datalist reads [Missing fonts].

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -470,7 +470,11 @@ define(function (require, exports, module) {
                 if (postScriptFamilyName) {
                     familyName = this._getPostScriptFontFamily(postScriptFamilyName);
                     if (!familyName) {
-                        placeholderText = "[" + postScriptFamilyName + "]";
+                        if (layers.size > 1) {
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/style/Type.jsx
```diff
@@ -470,7 +470,11 @@ define(function (require, exports, module) {
                 if (postScriptFamilyName) {
                     familyName = this._getPostScriptFontFamily(postScriptFamilyName);
                     if (!familyName) {
-                        placeholderText = "[" + postScriptFamilyName + "]";
+                        if (layers.size > 1) {
+                            placeholderText = strings.STYLE.TYPE.MIXED;
```
pineapplespatula: your comment said you it would show `[Missing Font]` but this is the string for "Mixed Fonts" . I just want to double check which one we want.

pineapplespatula:
back to @DivyaPrabhakar  with one question

DivyaPrabhakar:
@pineapplespatula thanks for the catch :sweat_smile: Back to you

Issue #1529, opened 2015-06-12
Min/Max constraints needed for radius field in rectangle properties


DesignSpaceBot:
Currently, any number is accepted in the field.
It should be from 0 to whatever allowed for the current rectangle dimension.

![screen shot 2015-06-04 at 3 02 29 pm](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @ktaki at [2015-06-04T22:08:09Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1529)

chadrolfs:
Nice to have if time in M3

PR #2987, opened 2015-10-16
added min and max for the rectangle radius

DivyaPrabhakar:
References issue #1529 

Note: The reason I added the variable ```maxRadiusInput``` is because currently the Range slider uses ```maxRadius``` and the maximum value is an integer. If you use maxRadius as the max value for the input, it stops at either an integer or a .5 decimal value. Therefore ```maxRadiusInput``` makes sure the Range slider and Input values are the same no matter which you use.

### DIFF
--- a/src/js/jsx/sections/style/Radius.jsx
+++ b/src/js/jsx/sections/style/Radius.jsx
@@ -99,6 +99,8 @@ define(function (require, exports, module) {
                 }, Immutable.List())
                 .min();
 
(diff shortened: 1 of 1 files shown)


mcilroyc:
Looks Good To Me :tm:

DivyaPrabhakar:
@mcilroyc alternative: Let's Get This Merged. 
Think about it, maybe a parking lot topic but if you feel like we should take it to corporate, I am with you 100%.

Issue #2989, opened 2015-10-16
Tool is always reset to select on startup


iwehrman:
It should attempt to use the current Photoshop native tool. This is a regression.

PR #2990, opened 2015-10-16
Fix startup tool inference

iwehrman:
We now have to infer tools based on the application of `nativeToolName` because it is now a function instead of a property.

Addresses #2989.


pineapplespatula:
good catch

Issue #2988, opened 2015-10-16
vector mask: could be trapped by pen tool while in mask mode


shaoshing:
1. create REC
2. enter vector mask mode
3. use select tool to make some selections (and nothing is selected)
4. switch to pen tool and start drawing, and you will probably stuck in pen tool mode

https://www.dropbox.com/s/e7ap2lvz3c8b78k/vector-mask-2.mov?dl=0

iwehrman:
FBNC 2 @chadrolfs

chadrolfs:
CNR from step 1 to step 2 bc you can't enter mask mode when a shape layer is targeted.

pineapplespatula:
@chadrolfs  @shaoshing  i believe meant draw a rectangular marquee selection with the vector superselect tool

PR #2991, opened 2015-10-16
Fix pen tools select handler

pineapplespatula:
fix for issue #2988

### DIFF
--- a/src/js/tools/pen.js
+++ b/src/js/tools/pen.js
@@ -66,25 +66,31 @@ define(function (require, exports, module) {
                     return descriptor.playObject(optionsObj);
                 });
 
-        if (this.vectorMode) {
-            var selectVectorMask = descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
-            return Promise.join(setPromise, selectVectorMask);
+        var deleteFn = function (event) {
+            event.stopPropagation();
+
+            return PS.performMenuCommand(_CLEAR_PATH)
+                .catch(function () {
+                    // Silence the errors here
+                });
+        };
+
+        // Disable target path suppression
+        var backspacePromise = this.transfer(shortcuts.addShortcut,
+                OS.eventKeyCode.BACKSPACE, {}, deleteFn, "penBackspace", true),
+            deletePromise = this.transfer(shortcuts.addShortcut,
+                OS.eventKeyCode.DELETE, {}, deleteFn, "penDelete", true),
+            disableSuppressionPromise = UI.setSuppressTargetPaths(false);
+
+        if (vectorMode) {
+            var selectVectorMask = UI.setSuppressTargetPaths(false)
+                .then(function () {
+                    return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
+                });
+
+            return Promise.join(setPromise, selectVectorMask,
+                disableSuppressionPromise, backspacePromise, deletePromise);
         } else {
-            var deleteFn = function (event) {
-                event.stopPropagation();
-
-                return PS.performMenuCommand(_CLEAR_PATH)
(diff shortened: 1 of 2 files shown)


iwehrman:
Failing CI checks.

src/js/tools/pen.js
```diff
@@ -66,25 +66,31 @@ define(function (require, exports, module) {
                     return descriptor.playObject(optionsObj);
                 });
 
-        if (this.vectorMode) {
-            var selectVectorMask = descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
-            return Promise.join(setPromise, selectVectorMask);
+        var deleteFn = function (event) {
+            event.stopPropagation();
+
+            return PS.performMenuCommand(_CLEAR_PATH)
```
iwehrman: 1. You don't need to return this promise, because the shortcut handler ignores the return value.
2. I think it would be preferable to execute (i.e., enqueue) the `menu.native` flux action here.

src/js/tools/pen.js
```diff
@@ -66,25 +66,31 @@ define(function (require, exports, module) {
...
+                .catch(function () {
+                    // Silence the errors here
+                });
+        };
+
+        // Disable target path suppression
```
iwehrman: Can you move this comment to the right place, or alternatively leave it here but add a note about what the first two shortcuts are for? Also, would it be better to call `shortcuts.addShortcuts` once instead? Finally, don't these delete/backspace shortcuts need to be removed when the pen tool is deselected?

src/js/tools/pen.js
```diff
@@ -66,25 +66,31 @@ define(function (require, exports, module) {
...
+                OS.eventKeyCode.DELETE, {}, deleteFn, "penDelete", true),
+            disableSuppressionPromise = UI.setSuppressTargetPaths(false);
+
+        if (vectorMode) {
+            var selectVectorMask = UI.setSuppressTargetPaths(false)
+                .then(function () {
+                    return descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
+                });
+
+            return Promise.join(setPromise, selectVectorMask,
```
iwehrman: Also, it would be better to factor this out so there is exactly one `return Promise.join` statement at the end.

iwehrman: (You can do this by having an if-then-else statement that initializes a variable to a promise returned by an adapter call in one branch and to `Promise.resolve()` in the other branch.)

iwehrman:
Also, just so I understand, delete didn't do anything in the pen tool before, right? This doesn't change the functionality of the pen tool when you're not in mask mode?

pineapplespatula:
correct. delete in the pen tool previously sent delete path native calls. this does the same.

src/js/tools/pen.js
```diff
@@ -66,28 +65,32 @@ define(function (require, exports, module) {
...
-                return PS.performMenuCommand(_CLEAR_PATH)
-                    .catch(function () {
-                        // Silence the errors here
-                    });
-            };
+        var backspacePromise = this.transfer(shortcuts.addShortcut,
+                OS.eventKeyCode.BACKSPACE, {}, deleteFn, "penBackspace", true),
+            deletePromise = this.transfer(shortcuts.addShortcut,
+                OS.eventKeyCode.DELETE, {}, deleteFn, "penDelete", true),
+            disableSuppressionPromise = UI.setSuppressTargetPaths(false),
```
iwehrman: It looks like you're calling `setSuppressTargetPaths(false)` twice if you're in mask mode and once if you're not. Do you want to call this once regardless of the mode? Or do you only want to call it once in mask mode?

pineapplespatula:
@iwehrman :loop: :repeat:

Issue #2984, opened 2015-10-16
Space no longer works to pan with Select tool.


chadrolfs:
new doc
Select tool
zoom out and hit Space + drag to pan

Result: marquee select.

Workaround: use a different tool.

chadrolfs:
00b81dbf1af41b2c2061ed2adfe2c8481018b6aa is the first bad commit
commit 00b81dbf1af41b2c2061ed2adfe2c8481018b6aa
Author: Ian Wehrman <wehrman@adobe.com>
Date:   Thu Oct 15 09:48:20 2015 -0700

    Remove arrow keydown policies for superselect/vectorselect, and set the default keyboard mode to FOCUS_PROPAGATE.

iwehrman:
@jsbache may also have inadvertently broken this with his scrubby zoom changes last night.

iwehrman:
May or may not be addressed with branch `iwehrman/issue-2984`. I can't repro on my current build and can't get the most recent one at the moment.

PR #2992, opened 2015-10-16
Never propagate space key to PS with select tool

iwehrman:
This used to work for free because we had a `NEVER_PROPAGATE` keyboard mode, but we've changed that recently so now it doesn't!

Addresses #2984.


iwehrman:
Closing this in favor of a more general fix for focus issues coming right up...

Issue #2956, opened 2015-10-14
Lost underscore in Export icon


ktaki:
Retina only

<img width="120" alt="screen shot 2015-10-14 at 2 58 27 pm" src="https://cloud.githubusercontent.com/assets/2029597/10499260/29506fc0-7284-11e5-98a3-cad66ce57535.png">

iwehrman:
Can you report the results of `window.screen` from the console with this display here?

ktaki:
Screen {orientation: ScreenOrientation, availWidth: 1440, availHeight: 841, availTop: 23, availLeft: 0…}
availHeight: 841
availLeft: 0
availTop: 23
availWidth: 1440
colorDepth: 24
height: 900
orientation: ScreenOrientation
pixelDepth: 24
width: 1440
__proto__: Screen

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix on Mac Book Pro and Surface Pro 3.

PR #2994, opened 2015-10-16
Icon header sizes adjusted 

DivyaPrabhakar:
References #2956  

Small changes to the icon sizes so that they don't get cut off and the search icon looks as big as the other icons.


iwehrman:
:+1: :+1: :+1:

Issue #2670, opened 2015-09-28
Use of Enter and Esc is confusing attempting to move/edit an existing layer mask


chadrolfs:
(layer with existing mask) hit m to enter mask mode so you can move the mask independent of the image it's masking, with Select tool move the mask, hit Esc to exit mask mode
Result: mask jumps back to where it was previously
Expect: exit mask mode
Workaround: hit Enter instead, then m to exit mask mode (don't hit Enter twice) - this is confusing.

placegraphichere:
Escape should not cancel the edit. Similar to text editing, ESC should exit the mode applying the changes. The edits are maintained. The same should be followed here with masks. Gonna plop this on someone else...

chadrolfs:
you heard the man.

pineapplespatula:
This is tied up in the keyboard policy changes that jesper just allowed us to make, so i should be able to finally make some progress on this

chadrolfs:
sucking in comments from #2682 and closing that one as a dupe since this one has more comments:

Press 'm' to enter the vector mask transform mode.
Hit Esc - Enters editing mode
Hit Esc - Enters transform mode
Hit Esc - Enters editing mode
Hit Esc - Exits vector mask mode
The doc title stays blue. <--- Redraw issue?

ktaki:
Here is the latest behavior.

Press 'm' to enter the vector mask transform mode.
'a' - Enters editing mode
Hit Esc - Enters transform mode
Hit Esc - Exits vector mask mode
mask mode icon stays blue. (issue #2890)

The Esc key works only superselect tool is selected. Shouldn't it also work when shape tools are selected?

chadrolfs:
Bouncing:

*With Select tool: hit m to enter mask mode, hit Esc (doesn't exit), hit Esc again then it exits*
With Rect and Ellipse tool: m to enter, hit Esc exits (as expected)
*With Pen tool: m to enter, Esc does not work to exit no matter how many times you hit it (related to #2988?)*
With Type and Sampler tool: you can't enter mask mode (as expected)

iwehrman:
Back to @chadrolfs to verify.

chadrolfs:
Looks good to me.  @ktaki wanna take a look and then close if you agree?

ktaki:
Thanks @chadrolfs. Yap! Works great. Closing.

PR #2995, opened 2015-10-16
add esc handler to pen and selection tools

pineapplespatula:
a fix for #2670

### DIFF
--- a/src/js/actions/mask.js
+++ b/src/js/actions/mask.js
@@ -34,6 +34,45 @@ define(function (require, exports) {
     var _CLEAR_PATH = 106;
 
     /**
+     * Handle escaping from vector mask for the selection based vector mask tools
+     * most tools can just call toolActions.changeVectorMaskMode directly
+     *
+     * @return {Promise}
+     */
+    var handleEscapeVectorMask = function () {
+        var appStore = this.flux.store("application"),
+            toolStore = this.flux.store("tool"),
+            currentDocument = appStore.getCurrentDocument();
+
+        if (!currentDocument) {
+            return Promise.resolve();
+        }
+
+        if (toolStore.getVectorMode() && toolStore.getModalToolState()) {
+            return this.transfer(toolActions.changeVectorMaskMode, false);
+        }
+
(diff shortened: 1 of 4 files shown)


src/js/actions/mask.js
```diff
@@ -34,6 +34,45 @@ define(function (require, exports) {
...
+
+        if (!currentDocument) {
+            return Promise.resolve();
+        }
+
+        if (toolStore.getVectorMode() && toolStore.getModalToolState()) {
+            return this.transfer(toolActions.changeVectorMaskMode, false);
+        }
+
+        var currentLayer = currentDocument.layers.selected.first();
```
iwehrman: This could be null if there are no selected layers. Probably the action should short circuit somehow in this case?

src/js/tools/pen.js
```diff
@@ -121,5 +124,20 @@ define(function (require, exports, module) {
...
+     * Handler for keydown events, installed when the tool is active.
+     *
+     * @param {CustomEvent} event
+     */
+    PenTool.prototype.onKeyDown = function (event) {
+        var flux = this.getFlux(),
+            toolStore = flux.store("tool"),
+            detail = event.detail;
+  
+        if (toolStore.getVectorMode() && detail.keyCode === OS.eventKeyCode.ESCAPE) {
```
iwehrman: There could be a reason I'm not thinking of, but wouldn't it be simpler to install an `ESCAPE` shortcut when selecting the pen tool and to remove it when deselecting?

src/js/tools/superselect/vector.js
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
```
iwehrman: :space_invader:

src/js/tools/pen.js
pineapplespatula: I don't think it would be simpler, we already have the key down handler built into the tool code path. adding additional handlers seems less simple.

pineapplespatula:
@iwehrman  :repeat_one:

iwehrman:
OK!

Issue #2986, opened 2015-10-16
Policy issue: keys fall through to classic PS


chadrolfs:
(using pgdev.780 mac or 761 win and master @ 965e872)
cmd/ctrl + n
t for Type tool
click on canvas and type
hit Enter to commit
hit f, or s

Result: you switch to a full screen mode or now have the clone stamp tool.

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4075279

iwehrman:
AFAICT, this does not have to do with editing type. Typing `f` with just the select tool takes us into full screen. This is definitely related to the switch from keyboard propagation mode of NEVER to FOCUS.

iwehrman:
Typing `y` takes you into the history brush. The key event flood gates are open. This is a major issue.

jsbache:
This is not a core problem (as far as I can see). Javascript sets keyboard propagation *mode* to 3 (focus) and this means that the "f" key is sent to PS when the canvas (or any other non HTML UI) has keyboard focus. I don't see this problem in perforce HTML, so I assume that this is related to a recent HTML change

tiburbage:
https://watsonexp.corp.adobe.com/#bug=4075279 Closed/Withdrawn/NotThisProduct

PR #2997, opened 2015-10-17
Stop propagating keyboard events to PS by default; propagate arrow key events by focus

iwehrman:
1. Revert default keyboard mode to `NEVER_PROPAGATE`.
2. Install (new from @jfitz-adobe!) `FOCUS_PROPAGATE` keyboard _policies_ in the select tool for arrow keys.

This restores the default keyboard routing behavior so that key events by default only go to CEF. If you want to send a keystroke to Photoshop from now on, you probably want to install a `FOCUS_PROPAGATE` policy for it. (The `FOCUS_PROPAGATE` policy action is new and requires at least Mac pgdev.790) In particular, you _only_ want to install an `ALWAYS_PROPAGATE` policy if you don't think CEF should get the key event _even when a text input is focused_, which is almost never because that would be a highly confusing interaction.

Addresses #2984 and #2986.  Requires Mac pgdev.790.

### DIFF
--- a/src/js/actions/policy.js
+++ b/src/js/actions/policy.js
@@ -320,7 +320,7 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var beforeStartup = function () {
-        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.FOCUS_PROPAGATE,
+        var defaultKeyboardMode = adapterUI.keyboardPropagationMode.NEVER_PROPAGATE,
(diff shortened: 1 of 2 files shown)


chadrolfs:
initial run through shows issues listed above as indeed fixed and I haven't run into anything else terrible

pineapplespatula:
the code looks good , @chadrolfs  does that mean you're giving this the all clear?

PR #3000, opened 2015-10-18
Disable possible mouse/keyboard interaction with unsupported document.

shaoshing:
When the current document is unsupported (3D / non-RGB color mode):

1. all mouse interaction is disabled by blocking mouse events to the canvas.
2. document's layers are deselected by default, to prevent keyboard events (e.g. using arrow keys to change layer positon)

Hopefully will address issue #2545 completely.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1076,7 +1076,7 @@ define(function (require, exports) {
         }
 
         // If document doesn't exist, or is a flat document
-        if (!document || document.unsupported || document.layers.all.size === 1 &&
+        if (!document || document.layers.all.size === 1 &&
             document.layers.all.first().isBackground) {
             return Promise.resolve();
         }
--- a/src/js/jsx/Scrim.jsx
+++ b/src/js/jsx/Scrim.jsx
@@ -283,7 +283,8 @@ define(function (require, exports, module) {
             
             var classNames = classnames({
                 "scrim": true,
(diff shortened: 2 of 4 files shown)


src/js/actions/layers.js
```diff
@@ -1076,7 +1076,7 @@ define(function (require, exports) {
         }
 
         // If document doesn't exist, or is a flat document
-        if (!document || document.unsupported || document.layers.all.size === 1 &&
```
shaoshing: I removed the `document.unsupported` checker here for deselecting all layers for unsupported document. I read the code where `deselectAll` and looks like it is safe.

src/js/jsx/Scrim.jsx
```diff
@@ -283,7 +283,8 @@ define(function (require, exports, module) {
             
             var classNames = classnames({
                 "scrim": true,
-                "scrim-drop": this.state.isDropTarget
+                "scrim-drop": this.state.isDropTarget,
+                "scrim__disabled": disabled
```
pineapplespatula: this is disabled on modal states with supported documents. is this intentional?

shaoshing: No Jeffrey. I just updated this to exclude modal state.

pineapplespatula:
the code LGTM . I wanted to test this before merging, but i'm heading out if someone would like to take that up this afternoon

pineapplespatula:
Looks good

Issue #2850, opened 2015-10-07
Panning the canvas should dismiss the sampler HUD


iwehrman:
It blinks away when panning due to the UI cloaking, but then it comes back, most likely positioned over the wrong object on the canvas.

chadrolfs:
M3 if there's time and fix is low risk, otherwise boot to M4.

iwehrman:
Over to @chadrolfs to verify.

PR #3001, opened 2015-10-18
sampler: hide sampler HUB on zoom, scroll, resize

shaoshing:
This PR addresses the follwoing sampler issues:

1. should hide the sampler HUB after zoom/scroll/resize (fixe #2850)
2. with sampler tool selected, hit space will throw error in no-doc mode.


iwehrman:
Yep, yep!

Issue #2530, opened 2015-09-18
Font menu sorted by captial letters first


ktaki:
'a' should be with 'A' not after 'Z'.

chadrolfs:
Should not be case sensitive sorted (like classic PS)

chadrolfs:
marking M3 unless we run out of time or fix is too risky.

iwehrman:
FBNC to @ktaki

ktaki:
Verified the fix.

PR #3002, opened 2015-10-18
Fix ordering of the typefaces list 

shaoshing:
This PR addresses the following issues:

1 typeface should ordered case insensitively. (#2530)

screen shot 2015-10-18 at 10 47 39 am

2 some typefaces are not ordered by title. This is becuase we ordered typeface by postScriptName, and it turns out that `title` does not always match with `postScriptName`. For example `title: Wide Latin, postScriptName: LatinWide`, `title: Wawati SC Regular, postScriptName: DFWaWaSC-W5`.

screen shot 2015-10-18 at 11 24 45 am

The fix is to sort typefaces by upper case title, instead of postScriptName. However, it breaks the binary search in `Select._findIndex`, as postScriptNames (which are the ids of of Select's options) are longer ordered. So another change I made is replacing the ID with typeface index (zero padding string). I have several solutions, and this is the one that requires least changes. Do let me know if you have better solution :D

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -37,16 +37,17 @@ define(function (require, exports, module) {
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
         NumberInput = require("jsx!js/jsx/shared/NumberInput"),
         SplitButton = require("jsx!js/jsx/shared/SplitButton"),
-        SplitButtonList = SplitButton.SplitButtonList,
-        SplitButtonItem = SplitButton.SplitButtonItem,
         LayerBlendMode = require("jsx!./LayerBlendMode"),
         Opacity = require("jsx!./Opacity"),
         Datalist = require("jsx!js/jsx/shared/Datalist"),
-        strings = require("i18n!nls/strings"),
-        collection = require("js/util/collection"),
-        Color = require("js/models/color"),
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/style/Type.jsx
```diff
@@ -142,12 +142,16 @@ define(function (require, exports, module) {
...
-         * Set the type face of the selected text layers from a font's postscript name.
+         * Set the type face of the selected text layers from a font's ID.
          * 
          * @private
-         * @param {string} postScriptName
+         * @param {string} typefaceID
          */
-        _handleTypefaceChange: function (postScriptName) {
+        _handleTypefaceChange: function (typefaceID) {
+            var typefaceIndex = parseInt(typefaceID), // typefaceID is equivalent to index in this.state.typefaces
```
iwehrman: Drive-by nit: we should never use `window.parseInt`. It's harmless here, but in most other cases the fact that it returns a `NaN` when the number doesn't parse is a real bummer. (We could probably talk JSHint or JSCS into flagging this as an error.) Try `parseNumber` from `js/util/math` instead.

shaoshing:
@iwehrman thanks for reviewing. I've replace `parseInt` with `math.parseNumber`. Now back to you.

iwehrman:
In the future, we should just provide a comparator to `Datalist` for this kind of thing. But, for now, this'll do.

PR #3004, opened 2015-10-18
Allow shortcuts when non-input elements have focus

iwehrman:
Currently, global shortcuts are disabled when any HTML element has focus. This makes sense when the focused element is a text input (typing `v` should insert a `v` character, not activate the select tool), but makes less sense for focused non-input elements like the color swatch. This prevents you, e.g., from dismissing the color picker (which focuses the swatch) and then immediately selecting a tool without first hitting escape, which seems pedantic.

This changes the shortcut matching algorithm to only be disabled for focused input elements. Furthermore, if a non-input element is focused and there is some shortcut that matches then the focused element is blurred before executing the handlers.

I'm pretty sure there was an open issue about this, but darned if I can find it.

### DIFF
--- a/src/js/actions/edit.js
+++ b/src/js/actions/edit.js
@@ -36,7 +36,8 @@ define(function (require, exports) {
         collection = require("js/util/collection"),
         headlights = require("js/util/headlights"),
         history = require("js/actions/history"),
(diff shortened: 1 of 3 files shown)


pineapplespatula:
Looks good

Issue #2714, opened 2015-09-29
Panning AB results in brief flash of the add AB icons


zoranagee:
Panning quickly to the left:

<img width="867" alt="screen shot 2015-09-29 at 3 50 24 pm" src="https://cloud.githubusercontent.com/assets/7904855/10180419/dc3a5d86-66c1-11e5-82e5-1700a0047b86.png"> 

<img width="789" alt="screen shot 2015-09-29 at 3 50 34 pm" src="https://cloud.githubusercontent.com/assets/7904855/10180420/dc3c864c-66c1-11e5-95e9-d8a2e06573a0.png">

<img width="706" alt="screen shot 2015-09-29 at 3 50 44 pm" src="https://cloud.githubusercontent.com/assets/7904855/10180418/dc389eba-66c1-11e5-8d72-f9ad6eb1af20.png">

shaoshing:
@chadrolfs I've verified the fix, but can you help double check?

chadrolfs:
This issue appears to be back.  Repro's intermittently on big moves.  Easier to repro on small drags of AB

iwehrman:
Yes, it seems to repro on move but not pan.

iwehrman:
Welp, at this point we've got `scroll` events battling with `autoCanvasShift` events. Each of the former results in an attempt to re-enable the overlays, whereas the latter starts with an attempt to disable the overlays. Currently stumped on this one.

iwehrman:
Giving up on this for now :person_frowning: If anyone else wants to take a crack, here's an almost comical branch which includes about 5 separate, failed attempts at hacking around the problem: https://github.com/adobe-photoshop/spaces-design/compare/iwehrman/issue-2714-again

Note that there are at least some obvious bugs fixed in that branch. For example, the debounced `autoCanvasShift` event handler is currently completely broken, in that for each event it creates a completely new debounced function and then calls it. Debouncing relies on successive calls of the _same_ function, so this is no different from just calling the function after some delay.

Other problems:

1. We know when the drag starts, so we can disable the overlays then, but we don't have a great way of knowing when the drag is complete to re-enable them. 
2. We get `autoCanvasShift` events throughout the drag, and we aren't exactly sure when to process them. Also, processing them by calling `resetBounds` automatically re-enables the overlays!

mcilroyc:
FBNC to @chadrolfs

chadrolfs:
CNR on Mac but still looks pretty buggy on Win

[...]

PR #3005, opened 2015-10-18
Disable overlays before setting the transform

iwehrman:
Disable overlays before setting the transform to prevent artboard buttons from flashing in the previous location at the end of a pan when the cloak ends. 

Addresses #2714.


shaoshing:
Looks good!

Issue #3003, opened 2015-10-18
Shape opacity isn't set via color input


iwehrman:
1. Create a rectangle.
2. Open the fill or stroke color picker.
3. Paste in `rgba(128, 128, 200, 0.1)`

The shape color changes but the opacity doesn't. The alpha slider does change.

chadrolfs:
you need to fix bugs, not verify fixes.  :)

PR #3006, opened 2015-10-18
Add missing onChange prop to ColorPicker

iwehrman:
The `ColorInput` has three change event handlers:

1. `onColorChange` for changes to the opaque (RGB) color;
2. `onAlphaChange` for changes to the alpha (and not RGB) value;
3. `onChange` for arbitrary color changes.

The latter wasn't correctly percolating from the `ColorInput` to the internal `ColorPicker` component, which prevented color changes in the picker's text input to not correctly propagate changes to the opacity. This fixes that.

One reason why this problem wasn't caught earlier is that the three handlers were not marked as required for either component, and the default value for `onChange` was the identity. To avoid this in teh future, all three handlers are now marked as required for both components.

These components have become fairly convoluted since moving the text input inside the picker itself. We should look into smoothing them out a bit after M3.

Addresses #3003.

### DIFF
--- a/src/js/jsx/shared/ColorPicker.jsx
+++ b/src/js/jsx/shared/ColorPicker.jsx
@@ -814,17 +814,14 @@ define(function (require, exports, module) {
 
         propTypes: {
             color: React.PropTypes.instanceOf(Color),
-            onChange: React.PropTypes.func,
-            onColorChange: React.PropTypes.func,
-            onAlphaChange: React.PropTypes.func
+            onChange: React.PropTypes.func.isRequired,
+            onColorChange: React.PropTypes.func.isRequired,
+            onAlphaChange: React.PropTypes.func.isRequired
         },
 
         getDefaultProps: function () {
             return {
-                color: Color.DEFAULT,
-                onChange: _.identity,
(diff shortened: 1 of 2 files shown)


src/js/jsx/shared/ColorPicker.jsx
```diff
@@ -949,8 +946,6 @@ define(function (require, exports, module) {
                         currentRgbaColor.b !== nextRgbaColor.b) {
                         this._changeColor(nextRgbaColor);
                     }
-
-                    this.props.onChange(nextRgbaColor);
```
pineapplespatula: Why is this being removed?

pineapplespatula:
one question. otherwise looks great.

src/js/jsx/shared/ColorPicker.jsx
iwehrman: I think the only callers of this function are changing only the opaque color or the alpha. When the text input changes, it calls `onChange` already, so it isn't needed here. But, also, see my comment above about how this is getting kind of gross... I agree that it's confusing at best.

pineapplespatula:
looks good

Issue #1538, opened 2015-06-12
Align, distribute, flip, swap on empty layers results in errors


DesignSpaceBot:
1. cmd/ctrl + n (empty layer selected by default), cmd/ctrl + c, cmd/ctrl + v (repeat).  Select empty layers in layers panel (note align and distribute icons light up), click on an align/distribute icon
or 
2. select an empty layer(s) and access menu commands for Flip/Swap

Result:
```Action transform.distributeY failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -1 Photoshop message: error: -1 message: Could not complete the command because none of the pixels in the target layer are more than 50% opaque.
    at eval (eval at undefined, <anonymous>:1:5)```

-- @chadrolfs at [2015-06-08T21:02:44Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/1538)

iwehrman:
Great error message, too:

    Photoshop message: error: -1 message: Could not complete the command because none of the pixels
    in the target layer are more than 50% opaque.

shaoshing:
Assign to Chad as this has his signature :D

PR #3007, opened 2015-10-18
Filter empty layers in swap, flip, align and distribute.

iwehrman:
What it says!

Addresses #1538. See notes there for info about how to repro the problem.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -430,18 +430,19 @@ define(function (require, exports) {
      *
      * @private
      * @param {Document} document Owner document
-     * @param {Array.<Layer>} layers An array of two layers
+     * @param {Immutable.Iterable.<Layer>} layers A list of two layers
      *
      * @return {Promise}
      */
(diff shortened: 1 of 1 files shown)


shaoshing:
@iwehrman This looks good and I don't see the error which clicking the buttons. Before merging, I want to confirm that it is your intention to enable the buttons (but act nothing when clicked) even thought some of the selected layers (in this case, empty layer) does not support the operation? For example, if I select an empty layer and a vector layer, the `swap` and `align` buttons are shown as enabled, and clicking on them make no changes to the layers, which feels like they are not functioning. If so, this seems inconsistent with the `flip` buttons, which are disabled if any of the selected layers can not be flipped.

iwehrman:
Yes, this is my intention in the short term, at least. I know that this isn't the ideal user experience, but this is a low-risk fix for M3 for a case that is very much on the edge. I would like later to think about a more holistic solution that will unify the enablement checks between UI, menus and actions.

shaoshing:
All right, let's do it

Issue #2448, opened 2015-09-14
Collapsed EFFECTS panel should expand when + icon is clicked and an effect is selected


ktaki:

iwehrman:
Also, it might be nice if the collapsed height were the same as the empty + expanded height? Basically, when empty, there should be no distinction between expanded and collapsed.

designedbyscience:
@iwehrman I'm concerned about confusion over "Is this open or is it empty?" if the states are exactly the same

iwehrman:
Mmm, yeah, you've got a point. Ok, I take it all back.

designedbyscience:
(I originally did it how you proposed and got confused myself)

chadrolfs:
reviewed - probably not M3

ktaki:
This fails at every other try.
1. Collapse Effects panel
2. Click on '+' and select an effect
3. Effects panel expands, but collapses quickly.
4. Click on '+' and select an effect
5. This time, the effects panel expand and stay as expanded.

1 ~ 5 will repeat. So, if you do it again:
6. Collapse Effects panel again
7. Click on '+' and select an effect
8. Again, the effects panel expands, but collapses quickly.

ktaki:
Verified the fix.

PR #3009, opened 2015-10-18
Force the effects panel open when adding an effect

iwehrman:
What it says! Works just like the Export panel.

Addresses #2448.


shaoshing:
LGTM :+1:

Issue #3008, opened 2015-10-18
Color overlays subheader is the wrong color


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/10567287/1988a91a-75c5-11e5-9f7b-aa2de9b22286.png)

CC @DivyaPrabhakar

iwehrman:
LGTM. Fixed by #3012.

PR #3012, opened 2015-10-19
Effects panel subheader colors corrected

DivyaPrabhakar:
References #3008 

Before: 

After: 

 @placegraphichere has :white_check_mark: the look of unsupported layer effects.


shaoshing:
@DivyaPrabhakar we have a fifth type of effects called "Unsupported", which is easy to overlook because you need to create at least one of these effects (listed in https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/models/effects/layereffect.js#L102) in classic to make it appear. Or, you can import my library http://adobe.ly/1KBVjAb and add apply the layer style from DS.

DivyaPrabhakar:
@shaoshing Thanks for pointing that out, I'll account for that now.

shaoshing:
Perfect!

PR #3024, opened 2015-10-19
Fix panel/column visibility toggle race

iwehrman:
This fixes a race condition in `PanelSet` when toggling panel and column visibility. Those functions would both call `setState` and `setPreferences`, both of which are asynchronous, without synchronization. In some interleavings, the toggled state is lost. (The details of the interleavings are complicated but boring.) This eliminates the _direct_ `setState` calls in favor of just setting the preferences, and then also having the component listen to the preferences store for updates.

Addresses #2448 again! @ktaki Please test, if you have a moment.

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -51,7 +51,7 @@ define(function (require, exports, module) {
         synchronization = require("js/util/synchronization");
 
     var PanelSet = React.createClass({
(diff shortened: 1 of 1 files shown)


ktaki:
It looks good now.

mcilroyc:
It's a little weird that object is still named `nextState`, but whatev.  merging!

PR #3025, opened 2015-10-20
Fix German shortcut issue.

fzhu1978:
https://watsonexp.corp.adobe.com/#bug=4075078
https://watsonexp.corp.adobe.com/#bug=4075062
https://watsonexp.corp.adobe.com/#bug=4075074

Changing shortcut according to linguist's suggestion.


iwehrman:
Looks OK to me! @ktaki will confirm that the original issue is resolved.

PR #3026, opened 2015-10-20
Typeface datalist updates with missing font name

DivyaPrabhakar:
There were some outstanding problems after #2983 got merged, this PR addresses them. @ktaki If you can double check below that these are the expected behaviors, that would be great. References issue #2917.

Expected behavior now:

| Case  | Placeholder Text |
| ------------- | ------------- |
| 1 missing font  | [name of the font]  |
| 2+ missing fonts    |  [Mixed Fonts]  |
| 1 present font    |  name of font  |
| 2+ present fonts    |  Mixed Fonts |
| 1+ missing and present fonts   |  [Mixed Fonts] |
| 1+ missing & all the same font   |  [name of the font] |
| 1+ present & all the same font   |  name of the font |

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -466,17 +466,22 @@ define(function (require, exports, module) {
                 styleTitle,
                 placeholderText;
 
+            var anylayerTextWarning = layers.some(function (layer) {
+                return layer.textWarningLevel > 0;
+            });
+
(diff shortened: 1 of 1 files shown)


ktaki:
Yes, that is a good spec. Add to that, if multiple text layers are selected and they use the same fonts, the font name should be shown.

DivyaPrabhakar:
@ktaki Good catch, added that case too.

iwehrman:
Even if this were crazy, I'd still merge it because of that great markdown table.

Issue #2811, opened 2015-10-05
Standard PS context menu available in mask mode


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/10295570/5c5363ba-6b76-11e5-8da2-fac38a9dfd8d.png)

pineapplespatula:
watson bug 4070997

iwehrman:
I can't repro this any longer, but I'm not sure why because that Watson bug hasn't budged...

iwehrman:
FBNC to @chadrolfs. pgdev.800 required.

PR #3027, opened 2015-10-20
disable right mouse button in vector mask mode

pineapplespatula:
fix for issue #2811 
requires jenkins build 800


iwehrman:
OK!

Issue #3021, opened 2015-10-19
Last Shift + Tab should deselect W


chadrolfs:
create a rect on canvas
click in H transform field
arrow up (verify value increments correctly)
Shift + tab to go to W
arrow up (verify value increments correctly)
Shift + tab again
arrow up 

Result: shape moves on canvas but the value in W field is still highlighted indicating it's selected.

PR #3030, opened 2015-10-20
Blur Width input when shift-tabbing beyond it

iwehrman:
What it says! Addresses #3021.


pineapplespatula:
You can no longer Shift + Arrow Up to increment this field  since pressing shift causes it to blur

iwehrman:
Derp. That is because I forgot to check for tab key. Fixed now. Back to @pineapplespatula.

pineapplespatula:
Looks good

PR #3035, opened 2015-10-20
More Contrast for Hover Color on Text/Icons 

placegraphichere:
- hover text color
- hover icon color


iwehrman:
So nice, I'd merge it twice.

Issue #2915, opened 2015-10-12
The transform system is getting borked when new artboards are created left/above


tiburbage:
1 - File > New > Artboard doc
2 - Select the artboard if it isn't already, then click on the left side "new artboard" widget
3 - Cmd+0

Notice where DS thinks the artboard is now (way offset to the left). The same thing will happen for the top, but not the right or bottom. I think we are in a pretty unstable state at this point. Repro's on Mac and Windows using the latest staging builds (10/12, late afternoon). I can switch back to standard and there is no sign of problems there, and on returning to DS, the wierd state seems to be corrected.

Please advise if on investigation you believe this to have origin on the core side and I'll open a Watson for it.

ktaki:
This issue has been reported as #2831. We still have border open policies issues.

iwehrman:
Possibly related to #2881. Can somebody try to regress this with the commit before #2881 landed to see if that broke it?

chadrolfs:
AB coords has been an issue for a while, see: #2754 #2756 and the one Taki mentioned above.

tiburbage:
I looked at Artboard bugs and didn't see anything that sounded like a dup...

iwehrman:
Eh, never mind, probably not related to #2881.

iwehrman:
This is high priority, but not a UIF stopper, so I'm swapping FIX NOW with High Pri.

iwehrman:
FBNC to @ktaki.

ktaki:
Creation of artboards to the top or left of current artboard works. There are still some undo issues that is tracked by #2831.

Issue #2831, opened 2015-10-06
Artboard coordinates don't update after undo or moving artboard after moving it's children


chadrolfs:
create a new doc
move the AB
cmd/ctrl + z

Result: coordinates are off

chadrolfs:
another way to get them separated is to drag assets into the AB, move them and then move the AB.

ktaki:
Another way is to add an artboard to the left or up.

iwehrman:
FBNC to @ktaki.

ktaki:
Works much better than before.

Over undoing will result in an error.

Create an artboard doc.
Move the artborad.
Undo
Undo

RESULT: Transfer from edit.undo to history.decrementHistory failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Select” is not currently available.

![screen shot 2015-10-20 at 4 30 46 pm](https://cloud.githubusercontent.com/assets/2029597/10623909/31ef1e7e-7748-11e5-892a-c704bc601776.png)

ktaki:
Creation of artboards towards top or left won't create frame policy issue anymore, but undoing it sometimes shows the issue.

![screen shot 2015-10-20 at 4 38 36 pm](https://cloud.githubusercontent.com/assets/2029597/10624101/39cd55aa-774a-11e5-88e0-84f16a9de423.png)

mcilroyc:
@ktaki regarding your last two comments:  

For the first one, I have found a bug with moving artboards IF it causes a canvas shift; FIP

I can not repro the second one where creating artboard + undo results in bounds issue.  If you can figure out the "sometimes" part of it... any clues appreciated.

ktaki:
One way to reproduce this is to create many artboards to the left or up, and undo one by one. For example, create 15 artboards to the left of the first artboard.

<img width="1173" alt="screen shot 2015-10-20 at 9 02 03 pm" src="https://cloud.githubusercontent.com/assets/2029597/10627451/712c1c42-776e-11e5-886f-153ef8f15580.png">

Then undo one by one. At some point, you will see something like:

<img width="1171" alt="screen shot 2015-10-20 at 9 03 09 pm" 
src="https://cloud.githubusercontent.com/assets/2029597/10627458/95a240ba-776e-11e5-8c80-b7ba4659172f.png">

[...]

PR #3036, opened 2015-10-20
Remove coord translation logic from handleCanvasShift…

mcilroyc:
…in favor if resetting all bounds, for compatibility with the history store.

This addresses #2915 and #2831 

Note that this trades some performance for correctness, and hopefully we can do better in the future.  On vermilion, this extra bounds reset costs an additional ~700ms.


iwehrman:
Seems to work well, and the performance penalty is probably as good as we can do in the short term. So, bombs away!

Issue #3029, opened 2015-10-20
Shouldn't be able to edit mask when not in mask mode


chadrolfs:
Open a test file and target a pixel layer
m to enter mask mode
r for Rect tool and drag out a rect mask
m to exit mask mode
v for Select tool
m to enter mask mode, m to exit mask mode
p for pen tool

Result: now you are able to see the mask anchor points and can edit with the pen tool (add/delete points).

iwehrman:
FBNC to @chadrolfs

PR #3037, opened 2015-10-20
 deselect paths when pen tool is not in vector mask mode

pineapplespatula:
fix for issue #3029 
requires https://github.com/adobe-photoshop/spaces-adapter/pull/159


iwehrman:
LGTM

Issue #1648, opened 2015-06-25
Instrument all external links with the Headlights API


iwehrman:
In particular, we failed to do this for the first-launch links to helpx, forums and Twitter.

volfied:
@DivyaPrabhakar Can you take care of this at some point? We have example uses of Headlights scattered across the app, so please find a fitting category/subcategory/event-name for each link.

DivyaPrabhakar:
@volfied Sure, will do

chadrolfs:
M3 if time, if not, boot to M4

DivyaPrabhakar:
@chadrolfs This can be M3

iwehrman:
Fixed by #3038.

PR #3038, opened 2015-10-20
Instrument all external links with Headlights API

DivyaPrabhakar:
References issue #1648 

Added calls to the Headlights API for the Introduction screen (Twitter, Github, and Forum) and the menu (Twitter, HelpX, and Forum) 

If we choose to merge this as is, then we need to update this wiki with the new headlights trio that is being used: https://github.com/adobe-photoshop/spaces-design/wiki/Headlights-Categories,-Subcategories,-and-Events

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -125,13 +125,16 @@ define(function (require, exports) {
     /**
      * Open a URL in the user's default browser.
      * 
-     * @param {{url: string}} payload
+     * @param {{url: string, category: string, subcategory: string, eventName: string}} payload
      * @return {Promise}
      */
     var openURL = function (payload) {
         if (!payload.hasOwnProperty("url")) {
             var error = new Error("Missing URL");
             return Promise.reject(error);
+        }
+        if (payload.category !== null && payload.subcategory !== null && payload.eventName !== null) {
+            headlights.logEvent(payload.category, payload.subcategory, payload.eventName);
         }
 
         return adapter.openURLInDefaultBrowser(payload.url);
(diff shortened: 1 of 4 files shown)


DivyaPrabhakar:
@iwehrman @mcilroyc There are no errors on this branch but I need help verifying whether the call to the API is getting logged or not. Let me know if you guys know of a way to make sure of this.

DivyaPrabhakar:
Also one question I have it: 

In src/js/util/ui.js:111 and src/js/actions/menu.js:131, do we need to make ```name``` and ```payload.category```, ```payload.subcategory```, ```payload.eventName``` required?

src/js/jsx/help/FirstLaunch.jsx
```diff
@@ -33,7 +33,7 @@ define(function (require, exports, module) {
         strings = require("i18n!nls/strings"),
         ui = require("js/util/ui"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon");
-        
+
```
iwehrman: :space_invader:

src/js/actions/menu.js
```diff
@@ -133,6 +133,8 @@ define(function (require, exports) {
             var error = new Error("Missing URL");
             return Promise.reject(error);
         }
+        
+        headlights.logEvent(payload.category, payload.subcategory, payload.eventName);
```
iwehrman: These extra parameters should be documented in the JSDoc.

iwehrman:
@DivyaPrabhakar Looks good besides a few nits. If we decided that the parameters to those actions are "required" then would other changes need to be made? We should avoid logging empty strings. So let's make the smallest additional change to enforce that.

DivyaPrabhakar:
@iwehrman Currently, the only calls to menu.openURL are in menu-actions.json for the Forum, helpX, and Twitter. There are no empty strings being logged as of now and I added the JSDoc to show that url, category, subcategory, and eventName are required so as of now so hopefully this will be indication in the future. I also added a small check to only log the event if all three fields are not null.

iwehrman:
Looks great, thanks!

Issue #2947, opened 2015-10-14
Artboard Prefix ordering with this file


placegraphichere:
Artboards have been named in order L->R top to bottom. However, export modal show a strange order when using the prefix. Layer order does not seem to explain it either. 

Shows AB 001 followed by AB 016 which doesn't really make sense even looking at the canvas.
<img width="1189" alt="screen shot 2015-10-14 at 1 39 28 pm" src="https://cloud.githubusercontent.com/assets/3229167/10493874/08256798-7279-11e5-805b-f8b6230e430d.png">

 

https://www.dropbox.com/s/l405mjchsbryz81/prefix-ordering.psd?dl=0

chadrolfs:
M3 if time permits, otherwise boot to M4 at your discretion

PR #3039, opened 2015-10-20
Artboard Sorting Issue with out-of-aligment rows

placegraphichere:
Artboards in a second row placed to the left of the first row would get the AB prefix ordering all messed up. This appears to fix that, but needs review. Fix is just adding an additional sort() to do first left-to-right, THEN sort() top-to-bottom.
- possible fix for #2947

Should fix this repro case shown here (and the file in the original issue):
screen shot 2015-10-20 at 3 44 40 pm

screen shot 2015-10-20 at 3 45 59 pm

screen shot 2015-10-20 at 3 48 36 pm

### DIFF
--- a/src/js/util/gridsort.js
+++ b/src/js/util/gridsort.js
@@ -110,6 +110,9 @@ define(function (require, exports, module) {
             .sort(function (a, b) {
                 return a.bounds.left - b.bounds.left;
             })
+            .sort(function (a, b) {
+                return a.bounds.top - b.bounds.top;
+            })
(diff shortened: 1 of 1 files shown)


mcilroyc:
Unfortunately, I think I found a counter example to your version of the sort: http://cl.ly/341a1h2o3F3y

placegraphichere:
yup, totally fails on that one. Good lookin out

placegraphichere:
not working. Will close this one.

Issue #3016, opened 2015-10-19
Sampling from a layer with one effect to a layer with multiple effects of the same type only replaces the 1st one


chadrolfs:
Create layer A with a drop shadow and 2 strokes
Create layer B with one stroke
Target layer A
select the Sampler tool
mouse over layer B, hit Space, and click on FX icon

Result: drop shadow is removed, the top stroke is replaced with the sampled stroke, and the 2nd stroke remains.

Expect: FX from layer B replace FX on layer A

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
:moneybag:

PR #3041, opened 2015-10-20
Sampling layer effects should delete extra effects.

shaoshing:
Prevously, sampling layer effects only handle the following cases:

1. delete all effects if the target has no layer effect
2. copy existing effects from source layer to target layer

This PR handles the 3rd case:

3. If target layer has more effects then the source layer, it should duplicate the existed effects from the source layer, and delete the extra effects.

Addresses #3016

### DIFF
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -410,11 +410,12 @@ define(function (require, exports) {
                     effects: effects
                 };
             });
-
+            
+        // 1. Build lists to update / insert the properties from source layer to target layers
+        // 
         // For each effectType
         //      For each effect at index i
         //          Build every layer's ID array, with index for that layer, and props for that layer's new effect
-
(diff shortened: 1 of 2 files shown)


chadrolfs:
@shaoshing does this PR also cover cases for copy/paste ( not covered in PR #3045 )?

iwehrman:
This works, but when I undo I see a number of different history states, as though the effects are being deleted/replace one at at time:
img

shaoshing:
This one is probably fixed in https://github.com/adobe-photoshop/spaces-design/pull/3062. but I will double check with that fix in this PR

shaoshing:
@iwehrman confirmed that it is fixed by #3062

shaoshing:
Rebased master with the history state fix, @iwehrman you should be able to undo the sampler effects now.

iwehrman:
OK! Looks great.

Issue #3022, opened 2015-10-19
Transform > Rotate 180/Right/Left on an empty layer results in error


chadrolfs:
cmd/ctrl + n (by default empty layer is selected)
Layer > Transform > Rotate 180/Right/Left

Result:  `Transfer from transform.rotateLayersInCurrentDocument to transform.rotate failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -1 Photoshop message: error: -1 message: Could not complete the command because the initial bounding rectangle is empty.`

chadrolfs:
Same thing happens on an empty group

mcilroyc:
fbnc2 @chadrolfs

PR #3043, opened 2015-10-20
Filter more layers in transform commands

iwehrman:
1. Filter empty layers (and adjustment layers and background layers) before playing `rotate` and `setRadius` actions. This addresses #3022.
2. Move all transform filtering into one helper function which filters empty layers, adjustment layers, backgrounds and (optionally) empty artboards.

As with #3007, this doesn't affect the UI, so some menu items or UI elements may still be enabled when the filtering will otherwise reduce the command to a no-op. This is something that it would be great to have a higher level, more unified solution for later, but for now this at least helps us avoid having unhandled exceptions.

Addresses #3022.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -60,6 +60,34 @@ define(function (require, exports) {
     };
 
     /**
+     * Filter out layers that can't generally be transformed like empty,
+     * adjustment and background layers.
+     *
+     * @private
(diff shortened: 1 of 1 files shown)


mcilroyc:
looks good, but needs rebase

iwehrman:
Rebased and ready 2 rock :rage3:

Issue #2755, opened 2015-10-02
Copy Style (Appearance) should include Opacity, Blend and Radius (if applicable)


chadrolfs:

chadrolfs:
expectation based on user feedback

iwehrman:
@placegraphichere: do you agree?

shaoshing:
Just encountered this. I can take this if we decide to include these attributes.

chadrolfs:
Do it!

shaoshing:
When #3100 is fixed, this should become FNBC

PR #3045, opened 2015-10-20
Fix for copy/paste buttons in appearance and effects panels

shaoshing:
This PR makes two changes to the copy/paste buttons:

1 copy/paste buttons in the appearances now include opacity, radius, and blend mode. This should apply for text layer, vector layer, and other layers (e.g. SO). You can checkout the video for more details of the bahaviors: https://www.dropbox.com/s/lao6v0xo6nezt0g/copy-styles.mov?dl=0. Addresses #2755 

2 copy/paste buttons in the effects panel should only apply to layer effects (used to include appearance styles). Same for the appearance panel, buttons should not copy/paste layer effects. 

###### Known issues that are not introduced from this PR (I will create separate issues later)

1 copy/paste appearance between two text layers with different opacity result in number change back and forth.

text-styles

2 undo paste-styles some times does not reset the appeanrace panel

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1544,19 +1544,22 @@ define(function (require, exports) {
      * @param {Document} document
      * @param {Immutable.Iterable.<Layer>} layers
      * @param {number} opacity Opacity as a percentage
-     * @param {boolean=} coalesce Whether to coalesce this operation's history state
+     * @param {object} options
+     * @param {boolean=} options.coalesce Whether to coalesce this operation's history state
      * @return {Promise}
      */
-    var setOpacity = function (document, layers, opacity, coalesce) {
+    var setOpacity = function (document, layers, opacity, options) {
         layers = layers.filterNot(function (layer) {
             return layer.isBackground;
         });
+        
+        options = _.merge({ coalesce: false }, options);
 
         var payload = {
                 documentID: document.id,
                 layerIDs: collection.pluck(layers, "id"),
                 opacity: opacity,
-                coalesce: !!coalesce
+                coalesce: !!options.coalesce
             },
             playObjects = layers.map(function (layer) {
                 var layerRef = [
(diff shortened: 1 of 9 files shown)


shaoshing:
@chadrolfs or @ktaki, when you have time, can you give this branch a try? I am not confident enough that I've tested all dark corners :P

src/js/actions/layers.js
```diff
@@ -1544,19 +1544,22 @@ define(function (require, exports) {
...
+     * @param {boolean=} options.coalesce Whether to coalesce this operation's history state
      * @return {Promise}
      */
-    var setOpacity = function (document, layers, opacity, coalesce) {
+    var setOpacity = function (document, layers, opacity, options) {
         layers = layers.filterNot(function (layer) {
             return layer.isBackground;
         });
+        
+        options = _.merge({ coalesce: false }, options);
```
iwehrman: I guess it doesn't really matter, but why is this line necessary? If you're doing `!!options.coalesce` below, this seems like a no-op.

iwehrman: Oh, maybe because you want a fresh object reference below. It still seems like the first param could be an empty object literal, but again it probably doesn't really matter.

iwehrman:
I do not see anything objectionable here! If @chadrolfs or @ktaki can do some functional testing, I'm happy to merge.

iwehrman:
Tentatively assigning to @chadrolfs for this, but feel free to bounce it elsewhere if that doesn't work.

chadrolfs:
Logged, #3049 and #3050

chadrolfs:
Looks like there is a bug similar to #3016 

shape A has (CC Library "layer effects": effects - a lot), shape B has one drop shadow
copy Effects from B, select A, paste
result, 2 drop shadows

chadrolfs:
^^ is fixed by PR 3041.  Merging.

Issue #3019, opened 2015-10-19
Unable to change position from the transform panel when a layer is locked


shaoshing:
Changing the X / Y does not update the layer's positon.

chadrolfs:
Shaoshing Lee: I found the cause but not sure how to fix
https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/actions/transform.js#L403

PR #3047, opened 2015-10-20
Fix updating position of locked layers

shaoshing:
Updating position from the transform panel is broken for a locked layer (#3019). I traced down our commits and found the cause: we set to run the descriptor asynchronously for `transforms.setPosition`, and it might break the lock overwrite mechanism (position is set before the layer is unlocked). The change was introduced from #2626, however, I am not familiar with transform enough to figure out the purpose of this change. I marked this for discussion, and let me know if you know the purpose of the `synchronous: false` setting, or the way to work around this in lock overwrite mode.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -423,12 +423,7 @@ define(function (require, exports) {
                 historyStateInfo: {
                     name: strings.ACTIONS.SET_LAYER_POSITION,
                     target: documentLib.referenceBy.id(document.id)
-                },
-                // Setting this to false allows PS to send us
-                // notifications that happen as a result of this
-                // action, in conjunction with `suppresPlayLevelIncrease`
-                // flag in the descriptor
-                synchronous: false
+                }
             };
 
         var dispatchPromise = this.dispatchAsync(events.document.history.optimistic.REPOSITION_LAYERS, payload)


pineapplespatula:
synchronous false allows CEF to continue running after the event is played, but before it resolves. @iwehrman  do you remember which bug we were seeing a frozen CEF with?

iwehrman:
It looks like that change came from @volfied in #2626. I'm guessing it had to do with making sure that we got auto-canvas-shift events when transforming sizes to ensure that, if the canvas expands, we know to update our bounds. @mcilroyc has drastically simplified this lately, but I'm not aware that his changes would obviate this.

One crazy idea would be to manually play the unlock / play / relock actions in separate play calls instead of jamming them all into one batchPlay call. 

CC @jsbache: here's a concrete use case for a per-command `synchronous` option.

iwehrman:
Short term I guess you can _try_ removing it, but you'll need to confirm that if you, e.g., resize layers in such a way that the canvas expands that somehow we're still getting the canvas-shift event we need. Or you can try my crazy idea above, which would cause _only_ the setSize command to be played synchronously. Or we can just defer this.

mcilroyc:
chiming in mostly just to say that I don't really understand what's going on here, specifically that first call with a bare payload `this.dispatchAsync(events.document.history.optimistic.REPOSITION_LAYERS, payload)`

Anyway, upon quick test, this branch seems to work.  My gut says that if this simple fix does *not* work, then it should be deferred.

iwehrman:
Agree with @mcilroyc that that empty array in the payload is weird.

Also, this bug is much worse than we realized, as mentioned in #3144. I guess now I do think we should try to rip out this `synchronous: false` option. I'm having a hard time figuring out exactly which bug it's presence originally solved though. My best guess is that it had something to do with layer movement causing an `autoCanvasShift` event, which was needed to ensure that layer bounds were updated. If that's the case, then I wonder if we can just call do what that handler does now instead, which is to just reset all bounds. (Maybe there is even something more efficient we can do, like only reset bounds when moving a layer to coordinate partially outside the document bounds? Might need to check with @timothynoel about that.)

iwehrman:
Well, my hypothesis may be incorrect because we still seem to get the `autoCanvasShift` events even without `synchronous: false`. So for all I know this PR is ready to go, but it's going to be hard to merge it unless we understand the original reasoning behind the `synchronous: false` flag.

chadrolfs:
this bug occurs on this branch: #2831

chadrolfs:
attempting to paste text from clipboard into a modal text layer errors and fails as well:
`Transfer from policy.suspendAllPolicies to policy.suspendPolicies failed: Error: Policies are already suspended`

copy any text (from an email)
enter DS, cmd/ctrl + n
t for Type tool, click on canvas, cmd/ctrl + v
result: fail (error above)

mcilroyc:
regarding the most recent comment, maybe this branch just needs to be rebased to master to pick up that related fix?

chadrolfs:
was wondering the same thing

chadrolfs:
rebase appears to have resolved the issues I was seeing above.  continuing to test...

mcilroyc:
let's do this

Issue #3040, opened 2015-10-20
Revert should exit mask mode


chadrolfs:
open a saved doc,
target a pixel layer, hit m to enter mask mode
r for Rect tool and drag out a rect mask
hit F12 or File > Revert
drag out another rect

Result: F12/Revert does not exit Mask Mode and rects are now created as paths instead of shapes

chadrolfs:
Review all menu commands to make sure their safe or not and disable accordingly.

iwehrman:
FBNC to @chadrolfs.

PR #3048, opened 2015-10-20
disable revert while in vector mask mode

pineapplespatula:
fix for issue #3040


iwehrman:
Could you also do a quick pass over all the other entries in menu-actions to see which ones _must_ be disabled in mask mode? (A best guess is better than nothing.)

pineapplespatula:
I did a quick pass. (certainly not exhaustive) but  I added a few more that needed to be disabled.

iwehrman:
Great, thanks.

PR #3051, opened 2015-10-21
move layer history fix

mcilroyc:
This is a follow-up fix for #2831

1. Allow handleCanvasShift action during modal state.  When dragging a layer, a canvas shift may occur during the drag.  It seems safe to me to allow that action to get queued up without breaking modal state.
1. A narrow fix of the history amendment process which should *not* overwrite the "rogue" flag (set when we get a history state event for an action that photoshop initiates).  The bug here is that if that rogue flag gets reset, then our follow-up action won't know it is safe to amend the roguely-created history state.


pineapplespatula:
LGTM

Issue #2004, opened 2015-08-12
Touch gesture undo does not update layers panel 


chadrolfs:
new doc, drag out several shape layers
perform touch gesture: slide 3 fingers on surface to left to undo until one or no shapes show on canvas.

Result: layer panel doesn't update/show results of undo

mcilroyc:
Created a core request to provide better event info when an out-of-bands undo occurs.  https://trello.com/c/PTpVWcx5

chadrolfs:
self-assigning to re-check since more touch stuff has come in from main and cory asked me to.  :)

chadrolfs:
Confirmed, layers panel does not update on undo/redo surface gestures (2- or 3- finger).

iwehrman:
When I slide three fingers to the left, the canvas disappears and I never see the touch UI... Windows pgdev.715.

mcilroyc:
Hey @chadrolfs I warned you I'd come-a-askin bout this.  I think you agreed to help.  Can you please run the first (all-events) snippet in dev tools listed on [this wiki page](https://github.com/adobe-photoshop/spaces-design/wiki/DevTools-Snippets), and then send the output when you swipe-to-undo on a touch device?  I'm looking to see what events PS throws at us.

chadrolfs:
yessir

chadrolfs:
![screenshot_fri_oct_09_16 18 42](https://cloud.githubusercontent.com/assets/8442631/10407553/b0ac7dce-6ea1-11e5-8059-45190f0a891b.png)
![Uploading screenshot_Fri_Oct_09_16.19.31.png…]()

iwehrman:
FBNC to @chadrolfs. I tested 3-finger-drag and 2-finger-swipe with success. Note that it ONLY works if the gestures happen on the PS canvas and not on the HTML UI.

chadrolfs:
ahhhh snap, it's merged now!  It's on!  Cool!

PR #3052, opened 2015-10-21
Handle step back/forward events that get initiated in photoshop

mcilroyc:
Handle touch-initiated time travel, by ~~attempting to load the given prev/next history state.~~ nuking history and updating the document.  This attempts such a maneuver only if a history state "select" event is soon followed by a historyState event for a prev/next history state.  These actions are also debounced so that 3-finger scrubbing will result in fewer updateDocument calls, hopefully just once when the scrubbing subsides.

The diff is a bit messy because I moved `handleHistoryState` further down in the file for organizational reasons, and factored out a `_loadHistory` helper function.

The crux of this PR is a short timer on the "select" handler (history state only) that, if it times out, will update the document.  Assuming that a historyState event follows quickly thereafter (<700ms) then it will  ~~attempt to navigate ONE step forward or ONE step back~~ nuke and pave.

This is not particularly pretty, but it should be limited to touch events, as we are not aware of other times that we would get history state select events.

~~It is very possible that scrubbing quickly through history states with touch will cause DS to freak out a little~~.  On net, I think this PR at least handles the simple case by keeping our model in sync.

A "best we can do" attempt at #2004

### DIFF
--- a/src/js/actions/history.js
+++ b/src/js/actions/history.js
@@ -77,51 +77,54 @@ define(function (require, exports) {
     queryCurrentHistory.modal = true;
 
     /**
-     * Given a history state event from photoshop, dispatch a flux event for the history store
+     * Helper function to load a state from history store based on a count offset
+     * and then to fix it up with the latest selection state, visibility state, and
+     * then reset border policies
      *
-     * @param {object} event a raw historyState event from photoshop
+     * @private
+     * @param {number} documentID
+     * @param {number} count
      * @return {Promise}
      */
-    var handleHistoryState = function (event) {
-        // This assumption of current document could be problematic
-        // but would require a core change to include document ID
-        var currentDocumentID = this.flux.store("application").getCurrentDocumentID();
+    var _loadHistory = function (documentID, count) {
+        return descriptor.getProperty(documentLib.referenceBy.id(documentID), "targetLayers")
+            .bind(this)
+            .catch(function () {
+                // no targetLayers property means no document is open
+                return [];
+            })
+            .then(function (targetLayers) {
+                var selectedIndices = _.pluck(targetLayers, "_index");
+                return this.dispatchAsync(events.history.LOAD_HISTORY_STATE, {
+                    documentID: documentID,
+                    count: count,
+                    selectedIndices: selectedIndices
+                });
+            })
+            .then(function () {
+                var currentDocument = this.flux.store("application").getCurrentDocument(),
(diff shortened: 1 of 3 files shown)


src/js/actions/history.js
```diff
@@ -77,51 +77,54 @@ define(function (require, exports) {
...
+     * and then to fix it up with the latest selection state, visibility state, and
+     * then reset border policies
      *
-     * @param {object} event a raw historyState event from photoshop
+     * @private
+     * @param {number} documentID
+     * @param {number} count
      * @return {Promise}
      */
-    var handleHistoryState = function (event) {
```
mcilroyc: This action was mostly just moved down in the file, but the "id" was also added to the payload.

src/js/actions/history.js
```diff
@@ -77,51 +77,54 @@ define(function (require, exports) {
...
-    handleHistoryState.writes = [locks.JS_HISTORY];
-    handleHistoryState.modal = true;
 
     /**
      * Go forward or backward in the history state by playing the appropriate photoshop action
      * and either loading a state from the history store's cache, or calling updateDocument
      *
      * @private
-     * @param {Document} document
+     * @param {number} documentID
```
mcilroyc: For clarity, changed a this to require only the ID because that's all it needs

mcilroyc:
@chadrolfs @iwehrman This is my "touch undo" PR.  I'm on the fence.  I think it is "better than what we have" and "generally safe", but it is also 1) effectively untested by me (hardware), 2) Known to be buggy with three-finger-history scrub.  Thoughts?

chadrolfs:
let me do another pass on it today on my surface and desktops to see if there is anything really bad.

chadrolfs:
after cory updated this branch to resolve rebase conflicts, layers panel no longer updates after gesture undo (2 or 3 finger).

iwehrman:
I'll test this when I'm back in the office tomorrow.

mcilroyc:
I pushed a second commit which gets rid of some of the smartness, and instead just debounces a call to nuke history and updateDocument.  Hopefully this will at least be safe for *rapid* undo/redo via touch, even though it is lame and slow in general.  I assigned it to @iwehrman so as to take him up on his offer to test it out.  But @chadrolfs FYI.

iwehrman:
Seems to work OK, and the patch looks about as gross as I'd expect!

Issue #2340, opened 2015-09-03
Time out error when exporting a 2.5 M JPEG file to PNG


ktaki:
Scale factors other than 1 failed. Failures were indicate by red ink in the Export dialog. It looks like the files were generated successfully.
```
export.js:163 Export Failed for asset 2 of layerID 2, documentID 313, with error 
Error: Generator call exportLayer has timed out {
  stack: "Error: Generator call exportLayer has timed out↵  …r_components/react/react-with-addons.js:16633:20)", 
  message: "Generator call exportLayer has timed out",
  __stackCleaned__: true
}
```

![img_0759](https://cloud.githubusercontent.com/assets/2029597/9672851/0396230c-5253-11e5-9268-dd4aa11ce1d1.jpeg)

mcilroyc:
quick note on this: We will soon be introducing a better feedback mechanism for "in progress" export, and will subsequently remove or significantly lengthen the timeout.

ktaki:
Got it. Keep this open for now. Thanks.

mcilroyc:
yes, for sure.  *It is a bug* I agree!

iwehrman:
FBNC to @ktaki

ktaki:
@mcilroyc Does this timeout occur per export session (i.e, all assets shown in Export dialog) or per an exporting asset? More assets you have, more likely you get timeout?

mcilroyc:
@ktaki : This timeout applies individually to each individual asset export call.  If you export two assets A and B, it is possible for B to complete successfully but A to timeout after its 2 minutes are up.  

But, since we make all the individual export calls for a "session" as quickly as possible (without waiting for the previous to complete), then it is possible that a session with a lot of assets will flood the queue and cause a greater chance for any one of them to timeout.

ktaki:
@mcilroyc I see. I saw cases if you export single asset, it is successful, but with multiple assets, exporting large assets (2x 3x) are failing. Since this operation is asynchronous, what is disadvantages of setting longer timeout vs short one?

mcilroyc:
You've seen them take longer than 2 minutes!? 

It is not unreasonable to increase the timeout.  As you said, the process is asynchronous.  Because we don't yet provide a "cancel" option, then I was trying to guard against the user being "locked out" of exports for an extended period because of some rogue long-running export.

What is your suggestion for a reasonable timeout value?

ktaki:
[...]

PR #3053, opened 2015-10-21
Increase the timeout for export ws calls to two minutes

mcilroyc:
Previous was 20 seconds, and it was fairly easy to cause this to bonk out for large exports.  For example:

1. Open a large JPG file in DS (my test case was 7MB)
1. Unlock the background layer, and add a PNG (default) asset
1. Export it - and note that you might see a console "timeout" message

Addresses (or reduces the frequency of) #2340


iwehrman:
Two minutes ought to be enough for anyone.

Issue #3033, opened 2015-10-20
Switching tools in mask mode with multi-layer selection errors


chadrolfs:
Open a test file with multiple layers and select two layers 
or 
cmd/ctrl + n
drag a cc library asset onto canvs
hold shift and click on Layer 1 so that both layers are selected
(by default Select tool is active)
hit m to enter mask mode
hit r to switch to Rect tool

Result: `Action tools.select failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “<unknown>” is not currently available.`

chadrolfs:
![screen shot 2015-10-20 at 11 41 58 am](https://cloud.githubusercontent.com/assets/8442631/10617433/681fd098-771f-11e5-8082-f7a3ba3170db.png)

pineapplespatula:
I cannot repro this in master. I'm not sure what changed.

pineapplespatula:
nevermind.. got it

iwehrman:
Back to @chadrolfs to verify.

PR #3054, opened 2015-10-21
vector mask with multiple layer selections now uses the selection dance 

pineapplespatula:
fix for issue #3033


iwehrman:
Looks good!

Issue #3034, opened 2015-10-20
Copying Effects then Undo does not always update canvas


placegraphichere:
- create two rects
- add multiple effects to one rect
- select the other rect and use the sampler to copy the effects
- CMD+Z to undo
- Result: Often the effects are not deleted from canvas, but nothing is in panel
- *Interestingly, if you save and reopen the doc, the effects will still be there.*

Example:
https://www.dropbox.com/s/93wd0bq4iv5ph80/undo-not-removing-canvas-effects.mov?dl=0

dev806

placegraphichere:
Continuing to undo will finally get you results, but it seems incorrect that one action does not equal one undo step.

chadrolfs:
Looks like there are multiple history states here.  If you do Update Document after undo, you'll see all the styles re-populate

mcilroyc:
FBNC to @chadrolfs for validation

PR #3062, opened 2015-10-21
Fix issues caused by merging on undefined/null options using `_.merge`.

shaoshing:
Instead of merging on action options, we should merge on a plain object for the following reasons:

1. the option could be `undefined`, and `_merge(options, {enabled: true})` will return `undefined`, instead of `{enabled: true}`

2. merging on the `options` directly will pollute the object, and we should avoid doing that as the `options` may be shared between multiple actions. for example: https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/actions/sampler.js#L583

So a recommneded / safer pattern would be:

```
_.merge({}, options, {enabled: true})
```

addresses issue #3034

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1553,7 +1553,7 @@ define(function (require, exports) {
             return layer.isBackground;
         });
         
-        options = _.merge({ coalesce: false }, options);
+        options = _.merge({}, options);
 
         var payload = {
                 documentID: document.id,
--- a/src/js/actions/layereffects.js
+++ b/src/js/actions/layereffects.js
@@ -68,7 +68,7 @@ define(function (require, exports) {
             },
             types = _.isArray(type) ? type : [type];
 
-        options = _.merge(options, syncOptions);
+        options = _.merge({}, options, syncOptions);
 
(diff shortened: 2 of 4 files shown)


src/js/actions/layers.js
```diff
@@ -1553,7 +1553,7 @@ define(function (require, exports) {
             return layer.isBackground;
         });
         
-        options = _.merge({ coalesce: false }, options);
+        options = _.merge({}, options);
```
mcilroyc: At first glance this seems like it might change the behavior ... but I guess maybe you confirmed is that there is no reason to supply a "default value" of false *here* because that is the default option further down the stack?

mcilroyc:
Just a quick question because we're in careful mode, but otherwise mergable

src/js/actions/layers.js
shaoshing: Sorry, I should've mentioned this to avoid confusion. these default options were added by me while I was working on another issue, and apparently we don't need them because we use `!!options. coalesce` to deal with `undefined` case below.

shaoshing:
@mcilroyc thanks for reviewing. Answered your question. Now back to you.

Issue #3057, opened 2015-10-21
mergeDeep call in resetBounds is slow, possibly unnecessary


iwehrman:
Here: https://github.com/adobe-photoshop/spaces-design/blob/e9b0ef739a3b864ebd8b9873701c27271c3a15de/src/js/models/layerstructure.js#L1092

In Vermillion, changing a font results in a call to `layers.resetBounds`, and it _seems_ like it's spending ~300ms inside this `mergeDeep` call.

shaoshing:
@iwehrman confirmed the test result in the PR, so issue closed.

PR #3064, opened 2015-10-21
Replace mergeDeep with merge in Layerstructure.prototype.resetBounds

iwehrman:
This replaces an apparently unnecessary `mergeDeep` call with a faster `merge` call in `Layerstructure.prototype.resetBounds`. Changing a font in Vermilion previously looked like this:
image

Now it looks like this:
image

That is, the time is now dominated by React instead of Immutable, saving ~250ms.

Addresses #3057.

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -1089,9 +1089,7 @@ define(function (require, exports, module) {
             }, this);
         }.bind(this));
 
-        return this.mergeDeep({
-            layers: nextLayers
-        });
+        return this.set("layers", nextLayers);
     };
 
     /**


src/js/models/layerstructure.js
```diff
@@ -1089,7 +1089,7 @@ define(function (require, exports, module) {
             }, this);
         }.bind(this));
 
-        return this.mergeDeep({
+        return this.merge({
             layers: nextLayers
         });
```
shaoshing: In this case, we need `replace` more than `merge`, as nextLayers is a complete set of next layers

```
this.set("layers", nextLayers)
```

If `nextLayers` is a sub-set of `this.layers`, then we should use `merge` on `this.layers`, otherwise `this.layers` will be replaced by an incomplete list of layers.

```
this.set("layers", this.layers.merge(nextLayers))
```

mcilroyc: that said, I'd prefer to see this converted to *not* use `withMutations`, which would then get us back to using `merge`.  Maybe that's too drastic a change for now.

iwehrman:
I was trying to make the smallest possible change for now. I agree `withMutations` should go, but let's do that later. For now, `set` is functionally equivalent to `merge`, but probably a touch faster, so I've updated the PR with that change. I also filed #3066 for later discussion. Back to @shaoshing.

shaoshing:
Perfect!

PR #3067, opened 2015-10-21
New variant of RESET_BOUNDS that amends history

mcilroyc:
This is used now by the call to `resetBounds` from `handleCanvasShift`.  This addresses bounds-separation-after-undo issues caused in some narrow cases when adding an artboard that causes a canvas shift, followed by one that does *not* cause canvas shift.  More discussion in issue #2831, which this PR is addressing.  again.

Note that I also added two unrelated "amendment" variants of events to menu store which I believe were omissions.


src/js/actions/layers.js
```diff
@@ -1642,8 +1647,8 @@ define(function (require, exports) {
      * though possibly out of order w.r.t. Photoshop's model.
      *
      * @param {Document=} document Document for which layers should be reordered, if undefined, use current document
-     * @param {boolean=} suppressHistory if truthy, dispatch a non-history-changing event.
-     * @param {boolean=} amendHistory if truthy, update the current state (requires suppressHistory)
+     * @param {boolean=} suppressHistory Optional. If true, emit an event that does NOT add a new history state
```
mcilroyc: reviewer note: I just changed this JS Doc to have the same wording as resetBounds because the two both use the same basic pattern.

pineapplespatula:
LGTM

PR #3070, opened 2015-10-21
Fixes document title scrolling behavior

designedbyscience:
The document that is currently selected should scroll into view in the document header. It wasn't and now it is. Also made it work more betterer at the expense of the getBoundingRect call


shaoshing:
LGTM!

shaoshing:
@designedbyscience does this address any issues?

designedbyscience:
Should be the reopened version of this bug: #2558

PR #3073, opened 2015-10-21
Adding calls to Headlights API

DivyaPrabhakar:
https://trello.com/c/jg3fH91Q/599-analytics-todo

Update: 

Currently Tracking:

*Search*
- What category was searched and if filters were used or not. Categories include: menu commands, layers (including the exact type), recent documents, current documents. 
- When a failed search occurs including the search term of the failed search. (This can be optimized in the future) 

*Export*
- When a preset is used and which one is used
- When user deletes an asset 
- When user creates a single asset 
- When user exports using the panel
- When user exports from the dialog
- Quality/format/setting of every asset exported 

*Swap*
- Artboards are swapped 
- Non-artboard layers are swapped
- Layers are flipped horizontally 
- Layers are flipped vertically 

*UI*
- When user collapse or expands a panel (2 column mode) and which panels are currently open at that time
- User goes back to PS 
- When user changes to 1 column mode or back to 2 column mode

*Sampler HUD*
- Layer effects applied via the Sampler HUD
- Graphic sampled via the Sampler HUD
- TypeStyle sampled via the Sampler HUD
- Stroke sampled via the Sampler HUD
- Fill color sampled via the Sampler HUD
- Sampler HUD invoked 

*Appearance*
- When color is changed
- When layer blend mode is changed and used
- When stroke color is changed
- When opaque color is changed
- When stroke is enabled and disabled
- Paste all styles is used in the Appearance Panel
- When alpha is changed
- When opacity is changed

*Tools*
- Tool used to create a layer 

*Effects*
- When an effect is added of any type (not tracking unsupported right now) 
- When style has been pasted via the icons in the panel
- When a blendmode is changed on a stroke and which one is used
- When user deletes a stroke
- When color-overlay is deleted
- Blendmode changed on a color-overlay and to what 
- Blendmode changed on a shadow and to what 
- When unsupported effect is deleted
- When shadow is deleted

*Mask*
- When the mask mode is enabled, what tool is selected, and the kind of layer

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -49,6 +49,7 @@ define(function (require, exports) {
         strings = require("i18n!nls/strings"),
         layerActionsUtil = require("js/util/layeractions"),
         system = require("js/util/system"),
+        headlights = require("js/util/headlights"),
         utilShortcuts = require("js/util/shortcuts"),
         EventPolicy = require("js/models/eventpolicy"),
         PointerEventPolicy = EventPolicy.PointerEventPolicy;
@@ -753,6 +754,8 @@ define(function (require, exports) {
                 });
             }
         }
+
+        headlights.logEvent("tools", "mask-mode-" + currentLayer.kind);
     };
     changeVectorMaskMode.reads = [locks.JS_APP, locks.JS_TOOL, locks.PS_DOC, locks.PS_TOOL];
     changeVectorMaskMode.writes = [locks.JS_TOOL, locks.PS_DOC];
--- a/src/js/jsx/IconBar.jsx
+++ b/src/js/jsx/IconBar.jsx
@@ -34,6 +34,7 @@ define(function (require, exports, module) {
     var Button = require("jsx!js/jsx/shared/Button"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon"),
         menu = require("i18n!nls/menu"),
+        headlights = require("js/util/headlights"),
         synchronization = require("js/util/synchronization");
 
     var PanelSet = React.createClass({
@@ -110,6 +111,7 @@ define(function (require, exports, module) {
          */
         _handleBackToPSClick: function () {
             this.getFlux().actions.menu.native({ commandID: 5999 });
+            headlights.logEvent("user-interface", "switch", "back-to-PS");
         }
     });
 
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -49,7 +49,8 @@ define(function (require, exports, module) {
         collection = require("js/util/collection"),
         strings = require("i18n!nls/strings"),
         synchronization = require("js/util/synchronization"),
-        system = require("js/util/system");
+        system = require("js/util/system"),
+        headlights = require("js/util/headlights");
 
     var PanelSet = React.createClass({
         mixins: [FluxMixin, StoreWatchMixin("application", "document", "preferences")],
@@ -206,6 +207,7 @@ define(function (require, exports, module) {
                 components = uiStore.components,
                 modifierState = modifierStore.getState(),
                 swapModifier = system.isMac ? modifierState.command : modifierState.control,
+                currentlyVisible,
                 nextState = {};
                 
             if (swapModifier) {
@@ -214,6 +216,28 @@ define(function (require, exports, module) {
             } else {
                 nextState[columnName] = !this.state[columnName];
             }
+
+            // For headlights
+            var layersLibPanel = this.state[components.LAYERS_LIBRARY_COL],
+                propertiesPanel = this.state[components.PROPERTIES_COL];
+
+            if (columnName === "layersLibrariesVisible") {
+                layersLibPanel = nextState[components.LAYERS_LIBRARY_COL];
+            } else {
+                propertiesPanel = nextState[components.PROPERTIES_COL];
+            }
+
+            if (layersLibPanel && propertiesPanel) {
+                currentlyVisible = "layersLibrary-properties-visible";
+            } else if (layersLibPanel) {
+                currentlyVisible = "layersLibrary-visible";
+            } else if (propertiesPanel) {
+                currentlyVisible = "properties-visible";
+            } else {
+                currentlyVisible = "none-visible";
+            }
+
+            headlights.logEvent("user-interface", "panels-visible", currentlyVisible);
 
             this.getFlux().actions.preferences.setPreferences(nextState);
         },
--- a/src/js/jsx/help/FirstLaunch.jsx
+++ b/src/js/jsx/help/FirstLaunch.jsx
@@ -31,7 +31,6 @@ define(function (require, exports, module) {
 
     var Carousel = require("jsx!js/jsx/shared/Carousel"),
         strings = require("i18n!nls/strings"),
-        ui = require("js/util/ui"),
         SVGIcon = require("jsx!js/jsx/shared/SVGIcon");
  
     var FirstLaunch = React.createClass({
@@ -60,6 +59,22 @@ define(function (require, exports, module) {
             if (_.isFunction(this.props.dismissDialog)) {
                 this.props.dismissDialog(event);
             }
+        },
+
+        /**
+         * Calls for the URL to be opened and then logs the headlights data
+         *
+         * @private
+         * @param {string} url
+         * @param {string} name
+         */
+        _handleClick: function (url, name) {
+            this.getFlux().actions.menu.openURL({
+                category: "user-interface",
+                subcategory: "introduction-link",
+                eventName: name,
+                url: url
+            });
         },
 
         render: function () {
@@ -105,7 +120,7 @@ define(function (require, exports, module) {
                             <ul className="carousel__slide__three__list">
                                 <li>
                                     <div
-                                        onClick={ui.openURL.bind(null, psDesignTwitterURL, "twitter")}>
+                                        onClick={this._handleClick.bind(this, psDesignTwitterURL, "twitter")}>
                                         <SVGIcon
                                             CSSID="bird"/>
                                         <h2>{strings.FIRST_LAUNCH.SLIDES[7].BODY_FIRST}</h2>
@@ -115,7 +130,7 @@ define(function (require, exports, module) {
                             <ul className="carousel__slide__three__list">
                                 <li>
                                     <div
-                                        onClick={ui.openURL.bind(null, githubURL, "github")}>
+                                        onClick={this._handleClick.bind(this, githubURL, "github")}>
                                         <SVGIcon
                                             CSSID="github"/>
                                         <h2>{strings.FIRST_LAUNCH.SLIDES[7].BODY_SECOND}</h2>
@@ -125,7 +140,7 @@ define(function (require, exports, module) {
                             <ul className="carousel__slide__three__list">
                                 <li>
                                     <div
-                                        onClick={ui.openURL.bind(null, psForumURL, "forum")}>
+                                        onClick={this._handleClick.bind(this, psForumURL, "forum")}>
                                         <SVGIcon
                                             CSSID="workspace"/>
                                         <h2>{strings.FIRST_LAUNCH.SLIDES[7].BODY_THIRD}</h2>
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -36,6 +36,7 @@ define(function (require, exports, module) {
     var collection = require("js/util/collection"),
         strings = require("i18n!nls/strings"),
         Datalist = require("jsx!js/jsx/shared/Datalist"),
+        headlights = require("js/util/headlights"),
         Button = require("jsx!js/jsx/shared/Button");
 
     /**
@@ -408,11 +409,26 @@ define(function (require, exports, module) {
                 return;
             }
 
+            var filters = this.state.filter,
+                filtersString,
+                idSplit = id.split("-"),
+                // Possible current types are Menu_Command, Current_Doc, Recent_Doc, All_Layers
+                type = idSplit[0],
+                category = "category-" + type;
+
+            // Seeing if filters are active or not for analytics
+            if (filters.length === 0) {
+                filtersString = "filter-inactive";
+            } else {
+                filtersString = "filter-active";
+            }
+            
             if (id !== PLACEHOLDER_ID) {
                 if (_.contains(this.state.filterIDs, id)) {
                     this._updateFilter(id);
                 } else {
                     this.props.executeOption(id);
+                    headlights.logEvent("search", filtersString, category);
                 }
             }
         },
@@ -446,6 +462,9 @@ define(function (require, exports, module) {
                     break;
                 }
                 case "Escape":
+                    if (this.refs.datalist.getSelected() === PLACEHOLDER_ID) {
+                        headlights.logEvent("tools", "search", "failed-search");
+                    }
                     this.props.dismissDialog();
                     return;
                 case "Backspace": {
--- a/src/js/jsx/sections/export/ExportAllPanel.jsx
+++ b/src/js/jsx/sections/export/ExportAllPanel.jsx
@@ -41,6 +41,7 @@ define(function (require, exports, module) {
         svgUtil = require("js/util/svg"),
         collection = require("js/util/collection"),
         gridSort = require("js/util/gridsort"),
+        headlights = require("js/util/headlights"),
         ExportAsset = require("js/models/exportasset");
 
     /**
@@ -181,6 +182,8 @@ define(function (require, exports, module) {
                         exportDisabled: false
                     });
                 });
+
+            headlights.logEvent("export", "export-selected", "export-dialog");
         },
 
         /**
@@ -202,6 +205,7 @@ define(function (require, exports, module) {
          */
         _setUseArtboardPrefix: function (event, enabled) {
             this.getFlux().actions.export.setUseArtboardPrefix(enabled);
+            headlights.logEvent("export", "batch-export", "prefix-enabled-" + enabled);
         },
 
         render: function () {
--- a/src/js/jsx/sections/export/ExportList.jsx
+++ b/src/js/jsx/sections/export/ExportList.jsx
@@ -38,6 +38,7 @@ define(function (require, exports, module) {
 
     var mathUtil = require("js/util/math"),
         collection = require("js/util/collection"),
+        headlights = require("js/util/headlights"),
         strings = require("i18n!nls/strings");
 
     /**
@@ -97,6 +98,8 @@ define(function (require, exports, module) {
                 index = this.props.index;
 
             this.getFlux().actions.export.deleteExportAsset(document, layers, index);
+
+            headlights.logEvent("export", "delete", "single-asset");
         },
 
         /**
--- a/src/js/jsx/sections/export/ExportPanel.jsx
+++ b/src/js/jsx/sections/export/ExportPanel.jsx
@@ -40,6 +40,7 @@ define(function (require, exports, module) {
         strings = require("i18n!nls/strings"),
         ExportAsset = require("js/models/exportasset"),
         synchronization = require("js/util/synchronization"),
+        headlights = require("js/util/headlights"),
         Dialog = require("jsx!js/jsx/shared/Dialog");
 
     var ExportPanel = React.createClass({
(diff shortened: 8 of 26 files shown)


src/js/jsx/IconBar.jsx
```diff
@@ -110,6 +111,7 @@ define(function (require, exports, module) {
          */
         _handleBackToPSClick: function () {
             this.getFlux().actions.menu.native({ commandID: 5999 });
+            headlights.logEvent("UserInterface", "switch", "back-to-PS");
```
iwehrman: For consistency, let's pick a particular syntax for all of these names and events. My suggestion: all lower case, all words separated by dashes. This will make our lives easier in the future.

src/js/jsx/PanelSet.jsx
```diff
@@ -204,8 +205,10 @@ define(function (require, exports, module) {
                 modifierStore = flux.store("modifier"),
                 components = uiStore.components,
                 modifierState = modifierStore.getState(),
+                currentlyVisible = "",
```
iwehrman: It doesn't seem like this initialization is needed?

src/js/jsx/PanelSet.jsx
```diff
@@ -204,8 +205,10 @@ define(function (require, exports, module) {
...
                 nextState = {};
                 
+
```
iwehrman: Also, no double newlines. I'm not sure why the CI isn't catching this. (Any ideas @shaoshing?)

src/js/jsx/IconBar.jsx
DivyaPrabhakar: Sounds good, I will change the ones used before and update the wiki.

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -141,6 +142,8 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.export.updateExportAsset(
                 this.props.document, this.props.layers, this.props.index, props);
+
+            headlights.logEvent("edit", "export-settings-quality", props);
```
iwehrman: You can't log objects. This needs to be a string. If nothing else, you can just concat the format and quality strings.

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -160,6 +163,8 @@ define(function (require, exports, module) {
                 keySuffix = this.props.faceKey,
                 scaleListID = "exportAsset-scale-" + keySuffix,
                 formatListID = "exportAsset-format-" + keySuffix;
+            
+            headlights.logEvent("export", "create-asset", formatOptionId);
```
iwehrman: I think it's a bad idea to put this call in the render method, because the render method may be called many times for the same asset. Can you put it in the click handler that adds the asset?

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -130,6 +131,10 @@ define(function (require, exports, module) {
             
             this._forceVisible();
             this.getFlux().actions.export.addAssetThrottled(document, documentExports, selectedLayers, props);
+            
+            if (preset) {
+                headlights.logEvent("export", "preset", preset);
```
iwehrman: I'm not sure if preset is a string here. Presumably it is, but there isn't any JSDoc. Can you confirm that it's a string and, if so, add the `param` annotation for this method?

src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -186,6 +187,8 @@ define(function (require, exports, module) {
             dialog.toggle(event);
             this._forceVisible();
             this.getFlux().actions.layerEffects.addEffect(this.props.document, layers, effectType);
+
+            headlights.logEvent("effect", "create", effectType);
```
iwehrman: The effect type here is an exception to my wish above to have all the categories and events be lower-case and dash-separated. Whatever effect type is can be jammed into the name field here as long as it's a string.

src/js/jsx/sections/style/Fill.jsx
```diff
@@ -70,6 +71,7 @@ define(function (require, exports, module) {
             this.getFlux().actions.shapes
                 .setFillColorThrottled(this.props.document, this.props.layers, color,
                     { coalesce: coalesce });
+            headlights.logEvent("edit", "color-input", "color-change");
```
iwehrman: Should this indicate that the color being changed is the fill?

iwehrman: Also note that the _opacityChanged handler is called by the color picker as well.

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -181,6 +184,8 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.shapes
                 .setStrokeColorThrottled(this.props.document, this.state.layers, color, options);
+
+            headlights.logEvent("edit", "stroke-input", "stroke-color-change");
```
iwehrman: This is an example of something that should only be logged if `!coalesce`.

src/js/jsx/sections/style/Fill.jsx
iwehrman: This is an example of something that should only be logged if `!coalesce`.

src/js/jsx/sections/style/UnsupportedEffectList.jsx
```diff
@@ -161,6 +163,8 @@ define(function (require, exports, module) {
                 );
             }
             
+            headlights.logEvent("effect", "create", "unsupported");
```
iwehrman: I think we need to get rid of all logging calls from `render` methods, even if that means we can't log some things.

iwehrman:
Looking good so far. High level suggestions:

1. Unify naming convenctions to `lower-case-dash-separated`
2. For color picker events, only log when `!coalesce`
3. To avoid duplicate logging, never log inside of a `render` method. Always prefer event handlers when possible.

src/js/jsx/sections/style/UnsupportedEffectList.jsx
DivyaPrabhakar: Can I add a componentDidMount method and have the headlights call be the only this in this method?

src/js/jsx/sections/export/ExportPanel.jsx
DivyaPrabhakar: Done

DivyaPrabhakar:
@iwehrman made PR changes and added some more calls. The last few I need to ask people about tomorrow.

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -141,6 +144,9 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.export.updateExportAsset(
                 this.props.document, this.props.layers, this.props.index, props);
+
+            headlights.logEvent("edit", "export-settings-quality",
+                "Setting: " + props.format + " Quality: " + props.quality);
```
iwehrman: This should just be `props.format + "-" + props.quality`.

src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -160,7 +166,7 @@ define(function (require, exports, module) {
                 keySuffix = this.props.faceKey,
                 scaleListID = "exportAsset-scale-" + keySuffix,
                 formatListID = "exportAsset-format-" + keySuffix;
-
+
```
iwehrman: :space_invader:

src/js/jsx/sections/style/ColorOverlayList.jsx
```diff
@@ -103,6 +104,7 @@ define(function (require, exports, module) {
                     this.props.index,
                     blendMode,
                     LayerEffect.COLOR_OVERLAY);
+            headlights.logEvent("edit", "colorOverlayBlendMode-input", blendMode);
```
iwehrman: `color-overlay-blend-mode`

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -181,6 +184,10 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.shapes
                 .setStrokeColorThrottled(this.props.document, this.state.layers, color, options);
+
+            if (!coalesce) {
+                headlights.logEvent("edit", "stroke-input", "stroke-color-change");
```
iwehrman: Above, there are two other handlers for color change.

src/js/jsx/sections/style/LayerBlendMode.jsx
```diff
@@ -60,6 +61,7 @@ define(function (require, exports, module) {
         _handleChange: function (mode) {
             this.getFlux().actions.layers
                 .setBlendModeThrottled(this.props.document, this.props.layers, mode);
+            headlights.logEvent("edit", "layerBlendMode-input", mode);
```
iwehrman: `layer-blend-mode`

src/js/jsx/sections/style/Fill.jsx
```diff
@@ -70,6 +71,9 @@ define(function (require, exports, module) {
             this.getFlux().actions.shapes
                 .setFillColorThrottled(this.props.document, this.props.layers, color,
                     { coalesce: coalesce });
+            if (!coalesce) {
+                headlights.logEvent("edit", "color-input", "fill-color-change");
```
iwehrman: Above, there is one other handler for color change.

src/js/jsx/sections/style/StrokeList.jsx
```diff
@@ -116,6 +117,7 @@ define(function (require, exports, module) {
                     this.props.index,
                     blendMode,
                     LayerEffect.STROKE);
+            headlights.logEvent("edit", "strokeBlendMode-input", blendMode);
```
iwehrman: `stroke-blend-mode`

iwehrman:
Back to @DivyaPrabhakar.

iwehrman:
Also, needs a rebase.

chadrolfs:
@DivyaPrabhakar I didn't see in the list above but maybe it is covered in UI, but are we tracking if users are have Single-Column Mode enabled or not?

src/js/jsx/sections/export/ExportList.jsx
DivyaPrabhakar: @mcilroyc and I talked about this one and it seems unnecessary to log when people change a format or quality, but rather just log what they end up exporting.

DivyaPrabhakar:
@chadrolfs Yup, it is incorporated now.

src/js/util/ui.js
```diff
@@ -111,7 +111,7 @@ define(function (require, exports) {
     var openURL = function (url, event, name) {
         adapter.openURLInDefaultBrowser(url);
 
-        headlights.logEvent("UserInterface", "introduction", name);
```
shaoshing: libraries links (e.g. share library/element) are also using `openURL`, so the subcategory should not be `introduction` here. Maybe a better way is putting the `logEvent` in the first launch element.

src/js/models/menubar.js
```diff
@@ -466,6 +467,12 @@ define(function (require, exports, module) {
             "TOGGLE_SINGLE_COLUMN_MODE": { "checked": (preferences.get("singleColumnModeEnabled", false) ? 1 : 0) }
         });
 
+        if (preferences.get("singleColumnModeEnabled")) {
+            headlights.logEvent("user-interface", "panels", "single-column-mode-enabled");
+        } else {
+            headlights.logEvent("user-interface", "panels", "single-column-mode-disabled");
+        }
+
```
shaoshing: These events are logged every time the menu mode updated, maybe consider moving these to the button components?

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -371,6 +373,7 @@ define(function (require, exports, module) {
                             // Apply the color to selected layers
                             fluxActions.sampler.applyColor(this.state.document, null, sample.value);
                             d3.event.stopPropagation();
+                            headlights.logEvent("tools", "sampler-hud", sample.type);
```
shaoshing: `tools > sampler-hud > fillColor` should be `tools > sampler-hud > fill-color`

you can use `_.kebabCase` to do the conversion.

src/js/jsx/sections/style/EffectsPanel.jsx
shaoshing: why not use `_.kebabCase` for effectType?

src/js/actions/tools.js
```diff
@@ -753,6 +754,20 @@ define(function (require, exports) {
                 });
             }
         }
+
+        if (vectorMaskMode) {
+            var layerKind = "";
```
iwehrman: Suggestion: just make the event equal to `"mask-mode-" + layer.kind`. We can figure out the mapping from number to name later, and then we'll catch all the layer kinds. (E.g., we're missing at least smart objects below.) Bonus: it's easier. Also, the tool can change during mask mode... But, I think we can either ignore that OR just not track the particular tool. Both seem fine.

src/js/jsx/help/FirstLaunch.jsx
```diff
@@ -62,6 +63,18 @@ define(function (require, exports, module) {
             }
         },
 
+        /**
+         * Calls for the URL to be opened and then logs the headlights data
+         *
+         * @param {string} url
+         * @param {string} name
+         *
```
iwehrman: `@private`. Also should This instead call the flux action `menu.openURL`? Then you get logging for free!

iwehrman: (To see an example, check out @shaoshing's open Libraries analytics PR. I stole the idea from him.)

src/js/jsx/search/SearchBar.jsx
```diff
@@ -388,6 +390,10 @@ define(function (require, exports, module) {
                 return optionList.take(this.props.maxOptions);
             }
 
+            if (optionList.size === 6 && searchTerm !== "") {
```
iwehrman: :police_car: :cop: :vertical_traffic_light:  We aren't allowed to log arbitrary strings, which I think is what's happening here with `searchTerm`. I also cannot abide by a magic number, 6, here without additional explanation. I think instead of this, you should just try to find the handler that is called when the search bar is called without executing any options, and log a generic `failed-search` event.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -396,6 +402,51 @@ define(function (require, exports, module) {
...
         },
+
+        /**
+         * Get the exact type of layer
+         *
+         * @param {string} id
+         *
+         * @return {string} layerKind
+         */
+        _findLayerType: function (id) {
```
iwehrman: I really don't like this. I actually think it would be fine to NOT log the layer kind here, and instead just log that a layer was searched for (or whatever the search category is). I do not want to add a bunch of code that exists only to reverse-engineer the search IDs just for the sake of analytics.

src/js/jsx/sections/export/ExportAllPanel.jsx
```diff
@@ -202,6 +205,7 @@ define(function (require, exports, module) {
          */
         _setUseArtboardPrefix: function (event, enabled) {
             this.getFlux().actions.export.setUseArtboardPrefix(enabled);
+            headlights.logEvent("export", "batch-export", "prefix-enabled-" + enabled);
```
iwehrman: I suspect these kinds of events will not be super useful... What are we going to count, the number of times someone toggled this setting? But... I guess I don't really have a better short-term suggestion. So it goes.

src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -34,7 +34,9 @@ define(function (require, exports, module) {
     var os = require("adapter/os"),
         LayerEffect = require("js/models/effects/layereffect"),
         strings = require("i18n!nls/strings"),
+        _ = require("lodash"),
```
iwehrman: This isn't your mess, but would you mind tidying these declarations up to match our code style? The typical way to do imports like this is in three different declaration blocks: 1) external libraries (lodash, bluebird, etc.) 2) adapter imports, 3) imports from within spaces-design.

iwehrman: Example: https://github.com/adobe-photoshop/spaces-design/blob/4f1a6c389f014500cde5ceb0b8ff2448bd1d7149/src/js/jsx/Toolbar.jsx#L27-L37

src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -186,6 +189,8 @@ define(function (require, exports, module) {
             dialog.toggle(event);
             this._forceVisible();
             this.getFlux().actions.layerEffects.addEffect(this.props.document, layers, effectType);
+
+            headlights.logEvent("effect", "create", _.kebabCase(effectType));
```
iwehrman: Kebab case! I had no idea that was what this was called. I love it.

src/js/util/headlights.js
```diff
@@ -44,6 +45,7 @@ define(function (require, exports) {
      */
     var logEvent = function (category, subcategory, event) {
         if (global.debug) {
+            log.info("headlights: %s > %s > %s", category, subcategory, event);
```
iwehrman: Remove this before merging. See my identical complaint in @shaoshing's PR #3097.

src/js/util/ui.js
```diff
@@ -106,13 +105,10 @@ define(function (require, exports) {
      *
      * @param {string} url
      * @param {SyntheticEvent?} event
-     * @param {string} name
      */
-    var openURL = function (url, event, name) {
+    var openURL = function (url, event) {
```
iwehrman: Hmm... it's weird that we have `openURL` in `js/util/ui` and also `js/actions/menu`...

iwehrman:
Back to @DivyaPrabhakar. Looking good; most of my suggestions are for radical simplifications, so there isn't much new code to be written. Happy to chat about this before the standup today!

src/js/actions/tools.js
iwehrman: TODO.

src/js/jsx/help/FirstLaunch.jsx
```diff
@@ -62,6 +61,23 @@ define(function (require, exports, module) {
             }
         },
 
+        /**
+         * Calls for the URL to be opened and then logs the headlights data
+         *
+         * @private
+         * @param {string} url
+         * @param {string} name
+         *
```
iwehrman: Kill this newline.

src/js/jsx/sections/style/AppearancePanel.jsx
```diff
@@ -144,6 +145,7 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.sampler.pasteLayerStyle(document, targetLayers);
             event.stopPropagation();
+            headlights.logEvent("style", "create", "paste-all-styles");
```
iwehrman: Might as well add a similar one for copy?

src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -161,6 +163,7 @@ define(function (require, exports, module) {
 
             this.getFlux().actions.sampler.pasteLayerEffects(document, targetLayers);
             event.stopPropagation();
+            headlights.logEvent("tools", "sampler", "paste-all-effects");
```
iwehrman: Ditto: should we also log `copy-all-effects`?

iwehrman:
- [ ] one dangling TODO
- [ ] one extra newline
- [ ] missing copy-styles event
- [ ] missing copy-effects event

src/js/jsx/sections/style/Stroke.jsx
```diff
@@ -110,6 +111,8 @@ define(function (require, exports, module) {
                 this.state.layers,
                 { enabled: isChecked }
             );
+
+            headlights.logEvent("edit", "stroke-activation", isChecked);
```
shaoshing: Shall we include "fill-activation" as well?

src/js/util/ui.js
shaoshing: That's my bad legacy. maybe we should switch to `menu.openURL` completely.

iwehrman: :white_check_mark:

iwehrman:
Let's clean up the `openUrl` after M3. I'd accept a separate PR for a `fill-activation` event. But I'm gonna merge this now so @DivyaPrabhakar can come up for air.

Issue #2560, opened 2015-09-21
Color Picker UI opens above swatch with smaller App window 


chadrolfs:
new doc, draw rect, add drop shadow, resize app window (or use an SP3) so that vertical is smaller.
click on color swatch

result: color picker opens above the swatch
![screen shot 2015-09-21 at 4 48 42 pm](https://cloud.githubusercontent.com/assets/8442631/10008193/f009b81c-6080-11e5-999f-a7198f686a43.png)

chadrolfs:
remove or change position of triangle if repositioned from default?

ktaki:
A bit of confused moment.

<img width="273" alt="screen shot 2015-09-30 at 9 15 55 am" src="https://cloud.githubusercontent.com/assets/2029597/10199117/1f3a6c92-6754-11e5-87e9-e6c65ec087c7.png">

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
yay!  nice work!

PR #3074, opened 2015-10-21
Change targeted dialog position via state, remove arrow of color picker when place above target

designedbyscience:
While attempting to work on #2560, I changed the targeted positioning of the dialogs to happen via state changes rather than the direct changing of properties

IT SEEMS TO WORK FINE. But that statement is a little worrisome, so more eyes/click to check would be better.

Also, that little arrow on the color picker goes away when it is placed above the target element

### DIFF
--- a/src/js/jsx/shared/ColorPicker.jsx
+++ b/src/js/jsx/shared/ColorPicker.jsx
@@ -1052,7 +1052,7 @@ define(function (require, exports, module) {
             });
 
             return (
-                <div className="color-picker">
+                <div>
                     <ColorType {...this.props}
                         ref="input"
                         color={color}
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -210,7 +210,8 @@ define(function (require, exports, module) {
                 if (this._target) {
                     dialogBounds = dialogEl.getBoundingClientRect();
 
-                    var clientHeight = window.document.documentElement.clientHeight,
+                    var placedAbove = false,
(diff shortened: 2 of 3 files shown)


placegraphichere:
Wondering if we can do the arrow on the bottom? here's what I mocked up:

*dialog.less*
`.dialog-pointer-bottom(@left) {
    content: "";
    position: absolute;
    top: auto;
    bottom: -0.9rem;
    left: @left;
    width: 1.5rem;
    height: 1.5rem;
    background: @bg-border;
    border: none;
    border-left: @hairline*1.5 solid @underlines;
    border-top: @hairline*1.5 solid @underlines;
    border-top-left-radius: 0.2rem;
    transform: rotate(225deg);
    -webkit-clip-path: polygon(0% 0%, 1.8rem 0%, 0% 1.8rem);
}`

*color-picker.less*
`&.placed-above:before {
    .dialog-pointer-bottom(1.6rem);
}`

screen shot 2015-10-21 at 6 23 14 pm

iwehrman:
I can't rationally see why your member/state change wouldn't work. But, also, the arrow elision doesn't seem to be working for me:
ar

placegraphichere:
Ahhh, that's probably why we shouldn't add it to the bottom. I think @designedbyscience is just adjusting for any variation that doesn't allow it to sit in its natural state.

designedbyscience:
Okay, now arrow disappears for both "out of natural state" conditions

iwehrman:
:hamburger: :fries:

Issue #3078, opened 2015-10-21
Error deleting next to last layer with Pen tool


chadrolfs:
cmd/ctrl + n
r for Rect tool, drag out a shape
p for Pen tool, draw a closed shape
hit delete

Result: `Transfer from tools.select to layers.deleteSelected failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Delete” is not currently available.`

chadrolfs:
related to #3011

chadrolfs:
![pendelete-error](https://cloud.githubusercontent.com/assets/8442631/10653278/c96358c2-7812-11e5-9d4b-615491233056.gif)

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
bouncing.  follow the steps above and after the last step, hit delete again.
Result: `Menu command not available: 106`

iwehrman:
That is a debug-only warning message. Is the behavior incorrect?

iwehrman:
Alternatively: try verifying on a compiled build.

chadrolfs:
yep, we're good on compiled and nothing bad happens in debug, just the console error.

PR #3080, opened 2015-10-22
Action `menu.native` should return the command result.

shaoshing:
`menu.native` should return the command result so that the pen tool is notified whether the path is deleted or not. Address issue #3078

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -96,17 +96,22 @@ define(function (require, exports) {
                 return ps.performMenuCommand(payload);
             })
             .then(function (success) {
-                // FIXME: After the M2 release, remove the debug conjunct
                 if (global.debug && !success) {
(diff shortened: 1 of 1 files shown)


src/js/actions/menu.js
```diff
@@ -96,17 +96,22 @@ define(function (require, exports) {
                 return ps.performMenuCommand(payload);
             })
             .then(function (success) {
-                // FIXME: After the M2 release, remove the debug conjunct
                 if (global.debug && !success) {
-                    throw new Error("Menu command not available: " + payload.commandID);
+                    log.error("Menu command not available: " + payload.commandID);
```
shaoshing: No all failed menu command are bad. For example, it should be acceptable for `clearPath` command to fail. So I replace the `throw` with `log.error` here.

iwehrman:
OK!

Issue #3075, opened 2015-10-21
Keyboard Arrow Keys Don't Seem to Nudge Shape Points


placegraphichere:
- Create Shape
- Double click into it
- select one point
- use arrow keys to nudge
- result: nothing

- Expected: nudging should work

iwehrman:
Is this a modal state @pineapplespatula / @mcilroyc? We may have suspended policies, which removed our arrow key policies. This could be an easy fix...

chadrolfs:
Verified this is fixed.

PR #3081, opened 2015-10-22
Allow nudging selected points in select vector-edit mode and pen tool

iwehrman:
This adds `FOCUS_PROPAGATE` arrow key policies in superselect vector mode and the pen tool to allow nudging selected points. Pretty cool.

Addresses #3075.


iwehrman:
Warning: we seem to be missing some events from PS when nudging points, and so I think we fail to update the path bounds. Investigating... CC @isonno @jsbache

iwehrman:
This should now be testable with Mac pgdev.820. Note the new `nudgePathPoints` event. The thing to verify is that the shape bounds are being updated

pineapplespatula:
looks good

Issue #2073, opened 2015-08-17
History actions are excessively chatty


iwehrman:
We should only be logging during truly interesting situations because otherwise this is just noise to most of the engineers.

iwehrman:
I would still like to clean these messages up, but now that they're `debug` level, they're no longer bothering Russell and only me, so pushing this to M4.

PR #3082, opened 2015-10-22
Reduced history-related log messages to "debug"

mcilroyc:
This addresses, or at least takes some sting off of #2073.  At the very least, this reduces the chattiness into the main PS log.  I believe that at the debug level, these messages are of ongoing usefulness.


iwehrman:
OK, sure!

Issue #3076, opened 2015-10-21
French Intro needs some repositioning of image and sentences


ktaki:
In French intro screens, images and sentences have been placed too high. Need to center them vertically.

Example:

![screen shot 2015-10-21 at 4 17 15 pm](https://cloud.githubusercontent.com/assets/2029597/10652943/734400e8-780f-11e5-933f-df91e18515a2.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Looks good now!

PR #3083, opened 2015-10-22
French introduction screen positioning fixed

DivyaPrabhakar:
References issue #3076 

Before: 

After:

I have also verified that all other screens are positioned correctly. The only one that needed to be changed was the Swap screen to avoid widows. That change is included in this PR also. 

Before: 

After:


iwehrman:
Beautiful!

Issue #3084, opened 2015-10-22
Sampling LLSO (Library Linked Smart Object) has one extra history state "Relink to Library Graphic"


shaoshing:
Watson bug: https://watsonexp.corp.adobe.com/#bug=4077984
Video: https://www.dropbox.com/s/w7456wh35dlbf4p/sampling-graphic.mov?dl=0

This cause:

1. take two undo to restore to the original SO
2. target layer does not reset to the correct bound (beucase the relink happens after the "resetLayer")

mcilroyc:
@shaoshing the bounds issue (part 2 in your description) seems to me to be unrelated to the history states.  That is, after I sample the graphic, immediately when I switch to SS tool I see that the bounds are incorrect. 
http://cl.ly/212E0320253I

can you confirm this?

iwehrman:
Should we call this FBNC? Or leave it open because we only checked in a superhack? Either way, handing it off to @chadrolfs for testing.

chadrolfs:
this is fixed but it uncovered or resulted in a secondary issue #3108

iwehrman:
For posterity, @jsbache has suggested that perhaps we can use `_spaces.ps.processQueuedCommands` instead of the delay hack we previously used to work around the issue.

Issue #3060, opened 2015-10-21
Sampling graphic from one CLSO to another CLSO takes multiple clicks to apply


chadrolfs:
Using "Test - OK to accidentally nuke" library
drag Climber onto canvas, Enter to place
drag needle-fireworks onto canvas, Enter to place
hit i for Sampler
mouse over Climber, hit Space, click on graphic icon (repeat)

Result: it take multiple clicks to apply

chadrolfs:
this also breaks undo which may be a separate bug/issue.

PR #3085, opened 2015-10-22
Make graphic button in the sampler HUB easier to click.

shaoshing:
Make graphic button in the sampler HUB easier to click, just like the Effects and Type buttons.

Addresses #3060

note: If you are testing with LLSO, you may encounter issue #3084


pineapplespatula:
(as an aside it is HUD which stands for heads up display)

pineapplespatula:
looks good.

shaoshing:
thanks for the heads up :P

Issue #3088, opened 2015-10-22
Swap-panels modifier click doesn't work on Windows


iwehrman:
I tried Ctrl, Win and Alt, but none of them do the right thing.

iwehrman:
Note that, even with a JS patch, this is still unreliable on Win due to [Watson 4078041](https://watsonexp.corp.adobe.com/#bug=4078041).

mcilroyc:
The engineer has reported this as being partially solved by 3091

iwehrman:
The @jfitz-adobe has reported that this is fixed by CL 1049772. Over to @ktaki to verify.

ktaki:
Ctrl key works now on Windows. Cmd key for Mac.

PR #3091, opened 2015-10-22
Make the swap-panels modifier command on Mac and control on Win.

iwehrman:
What it says! Partially addresses #3088. 

I notice, however, that the on Windows (with the SP3 attached keyboard at least) we do not seem to reliably get keydown events with modifier changes. It seems to work ~50% of the time... So, better than nothing?


iwehrman:
[Watson 4078041](https://watsonexp.corp.adobe.com/#bug=4078041)

iwehrman:
This is a low-risk fix, btw.

mcilroyc:
booya

Issue #3049, opened 2015-10-20
copy/paste appearance between two text layers with different opacity will flash the number


chadrolfs:
https://cloud.githubusercontent.com/assets/118264/10621526/34bf23ea-7736-11e5-8d10-37ed838fb837.gif

chadrolfs:
Only if there's time and fix is low risk.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
no longer flashes, but doesn't copy the correct value.  see #3100

PR #3092, opened 2015-10-22
Coalesce multiple text attribute updates when pasting/sampling text styles

shaoshing:
This PR addresses the following issues:

1. copy/paste text styles will flash the opacity field. Fix for #3049 
2. sampling text styles should include opacity.


iwehrman:
Looks good, works good!

Issue #3050, opened 2015-10-20
copy/paste appearance from Rect w/radius to Ellipse flashes Radius on paste


chadrolfs:

chadrolfs:
![paste-radiusflash](https://cloud.githubusercontent.com/assets/8442631/10624093/2d2ec3e2-774a-11e5-8c19-9034bb2959bd.gif)

chadrolfs:
Only if there's time and fix is low risk.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Looks fixed on Win but not on Mac (10.10.5, haven't tried 10.11.1 yet).  pgdev.826 with master @ 35bf347

chadrolfs:
![paste-radiusflash2](https://cloud.githubusercontent.com/assets/8442631/10680165/11ff6e58-78d2-11e5-95a0-6b76a07f81ae.gif)

shaoshing:
Hmm, that's weird, beucase the fix is produced on MAC. Let me double check.

chadrolfs:
restested this morning with latest jenkins build cotaining the updated compiled js and it looks fixed.

chadrolfs:
yesterday was nuts, i might have been on a branch other than master.  I trust myself more this morning.  :stuck_out_tongue_winking_eye:

PR #3093, opened 2015-10-22
`transform.setRadius` should exclude layers that does not support radius.

shaoshing:
Otherwise, the document store will try to update the layer's radii attribute and then get reverted by the `resetLayer` action, which flash the radius slider when copy/paste layer styles. 

Addresses issue #3050


iwehrman:
OK!

Issue #1605, opened 2015-06-17
Cannot move layers under vector mask with select tool


ktaki:
Group 1 has a vector mask and is inside Group 2. Ellipse 1 layer and Select Me text layer cannot be selected and moved with the super select tool.

![screen shot 2015-06-17 at 3 36 48 pm](https://cloud.githubusercontent.com/assets/2029597/8220374/f5a9f3ec-1506-11e5-95ce-ece8ae7405eb.png)

The test file is posted at:
smb://sjshare/photoshoptestfiles/TestFiles/_Photoshop_Feature_Testing/Playground-Recess/bugs/CannotSelectLayers.psb
Open CannotSelectLayers.psb and select "Group 2"

chadrolfs:
Interesting.  It's the top level group that is preventing the selection.  If you double-click to drill down, you can then select the ellipse/text.  Might be as designed.

ktaki:
Can you move them?

chadrolfs:
Yes.

chadrolfs:
here's what i'm seeing/doing; double-click to drill down then move
https://www.dropbox.com/s/04tv86tbr9w9n5a/dclick-then-move.mov?dl=0

ktaki:
Interesting. I cannot do what you are doing in the video. Trying to move the ellipse will move the Group 2. When Artboard 1 is selected, double clicking the ellipse should not select Group 1 in the first place. Anyway, this is a bug.

chadrolfs:
@ktaki please re-test.  I think we can close this?

ktaki:
No, this is still an issue. You can click on it to select, but cannot drag it to move. @pineapplespatula Let us know your assessment of fixing this bug. Fixable by M3 or possibly deferring to M4?

pineapplespatula:
This is a core issue with the hit test we are using. It returns groups to us despite there not being any visible pixels under the pointer location that is sent to the hit test. Filing a core bug

pineapplespatula:
https://watsonexp.corp.adobe.com/#bug=4077991

chadrolfs:
core bug deferred, moving to M4

tomwmcrae:
Doing the following:
1. Select the ellipse layer (or the text layer) in the layers panel.
2. On the canvas, try to move the selected item.
results in the image being moved. It's as if the parts of the image that are masked out are still selectable from clicking on the canvas.

pineapplespatula:
[...]

PR #3095, opened 2015-10-22
add mask awareness to superselect click routine

pineapplespatula:
fix for issue #1605

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -396,8 +396,17 @@ define(function (require, exports) {
 
                 if (!clickedSelectableLayerIDs.isEmpty()) {
                     // due to way hitTest works, the top z-order layer is the last one in the list
-                    var topLayerID = clickedSelectableLayerIDs.last(),
-                        topLayer = layerTree.byID(topLayerID),
+                    var topLayerID = clickedSelectableLayerIDs.findLast(function (layerID) {
+                        var layer = layerTree.byID(layerID);
+                        if (layerTree.minumumChildBounds(layer).contains(coords.x, coords.y)) {
+                            return true;
+                        }
+                    });
+
+                    if (!topLayerID) {
+                        return Promise.resolve(false);
+                    }
+                    var topLayer = layerTree.byID(topLayerID),
                         modifier = "select";
 
                     if (add && topLayer.selected) {
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -398,9 +398,7 @@ define(function (require, exports) {
                     // due to way hitTest works, the top z-order layer is the last one in the list
                     var topLayerID = clickedSelectableLayerIDs.findLast(function (layerID) {
                         var layer = layerTree.byID(layerID);
-                        if (layerTree.minumumChildBounds(layer).contains(coords.x, coords.y)) {
-                            return true;
-                        }
+                        return (layer && layerTree.minumumChildBounds(layer).contains(coords.x, coords.y));
                     });
 
                     if (!topLayerID) {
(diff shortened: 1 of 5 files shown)


src/js/models/layer.js
```diff
@@ -504,6 +510,7 @@ define(function (require, exports, module) {
             proportionalScaling: false,
             isArtboard: !!isArtboard,
             bounds: isArtboard ? new Bounds(boundsDescriptor) : null,
+            maskbounds: isArtboard ? new Bounds(boundsDescriptor) : null,
```
iwehrman: Non-artboard groups can't have mask bounds?

iwehrman: This line confuses me for a number of reasons actually. Why would `maskBounds` be the same as `bounds` here, but for artboards? It seems like the mask itself is being ignored.

src/js/actions/superselect.js
```diff
@@ -396,8 +396,17 @@ define(function (require, exports) {
 
                 if (!clickedSelectableLayerIDs.isEmpty()) {
                     // due to way hitTest works, the top z-order layer is the last one in the list
-                    var topLayerID = clickedSelectableLayerIDs.last(),
-                        topLayer = layerTree.byID(topLayerID),
+                    var topLayerID = clickedSelectableLayerIDs.findLast(function (layerID) {
+                        var layer = layerTree.byID(layerID);
+                        if (layerTree.minumumChildBounds(layer).contains(coords.x, coords.y)) {
```
iwehrman: `layerTree.minimumChildBounds(layer)` can be null here. Also, if it's not null, I think you can just return the result of `.contains(x,y)`.

iwehrman: It's also possible (for a variety of reasons) that `layer` will be null. If you pass a null layer into `minimumChildBounds` it will throw. I suggest just bailing out early here if `layer` is null.

src/js/models/layerstructure.js
```diff
@@ -854,6 +854,39 @@ define(function (require, exports, module) {
...
+                if (layer.isArtboard) {
+                    return layer.maskBounds;
+                }
+
+                var childBounds = this.children(layer)
+                    .map(this.minumumChildBounds, this)
+                    .filter(function (bounds) {
+                        return bounds && bounds.area > 0;
+                    }),
+                    unionChildBounds = Bounds.union(childBounds);
```
iwehrman: This should not be computed unless it will be used.

iwehrman:
Back to @pineapplespatula. This seems reasonable, but I'm also a little concerned about the riskiness of this change so late. CC @mcilroyc for a second opinion.

iwehrman:
Also, CC @chadrolfs for testing. The bug this fixes is definitely worth fixing, so I'd love to know whether you think this kills it 100% dead.

chadrolfs:
Taking a look...

mcilroyc:
While testing this branch I noticed that after moving the circle, portions of the circle are not "SS-able".  http://cl.ly/383l1C1L3I08
In the video, when clicking on the right side of the circle, it won't select.

chadrolfs:
I can repro what @mcilroyc is seeing as well ^^

chadrolfs:
Revert appears to be broken

chadrolfs:
Seems to be something funky going on with AB moving/selection.  I can't seem to select the AB by clicking on it's title.  (workaround: select AB via layers panel, then arrow key to nudge/force update of bounds)  Once you can move the AB, other things become unselectable.

chadrolfs:
This feels too risky for this late in the game for M3 given the issues I'm seeing ^^

src/js/models/layerstructure.js
pineapplespatula: It will always be used.

pineapplespatula:
I did not update reset bounds. (fix coming shortly)

PR #3097, opened 2015-10-23
Log analytic events for CC Libraries

shaoshing:
Here are the events logged:

* create/delete/update (only rename)/switch library
* create/delete/update (only rename) element
* apply color / color-theme / layerstyle / characterstyle
* place graphic
* open/update graphic
* open library links: share library, share public link, view online
* open element links: share public link, view online
* open Adobe Stock link

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -46,6 +46,7 @@ define(function (require, exports) {
         layerActionsUtil = require("js/util/layeractions"),
         collection = require("js/util/collection"),
         librariesUtil = require("js/util/libraries"),
+        headlights = require("js/util/headlights"),
         layerActions = require("./layers"),
         exportActions = require("./export"),
         documentActions = require("./documents"),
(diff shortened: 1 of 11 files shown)


src/js/actions/libraries.js
```diff
@@ -1199,6 +1200,36 @@ define(function (require, exports) {
...
+                    
+                    switch (event) {
+                        case "createLibrary":
+                            headlights.logEvent("libraries", "library", "create");
+                            break;
+                        case "deleteLibrary":
+                            headlights.logEvent("libraries", "library", "delete");
+                            break;
+                        case "createElement":
+                            //
```
iwehrman: Missing helpful comment here? Or stray, stale comment?

src/js/util/headlights.js
```diff
@@ -44,6 +45,7 @@ define(function (require, exports) {
      */
     var logEvent = function (category, subcategory, event) {
         if (global.debug) {
+            log.info("headlights: %s > %s > %s", category, subcategory, event);
```
iwehrman: I think this should be removed before this is merged. I know it only happens in debug mode only, but I'm still interested in keep the console output readable for developers, which means trying to limit the output to that which is widely useful. I think, outside of testing this PR, this probably won't be widely useful.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -253,6 +254,20 @@ define(function (require, exports, module) {
             
             this.setState({ command: null });
         },
+        
+        /**
+         * Handle library button clicked
+         *
+         * @private
+         */
+        _handleLibraryButtonClicked: function (event, url) {
```
iwehrman: Missing `@param` annotations.

src/js/jsx/sections/libraries/assets/AssetSection.jsx
```diff
@@ -162,6 +161,20 @@ define(function (require, exports, module) {
         _handleEndEditingTitle: function () {
             this._isEditingTitle = false;
         },
+            
+        /**
+         * Handle element button clicked
+         *
+         * @private
+         */
+        _handleElementButtonClicked: function (event, url) {
```
iwehrman: Missing `@param` annotations.

src/js/jsx/sections/libraries/LibraryBar.jsx
```diff
@@ -169,14 +169,28 @@ define(function (require, exports, module) {
...
         _handleSyncLibraries: function () {
             this.getFlux().actions.libraries.syncLibraries();
         },
+        
+        /**
+         * Handle stock button clicked
+         *
+         * @private
+         */
+        _handleStockButtonClicked: function () {
```
iwehrman: Missing `@param` annotations.

iwehrman:
Fantastic! Just a few minor changes suggested. We'll get this merged before the standup. Back to @shaoshing.

iwehrman:
Also, just want to confirm that you explicitly opted not to use `CCLibraries.analyticsUtils.reportUseElement` for a good reason. Like, e.g., it doesn't actually serve any purpose?

shaoshing:
@iwehrman thanks for reviewing. I've addressed your comments. I actually missed the `reportUseElement` util function; but my first looks shows it is similar to using the `headlights.logEvent`, and require more codes to emit the events and then handle them together in the `reportEvent` callback, so probably not worth a lot to switch to it. Please let me know if you disagree.

iwehrman:
Based on my limited understanding, I don't disagree with your decision to skip `reportUseElement`.

PR #3099, opened 2015-10-23
Fix for sampler/replaceGraphic to workaround a core bug

mcilroyc:
See the comment in the code for more details.  I'm not really proud of this, but I think it is safe and works better than the current functionality.  Addresses (hacks around) #3084


iwehrman:
OK!

Issue #3101, opened 2015-10-23
Cloaking rectangle is incorrect after display configuration changes


iwehrman:
I've seen this a few times in the last two weeks, but can't figure out how to repro it. One concrete observation is that it is _always_ off on the right-hand side by a width that seems equivalent to the toolbar or icon bar width:

![img](https://s3.amazonaws.com/f.cl.ly/items/1v3j382D2m1a1c2p1N17/Screen%20Recording%202015-10-23%20at%2011.40%20AM.gif)

iwehrman:
In particular, it seems like the iconBarWidth is not being set correctly:
![img](https://s3.amazonaws.com/f.cl.ly/items/2j1a10200Q3r370T1E0z/Image%202015-10-23%20at%2011.43.45%20AM.png)

PR #3103, opened 2015-10-23
Fix cloaking rectangle bug when display configuration changes

iwehrman:
This fixes a bug in which the `displayConfigurationChanged` event handler's event argument would be interpreted as a truthy argument to `_updatePanelSizes`, preventing `iconBarWidth` from being correctly set, thus making the cloaking rectangle too large and causing the UI to flash out when panning the canvas.

Addresses #3101. Low risk, AFAICT.

### DIFF
--- a/src/js/jsx/IconBar.jsx
+++ b/src/js/jsx/IconBar.jsx
@@ -73,7 +73,10 @@ define(function (require, exports, module) {
         _updatePanelSizesDebounced: null,
 
         componentDidMount: function () {
-            this._updatePanelSizesDebounced = synchronization.debounce(this._updatePanelSizes, this, 500);
+            this._updatePanelSizesDebounced = synchronization.debounce(function () {
+                return this._updatePanelSizes();
+            }, this, 500);
+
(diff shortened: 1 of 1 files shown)


src/js/jsx/IconBar.jsx
```diff
@@ -74,12 +84,12 @@ define(function (require, exports, module) {
 
         componentDidMount: function () {
             this._updatePanelSizesDebounced = synchronization.debounce(this._updatePanelSizes, this, 500);
```
mcilroyc: an alternate suggestion is to just stuff this inside an anon function right here, instead of creating `_handleDisplayConfigurationChanged`

mcilroyc:
Your diagnosis and fix seem correct.  WFM.  low risk.

src/js/jsx/IconBar.jsx
iwehrman: The problem with that is that I have to remove the function later. Which means I need a reference to it in a different method.

iwehrman: Derp, I get it. Commit amended.

Issue #3100, opened 2015-10-23
Copy/Paste text Appearance doesn't paste the correct Opacity


chadrolfs:
![copy-paste-typeappearnce](https://cloud.githubusercontent.com/assets/8442631/10699114/6fd14da2-7969-11e5-980c-b9fef771b76e.gif)

chadrolfs:
related to #3092 which fixed #3049 (related)

chadrolfs:
Notice, the leading did not paste correctly either.

chadrolfs:
After I moved the layers and repeated the copy/paste, the opacity pasted correctly, but not the leading (Letterspacing)

chadrolfs:
also related to #2755

iwehrman:
Marked Fix-in-progress. If that's now the case @shaoshing, lemme know.

iwehrman:
FBNC to @chadrolfs.

PR #3104, opened 2015-10-23
Copy/Past text should include opacity and tracking - edge cases

shaoshing:
These two styles are included in the previous PR, and this PR covers some edge cases:

1. copy/paste should include 0 tracking.
2. opacity is not copied after changing. Addresses #3100 
3. bonus: fix changing type weight.

### DIFF
--- a/src/js/jsx/sections/style/Type.jsx
+++ b/src/js/jsx/sections/style/Type.jsx
@@ -157,6 +157,26 @@ define(function (require, exports, module) {
                 return;
             }
 
+            this._setPostScript(postScriptName);
+        },
+        
+        /**
+         * Handle change of type weight.
+         *
+         * @private
+         * @param {string} postScriptName
+         */
+        _handleTypeWeightChange: function (postScriptName) {
+            this._setPostScript(postScriptName);
+        },
+        
+        /**
+         * Set the type face of the selected text layers.
+         *
+         * @private
+         * @param {string} postScriptName
+         */
+        _setPostScript: function (postScriptName) {
             var document = this.props.document,
                 layers = document.layers.selected.filter(function (layer) {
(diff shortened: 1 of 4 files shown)


src/js/stores/font.js
```diff
@@ -228,7 +228,7 @@ define(function (require, exports, module) {
                 };
             }
 
-            if (textStyle.tracking) {
+            if (Number.isFinite(textStyle.tracking)) {
```
shaoshing: ```!!0 === false``` cause 0 tracking is been left behind.

src/js/jsx/sections/style/Type.jsx
```diff
@@ -157,6 +157,26 @@ define(function (require, exports, module) {
...
             }
 
+            this._setPostScript(postScriptName);
+        },
+        
+        /**
+         * Handle change of type weight.
+         *
+         * @private
+         * @param  {string} postScriptName
```
iwehrman: Extra space here after `@param`

src/js/stores/font.js
iwehrman: :person_frowning:

iwehrman:
I set all the properties on a type layer and then sampled it, but only a few made it over:
img

chadrolfs:
@iwehrman it looks like you're using the Sampler.  #3100 is specific to copy/pasting via the panel icons.  I think click with Sampler just copies fill color.

iwehrman:
Yes, sorry. I was confused. In this case at least it seems to be working completely correctly.

iwehrman:
After discussion with @shaoshing and @placegraphichere, I'm satisfied. I'll wait to get a thumbs-up from @chadrolfs before merging.

iwehrman:
@chadrolfs gives us :+1: :+1:, so bombs away!

Issue #3107, opened 2015-10-23
List of issues / suggested chagnes for headlights tracking


shaoshing:
(Feel free to expand the list or re-order)

#### Must Fix

- [x] command-click on the panel icon send
`user-interface > panels-visible > layersLibrary-visible`
should be
`user-interface > panels-visible > none-visible`

- [x] Mouse over blend mode (layer, shadow, color-overlay, etc.) list result in many event logs. Should only emit the final blend mode.

- [x] Clicking the + icon in the export panel sent
`export > preset > SyntheticMouseEvent`

- [x] Should have events for toggling panel (double-click on panel title)

- [x] I don't see any mask mode tracking, and this seems missing event name? ```headlights.logEvent("tools", "mask-mode-" + currentLayer.kind); ```

- [x] Should track sampler primary (mouse click on layer) and secondary click (mouse click on layer with shift key)

#### If time permitted

- [ ] Should have `fill-activation` for fill colors

- [x] Should track for “export”, “search", and "vector mask" buttons in the headers, and dismiss of search dialog
(I don't see the `tools > search > failed-search` after dismissing the dialog)

- [x] Should have better naming for export configuration.
`export > asset > scale-1-suffix-@1x-setting-png`
maybe
`export > asset > scale-1/suffix-@1x/setting-png`

**Divya's thoughts** The current standard is to separate all words with "-". We should decided on a overarching syntax if we make the above change and then make sure all the other logEvents are consistent with this. 

- [ ] Sending two events if copy layer via shortcut
`menu > EDIT > COPY`
`edit > layers > copy-layers`
Same for pasting layer
`menu > EDIT > PASTE`
`edit > layers > paste-layers`
may be we should get ride of the menu logging for copy/past layer?

**Divya's thoughts** I think we should keep this to see how users navigate to copy/paste because the trail to copy or paste in each case is different.

- [x] Create text layer sent
`tools > create > typeCreateOrEdit`

- [x] Clicking menu command send
`menu > WINDOW > TOGGLE_TOOLBAR`

- [x] Many `_logSuperselect` names are not in kebob case, for example: `click_select`, `select_all`, `marqueeSelect`

- [x] Confirm search sent
`search > filter-inactive > category-ALL_LAYER

- [ ] Copy/Paste style
`headlights: style > create > copy-all-styles`
should have more details
`headlights: style > create > copy-text-styles`
`headlights: style > create > copy-vector-styles`
[...]

PR #3113, opened 2015-10-23
Headlights tracking more

DivyaPrabhakar:
References #3107 -- fix now 

With this: 

- Command + click on the panel icons logs the correct event (none-visible) 
- Mouse over blend mode in stroke, layer, shadow, and color-overlay now should track only track the final blendmode chosen
- Export panel + icon logs the correct event -- preset or no preset 
- Toggling individual panels are now tracked
- Document header icons are all tracked 
- Primary sampler and secondary sampler are tracked

### DIFF
--- a/src/js/actions/sampler.js
+++ b/src/js/actions/sampler.js
@@ -45,6 +45,7 @@ define(function (require, exports) {
         transformActions = require("./transform"),
         toolActions = require("./tools"),
         layerActionsUtil = require("js/util/layeractions"),
+        headlights = require("js/util/headlights"),
         strings = require("i18n!nls/strings");
 
     /**
@@ -345,6 +346,12 @@ define(function (require, exports) {
             docStore = this.flux.store("document"),
             coords = uiStore.transformWindowToCanvas(x, y),
             layerTree = doc.layers;
+        
+        if (secondary) {
+            headlights.logEvent("tools", "eyedropper-secondary", "copy-all-effects");
+        } else {
+            headlights.logEvent("tools", "eyedropper-primary", "copy-all-effects");
+        }
         
         return uiUtil.hitTestLayers(doc.id, coords.x, coords.y)
             .bind(this)
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -755,7 +755,7 @@ define(function (require, exports) {
             }
         }
 
-        headlights.logEvent("tools", "mask-mode-" + currentLayer.kind);
+        headlights.logEvent("tools", "mask-mode", String(currentLayer.kind));
     };
     changeVectorMaskMode.reads = [locks.JS_APP, locks.JS_TOOL, locks.PS_DOC, locks.PS_TOOL];
     changeVectorMaskMode.writes = [locks.JS_TOOL, locks.PS_DOC];
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -227,6 +227,11 @@ define(function (require, exports, module) {
                 propertiesPanel = nextState[components.PROPERTIES_COL];
             }
 
+            if (swapModifier) {
+                layersLibPanel = !layersLibPanel;
+                propertiesPanel = !propertiesPanel;
+            }
+
             if (layersLibPanel && propertiesPanel) {
                 currentlyVisible = "layersLibrary-properties-visible";
             } else if (layersLibPanel) {
@@ -253,6 +258,7 @@ define(function (require, exports, module) {
             nextState[panelName] = !this.state[panelName];
 
             this.getFlux().actions.preferences.setPreferences(nextState);
+            headlights.logEvent("user-interface", "panel-toggle", panelName);
         },
 
         render: function () {
--- a/src/js/jsx/sections/style/ColorOverlayList.jsx
+++ b/src/js/jsx/sections/style/ColorOverlayList.jsx
@@ -33,15 +33,25 @@ define(function (require, exports, module) {
     var LayerEffect = require("js/models/effects/layereffect"),
         collection = require("js/util/collection"),
         strings = require("i18n!nls/strings"),
+        synchronization = require("js/util/synchronization"),
         headlights = require("js/util/headlights");
 
     var BlendMode = require("jsx!./BlendMode"),
         ColorInput = require("jsx!js/jsx/shared/ColorInput"),
         ToggleButton = require("jsx!js/jsx/shared/ToggleButton");
 
+    /**
+     * Debounced version of headlights.logEvents to help prevent false changes from 
+     * datalists from being logged. 
+     *
+     * @private
+     * @type {function(*)}
+     */
+    var logEventDebounced = synchronization.debounce(headlights.logEvent, null, 10000);
+
     var ColorOverlay = React.createClass({
         mixins: [FluxMixin],
-        
+
         shouldComponentUpdate: function (nextProps) {
             var sameLayerIDs = collection.pluck(this.props.layers, "id")
                 .equals(collection.pluck(nextProps.layers, "id"));
@@ -104,7 +114,8 @@ define(function (require, exports, module) {
                     this.props.index,
                     blendMode,
                     LayerEffect.COLOR_OVERLAY);
-            headlights.logEvent("edit", "color-overlay-blendmode-input", blendMode);
+
+            logEventDebounced("edit", "color-overlay-blendmode-input", blendMode);
         },
 
         /**
--- a/src/js/jsx/sections/style/ShadowList.jsx
+++ b/src/js/jsx/sections/style/ShadowList.jsx
@@ -34,13 +34,23 @@ define(function (require, exports) {
     var LayerEffect = require("js/models/effects/layereffect"),
         collection = require("js/util/collection"),
         headlights = require("js/util/headlights"),
-        strings = require("i18n!nls/strings");
+        strings = require("i18n!nls/strings"),
+        synchronization = require("js/util/synchronization");
 
     var Label = require("jsx!js/jsx/shared/Label"),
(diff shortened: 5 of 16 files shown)


src/js/actions/sampler.js
```diff
@@ -346,6 +347,12 @@ define(function (require, exports) {
             coords = uiStore.transformWindowToCanvas(x, y),
             layerTree = doc.layers;
         
+        if (secondary) {
```
iwehrman: This seems backwards.

src/js/actions/tools.js
```diff
@@ -755,7 +755,7 @@ define(function (require, exports) {
             }
         }
 
-        headlights.logEvent("tools", "mask-mode-" + currentLayer.kind);
```
iwehrman: Ida left the dash but whatev!

src/js/jsx/PanelSet.jsx
```diff
@@ -227,6 +227,16 @@ define(function (require, exports, module) {
                 propertiesPanel = nextState[components.PROPERTIES_COL];
             }
 
+            if (modifierState.command) {
```
iwehrman: You probably want to use `swapModifier` instead. Also, I think the logic is incorrect. If the swap modifier is present, then _each_ panel's state is swapped.

src/js/actions/tools.js
shaoshing: This should be ```headlights.logEvent("tools", "mask-mode", currentLayer.kind);```?

src/js/jsx/sections/style/ShadowList.jsx
```diff
@@ -219,7 +219,7 @@ define(function (require, exports) {
                     blendMode,
                     this.props.type);
             // Currently does not differentiate between inner and drop shadows
-            headlights.logEvent("edit", "shadow-blendmode-input", blendMode);
+            headlights.logEventDebounced("edit", "shadow-blendmode-input", blendMode);
```
iwehrman: I don't think `logEventsDebounced` exists??

iwehrman: Oh, nevermind....

src/js/util/headlights.js
```diff
@@ -49,6 +51,9 @@ define(function (require, exports) {
             return adapterPS.logHeadlightsEvent(category, subcategory, event);
         }
     };
+
+    var logEventDebounced = synchronization.debounce(logEvent, null, 1500);
```
iwehrman: Needs JSDoc.

src/js/jsx/sections/style/ShadowList.jsx
iwehrman: I don't think this is a good pattern, btw. If you call `logEventDebounced` in multiple places with different arguments at nearly the same time, all but the last call will be dropped... It might be OK in this case, but...

shaoshing: I agree. Ideally we should log the event based on the last selection in DataList, but that seems risky to make the change at the last moment, compare to using debounced logEvent. We may consider lowering the delay time from 1500 to 1000.

src/js/actions/tools.js
iwehrman: Ah, yes, what @shaoshing said. The event name is missing below.

src/js/jsx/sections/style/ColorOverlayList.jsx
```diff
@@ -104,7 +105,11 @@ define(function (require, exports, module) {
                     this.props.index,
                     blendMode,
                     LayerEffect.COLOR_OVERLAY);
-            headlights.logEvent("edit", "color-overlay-blendmode-input", blendMode);
+
+            if (!this._logEventDebounced) {
+                this._logEventDebounced = synchronization.debounce(headlights.logEvent, null, 1000);
```
iwehrman: It would be better to just make this a module-level variable instead of a member of the class.

src/js/jsx/sections/style/ShadowList.jsx
```diff
@@ -219,7 +220,10 @@ define(function (require, exports) {
                     blendMode,
                     this.props.type);
             // Currently does not differentiate between inner and drop shadows
-            headlights.logEvent("edit", "shadow-blendmode-input", blendMode);
+            if (!this._logEventDebounced) {
```
iwehrman: Ditto for all the rest.

src/js/util/headlights.js
```diff
@@ -27,7 +27,7 @@ define(function (require, exports) {
     var Promise = require("bluebird"),
         adapterPS = require("adapter/ps"),
         global = require("./global");
-
+
```
iwehrman: :space_invader:

src/js/jsx/sections/style/StrokeList.jsx
```diff
@@ -117,7 +118,11 @@ define(function (require, exports, module) {
                     this.props.index,
                     blendMode,
                     LayerEffect.STROKE);
-            headlights.logEvent("edit", "stroke-blendmode-input", blendMode);
+
+            if (!this._logEventDebounced) {
+                this._logEventDebounced = synchronization.debounce(headlights.logEvent, null, 1000);
```
iwehrman: It might actually be better to make the timeout on this larger? How long does it take you to browse through the menu to find your next blend mode... Maybe 10 seconds?

src/js/actions/tools.js
iwehrman: And if you take @shaoshing's suggestion to make the event be equal to the ID, make sure you to coerce that number into a string first.

iwehrman:
Back to @DivyaPrabhakar for a few more nits.

src/js/actions/tools.js
```diff
@@ -755,7 +755,7 @@ define(function (require, exports) {
...
+        headlights.logEvent("tools", "mask-mode", (currentLayer.kind).toString());
```
iwehrman: Err... I'm concerned that `Layer.prototype.kind` might be null, in which case this will throw an error. (Note [here](https://github.com/adobe-photoshop/spaces-design/blob/084d00c6ebe4be31d2b51bc94cf46eac812f5398/src/js/actions/layers.js#L92) that the `layerKind` property on the layer descriptor is marked "optional", which almost certainly means that there is some sort of layer which does not have a kind. E.g., background layers? Who knows.) Alternatives: `String(currentLayer.kind)` or `"" + currentLayer.kind`.

src/js/jsx/sections/style/ColorOverlayList.jsx
```diff
@@ -41,6 +42,7 @@ define(function (require, exports, module) {
 
     var ColorOverlay = React.createClass({
         mixins: [FluxMixin],
+        _logEventDebounced: synchronization.debounce(headlights.logEvent, null, 10000),
```
iwehrman: This isn't a module-level variable declaration :)

iwehrman:
OK, this is technically correct now, but I have one last nit. Our code style involves having require statements declared separately (i.e., in different "declaration blocks") from other variable declarations. So, the `logEventDebounced` variable declarations should come on a new line with `var logEventDebounced = ... `. Also, there should be a JSDoc decoration on these variable declarations, which includes a one-line explanation of the defined function's purpose (something about how it's a debounced version of `headlights.logEvents` to prevent false, transient changes from datalists from being logged), as well as a `@type {function(string, string, string)}` annotation.

In general, we require JSDoc for variables and properties that aren't "local". (That's a fuzzy distinction. Basically, if you're going to declare a variable one place and then use it some other place, we try to document it to make it easier for the next person who reads the code to put it all together. We also exempt variables which hold references to imported modules, i.e., `var Promise = ("bluebird")`.)

ktaki:
FYI: How the data currently look like with very short session of DS.

<EventRec doc="0" category="UserInterface" subcategory="Playground" name="switch to 01" time="2015-10-24T19:57:51Z" seq="192" />
<EventRec doc="0" category="menu" subcategory="edit" name="Preferences Changed" time="2015-10-24T19:57:52Z" seq="193" />
<EventRec doc="0" category="menu" subcategory="edit" name="Preferences Changed" time="2015-10-24T19:58:29Z" seq="194" />
<EventRec doc="0" category="history" subcategory="en_US" name="New" time="2015-10-24T19:58:34Z" seq="195" />
<EventRec doc="0" category="history" subcategory="en_US" name="New" time="2015-10-24T19:58:34Z" seq="196" />
<EventRec doc="0" category="CCXWelcome" subcategory="Interaction" name="DialogAutoDismiss" time="2015-10-24T19:58:34Z" seq="197" />
<EventRec doc="0" category="history" subcategory="Artboards" name="Used Artboards in Session" time="2015-10-24T19:58:37Z" seq="198" />
<EventRec doc="0" category="UserInterface" subcategory="PropertyPanel" name="No Properties" time="2015-10-24T19:58:38Z" seq="199" />
<EventRec doc="0" category="history" subcategory="en_US" name="Rectangle Tool" time="2015-10-24T19:58:38Z" seq="200" />
<EventRec doc="0" category="history" subcategory="en_US" name="Updated Export Asset Configurations" time="2015-10-24T19:58:43Z" seq="201" />
<EventRec doc="0" category="history" subcategory="en_US" name="Add Vector Mask" time="2015-10-24T19:58:59Z" seq="202" />
<EventRec doc="0" category="history" subcategory="en_US" name="Delete Vector Mask" time="2015-10-24T19:59:01Z" seq="203" />
<EventRec doc="0" category="flashPanel" subcategory="com.adobe.DesignLibraries.angular" name="show" time="2015-10-24T20:01:32Z" seq="204" />
<EventRec doc="0" category="UserInterface" subcategory="Playground" name="switch to 00" time="2015-10-24T20:01:32Z" seq="205" />

iwehrman:
:joy:

Issue #3112, opened 2015-10-23
Text for suffix in Export panel appears too low on Surface Pro 3


ktaki:
<img width="255" alt="screen shot 2015-10-23 at 2 48 29 pm" src="https://cloud.githubusercontent.com/assets/2029597/10705884/5ec064f8-7996-11e5-80d9-bde01674e94f.png">

Is this easy to fix? If yes, then M3. If not M4.

iwehrman:
I can repro his. @placegraphichere: does anything look weird to you in the CSS here? If you have a small, suggested fix, I will test.

iwehrman:
This `input` needs `font-size: 1.4rem`, but I'm not sure where to put it.

ktaki:
Verified the fix.

PR #3114, opened 2015-10-23
fix for #3112

placegraphichere:


iwehrman:
Tiny: 
image

Huge: 
image

iwehrman:
Which is to say, this works great. Verified at multiple resolutions and at the native SP3 resolution. I think this should be merged for an M3 push on Monday morning. Assigning to @shaoshing to get a second opinion, and a second test. Also, @ktaki, if you're out there, please feel free to test as well!

ktaki:
Yes, this looks good.

Issue #3115, opened 2015-10-24
Recent history changes break the opacity slider in the color picker


shaoshing:
Open any color picker, drag the opacity handle and start sliding: receiving errors related to history state id unmatch. 

<img width="1028" alt="screen shot 2015-10-23 at 5 24 06 pm" src="https://cloud.githubusercontent.com/assets/118264/10707673/e6d3f4d6-79aa-11e5-89c2-d0612d67a648.png">

As far as I know, the first related commit is from PR #3052, so I assign to @mcilroyc to take a look.

iwehrman:
@mcilroyc @shaoshing: I'm not sure if this is actually FBNC based on @mcilroyc's comments in #3116. Assigning back to @shaoshing to work it out with @mcilroyc.

mcilroyc:
Note: I believe @iwehrman's #3119 further addresses this issue.  The concerns I mentioned in #3116 (invalidating the history cache) seem to be fixed with 3119

PR #3116, opened 2015-10-24
Remove the capture of ID from historyState events

mcilroyc:
Reduces the impact of #3115.  In PR #3052 I began capturing "ID" from the historyState events from photoshop.  But, given the subsequent simplification of that PR (which attempts to handle touch undo events, and was eventually simplified), it is not strictly necessary to track ID.  Thus, this PR backs out that single line of code, which should prevent the errors described in 3115.

That said, the color picker is still not working correctly wrt history; it sometimes seems to invalidate the history states.  I believe this regression happened independent of 3052, possibly much earlier.

### DIFF
--- a/src/js/actions/history.js
+++ b/src/js/actions/history.js
@@ -296,7 +296,6 @@ define(function (require, exports) {
         var payload = {
             source: "listener", // for human convenience
             documentID: documentID,
-            id: event.ID,
             name: event.name,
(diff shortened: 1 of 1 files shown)


iwehrman:
I also notice that after undoing any change in the color picker, the picker sliders are not updated (though the swatch in the main panel is.)

Issue #3117, opened 2015-10-24
Undo does not reset color picker sliders


iwehrman:
1. Open a color picker.
2. Drag either slider.
3. Cmd-z to undo. 

The color change is reversed on canvas and in the swatch in the main panel UI, but the color picker sliders do not update.

Noted in #3116

iwehrman:
It works correctly if you just click on a slider to change the color.

iwehrman:
CC @designedbyscience in case he has any ideas. I don't think this is related to incorrect history states, just incorrect state dynamics inside `ColorInput`.

mcilroyc:
FBNC - assigning to @chadrolfs for independent review.

PR #3119, opened 2015-10-25
Reset color picker sliders on undo/redo

iwehrman:
This fixes two distinct problems with the color picker and undo/redo:

1. The `ColorInput` listens for `timetravel` events and uses them to forcefully reset the positions of the sliders to an up-to-date value. But, currently, it is simply resetting the sliders to their current position, which achieves nothing. This fixes that by using a value derived by instead re-parsing the input props to determine the correct color to which the sliders should be reset. This involves factoring out the parsing logic currently in the `render` method (which is totally gross) into a separate method which can be called by `render` and by this `timetravel` handler. Mostly this is just a cut-and-paste refactoring, though I did have to make some minor changes to account for `swatchClassProps` and friends.
2. The history store currently emits the `timetravel` event when "adjusting" the history state, but in this case it also seems to need to call `updateDocument` in order to update the document model. But this means that the `timetravel` event currently happens well before the document model has been updated, which means even with fix 1) above, the `ColorInput` re-parses its input props before they've been correctly updated. My proposed workaround (which is not intended to be the last word, but instead to just be a minimal fix) involves an additional history event `FINISH_ADJUSTING_HISTORY_STATE`, which is emitted after the history-related `updateDocument` call is complete, and to have the history store emit its `timetravel` event when handling this, instead of earlier in the `ADJUST_HISTORY_STATE` event.

Addresses #3117. 

@mcilroyc is probably the best person to review this, but I welcome analysis and testing from anyone else!

### DIFF
--- a/src/js/jsx/shared/ColorInput.jsx
+++ b/src/js/jsx/shared/ColorInput.jsx
@@ -213,7 +213,9 @@ define(function (require, exports, module) {
         _handleHistoryStateChange: function () {
             var dialog = this.refs.dialog;
             if (dialog.isOpen()) {
-                var color = this.refs.colorpicker.props.color;
+                var info = this._extractColorInfo(),
+                    color = info.color;
+
                 this.updateColorPicker(color, true); // don't emit a change event
             }
         },
@@ -303,13 +305,13 @@ define(function (require, exports, module) {
             this.refs.colorpicker.focusInput();
         },
 
-        render: function () {
-            var swatchClassSet = null,
-                swatchClassProps = {
-                    "color-input": true
-                };
-
-            // Normalize to arrays
+        /**
+         * Extract color information from the value supplied to the component.
+         *
+         * @private
+         * @return {{value: ?Color, label: string, color: Color, colorTiny: object=}}
+         */
+        _extractColorInfo: function () {
             var defaultValue = this.props.defaultValue,
                 valueArray = !Immutable.Iterable.isIterable(defaultValue) ?
                     Immutable.List.of(defaultValue) : defaultValue,
@@ -323,7 +325,6 @@ define(function (require, exports, module) {
                 if (typeof value === "string") {
                     label = value;
                     color = Color.DEFAULT;
-                    swatchClassProps["color-input__invalid-color"] = true;
                 } else {
                     if (this.props.opaque) {
                         value = value.opaque();
@@ -336,7 +337,6 @@ define(function (require, exports, module) {
                 }
             } else {
                 label = strings.TRANSFORM.MIXED;
-                swatchClassProps["color-input__mixed"] = true;
 
                 // The colors aren't completely uniform, but maybe the opaque colors are?
                 value = collection.uniformValue(valueArray.map(function (color) {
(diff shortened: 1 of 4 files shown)


shaoshing:
Ah, that's an excellent attempt on fixing the issue! I happened to looking at the same issue but did not claim it because I was not confident and don't want to slow it down. I ended up taking another path: since `ColorPicker#props.color` reflects the current value, or the value after undo/redo, the component should be able to re-render automatically. Currently, we took a different way by updating the ColorPicker manually based on history change, as we also want to avoid unnecessary re-rendering when it receives a new color prop (https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/jsx/shared/ColorInput.jsx#L183), because apparently the new color prop is always the same as the current color in the `ColorPicker`, except for undo/redo. However, the optimization is not working as ``ColorPicker` has [PureRenderMixin](https://facebook.github.io/react/docs/pure-render-mixin.html): it will always re-render upon receiving new props that are different with the current props:

screen shot 2015-10-25 at 12 03 06 pm
(ColorPicker is rendered twice for a single change)

To make it works, we need a custom SCU for `ColorPicker` that updates when `this.state.color !== nextProps.color` (it need more handling here, because `this.state.color` is in HSVA, and `nextProps.color` is in RGBA), which deals with undo/redo cases nicely and allows the parent component `ColorInput` to get rid of history event listener. 

I will create a PR for the proposed changes for discussion, but it is not necessarily needed for M3 as it might be risky and this PR fixes the issue perfectly.

iwehrman:
I think the historical reason that `ColorPicker` was built this way is because we didn't want it to update the slider position whenever the document's (fill, stroke, whatever) color changed. This is becuase when you're dragging the slider around, the document's color always lags behind the the internal color state a bit, and if it worked in the usual way, re-rendering whenever there was a fresh `color` prop, the slider would always bounce back to where it just was. At least I _think_ this is what you're seeing; I could certainly be wrong. It may well be that in my part 1) above there is a simpler way to get the current document color. If you post your PR, I'll take a look :)

mcilroyc:
At first glance, I agree with the following:

1. The history change (as a first step)
1. @iwehrman's response to @shaoshing regarding the historical reason for the color picker approach
1. The willingness to see @shaoshing's alternate approach!

I'll take another look at the color picker changes in this PR first thing in the morning with fresher eyes.

src/js/jsx/shared/ColorInput.jsx
```diff
@@ -358,10 +358,36 @@ define(function (require, exports, module) {
...
+                colorTiny = colorInfo.colorTiny;
+
+            var swatchClassProps = {
+                    "color-input": true
+                };
+
+            // setup text and swatch based on the mixed-ness of the inputs
+            if (value) {
+                if (typeof value === "string") {
+                    swatchClassProps["color-input__invalid-color"] = true;
```
mcilroyc: This is just a nit, but it seems like it would be even cleaner to just derive these two booleans (invalid, mixed) in `_extractColorInfo` since it is already doing similar logic, and then assign them to a single clean `swatchClassProps` here.

mcilroyc:
Alright, this seems correct and safe.

PR #3120, opened 2015-10-25
Set delay for throttled and debounced actions

shaoshing:
We, somehow, did not set a delay for throttled and debounced actions, which make thees functions false throttled/debounced as the default delay is 0. To compare before and after, open the color picker and start dragging; it is much more responsive with the real throttled actions.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -111,7 +111,7 @@ define(function (require, exports, module) {
             allStores = _.merge(stores, testStores || {});
 
         this._flux = new Fluxxor.Flux(allStores, actions);
-        this._resetHelper = synchronization.throttle(this._resetWithDelay, this);
+        this._resetHelper = synchronization.throttle(this._resetWithDelay, this, 100);
         this._actionReceivers = this._createActionReceivers(this._flux);
     };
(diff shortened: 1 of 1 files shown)


iwehrman:
I do not want to change the default value of 0 for the throttled function. I think 0 for a default there makes perfect sense. I agree, however, that 0 as a default for the debounced function makes no sense. However, I'm not sure that _any_ default should be provide here. Basically, I think we should just remove the `Debounced` version of each action, and force callers to explicitly choose their own debounce timeout.

iwehrman:
(I'm also totally open to changing the throttle timeout for the color picker, type face changes, etc.)

shaoshing:
Hmm, I don't see the way for the callers to set a custom delay (e.g. `colors#setFillOpacityThrottled`), unless I overwrite the default throttled / debounced function, maybe I am missing something here?

iwehrman:
In `componentWillMount`, something like `this._setFillOpacityThrottled = synchronization.throttle(this.flux.actions.setFillOpacity, this, 500)`.

shaoshing:
Ah. Ok, I will create separate throttled function.

shaoshing:
I am clsoing this and will create a separate one.

PR #3124, opened 2015-10-26
Increase the number of generator connection retries during 'debug' mode

mcilroyc:
This helps generator crash crash recovery in debug mode, and should have no effect in non-debug mode.

When in debug mode, we do an abbreviated attempt to connect to generator in order to find the possible existence of a remote generator.  This is simply to make it convenient for developers.  The problem is that this leaves the connection in "super quick mode" which gives up too quickly after a crash-restart.  This quick-fix simply makes it "kinda quick mode".  

@chadrolfs - you may want to use this branch when initially testing the generator crash-restart functionality which is available as of dev mac 842.


iwehrman:
Very, very low risk.

Issue #3122, opened 2015-10-26
Copy/Paste text Appearance doesn't paste the correct Size or Linespacing on Res > 72


chadrolfs:
cmd/ctrl + alt/opt + n, set Resolution to n > 72 (ex. 150)
create 2 text layers and change all defaults for appearance for one of them
Click on Copy Style icon in Appearance panel
select the other text layer and click on the Paste Style icon

Result: Size and Linespacing don't paste correctly due to resolution dependency

chadrolfs:
![copypaste-hires](https://cloud.githubusercontent.com/assets/8442631/10734579/33bf1f60-7bc1-11e5-9c6a-f8c22801ea60.gif)

chadrolfs:
related to #3121

PR #3127, opened 2015-10-26
Font sizes and line heights extracted from layers are in pixels, not points.

iwehrman:
Please test w.r.t. font size and line height on documents of multiple resolution. Affects the sampler and copy/paste styles workflows.

Addresses #3122.


pineapplespatula:
after testing and review this LGTM.

Issue #3123, opened 2015-10-26
Copy/Paste Appearance or using Sampler with multi-run text fails with console errors


chadrolfs:
Create a Type layer that is multi-run (contains chars with different Typeface and/or color or size)
Copy Style (appearance panel icon)
create a new type layer
Paste Style

Result: `Action sampler.pasteLayerStyle failed: TypeError: undefined is not a function`

or 
Create scenario above and use Sampler tool to sample Typeface
Result: `Action type.applyTextStyle failed: TypeError: undefined is not a function`

PR #3128, opened 2015-10-26
Fix sampler and copy/paste text layer with multiple text styles

shaoshing:
Addresses issue #3123 

If a text has multiple text styles (multiple text ranges with different text style), we will use the first character style (and the stroke color) in these tools:

* sampler
* copy/paste appearance
* create text style in libraries panel

### DIFF
--- a/src/js/stores/font.js
+++ b/src/js/stores/font.js
@@ -234,7 +234,7 @@ define(function (require, exports, module) {
                 };
             }
 
-            if (textStyle.leading >= 0) {
+            if (Number.isFinite(textStyle.leading) && textStyle.leading >= 0) {
                 obj.lineHeight = {
                     type: "pt",
                     value: textStyle.leading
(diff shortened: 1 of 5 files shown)


src/js/stores/font.js
```diff
@@ -237,7 +234,7 @@ define(function (require, exports, module) {
                 };
             }
 
-            if (textStyle.leading >= 0) {
```
shaoshing: This will filter out `null` value in `textStyle.leading`, which is also `null >= 0`

iwehrman:
This looks OK to me so far.

iwehrman:
Works OK for me so far.

iwehrman:
Well, I'm satisfied. Someone else should take a look and test independently though.

pineapplespatula:
looks good. i'll test a little more and then merge it

shaoshing:
Thanks everyone

Issue #3126, opened 2015-10-26
Unable to commit text with Esc or Enter or continue to type after Paste


chadrolfs:
create a type layer
(in edit mode) copy some of the on-canvas text (cmd + c), paste (cmd + v)
hit Enter or Esc

Result: Nothing

Workaround: select a different tool or switch focus away from app

chadrolfs:
related to fix for #3020

chadrolfs:
it’s actually worse than just Esc/Enter not committing the text
you can’t continue typing as well unless you hit a tool shortcut key

iwehrman:
The fact that something changes when you hit a tool shortcut key means that, somehow, keyboard focus has gone back to CEF even while we're in a tool modal state. This is not how `FOCUS_PROPAGATE` mode is supposed to work. I suspect that this is at least partially a core issue.

tiburbage:
I'll got a core bug to sync with this. John will need to investigate. I'm assuming this is a regression, so meanwhile, it will be useful to try to determine when this regression occurred. If you explicitly re-acquire focus after doing a paste, does that improve the outcome?

tiburbage:
https://watsonexp.corp.adobe.com/#bug=4079392

iwehrman:
FBNC to @chadrolfs. @tiburbage: the Watson bug can be closed. There are no new focus issues here; that was a bad early guess from me.

jsbache:
Watson bug closed. I will remove the ps core tags

PR #3129, opened 2015-10-26
Restore policies after nativePaste only if not currently in modal tool state

mcilroyc:
…by toolModalStateChanged - even though those are received only at the end of modal placement.

The theory is that, as a workaround for the un-timely toolModalStateChanged events, we must suspend policies at the beginning of nativePaste.  But, because we DO end up receiving toolModalStateChanged (both enter and exist, in quick succession) after the user places the object, then we do not strictly need to restore policies "finally" as part of nativePaste.

This hopes to address #3126 which was restoring policies while still in text edit modal state.

### DIFF
--- a/src/js/actions/edit.js
+++ b/src/js/actions/edit.js
@@ -98,9 +98,11 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var nativePaste = function () {
-        // FIXME: the suspend/restore policies hack is for pasting smart object from other sources (e.g. Illustrator).
+        // FIXME: this suspend policies hack is for pasting smart object from other sources (e.g. Illustrator).
         // The cause is similar to the place-object menu commands: DS is not receiving "toolModalStateChanged" events
         // until the object is committed.
+        // To avoid restoring policies while in a modal state, text in particular, first check the tool store
+        // modal state before restoring policies.  Super. Hack.
         var policyStore = this.flux.store("policy"),
             suspendPromise;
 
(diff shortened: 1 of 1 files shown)


iwehrman:
Can you sprinkle a comment about this in here along the lines of "This looks like a bug because we aren't restoring policies, but it isn't because...."

mcilroyc:
comment elaborated, commit amended.

pineapplespatula:
the code for this looks fine, but i would like to wait for approval by @chadrolfs or @ktaki

iwehrman:
This fails when attempting to paste text into the PS canvas (not in a modal tool state).

iwehrman:
Pasting an image copied from a web page also fails in a slightly different way.

ktaki:
"This fails when attempting to paste text into the PS canvas (not in a modal tool state)." After this, the shortcut keys for the tools selects tools for standard Ps.

iwehrman:
I think this is the fix we want to go with for M3. It works for me so far.

iwehrman:
For posterity, I'll note that we have to suspend policies in `nativePaste` in the first place due to this core bug: 
[https://watsonexp.corp.adobe.com/#bug=4074688](https://watsonexp.corp.adobe.com/#bug=4074688)

pineapplespatula:
this should be rebased to 1 PR

mcilroyc:
re-ba-ba-based

iwehrman:
Cross your fingers.

Issue #3131, opened 2015-10-26
Mixed states don't show correctly 


chadrolfs:
create two rects with different fill colors
select them both

Result:
![screen shot 2015-10-26 at 1 36 10 pm](https://cloud.githubusercontent.com/assets/8442631/10741714/67b93d44-7be6-11e5-9b15-75c9123e7fd0.png)

Expected:
![screen shot 2015-10-26 at 1 36 32 pm](https://cloud.githubusercontent.com/assets/8442631/10741723/6ebd37a8-7be6-11e5-804e-ca9e4c37c4c2.png)

chadrolfs:
81520b749623955e59eca0551574d4b207bdd7fb is the first bad commit
commit 81520b749623955e59eca0551574d4b207bdd7fb
Author: Ian Wehrman <wehrman@adobe.com>
Date:   Sun Oct 25 09:22:28 2015 -0700

    Upon receiving a timetravel event from the history store, re-parse the color props in order to reset the color picker sliders.

iwehrman:
FBNC to @chadrolfs.

PR #3133, opened 2015-10-26
Addresses bug where mixed class was not set properly on swatches

designedbyscience:
`colorInfo.value` was never null after last `colorInfo` changes. 

Addresses #3131


iwehrman:
LGTM.

PR #3135, opened 2015-10-26
Updated by ALF automation.

fzhu1978:
Translation update. 

@iwehrman Please help to review and merge!


src/nls/ja/strings.json
```diff
@@ -237,7 +237,7 @@
         "SECTION_COLLAPSE": ": ダブルクリックして折りたたむ",
         "SECTION_EXPAND": ": ダブルクリックして展開",
         "PANEL_COLUMN_COLLAPSE": ": クリックして折りたたむ",
-        "PANEL_COLUMN_EXPAND": ": クリックして展開",
+        "PANEL_COLUMN_EXPAND": " : クリックして展開",
```
iwehrman: @fzhu1978 is this correct? it seems inconsistent with the other strings.

joelrbrandt: This doesn't seem like a whitespace problem -- it seems like an "insertion of the replacement string" problem. I can't imagine the `: ' should be in the string at all (for lines 237-240, not just 240). It looks like somehow the `":` that closes the key is getting copied to the value.

Just my $0.02...

fzhu1978: I sent mail to iDev to see why they update the translation.

Feng

iwehrman: The `:` should be in the string. It's so that we can concatenate two strings and get something like `Title: Header`.

iwehrman: But, as-is, we'll end up with `Title : Header` instead of `Title: Header`.

iwehrman:
@fzhu1978: since we only have ~1 hour until our RCS deadline, I may just merge this and then remove the space character in a separate commit.

fzhu1978:
It's out of work time in China, I will revert the translation and create new pull request.

src/nls/ja/strings.json
fzhu1978: The core string is $B!H(B: click to expand$B!I(B, I think the linguists follows the rules.

fzhu1978:
I removed the space.  I will delete localization branch and create new clean pull request.  

Close it.

Issue #3134, opened 2015-10-26
Style copy-paste from vector to type doesn't replace the color


iwehrman:
It works from type color to vector fill color, but not the other way around.

chadrolfs:
verified fix on Mac/Win, closing

chadrolfs:
@pineapplespatula logged related issue #3141

PR #3138, opened 2015-10-26
copy/paste: set text color by fill color, if style is not copied from text layer.

shaoshing:
Addresses issue #3134 . 

Also fixed bugs related to out of sycn between `Text#firstCharacterStyle` and `Text#characterStyle`: after updating text style (size, color, tracking), clicking copy button will copy the out-dated style.

### DIFF
--- a/src/js/actions/sampler.js
+++ b/src/js/actions/sampler.js
@@ -598,6 +598,8 @@ define(function (require, exports) {
             textStylePromise = (!style.typeStyle || textLayers.isEmpty()) ? Promise.resolve() :
                 this.transfer(typeActions.applyTextStyle, document, textLayers, style.typeStyle, actionOpts,
                 { ignoreAlpha: false }),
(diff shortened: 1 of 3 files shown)


shaoshing:
I also tested copy/paste between text layer, vector layer, and graphic layer (only has blend mode and opacity attributes), and they are all working correctly now

pineapplespatula:
@shaoshing  this doesnt seem to replace the opacity or the text color

iwehrman:
I'm seeing it replace the color both directions. What are your repro steps to fail to catch teh color?

iwehrman:
I'm seeing it replace opacity too.

DivyaPrabhakar:
I'm with Ian -- I am seeing it replace opacity, color and size when it is copied from one text layer to another. When copying from type to vector, it copies color and opacity. The only thing not copied is the effects but I don't think it is supposed to copy effects.

pineapplespatula:
it only copies layer opacity. not fill color opacity ( which is because text layers only have layer opacity and not fill opacity), 
but we can defer that bug for now

Issue #3130, opened 2015-10-26
Place fails for 2nd asset when modifier is used


chadrolfs:
cmd/ctrl + n
drag a cc library asset onto canvas (don't commit)
drag a 2nd asset onto canvas hold down opt/alt

Result: only 1st asset is placed.  nudge AB will force update the layers panel but will result in 
`Action layers.resetIndex failed: Error: Too many layers in layer index`

chadrolfs:
![place-error](https://cloud.githubusercontent.com/assets/8442631/10741013/b5b8cf5e-7be2-11e5-970d-4ab8254e60de.gif)

chadrolfs:
there is a secondary intermittent issue where you end up with two children with the same key
to try to repro that follow these steps:
cmd/ctrl + n
drag a cc library asset onto canvas (don't commit)
click once in the Lib panel (places 1st asset) then drag a 2nd asset onto canvas hold down opt/alt
select them both
Result: swap is greyed out

![screen shot 2015-10-26 at 1 13 32 pm](https://cloud.githubusercontent.com/assets/8442631/10741183/a5eb9a10-7be3-11e5-90f3-8c7b5ad9c513.png)

iwehrman:
As of #3140, this is "fixed" by simply preventing the drag from libraries while in a tool modal state. If you think we'd like to file a bug about providing a somewhat better UX for this, go for it!

FBNC to @chadrolfs.

PR #3140, opened 2015-10-26
Temporarily disable drag-and-drop for libraries graphic while in modal tool state

shaoshing:
Addresses issue #3130 

This is a wrok around due to the reason that modifiers are not updated while PS is in modal tool state.


mcilroyc:
seems reasonable to me.

iwehrman:
I think this is great given the time constraints. Later we can spruce this up to improve the UX a little, e.g., to make the cursor disabled (or something) when the drag operation is guaranteed to fail. But, for now, I think this is perfect.

chadrolfs:
Checking this out, looking good so far.  checking on Windows now

Issue #3146, opened 2015-10-27
Swap enabled for single selected CC Library asset


chadrolfs:
drag a CC Lib asset onto canvas (hit Enter to place)
Result: Swap menu cmd and transform panel icon is enabled

Expect: Swap should only be enabled when only 2 valid objects are selected

chadrolfs:
drag out and select a shape, result: swap is greyed out.

chadrolfs:
Currently, I can only repro on resolution >72.

Workaround: drag LLSO onto canvas, hit Swap icon, all is set right with the world (apparently).

shaoshing:
@chadrolfs I think this is fixed by #3150, can you verify on that branch?

iwehrman:
FBNC to @chadrolfs.

Issue #3145, opened 2015-10-27
Dragging assets onto canvas sometimes results in two children with the same key; unable to swap


chadrolfs:
cmd/ctrl + n (or cmd/ctrl + alt/opt + n and set Resolution to >72)
drag one asset from CC Lib onto canvas (repeat 2x)
select both images

Result: swap is dimmed and console outputs
Warning: flattenChildren(...): Encountered two children with the same key, `.$3283=15`. Child keys must be unique; when two children share a key, only the first child will be used.

chadrolfs:
this is an intermittent bug, not always reproducible but seems easier for me to repro on docs with resolution >72

chadrolfs:
<img width="1044" alt="screen shot 2015-10-27 at 8 45 38 am" src="https://cloud.githubusercontent.com/assets/8442631/10763654/0c423c66-7c88-11e5-8ef7-fa3bbb4e2c2d.png">

shaoshing:
I found that if you expand a graphic that is a embedded SO, you will end up adding the layer twice and trigger the error (also #3146)

Reproduce: drag out "Embedded" from "Test - OK" with ALT key pressed

iwehrman:
FBNC to @chadrolfs.

PR #3150, opened 2015-10-27
Fix for duplicate layers due to adding graphic from cc libraries.

shaoshing:
This PR fixes duplicate layers in our LayerStructure model due to adding a graphic with ALT key pressed. Usually, when the ALT key is pressed, PS will add the layers contained in the graphic, instead of creating LLSO, and will not enter modal tool state. However, if the graphic contains only one layer and it is an embedded SO, PS will enter modal state tool, which causes adding the duplicate layer when we handle the `placeEvent`.

So the solution is to check whether the layer is existed when handling `placeEvent`. I also added cheker to throw error in the `layers.addLayer`, if we are trying to add duplicate layers. 

To test: drag graphic `Embedded` from library `Test - OK ...` (or import the graphic from http://adobe.ly/1Mh2XAM) onto the canvas while holding the ALT key.

Addresses issues: #3145 #3146 

While testing, you may see the following error, which is because the layer is not ready for reading (same problem again, and I remember we have watson for it) but it is totally ok for this case because we will add the layer later when handling the `placeEvent`.

screen shot 2015-10-27 at 3 21 50 pm

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -381,6 +381,14 @@ define(function (require, exports) {
         if (typeof layerSpec === "number") {
             layerSpec = [layerSpec];
         }
+        
+        layerSpec.forEach(function (layerID) {
+            var layer = document.layers.byID(layerID);
+
+            if (layer) {
+                throw new Error("Trying to add a layer that is already existed: " + layerID + " " + layer.name);
+            }
+        });
 
         if (selected === undefined) {
             selected = true;
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -386,7 +386,7 @@ define(function (require, exports) {
             var layer = document.layers.byID(layerID);
 
             if (layer) {
-                throw new Error("Trying to add a layer that is already existed: " + layerID + " " + layer.name);
+                throw new Error("Trying to add a layer that already exists: " + layerID + " " + layer.name);
             }
         });
 
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -850,6 +850,7 @@ define(function (require, exports) {
                         //
                         // FIXME: we should instead get IDs back from Photoshop when layers are placed with modifier, 
                         //        so we don't have to get all the layer IDs.
+                        //        https://watsonexp.corp.adobe.com/#bug=4080071
                         if (hasAlt) {
                             return this.transfer(layerActions._getLayerIDsForDocumentID, currentDocument.id)
                                 .bind(this)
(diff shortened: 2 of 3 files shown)


chadrolfs:
I want to include this in the PR 3047 push as well

src/js/actions/libraries.js
```diff
@@ -850,6 +850,7 @@ define(function (require, exports) {
                         //
                         // FIXME: we should instead get IDs back from Photoshop when layers are placed with modifier, 
                         //        so we don't have to get all the layer IDs.
+                        //        https://watsonexp.corp.adobe.com/#bug=4080071
```
shaoshing: a new watson bug that could free us from this hack.

pineapplespatula:
This looks good to me @chadrolfs have you tested this branch already?

iwehrman:
Can someone please repro this bug in compiled mode to check whether or not we can recover from it automatically? This is in case we don't get BRC approval for the PR. Or, if we _don't_ recover automatically, that will help make a stronger case to the BRC.

shaoshing:
I haven't test this in compiled mode. But adding duplicate layers is not going to throw any errors, and you will see the swap button (and other buttons involve 2 layers) enabled if you selected the layer, and also a bunch of warnings when the layers panel re-rendered. However, if you click to swap, DS will throw error, and that gives us a chance to recover everything. Base on this, I would say this is *not recoverable*

iwehrman:
It sounds like you're saying that, once the swap action fails it _will_ recover. Is there some scenario in which we explicitly do not recover? Don't get me wrong, I definitely want to fix this :grimacing: I just want to confirm that this is in the class, "temporarily out of date layers panel which will fix itself once the inconsistency causes a problem."

src/js/actions/layers.js
```diff
@@ -381,6 +381,14 @@ define(function (require, exports) {
         if (typeof layerSpec === "number") {
             layerSpec = [layerSpec];
         }
+        
+        layerSpec.forEach(function (layerID) {
+            var layer = document.layers.byID(layerID);
+
+            if (layer) {
+                throw new Error("Trying to add a layer that is already existed: " + layerID + " " + layer.name);
```
iwehrman: Grammar nit: this should either read "Trying to add a layer that already exists" or "Trying to add a layer this is already present".

iwehrman:
The patch looks great, besides the grammar nit. If we can get this past the BRC, I am definitely in favor of merging it.

shaoshing:
Yes, then it falls into the class "temporarily out of date layers panel which will fix itself once the inconsistency causes a problem."

iwehrman:
I think we've decided against taking this for M3, but we surely want it in master!

Issue #3151, opened 2015-10-27
Multiple selection: setting W also sets Y to a standard 


chadrolfs:
cmd/ctrl + n
draw out several rects of different sizes (both W & H)
select them all with Select tool (W/H/X/Y are all mixed values so fields show --)
enter a value into W

Result: Y value also gets set, making the shapes all the same size

chadrolfs:
![setx-fixesy](https://cloud.githubusercontent.com/assets/8442631/10775573/df46d282-7cc6-11e5-9b81-220f5feb76fc.gif)

chadrolfs:
wondering if this will be addressed by #3047 since it deals with Transformation fields.

chadrolfs:
This broke sometime after the Sprint 33 build ( 20f8742f7188333db70ddac6688838d118723917 )

mcilroyc:
chad, it seems broken to me on the commit you referenced (20f8742)

mcilroyc:
by bisecting starting with 32.1 sprint release, I found:
0da3c2f81799372c8696d801be21e64bad508c3f is the first bad commit

mcilroyc:
stealing this for now - but @shaoshing (original assignee) let me know if you want it back, or whatnot

chadrolfs:
Thanks Cory.

iwehrman:
FBNC to @chadrolfs.

PR #3152, opened 2015-10-28
Fix transform setSize process to avoid newly derived size values from…

mcilroyc:
… being propagated forward when multiple layers selected.  Addresses #3151.  

The reference point feature introduced this bug where, when changing "W" value of a set of layers, the "H" value of the first layer was being propagated down to each subsequent layer in the selection.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -253,13 +253,13 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer} targetLayer
-     * @param {{w: number=, h: number=}} size
+     * @param {{w: number=, h: number=}} nextSize
      * @param {Array.<{layer: Layer, w: number, h: number}>} resizeResults payload for updating each layer's bounds
      * @param {string} refPoint in the format of vertical direction then horizontal
      *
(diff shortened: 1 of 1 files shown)


iwehrman:
@DivyaPrabhakar should review this.

chadrolfs:
just hit this
ice_screenshot_20151028-093013

mcilroyc:
@chadrolfs can you please file a separate bug for that, with repro steps

chadrolfs:
yeah, ok, not sure if it happens on master or not.  Checking...  Steps are pretty easy.

chadrolfs:
It's actually fallout from issue 3144.  Once that is fixed, this shouldn't happen.  Testing on that branch now.  Gotta love cross-branch-testing.

chadrolfs:
yeah, that issue doesn't happen on PR 3047 branch

chadrolfs:
What should happen if you enter negative numbers?

mcilroyc:
It should set the value to the minimum allowable value, I think

iwehrman:
This doesn't change the logic; it just corrects a programming error. So, I feel perfectly confident in merging it without having additional reviewers.

Issue #3153, opened 2015-10-28
Entering negative transform value for LLSO (and other layers, i.e. Type) results in badness


chadrolfs:
cmd/ctrl + n
drag a library asset onto canvas
click in Transform field (W) and enter a negative number (-50), Enter

Result: all fields are now blank.  If you click on Layer 1 you'll error
`Action transform.setSize failed: Error: Photoshop returned an error when playing command number: 2 Photoshop code: -1 Photoshop message: error: -1 message: Could not complete the command because the initial bounding rectangle is empty.`

chadrolfs:
![screen shot 2015-10-28 at 9 53 42 am](https://cloud.githubusercontent.com/assets/8442631/10796137/0d195872-7d5a-11e5-8395-331aa7921baa.png)

chadrolfs:
![screen shot 2015-10-28 at 9 57 37 am](https://cloud.githubusercontent.com/assets/8442631/10796156/1f1dcb5c-7d5a-11e5-83a2-2a818897d8a8.png)

shaoshing:
This is probably a core issue. The valid range for size is 0.1 ~ 32000 in DS. However, if we set the width / height to 0.1 for some graphics (for example, `Test - OK` > `Climber`), they will turn into invisble with null w/h/x/y.

shaoshing:
Watson: https://watsonexp.corp.adobe.com/#bug=4080692

ktaki:
This issue still exists, but console does not show any errors now. See the #3526.

ktaki:
Opened a bug in JIRA. https://jira.corp.adobe.com/browse/PS-884

chadrolfs:
I'm still able to hit this:

![negxformvalues](https://cloud.githubusercontent.com/assets/8442631/13188538/e35614a6-d706-11e5-9e8d-ea20c3b4b3f2.gif)

chadrolfs:
using pgdev.1081 w/master @ 58fc655e7f74cf28b2550a5c3f87d4060b585a5e

iwehrman:
Isn't this the different bug that has to do with proportional locking? See #3725.

chadrolfs:
Oh right, my bad.  As written, this is fixed.

PR #3155, opened 2015-10-28
Increase the minimum layer size for a core issue

shaoshing:
Addresses #3153: if you set a LLSO's size to less than 0.5, it will turn into a bad state with null X/Y/W/H. More testings show that this is not limited to LLSO: any pixel layers and SOs that link to an image will have the same issue. 

Reproduce:

drag-n-drop an image from browser to create a pixel layer, and set the width / height to less than 0.5

I created a watson bug for this (https://watsonexp.corp.adobe.com/#bug=4080692), and the most risk-free work around to me is increasing the minimum size to 0.5 pixel. Please let me know if you disagree with the solution (breaking UX?) and we can come up with a better one.

### DIFF
--- a/src/js/jsx/sections/transform/Size.jsx
+++ b/src/js/jsx/sections/transform/Size.jsx
@@ -36,8 +36,12 @@ define(function (require, exports, module) {
         strings = require("i18n!nls/strings"),
         collection = require("js/util/collection");
 
+    // The minimum layer size in PS is 0.1 pixel, but it has an issue with some smart objects and graphic layers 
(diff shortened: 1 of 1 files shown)


iwehrman:
Is it even worth making this change if this doesn't go in for M3? Maybe we should just leave the correct JS code in place and wait for core to fix things on their side...

shaoshing:
@iwehrman ok, that sounds good. I will update #3153  with core labels, and close this PR (we can repon if necessary.)

PR #3156, opened 2015-10-28
refactor code to add handler to reset border policies

pineapplespatula:
This shouldn't be merged until after we GM. Adding a review only label

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -441,15 +441,11 @@ define(function (require, exports) {
             .then(function (result) {
                 // After setting everything, dispatch to stores
                 this.dispatch(events.tool.SELECT_TOOL, result);
-
-                if (!toolStore.getVectorMode() || !nextTool.handleVectorMaskMode) {
-                    return this.transfer(resetBorderPolicies);
-                }
             });
     };
     selectTool.reads = [];
     selectTool.writes = [locks.JS_TOOL, locks.PS_TOOL];
-    selectTool.transfers = [resetBorderPolicies, installShapeDefaults, shortcuts.addShortcut,
+    selectTool.transfers = [installShapeDefaults, shortcuts.addShortcut,
         shortcuts.removeShortcut, "layers.deleteSelected", "layers.resetLayers",
         policy.removePointerPolicies, policy.removeKeyboardPolicies, policy.addPointerPolicies,
         policy.addKeyboardPolicies, policy.setMode];
@@ -553,6 +549,7 @@ define(function (require, exports) {
      * @type {function()}
      */
     var _toolModalStateChangedHandler,
+        _documentChangeHandler,
         _vectorMaskHandler,
         _vectorSelectMaskHandler;
 
@@ -800,6 +797,12 @@ define(function (require, exports) {
             toolStore = this.flux.store("tool"),
             tools = toolStore.getAllTools();
 
+        var _documentChangeHandler = function () {
+            this.transfer(resetBorderPolicies);
+        }.bind(this);
+
+        this.flux.store("document").on("change", _documentChangeHandler);
+
         // Listen for modal tool state entry/exit events
         _toolModalStateChangedHandler = this.flux.actions.tools.handleToolModalStateChanged.bind(this);
         descriptor.addListener("toolModalStateChanged", _toolModalStateChangedHandler);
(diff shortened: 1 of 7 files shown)


src/js/actions/tools.js
```diff
@@ -793,6 +790,12 @@ define(function (require, exports) {
             toolStore = this.flux.store("tool"),
             tools = toolStore.getAllTools();
 
+        var _documentChangeHandler = function () {
+            this.transfer(resetBorderPolicies);
```
iwehrman: You should never transfer inside a bare event handler. Instead, this should just call a flux action.

iwehrman: Also, we can do a lot better here. In particular, the handler should be debounced by a small amount (100 or 200ms), and ideally would only be triggered if the selected layers in the active document changes. Currently, I do not think that this will trigger if you simply switch documents because the document store doesn't know anything about which document is currently active.

iwehrman:
A little more iteration is required on the change listener, but this is definitely going in the right direction! I wonder if there are other places where we can take this approach? What about updating the recent files list?

pineapplespatula:
We need this to be throttled and not debouced so as to not delay calling reset border policies

src/js/actions/tools.js
```diff
@@ -793,6 +791,14 @@ define(function (require, exports) {
             toolStore = this.flux.store("tool"),
             tools = toolStore.getAllTools();
 
+        var debouncedResetBorderPolicies =
+            synchronization.debounce(this.flux.actions.tools.resetBorderPolicies, this, 100),
+            _documentChangeHandler = function () {
+                debouncedResetBorderPolicies();
```
pineapplespatula: This should be throttled and not debounced

src/js/actions/tools.js
```diff
@@ -881,7 +887,7 @@ define(function (require, exports) {
      */
     var onReset = function () {
         descriptor.removeListener("toolModalStateChanged", _toolModalStateChangedHandler);
-
+        this.flux.store("document").removeListener("change", _documentChangeHandler);
```
iwehrman: Need to listen to application store changes as well to account for changes to the active document.

iwehrman:
One more gripe. Looking great, otherwise. Can't wait to merge!

pineapplespatula:
back to @iwehrman

iwehrman:
Actually, we may still need to do the thing where we keep track of what the last set of selected layers was before calling the throttled action. Look at what happens when you drag the color picker around...

@volfied: do you have an opinion about this?

volfied:
It might be to our benefit to do a basic check in resetBorderPolicies similar to a shouldComponentUpdate. Only disadvantage is we need to store the last call's layer models so we can quickly compare things.

src/js/actions/tools.js
```diff
@@ -791,8 +789,24 @@ define(function (require, exports) {
...
+            appStore = flux.store("application"),
             toolStore = this.flux.store("tool"),
-            tools = toolStore.getAllTools();
+            tools = toolStore.getAllTools(),
+            documentLayerBounds;
+
+        var throttledResetBorderPolicies =
+            synchronization.throttle(this.flux.actions.tools.resetBorderPolicies, this, 100),
+            _documentOrApplicationChangeHandler = function () {
+                var nextDocuemntLayerBounds = appStore.getCurrentDocument().layers.selectedChildBounds;
```
iwehrman: If there is no current document, this will throw an unhandled exception. Also, there's a typo in the variable name.

pineapplespatula:
rebased and back to @iwehrman

src/js/actions/tools.js
```diff
@@ -797,8 +795,28 @@ define(function (require, exports) {
...
-            tools = toolStore.getAllTools();
+            tools = toolStore.getAllTools(),
+            documentLayerBounds;
+
+        var throttledResetBorderPolicies =
+            synchronization.throttle(this.flux.actions.tools.resetBorderPolicies, this, 100),
+            _documentOrApplicationChangeHandler = function () {
+                var currentDocument = appStore.getCurrentDocument();
+
+                if (currentDocument) {
```
iwehrman: Seems like you want to clean up the `documentLayerBounds` if there is no current document?

pineapplespatula:
good catch. rebased and back to @iwehrman

iwehrman:
Let's give it a try!

Issue #3159, opened 2015-10-28
Export panel should be bottom aligned


iwehrman:
It doesn't make any sense for the Export panel to both be nearly clipped on the bottom and top-aligned:
![image](https://cloud.githubusercontent.com/assets/1075154/10805517/7e9e219a-7d8b-11e5-9a0b-693a24d5aea8.png)

More context: http://cloud.wehrman.org/image/3X0t00343o0s

mcilroyc:
FWIW, the export panel behaves differently for the Nth (rightmost) document when multiple documents are open

iwehrman:
This is now happening in the no-doc UI as well:
![image](https://cloud.githubusercontent.com/assets/1075154/13097214/3fec9e4a-d4d4-11e5-9d31-29047203c090.png)

PR #3160, opened 2015-10-28
Updated README screenshot

iwehrman:
What it says! You can click the "Display Rich Diff" button below to see the before/after. (I also filed #3159 in the process of making this.)


mcilroyc:
correct.

PR #3162, opened 2015-10-28
switch to using the doc id.

pineapplespatula:
switch to using doc IDs instead of index to avoid core bug where index could include non visible docs

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -278,21 +278,29 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var initInactiveDocuments = function (currentDocID, docCount) {
-        var otherDocPromises = _.range(1, docCount + 1)
+        var docReferences = _.range(1, docCount + 1)
             .map(function (index) {
-                var indexRef = documentLib.referenceBy.index(index);
-
-                // Only load essential properties for inactive documents
-                return _getInactiveDocumentByRef(indexRef)
-                    .bind(this)
-                    .then(function (document) {
-                        if (document.documentID !== currentDocID) {
+                return documentLib.referenceBy.index(index);
+            }),
+            otherDocPromises = descriptor.batchGetProperty(docReferences, "documentID")
+                .bind(this)
+                .then(function (documentIDs) {
+                    return documentIDs.filter(function (documentID) {
+                        return documentID !== currentDocID;
+                    });
+                })
+                .map(function (documentID) {
+                    var docRef = documentLib.referenceBy.id(documentID);
+                   
+                    // Only load essential properties for inactive documents
+                    return _getInactiveDocumentByRef(docRef)
+                        .bind(this)
+                        .then(function (document) {
                             this.dispatch(events.document.DOCUMENT_UPDATED, {
                                 document: document
                             });
-                        }
-                    });
-            }, this);
+                        });
+                });
 
         return Promise.all(otherDocPromises)
             .bind(this)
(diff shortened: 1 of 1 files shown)


src/js/actions/documents.js
```diff
@@ -273,15 +273,12 @@ define(function (require, exports) {
...
      *
-     * @param {number} currentIndex
+     * @param {number} currentDocID
      * @param {number} docCount
      * @return {Promise}
      */
-    var initInactiveDocuments = function (currentIndex, docCount) {
+    var initInactiveDocuments = function (currentDocID, docCount) {
         var otherDocPromises = _.range(1, docCount + 1)
-            .filter(function (index) {
```
iwehrman: You don't want to call `_getInactiveDocumentByRef` on the active document. This is the reason why we used the index in the first place. For this reason (i.e., to avoid an extra round-trip call to PS) I'm not sure we want to take this change.

pineapplespatula:
that is fair, we'll see how far we get on fixing this on the core side.

src/js/actions/documents.js
```diff
@@ -273,27 +273,36 @@ define(function (require, exports) {
...
+                })
+                .bind(this)
+                .map(function (documentID) {
+                    var indexRef = documentLib.referenceBy.id(documentID);
+                   
+                    // Only load essential properties for inactive documents
+                    return _getInactiveDocumentByRef(indexRef)
+                        .bind(this)
+                        .then(function (document) {
+                            if (document.documentID !== currentDocID) {
```
iwehrman: This if statement isn't necessary, right?

src/js/actions/documents.js
```diff
@@ -273,27 +273,36 @@ define(function (require, exports) {
...
+                return documentLib.referenceBy.index(index);
+            }),
+            otherDocPromises = descriptor.batchGetProperty(docReferences, "documentID")
+                .then(function (documentIDs) {
+                    return documentIDs.filter(function (documentID) {
+                        return documentID !== currentDocID;
                     });
-            }, this);
+                })
+                .bind(this)
```
iwehrman: For consistency, move the `.bind(this)` up to being the first method called on the original promise, after the `descriptor` call.

src/js/actions/documents.js
```diff
@@ -273,27 +273,36 @@ define(function (require, exports) {
...
+                .map(function (documentID) {
+                    var indexRef = documentLib.referenceBy.id(documentID);
```
iwehrman: This is no longer an "index". I suggest naming this variable `docRef`.

iwehrman:
Back 2 @pineapplespatula 4 nits.

chadrolfs:
Quick testing on mac - looks like it's fixed.  Validating on Windows now...

chadrolfs:
looks like this is still failing on Windows

src/js/actions/documents.js
```diff
@@ -273,27 +273,38 @@ define(function (require, exports) {
...
-            }, this);
+                })
+                .map(function (documentID) {
+                    var docRef = documentLib.referenceBy.id(documentID);
+                   
+                    // Only load essential properties for inactive documents
+                    return _getInactiveDocumentByRef(docRef)
+                        .bind(this)
+                        .then(function (document) {
+                            if (document.documentID === currentDocID) {
```
iwehrman: Just get rid of this check. I don't want to clutter the code with checks for every possible crazy thing Photoshop could possibly do, and there's no reason why you're `filter` step above wouldn't work reliably.

pineapplespatula:
i think we should pull this change since the PS change fixes it for us

iwehrman:
I agree. I think post M3 we should also request a new core API which is "list the open document IDs." That would save us one round-trip to PS on startup w.r.t. the current state of the art.

Issue #1946, opened 2015-08-07
LESS Variable Refactoring


designedbyscience:
Right now, we use `@warm-black` and such throughout the code rather than a variable named for the context (such as `@label-primary-color` in my new layers work). We should be replace non-context variables outside of colors.less.

iwehrman:
I believe that @placegraphichere has done this with his theming work, and it's amazing. If I'm missing something, please re-open!

placegraphichere:
@designedbyscience is right, we should leave this open. It should be more of a non-context name. So it will not be theme specific. I think I missed this one in the work I did.

iwehrman:
master is open for pull requests!

placegraphichere:
I should add, we need this applied to `@warm-grey`. That's the one I missed.

iwehrman:
Fixed by #3171.

PR #3171, opened 2015-10-30
fix for #1946

placegraphichere:
Thrilling PR changing the name of an older variable names to be less specific for colors.


iwehrman:
Finally!

PR #3176, opened 2015-11-02
Factor out a DocumentIndex model

iwehrman:
This is a code cleanup that factors out a `DocumentIndex` model from the application store. It should preserve the existing functionality, but in a way (i.e., it is immutable) that should make writing `shouldComponentUpdate` functions a little easier. I also include a slight simplification of the `PanelSet` logic to fetch the set of initialized documents, reducing duplication of that logic. I haven't gone through the rest of the JSX to clean up more of this logic, but I thought it would be worthwhile to get this in now instead of waiting for more.

### DIFF
--- a/src/js/models/documentindex.js
+++ b/src/js/models/documentindex.js
@@ -1,0 +1,200 @@
+/*
+ * Copyright (c) 2015 Adobe Systems Incorporated. All rights reserved.
+ *  
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"), 
+ * to deal in the Software without restriction, including without limitation 
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
+ * and/or sell copies of the Software, and to permit persons to whom the 
+ * Software is furnished to do so, subject to the following conditions:
+ *  
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *  
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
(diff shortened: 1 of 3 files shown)


src/js/models/documentindex.js
```diff
@@ -0,0 +1,200 @@
...
+    var DocumentIndex = Immutable.Record({
+        /**
+         * An ordered list of document IDs
+         * @type {Immutable.List.<number>}
+         */
+        openDocumentIDs: Immutable.List(),
+
+        /**
+         * The index of the currently active document; negative if there are none
+         * @type {?number}
```
volfied: Semantic argument: Shouldn't this be `{number}` since it's not supposed to be null.

src/js/models/documentindex.js
```diff
@@ -0,0 +1,200 @@
...
+     * @param {number} documentID
+     * @param {number} position
+     * @param {boolean=} active
+     * @return {DocumentIndex}
+     */
+    DocumentIndex.prototype.setPosition = function (documentID, position, active) {
+        var nextIndex = this._updateDocumentPosition(documentID, position);
+
+        if (active) {
+            nextIndex = nextIndex.setActive(documentID, position);
```
volfied: This will not run if active is false. We may want to distinguish between undefined and false here.

src/js/models/documentindex.js
```diff
@@ -0,0 +1,200 @@
...
+
+        var nextActiveDocumentIndex;
+        if (hasOpenDocuments) {
+            nextActiveDocumentIndex = nextDocumentIDs.indexOf(nextActiveDocumentID);
+            if (nextActiveDocumentIndex < 0) {
+                throw new Error("Next active document ID not found in index: " + documentID);
+            }
+        } else {
+            nextActiveDocumentID = null;
+            nextActiveDocumentIndex = null;
```
volfied: You probably want `= -1` here.

volfied:
Got couple comments @iwehrman, otherwise looks good.

src/js/models/documentindex.js
iwehrman: I think that's the intended behavior. It only updates the active ID and index during setPosition if `active === true`.

iwehrman:
Nits fixed, methinks.

volfied:
:cake:

Issue #3066, opened 2015-10-21
Re-consider all mergeDeep calls


iwehrman:
The Immutable `mergeDeep` function seems to be a fertile source of both correctness and performance issues. We should do a code-cleanup pass of all our uses of this function to see if they can be replaced with faster, simpler data structure update calls. (See, e.g. #3057.)

iwehrman:
Fixed by #3177.

PR #3177, opened 2015-11-02
Replace mergeDeep calls in LayerStructure with shallow merge calls

iwehrman:
In the past, incorrect or unnecessary `mergeDeep` calls have caused both functional correctness and performance issues. This replaces all the `mergeDeep` calls inside of `LayerStructure`, where most of them exist. In the process, I noticed that the `LAYER_BOUNDS_CHANGED` event was unused, so I also deleted that, as well as its handler in the document store.

Partially addresses #3066. "Partially" because some calls still remain. In particular, one in the `dialog` store and a few in the `documentexports` model. Any ideas about those, @mcilroyc?

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1453,16 +1453,12 @@ define(function (require, exports) {
      * @param {Document} document
      * @param {Layer} layer
      * @param {boolean} visible Whether to show or hide the layer
-     * 
      * @returns {Promise}
      */
     var setVisibility = function (document, layer, visible) {
         var payload = {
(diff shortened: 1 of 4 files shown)


shaoshing:
This looks good to me, and all related features are working correctly. Back to @iwehrman for rebase.

iwehrman:
Rebased and ready 2 rock.

mcilroyc:
Looks good to me too, so I'll pull the trigger.  @iwehrman if there are still some mergeDeep instances that you want me to clean up, then assign the issue to me?

Issue #3132, opened 2015-10-26
Select overlay doesn't highlight siblings on hover


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/1n1y1M0z2h0N2O470V0z/Screen%20Recording%202015-10-26%20at%2001.52%20PM.gif)

chadrolfs:
related to #3110

volfied:
The hitTest order of layers have changed. Fix for this will be in SuperSelectOverlay.jsx updateMouseOverHighlights function. We need to filter out non-superselectable layers similar to what we do in superselect.click

iwehrman:
FBNC to @chadrolfs. Is there any way that we can sneak this into M3?

chadrolfs:
Probably not in the least.

PR #3178, opened 2015-11-02
Fix sibling highlighting when in a grouped layer

volfied:
Highlighting wouldn't work right when you were in a group and tried highlighting the siblings. Problem was, we would use whatever the hitTest was returning us, and with recent changes (or maybe it was wrong for the last two months) the order was different, and the layer we want to highlight would be preceded by it's owner group.

Now I consider what's super selectable, and only use the intersection of those with what PS tells us we're hovering over.

Addresses #3132


iwehrman:
Looks good! I sure wish we hadn't broken this for M3 :( @chadrolfs do you think there's any chance that we could get this past the BRC?

chadrolfs:
I really don't think so.  :(

PR #3180, opened 2015-11-02
Fix non-modal transfers

iwehrman:
Currently if an action is marked as `modal: true`, which means it may execute while Photoshop is in a modal state, but then transfers to an action with `modal: false`, which means it must not execute while Photoshop is in a modal tool state, the modal tool state will not be committed because the controller only handles this flag at the action-enqueue level and not the transfer level. This PR changes the controller so that the `modal` flag is handled uniformly for both cases, inside `FluxController.prototype._applyAction`.

Addresses: concerns.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -552,11 +552,11 @@ define(function (require, exports, module) {
      * @return {Promise}
      */
     FluxController.prototype._applyAction = function (action, actionReceiver, params, parentActionName) {
-        var lockUI = action.lockUI,
+        var flux = this.flux,
+            lockUI = action.lockUI,
             post = action.post,
+            modal = action.modal || false,
             actionName = this._actionNames.get(action),
-            preferences = this.flux.store("preferences").getState(),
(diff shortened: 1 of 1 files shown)


src/js/fluxcontroller.js
```diff
@@ -552,11 +552,11 @@ define(function (require, exports, module) {
...
      */
     FluxController.prototype._applyAction = function (action, actionReceiver, params, parentActionName) {
-        var lockUI = action.lockUI,
+        var flux = this.flux,
+            lockUI = action.lockUI,
             post = action.post,
+            modal = action.modal || false,
             actionName = this._actionNames.get(action),
-            preferences = this.flux.store("preferences").getState(),
-            checkPost = preferences.get("postConditionsEnabled"),
```
pineapplespatula: I'm not clear on why you removed checkPost ( or what it did )

volfied:
Changes seem good, of course, we may get some use errors from transfers.

iwehrman:
We should get _fewer_ errors.

Issue #3169, opened 2015-10-30
Color of shape resets to default


ktaki:
In Standard Ps, create a new doc, and draw a rectangle. Change the color. With the shape layer selected, switch to DS. RESULT: The color of the rectangle is reset to default (gray).

pineapplespatula:
We are setting the default of shapes intentionally. assigning to @placegraphichere for discussion

iwehrman:
I don't think @ktaki means that the default is changing. I think he means that the color of the existing layer is changing.

ktaki:
Sorry for the confuse on the title. Yes, color will reset.

volfied:
Adding to STR: You need to leave rect tool selected.

volfied:
Gets reset because tools.resetShapeDefaults gets called before document is initialized, which skips the selection dance and forces the selected layer's colors.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4087965

PR #3182, opened 2015-11-02
Prioritize documents.beforeStartup higher than tools.beforeStartup

volfied:
Problem was, tools.installShapeDefaults would directly change the current active fill/stroke color if there was no document model loaded. And if we had rectangle tool selected in PS, during `tools.beforeStartup`, we would initialize the rectangle tool, and since `documents.beforeStartup` was not yet called, reset the defaults.

Solution was to move initTool call to tools.afterStartup, so document can be initialized in our models before we try to set shape defaults.

Addresses #3169

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -359,7 +359,7 @@ define(function (require, exports) {
             });
     };
     initActiveDocument.reads = [locks.PS_DOC];
-    initActiveDocument.writes = [locks.JS_DOC];
+    initActiveDocument.writes = [locks.JS_DOC, locks.JS_APP];
     initActiveDocument.transfers = [historyActions.queryCurrentHistory, "layers.deselectAll"];
 
     /**
@@ -1166,5 +1166,11 @@ define(function (require, exports) {
     exports.onReset = onReset;
 
     exports._getDocumentByRef = _getDocumentByRef;
-    exports._priority = -99;
+    
+    // This module needs to have a higher priority than tools
+    // Otherwise, if rectangle tool is selected, we set shape defaults
+    // which causes PS to set the fill color, and if there is an active document
+    // with a shape layer selected, this will reset the color of that shape
+    // But if document is initialized, we do a selection dance, avoiding this situation
+    exports._priority = 1;
 });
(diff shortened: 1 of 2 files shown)


iwehrman:
Hmmm, I don't like this solution because we use `beforeStartup` to know when the UI should be unlocked. We don't want it unlocked with the brush tool. There's already a mechanism to ensure that one action module's `beforeStartup` method runs (synchronously) before another one using `_priority`. Would that be sufficient? Or is there some other solution here?

volfied:
We have a comment in tools.priority 
```
// This module must have a higher priority than the document module to avoid
    // duplicate current-document updates on startup, but lower priority than the
    // ui module so that defaults, which tool select handlers rely on, can be set.
```

If you think we've cleared up enough of the document updates, just setting priorities will be sufficient.

iwehrman:
Well, yeah. I seem to recall that that was just a small performance optimization, which may no longer even be valid. In any case, I think the `beforeStartup` meaning "UI is ready to use" thing is important to preserve.

volfied:
@iwehrman I've rebased a new change. Removed the stale comment, and edited priorities.

src/js/actions/documents.js
```diff
@@ -1099,7 +1099,7 @@ define(function (require, exports) {
 
         return this.transfer(initActiveDocument);
     };
-    beforeStartup.reads = [locks.PS_DOC];
+    beforeStartup.reads = [locks.PS_DOC, locks.JS_APP];
```
iwehrman: I don't see that this actually reads `JS_APP`. It gets a reference to the store up front, but that's it. Is this an important part of the fix?

volfied: It's a safety measure to make sure it waits for tools.beforeStartup to finish.

iwehrman: If action A transfers to an action B which holds lock X, then action A executes with lock X as well.

iwehrman:
Related: is this just something that we should fix in PS?

src/js/actions/documents.js
iwehrman: So, maybe just a comment would be better here? If we just leave this superfluous lock, it will be confusing later.

volfied:
@iwehrman Please re-review.

iwehrman:
I still think we should file a Watson bug for this. Imagine a third-party developer trying to figure this out... In the meantime, this works.

PR #3183, opened 2015-11-03
3132 Fix sibling highlighting when in a grouped layer

volfied:
In case we decide to go to BRC with this, this fix will address #3132 on release.


volfied:
During the stand up, we decided to wait on this fix for M3.

Issue #3168, opened 2015-10-30
A library sticks every time coming back to DS


ktaki:
DS started with Library A selected. Select Library B. Switch to standard Ps, and switch back to DS. Library A is selected. I expect Library B is selected. This behavior seems started happening recently.

chadrolfs:
related to if not a dupe of #2360

ktaki:
Yes, it might be related, but here, the issue is more towards within DS. The stickiness to a particular lib happens when quitting Ps in DS mode and starting Ps in DS mode.

volfied:
FNBC to Taki

ktaki:
Verified the fix.

Issue #3166, opened 2015-10-29
actions/sampler.js:564 has two new lines in master


pineapplespatula:
lintspaces:src seems to have not caught it

shaoshing:
Close as those lines are removed in master

PR #3184, opened 2015-11-03
Fix for wrong last selected library.

shaoshing:
Addresses issue #3168, which your last selected library was wrong due to reading the library state at a wrong timing. 

~~Removed extra lines #3166~~

Added right padding to dropdown input to avoid long title overlapping with the arrow icon.

screen shot 2015-11-03 at 9 23 48 am


volfied:
Beautiful

PR #3185, opened 2015-11-03
Fix modal flag of transferred actions during type tool use

volfied:
These actions were not modal, so during type tool use, we were killing modal state. Fixes type tool use


mcilroyc:
I can't see why we would currently be relying on these actions to prevent modal ... so OK

PR #3186, opened 2015-11-03
Make color picker more responsive

shaoshing:
1. Use throttled and delayed actions related to color/alpha in the components to improve color picker's reponse time. I defined a separate set of throlleted actions with `50` delay, instead of creating separate functions in each component, as we are likely to change / revmoe this fix in the future, but please let me know if you disagree. I am open to better solutions.
2. Show the mouse pointer when dragging within the color palette. 
3. Fix the position of the circular icon: its center should be the location of the mouse.

### DIFF
--- a/src/js/jsx/shared/ColorPicker.jsx
+++ b/src/js/jsx/shared/ColorPicker.jsx
@@ -420,13 +420,12 @@ define(function (require, exports, module) {
 
         render: function () {
             var classes = classnames({
-                "color-picker-map": true,
-                "color-picker-map__active": this.state.active
-            });
-
+                "color-picker-map": true
+            }, this.props.className);
+            
             return (
                 <div
-                    className={this.props.className + " " + classes}
+                    className={classes}
                     onMouseUp={this._handleMouseUp}
                     onMouseDown={this._handleMouseDown}
                     onTouchStart={this._startUpdates}>
--- a/src/style/sections/libraries/libraries-section.less
+++ b/src/style/sections/libraries/libraries-section.less
@@ -291,7 +291,7 @@
                 &:focus, &:focus:hover, &:focus:disabled {
                     align-self: stretch;
                     color: @item;
-                    user-select: none;
+                    -webkit-user-select: none;
                     border-bottom: @hairline solid @focus-highlight;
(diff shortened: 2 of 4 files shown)


src/js/fluxcontroller.js
```diff
@@ -708,13 +709,16 @@ define(function (require, exports, module) {
             }
 
             var throttledName = name + THROTTLED_ACTION_SUFFIX,
+                throttledWithDelayName = name + THROTTLED_WITH_DELAY_ACTION_SUFFIX,
```
iwehrman: I don't think we should make a new global suite of throttled actions. Instead, in each module, we should just instantiate a new throttled function which calls the action.

iwehrman:
I've gone down this route before and honestly I'm not sure if it's better than what we have currently. When you add a delay like this, I think it makes the pointer feel jumpy instead of sluggish... Which is different, but... I dunno. Does anybody else have an opinion about this? @placegraphichere or @designedbyscience?

iwehrman:
Over to @designedbyscience for a second opinion. To be clear, I completely agree that the color picker performance is unacceptable. But, I'm more interested in pushing harder on the core team to help us find a better solution at this point in the development cycle. CC @jsbache.

designedbyscience:
It certainly isn't yet to the point of feeling good.

The CSS changes seem useful

Given where we are at the moment, I think choosing throttling versus the old option comes down to: _Which would be preferred if the core side of this was really fast?_ I don't think we'd want throttling until we could measure exactly how fast/slow it is.

placegraphichere:
@shaoshing Any chance we can just get the CSS changes for now? I also want to take a look at better handling the visual placement of the sliders and their bounds, and these changes seem useful to build on.

iwehrman:
It was throttled before and after, but after there is also an enforced 50ms delay between calls. Previously, it throttled by calling as fast as possible without overlapping the calls.

shaoshing:
Looks like we are not big fan of setting delay for the actions :D I've dropped that commit so that we can just pull in the css changes only. Back to you @designedbyscience

designedbyscience:
Looks good

Issue #3164, opened 2015-10-28
Resizing multiple artboards via Transform fields can have unexpected results


chadrolfs:
open explore-mt-hood.psd
select ~~all 3~~ an artboards
click in H field and Shift+up arrow a few times
click in W field and Shift+up arrow a few times

Result: artboards jump extremely left

Note: reference point is bottom left

chadrolfs:
![ab-jump-jump](https://cloud.githubusercontent.com/assets/8442631/10806637/4521d782-7d94-11e5-963c-a849fc9facff.gif)

chadrolfs:
can repro with just one AB selected (don't need to select all 3)

iwehrman:
I think standard PS explicitly removed support for on-canvas resizing of multiple artboards at once. We should probably just enforce that same constraint in our transform panel as well.

volfied:
I can't repro this today.

chadrolfs:
make sure the ref point is bottom right.

iwehrman:
Disable the resize of multiple artboards to downgrade this to workflow.

ktaki:
I think the original jumping issue was fixed, but the canvas is not expanded until artboard is moved.

![screen shot 2016-02-18 at 3 12 00 pm](https://cloud.githubusercontent.com/assets/2029597/13161368/141fee5a-d652-11e5-8c5a-eca95cf38bc8.png)

ktaki:
I opened a new issue #3726 for the above.

ktaki:
@chadrolfs: I think the original problem was fixed. Do you see any issues with resizing multiple ABs?

chadrolfs:
Yep, verified this is fixed.  Thanks for logging the other issue.

PR #3188, opened 2015-11-03
Do the autoExpand dance for setSize

volfied:
For similar reasons to Swap, if autoExpand is enabled, and we try to resize 3 artboards at once, the canvas shift would break things. SetSize has been updated to temporarily turn off auto expand during it's operation.

Addresses #3164 

As far as I can tell, for three artboards, the setSize time went from 90 ms to 150 ms, so it's not a bad loss.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -583,14 +583,42 @@ define(function (require, exports) {
             dispatchPromise = this.dispatchAsync(events.document.history.optimistic.RESIZE_DOCUMENT, payload);
             sizePromise = descriptor.playObject(resizeObj);
         } else {
-            var resizeLayerActions = layerSpec.reduce(function (actions, layer) {
-                var layerActions = _getResizeLayerActions.call(this, document, layer, size, payload.sizes, refPoint);
-                return actions.concat(layerActions);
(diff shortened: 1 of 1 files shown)


src/js/actions/transform.js
```diff
@@ -583,14 +583,42 @@ define(function (require, exports) {
...
+                        });
+
+                        return descriptor.playObject(setObj);
+                    }
+                })
+                .then(function () {
+                    return layerActionsUtil.playLayerActions(document, resizeLayerActions, true, options);
+                })
+                .then(function () {
+                    if (autoExpandEnabled) {
```
iwehrman: I kind of wonder if `playLayerActions` could take some extra, optional parameters `before` and `after` which are prepended resp. appended to the array of play objects that are eventually batchPlay'd? That way we in this case we could still execute a single play call instead of 1 + 2 = 3.

iwehrman:
If you don't like that idea you can tell me to take a hike. But, in the meantime, assigning this back to @volfied for a judgment.

volfied:
Eventually, it may be a good idea. But I'd like to go explicit here.

iwehrman:
I filed #3203 so this feature request isn't lost.

PR #3190, opened 2015-11-04
Initialize documents by ID

iwehrman:
Initialize documents by ID instead of by `numberOfDocuments` and then by index. We wanted to do this a few weeks ago to work around a PS bug in which `itemIndex` would have crazy values, but we couldn't because the appropriate multiGetRef descriptor didn't quite work right. As of Mac pgdev.864 it does. Even with `itemIndex` bandaged over inside PS, I suggest that this is a less surprising way to initialize documents.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -30,7 +30,8 @@ define(function (require, exports) {
     var photoshopEvent = require("adapter").lib.photoshopEvent,
         descriptor = require("adapter").ps.descriptor,
         documentLib = require("adapter").lib.document,
-        selectionLib = require("adapter").lib.selection;
+        selectionLib = require("adapter").lib.selection,
+        appLib = require("adapter").lib.application;
 
     var guideActions = require("./guides"),
         exportActions = require("./export"),
@@ -273,20 +274,20 @@ define(function (require, exports) {
      * Initialize document and layer state, emitting DOCUMENT_UPDATED events, for
      * all the inactive documents.
      *
-     * @param {number} currentIndex
-     * @param {number} docCount
+     * @param {number} activeDocumentID
+     * @param {Array.<number>} openDocumentIDs
      * @return {Promise}
      */
-    var initInactiveDocuments = function (currentIndex, docCount) {
-        var otherDocPromises = _.range(1, docCount + 1)
-            .filter(function (index) {
-                return index !== currentIndex;
+    var initInactiveDocuments = function (activeDocumentID, openDocumentIDs) {
+        var otherDocPromises = openDocumentIDs
+            .filter(function (documentID) {
+                return documentID !== activeDocumentID;
             })
-            .map(function (index) {
-                var indexRef = documentLib.referenceBy.index(index);
+            .map(function (documentID) {
+                var docRef = documentLib.referenceBy.id(documentID);
 
                 // Only load essential properties for inactive documents
-                return _getInactiveDocumentByRef(indexRef)
+                return _getInactiveDocumentByRef(docRef)
                     .bind(this)
                     .then(function (document) {
                         this.dispatch(events.document.DOCUMENT_UPDATED, {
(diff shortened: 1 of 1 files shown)


src/js/actions/documents.js
```diff
@@ -307,13 +307,23 @@ define(function (require, exports) {
     /**
      * Initialize document and layer state, emitting DOCUMENT_UPDATED.
      *
-     * @return {Promise.<{currentIndex: number, docCount: number}>}
+     * @return {Promise.<{activeDocumentID: number, openDocumentIDs: Array.<number>}=>}
      */
     var initActiveDocument = function () {
-        return descriptor.getProperty("application", "numberOfDocuments")
+        var appRef = {
```
pineapplespatula: this looks like code that should be in spaces/adapter

iwehrman:
Actually, let's hold off on this a bit. I realized on my bike ride home that we can do a little better than this. In particular, there is no need to query all the open documents in `initActiveDocument`; instead, we can wait for that until `initInactiveDocuments`.

src/js/actions/documents.js
iwehrman: Yes, very true. Even better perhaps would be to fix `descriptor.multiGetProperty` so that it correctly wraps an `"application"` reference like `descriptor.get` does.

iwehrman:
Actually I decided to not do anything "better". We could omit the initial call to `numberOfDocuments`, but the tradeoff is that when there are _no_ open documents we make a number of failed calls, which delays the no-doc case about as much as it improves the some-doc case. Instead, I just fixed @pineapplespatula's nit by using `appLib.referenceBy.current` instead of spelling out action reference. To support this, **I pushed commit adobe-photoshop/spaces-adapter@abe0d1c697861835ee629d4967ee0f8b98382664 directly to the adapter, so be sure to pull that before testing.**

iwehrman:
Rebased and ready 2 rock.

src/js/actions/documents.js
```diff
@@ -273,20 +274,20 @@ define(function (require, exports) {
      * Initialize document and layer state, emitting DOCUMENT_UPDATED events, for
      * all the inactive documents.
      *
-     * @param {number} currentIndex
-     * @param {number} docCount
+     * @param {number} activeDocumentID
+     * @param {Array.<number>} openDocumentIDs
      * @return {Promise}
      */
-    var initInactiveDocuments = function (currentIndex, docCount) {
```
pineapplespatula: Should initInactiveDocuments be private? is it used anywhere else?

src/js/actions/documents.js
```diff
@@ -273,20 +274,20 @@ define(function (require, exports) {
...
-        var otherDocPromises = _.range(1, docCount + 1)
-            .filter(function (index) {
-                return index !== currentIndex;
+    var initInactiveDocuments = function (activeDocumentID, openDocumentIDs) {
```
pineapplespatula: Should `initInactiveDocuments` be private? is it used anywhere else?

src/js/actions/documents.js
```diff
@@ -307,16 +308,22 @@ define(function (require, exports) {
     /**
      * Initialize document and layer state, emitting DOCUMENT_UPDATED.
      *
-     * @return {Promise.<{currentIndex: number, docCount: number}>}
+     * @return {Promise.<{activeDocumentID: number, openDocumentIDs: Array.<number>}=>}
      */
     var initActiveDocument = function () {
```
pineapplespatula: should `initActiveDocument` also be private?

pineapplespatula:
a couple of nits. back to @iwehrman

iwehrman:
Those methods are exported, so they aren't technically private. I could imagine separately adding a feature to the controller which allows actions to be marked with `.private = true`, which would prevent them from being called directly or transferred to from different modules. But, for now, not private.

pineapplespatula:
right, I was suggesting that they be taken out of the export list and appended with an underscore, since i dont know if we want other external functions to call them

pineapplespatula:
:fist:  @iwehrman

iwehrman:
We aren't doing that anywhere else, so I don't want to do that in one place right here. I'd rather wait for a more complete cleanup.

pineapplespatula:
I'm not sure what you mean about not doing it anywhere else. We do it for functions like _getInactiveDocumentByRef and _getLayersForDocument in this file

iwehrman:
What I'm saying is: it's exported, so it's not private. We use that rule everywhere, including for actions.

pineapplespatula:
right.. why are we exporting it?

iwehrman:
Because that's how actions work. The controller needs to understand the locks, transfers, etc. of actions in order for it to provide all of the features it provides.

pineapplespatula:
after chatting with Ian about the current limitations of our action model i'm going to merge this and set up a new branch to address the lack of privacy and modularity in our actions.

Issue #3195, opened 2015-11-04
Can't add layers that are within the bounds of current selection to the selection (shift+click)


placegraphichere:
- create two rects, one larger than the other
- put the smaller rect fully within the larger one
- the smaller layer should be on top
- select the larger rect, then attempt to select the smaller rect
- result: unable to select the smaller layer, even though it is visible and z-ordered above 

- Expected: should be able to multi-select objects that are visibly z-ordered above others
![screen shot 2015-11-04 at 12 31 12 am](https://cloud.githubusercontent.com/assets/3229167/10931241/6592469a-828b-11e5-94ad-48216b5a01d0.png)

placegraphichere:
![screen shot 2015-11-04 at 12 36 35 am](https://cloud.githubusercontent.com/assets/3229167/10931332/20b45f08-828c-11e5-8c61-cd75d1f57b79.png)

Here's a really practical repercussion of this issue. In this example I've selected the red boxes first, and then want to attempt to select the text layer. This is not possible. It forces the selection order to be in  perfect order where each selection is below the last.

volfied:
Note: this works for single click, doesn't work for shift+click.

iwehrman:
FBNC to @chadrolfs.

PR #3198, opened 2015-11-04
Add an inside policy for shift modifier to border policies

volfied:
The outside propagate_by_alpha policy was kicking in because we didn't handle shift key on the inside area as we should. So holding down shift and clicking within the selection would send the click to PS.

Addresses #3195


iwehrman:
Yep.

Issue #3191, opened 2015-11-04
Opening/Closing Folders shows changed file status without ability to save


placegraphichere:
- Create new file with some folder/artboards
- Save the file
- Open/close the AB or folder a few times using the layers panel
- Get into a state where the file has the asterisk, but the file menu does not allow a save. 
- Exiting Ps will prompt a file-has-changed save tho

- expected: saving file menu and CMD+S should be in sync with the folder changes

![screen shot 2015-11-03 at 8 12 46 pm](https://cloud.githubusercontent.com/assets/3229167/10927716/b14b6f9a-8267-11e5-916c-2e03f3f4bb18.png)

iwehrman:
FBNC to @chadrolfs.

PR #3199, opened 2015-11-04
Update menu items on group expansion/collapse

volfied:
We didn't, so doc would be set dirty, but save would be disabled. Which is #3191.


iwehrman:
Yep.

PR #3200, opened 2015-11-04
Begin history refactor: normalize resetBounds and resetLayers

mcilroyc:
1. `resetBounds` was, in several cases, used as the "history changing" action.  This was confusing so is being phased out.  This PR changes those actions to explicitly create a history state early, and then let resetBounds amend the new history state.  Additionally, during document select and dispose, make sure history is queried before resetting bounds
1. `resetLayers` should always amend history.  This required changing several actions to explicitly create history states early (instead of waiting for a historyState event from photoshop*) so that the correct state is amended.
1. Add a new generic newHistoryState action to help some of the weird cases from above
1. Introduces a new "unifiedHistory" sub-class of events to help the slow march towards a more options-in-the-payload approach.  *Eventually* this will get factored out completely, but until it is standardized, this new sub-class of events will be a holding pen.
1. Removed some dead code

Testing notes: This touches (though transparently, hopefully) the following user actions:
- flip*
- align*
- distribute
- rotate
- combine*
- strokes
- text updates (modal AND non modal)
- select document that hasn't been active before
- select document that HAS been active before
- dispose document
- apply styles and graphics with sampler
- apply styles from library


src/js/actions/layers.js
```diff
@@ -662,7 +656,7 @@ define(function (require, exports) {
         if (linkedLayers.isEmpty()) {
             return Promise.resolve();
         }
-        return this.transfer(resetBounds, document, linkedLayers, true);
+        return this.transfer(resetBounds, document, linkedLayers, true, true);
```
pineapplespatula: for readability i would prefer there be named variables suppressHistory amendHistory which are set to true instead of just passing booleans ( this is true for a number of calls)

mcilroyc: for point, but to clarify: these will soon be completely refactored out.  ResetBounds will no longer have these arguments at all.

mcilroyc: I'll probably do that as another commit on this PR, FYI

src/js/actions/shapes.js
```diff
@@ -338,6 +325,22 @@ define(function (require, exports) {
...
     /**
+     * Sets the enabled flag for all selected Layers on a given doc.
+     *
+     * @param {Document} document
+     * @param {Immutable.List.<Layer>} layers list of layers being updating
+     * @param {object} options
+     * @param {boolean=} options.enabled
+     * @return {Promise}
+     */
+    var setStrokeEnabled = function (document, layers, options) {
```
mcilroyc: I moved this down in the file so that it could transfer to `setStrokeColor`

src/js/actions/documents.js
```diff
@@ -472,10 +472,14 @@ define(function (require, exports) {
 
                 this.dispatch(events.document.CLOSE_DOCUMENT, payload);
 
-                var newDocument = this.flux.store("application").getCurrentDocument(),
-                    resetLinkedPromise = this.transfer(layerActions.resetLinkedLayers, newDocument),
-                    resetHistoryPromise = newDocument ?
-                        this.transfer(historyActions.queryCurrentHistory, newDocument.id) : Promise.resolve(),
+                return this.flux.store("application").getCurrentDocument();
+            })
+            .tap(function (newDocument) {
```
mcilroyc: Moved this queryCurrentHistory up in the chain so it will be finished before `resetLinkedLayers` which then does an resetBounds, which amends history.  Same thing in the next diff in this file.

src/js/actions/type.js
```diff
@@ -599,31 +623,6 @@ define(function (require, exports) {
...
 
     /**
-     * Duplicates the layer effects of the source layer on all the target layers
-     *
-     * @param {Document} document
-     * @param {Immutable.Iterable.<Layer>} targetLayers
-     * @param {Layer} source
-     * @return {Promise}
-     */
-    var duplicateTextStyle = function (document, targetLayers, source) {
```
mcilroyc: dead code

src/js/events.js
```diff
@@ -39,6 +39,19 @@ define(function (require, exports, module) {
         },
         document: {
             history: {
+                unifiedHistory: {
```
mcilroyc: as mentioned in the PR description, eventually when ALL the history events get migrated to the new system, then "unifiedHistory" will just get bumped up a level and we'll remove the other three.

mcilroyc:
Alright, I've made a few more fixes, rebased to one commit, and I'm now removing the 'review only' label

src/js/stores/history.js
```diff
@@ -638,6 +647,21 @@ define(function (require, exports, module) {
...
 
         /**
+         * New and improved "unified" history event handler.  payload.history determines
+         * whether or not history state is created or amended.
+         *
+         * @param {object} payload
+         */
+        _handleUnifiedHistory: function (payload) {
+            if (payload.history && payload.history.newState) {
+                log.debug("UnifiedHistory: Creating new history state");
```
iwehrman: I don't want to commit more debug-only log statements like this unless you really, really think this will be useful later. If so, can I'd like to group them all somehow under a "history" tag kind of like CC Libraries does.

src/js/tools/type.js
```diff
@@ -124,7 +126,15 @@ define(function (require, exports, module) {
                 var documentStore = this.flux.store("application"),
                     currentDocument = documentStore.getCurrentDocument();
 
-                this.flux.actions.layers.resetBounds(currentDocument, currentDocument.layers.allSelected);
+                // Debounce to allow history state event to precede the call to resetBounds
+                var debouncedMoveHandler = synchronization.debounce(function () {
```
iwehrman: This is a little scary, but... OK? I guess it's just swapping one frightening thing for a different one.

iwehrman:
Looks great. I guess I just want to force you to type one more comment in here before I merge that `log.debug` statement.

src/js/tools/type.js
mcilroyc: Yes, this is scary, and I am planning to investigate a better, more general solution to this problem of ordering history state events in the "PS initiated" flow.  But, this is a technique used elsewhere so is at least somewhat consistent.  for now.

src/js/stores/history.js
mcilroyc: I removed this debug log AND added a tag-like prefix.  Maybe someday we could have a way of enabling tag-level logging...

mcilroyc:
Comments addressed.  Also rebased.  There was a conflict after your tools refactoring ... but it was just some spelling corrections in superselect!

iwehrman:
:ship: :it:

Issue #3202, opened 2015-11-04
Turning off Stroke visibility results in Unhandled rejection Error


chadrolfs:
On compiled .js only - 

new doc, drag out a rect, turn off stroke visibility
result:
`Unhandled rejection Error: Attempt to extract key: event as: string actual type is: boolean`
  `at eval (eval at undefined, <anonymous>:1:5)`
   `at eval (native)`

CNR on release branch and master
Shipped in 16.1

iwehrman:
This is a headlights bug. If you enable debug-time headlights logging, you get this:
```
Error: Attempt to extract key: event as: string actual type is: boolean
    at eval (eval at undefined, <anonymous>:1:5)
    at eval (native)
From previous event:
    at Object.logHeadlightsEvent (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps.js:87:20)
    at Object.logEvent (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/headlights.js:49:30)
    at React.createClass._toggleStrokeEnabled (file://./js/jsx/sections/style/Stroke.jsx:115:24)
    at React.createClass.handleClick (file://./js/jsx/shared/ToggleButton.jsx:95:28)
    at Object.executeDispatch (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:3311:21)
    at SimpleEventPlugin.executeDispatch (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:17428:40)
    at forEachEventDispatch (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:3299:5)
    at Object.executeDispatchesInOrder (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:3320:3)
    at executeDispatchesAndRelease (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/react/react-with-addons.js:2693:22)
    at Array.forEach (native)
[...]

PR #3204, opened 2015-11-05
Converts isChecked Boolean to String

DivyaPrabhakar:
Fix for #3202


iwehrman:
Even though it says the parameter is a boolean, it could instead be something like null or undefined. In which case, this will throw an exception. As before, something like `String(isChecked)` might be safer. But, I wonder instead if we should just do this inside of `logEvent` so that we don't keep having to fix bugs like this?

iwehrman:
LGTM!

PR #3205, opened 2015-11-05
Tool/policy clean-up grab bag

iwehrman:
1. Allow all policy actions to delay committing by passing in a trailing `true`. This adds the behavior to more of the actions.
2. Make policy actions all commit by default. This reverses the the current behavior which does not commit by default.
3. Provide a `syncAllPolicies` action to sync policies only if they're dirty. In fact, all policy syncing is now a no-op when the policies aren't dirty. This removes a number of redundant adapter calls.
4. Factor `select` and `deselect` actions out of tools and into their own action modules.
5. Clean up `tools.select` to transfer to tool select/deselect handlers instead of `.call`ing them.
6. Make it possible to transfer to an action specified by it's path, i.e., `this.transfer("policy.setMode")`.

More comments inline below.


src/js/actions/policy.js
```diff
@@ -42,9 +42,16 @@ define(function (require, exports) {
      * @return {Promise} Resolves when polices are set in photoshop
      */
     var syncPolicies = function (kind) {
-        var policyStore = this.flux.store("policy"),
-            masterPolicyList = policyStore.getMasterPolicyList(kind),
-            dispatchPromise = this.dispatchAsync(events.policies.POLICIES_INSTALLED),
+        var policyStore = this.flux.store("policy");
+
+        if (!policyStore.isDirty(kind)) {
```
iwehrman: Policy syncing is a no-open if the policy kind is not dirty.

src/js/actions/tool/sampler.js
```diff
@@ -0,0 +1,89 @@
...
+    /** @ignore */
+    var _mouseMoveHandler;
+
+    /**
+     * Track mouse movements and set the pointer propagation mode.
+     *
+     * @return {Promise}
+     */
+    var select = function () {
+        _mouseMoveHandler = _.throttle(function (event) {
```
iwehrman: This previously was a local variable, but now it's tracked in the UI store because the reader is in the model and the writer is in the handler.

src/js/actions/tool/sampler.js
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Stop tracking mouse movements and hide the HUD.
+     *
+     * @return {Promise}
+     */
+    var deselect = function () {
+        OS.removeListener("externalMouseMove", _mouseMoveHandler);
+
+        var mousePositionPromise = this.dispatchAsync(events.ui.MOUSE_POSITION_CHANGED, null),
```
iwehrman: The mouse position in the store is cleared when we stop tracking it.

src/js/actions/tool/type.js
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Handles updateTextProperties events, which are emitted during the type modal
+     * tool state to indicate the text properties of the current text selection or
+     * cursor position.
+     *
+     * @param {object} event
+     * @return {Promise}
+     */
+    var updateTextPropertiesHandler = function (event) {
```
iwehrman: This was a helper function used by the type tool and the select type subtool, but now it's an action.

src/js/actions/tools.js
```diff
@@ -243,7 +243,7 @@ define(function (require, exports) {
 
         // if we are in vector mask mode we should not change pointer policies 
         if (toolStore.getVectorMode()) {
-            return Promise.resolve();
+            return this.transfer(policy.syncAllPolicies);
```
iwehrman: Policies are only synced at the end points of this function. Note above that the `removePointerPolicies` call does not sync the policies.

src/js/actions/tools.js
```diff
@@ -321,7 +330,7 @@ define(function (require, exports) {
      *         }>}
      *         Resolves to the new tool and it's policy IDs
      */
-    var _swapPolicies = function (nextTool) {
+    var swapPolicies = function (nextTool) {
```
iwehrman: Now an action.

src/js/actions/tools.js
```diff
@@ -423,13 +441,13 @@ define(function (require, exports) {
                 }
 
                 // Calls the select handler of new tool
-                return selectHandler.call(this, nextTool);
+                return this.transfer(selectHandler, nextTool);
```
iwehrman: Note that the `selectHandler` and `deselectHandler` variables are strings now.

src/js/actions/tools.js
```diff
@@ -442,10 +460,16 @@ define(function (require, exports) {
     };
     selectTool.reads = [];
     selectTool.writes = [locks.JS_TOOL, locks.PS_TOOL];
-    selectTool.transfers = [resetBorderPolicies, installShapeDefaults, shortcuts.addShortcut,
-        shortcuts.removeShortcut, "layers.deleteSelected", "layers.resetLayers",
-        policy.removePointerPolicies, policy.removeKeyboardPolicies, policy.addPointerPolicies,
-        policy.addKeyboardPolicies, policy.setMode];
+    selectTool.transfers = [resetBorderPolicies, swapPolicies, policy.removePointerPolicies,
+        "toolSuperselect.select", "toolSuperselect.deselect",
```
iwehrman: This is kind of gross, but I think it's way less gross than before.

src/js/fluxcontroller.js
```diff
@@ -409,16 +409,24 @@ define(function (require, exports, module) {
...
              *
-             * @param {Action} nextAction
+             * @param {string|Action} nextAction
              * @return {Promise} The result of executing the next action
              */
             transfer: {
                 value: function (nextAction) {
+                    var nextActionName;
+
+                    if (typeof nextAction === "string") {
```
iwehrman: The first parameter of transfer can now be a string.

src/js/models/tools/sampler.js
```diff
@@ -59,26 +44,8 @@ define(function (require, exports, module) {
...
-                .bind(this)
-                .then(function () {
-                    return this.transfer(policyActions.setMode, PolicyStore.eventKind.POINTER,
-                        UI.pointerPropagationMode.PROPAGATE_BY_ALPHA);
-                });
-        };
-
-        this.selectHandler = selectHandler;
-        this.deselectHandler = deselectHandler;
+        this.selectHandler = "toolSampler.select";
```
iwehrman: Again, note that this property is now a string that refers to an action.

src/js/stores/policy.js
```diff
@@ -96,8 +104,13 @@ define(function (require, exports, module) {
 
         /**
          * Emits a change event - used by debug policy overlay
+         *
+         * @private
+         * @param {{kind: string}} payload
          */
-        _emitChange: function () {
+        _handlePoliciesInstalled: function (payload) {
+            delete this._dirtyPolicies[payload.kind];
```
iwehrman: This event means we just sync'd policies, which means this policy kind is no longer dirty.

iwehrman:
Rebased and ready 2 rock. This PR isn't as bad as it looks, I swear :)

iwehrman:
Rebased again!

src/js/actions/tools.js
```diff
@@ -446,10 +464,16 @@ define(function (require, exports) {
     };
     selectTool.reads = [];
     selectTool.writes = [locks.JS_TOOL, locks.PS_TOOL];
-    selectTool.transfers = [installShapeDefaults, shortcuts.addShortcut,
-        shortcuts.removeShortcut, "layers.deleteSelected", "layers.resetLayers",
-        policy.removePointerPolicies, policy.removeKeyboardPolicies, policy.addPointerPolicies,
-        policy.addKeyboardPolicies, policy.setMode];
+    selectTool.transfers = [swapPolicies, policy.removePointerPolicies,
```
volfied: Eep.

mcilroyc: Each of these various select/deselect transfers has roughly the same scope, so other than being an "ugly long list" this doesn't really have much down side.  eh?

iwehrman: It would be great if there were a way to say "all the tools? the select/deselect actions on those". The only downside is that the list of tools' actions has to be specified up front. So, I think there's definitely room for improvement, but I think it is clearly preferable to what is exists on master today.

mcilroyc:
Seems like a worthwhile refactoring.  doin' it to it.

PR #3206, opened 2015-11-05
First draft of HUD for vector mask tool

pineapplespatula:
Adding a hud for the vector mask super select. needs some more work with seth before its ready to merge though


src/js/actions/mask.js
```diff
@@ -139,6 +144,80 @@ define(function (require, exports) {
...
+
+        if (currentLayers.isEmpty()) {
+            return Promise.resolve();
+        }
+
+        var currentLayer = currentLayers.first(),
+            bounds = currentLayer.bounds;
+            
+        return descriptor.batchPlayObjects([vectorMaskLib.deleteVectorMask(),
+            vectorMaskLib.makeCircularBoundsWorkPath(bounds),
```
pineapplespatula: This does not handle history properly currently. @mcilroyc what do i need to add to make it history respectful

shaoshing: Have you try adding the `historyStateInfo` option? I see multiple histories in classic after applying a rectangle or eclipse mask.

mcilroyc: what @shaoshing said.

pineapplespatula:
@placegraphichere is working on this for a bit, I'll open it back up when its ready to be PRed

PR #3213, opened 2015-11-06
Scrubbing of Positions works now

pineapplespatula:
I tied the scrubby to the zoom level, which i think i like. this probably needs some tweaking

### DIFF
--- a/src/js/jsx/mixin/Scrubby.js
+++ b/src/js/jsx/mixin/Scrubby.js
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2015 Adobe Systems Incorporated. All rights reserved.
+ *  
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"), 
+ * to deal in the Software without restriction, including without limitation 
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
+ * and/or sell copies of the Software, and to permit persons to whom the 
+ * Software is furnished to do so, subject to the following conditions:
+ *  
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *  
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER 
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING 
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
+ * DEALINGS IN THE SOFTWARE.
+ * 
+ */
+
+define(function (require, exports, module) {
+    "use strict";
+
+    var React = require("react"),
+        _ = require("lodash"),
+        log = require("js/util/log");
+
+    /**
+     * Mixin for drag-and-drop functionality. Clients should call _startUpdates
+     * on mousedown/touchstart and implement the abstract method _updatePosition.
+     * CLinets can optinally implement _updatePositionBegin and _updatePositionEnd
+     */
+    module.exports = {
+
+        propTypes: {
+            onChange: React.PropTypes.func,
+            max: React.PropTypes.number
+        },
+
+        getDefaultProps: function () {
+            return {
+                onChange: _.identity,
+                max: 1
+            };
+        },
+
+        getInitialState: function () {
+            return {
+                active: false
+            };
+        },
+
+        componentDidMount: function () {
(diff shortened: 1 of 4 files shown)


src/js/jsx/mixin/Scrubby.js
```diff
@@ -0,0 +1,164 @@
...
+                e = e.touches[0];
+            }
+
+            return {
+                x: e.clientX,
+                y: e.clientY
+            };
+        },
+
+        /**
```
volfied: These functions are unused and should be taken out.

src/js/jsx/shared/Label.jsx
```diff
@@ -25,10 +25,48 @@ define(function (require, exports, module) {
...
     
     var Label = React.createClass({
-        mixins: [React.addons.PureRenderMixin],
+        mixins: [FluxMixin, ScrubbyMixin],
+        
+        propTypes: {
+            size: React.PropTypes.string,
+            onScrub: React.PropTypes.func,
+            onScrubStart: React.PropTypes.func,
+            scrubby: React.PropTypes.bool
```
volfied: We can probably get away with removing scrubby and just checking for existence of onScrub function (onScrubStart/onScrubEnd can be optional)

src/js/jsx/mixin/Scrubby.js
```diff
@@ -0,0 +1,164 @@
...
+
+define(function (require, exports, module) {
+    "use strict";
+
+    var React = require("react"),
+        _ = require("lodash");
+
+    var math = require("js/util/math");
+
+    module.exports = {
```
volfied: There was a REALLY useful comment in ColorPicker that explained how to use this mixin. It stated that the user had to provide their own `_updatePosition` function. 

Can you make a similar comment here on how to use this mixin, what the user of this mixin has to provide with function signatures?

volfied:
This is a good start, but needs some cleanup. Since we're providing onScrubStart and onScrub.. we should also provide onScrubEnd, which returns the overall scrub difference so in cases where we don't want to live scrub, we can use onScrubEnd to submit an action.

Great work!~

src/js/jsx/mixin/Scrubby.js
```diff
@@ -0,0 +1,164 @@
...
+
+        propTypes: {
+            onChange: React.PropTypes.func,
+            max: React.PropTypes.number
```
volfied: Not sure if this is possible, but can you look into making the _updatePosition that the owner has to implement a required prop so React complains at us if we decide to use this mixin but not provide that function?

pineapplespatula: it doesn't look like that is possible

src/js/jsx/sections/transform/Position.jsx
```diff
@@ -45,6 +46,9 @@ define(function (require, exports, module) {
             referencePoint: React.PropTypes.string.isRequired
         },
 
+        xScrubbyValue: null,
```
iwehrman: It seems like these should be JSDoc's private and also underscore prefixed?

src/js/jsx/mixin/Scrubby.js
volfied: If that's not possible, during componentDidMount, please check for existence of `_updatePosition` and show a warning/error and skip calling it below so it doesn't throw.

src/js/jsx/shared/Label.jsx
```diff
@@ -25,10 +25,47 @@ define(function (require, exports, module) {
...
+                rect = React.findDOMNode(this).getBoundingClientRect();
+
+            var value;
+            if (this.props.vertical) {
+                value = (rect.bottom - clientY) / rect.height;
+            } else {
+                value = (clientX - rect.left) / rect.width;
+            }
+
+            this.props.onScrub(uiStore.zoomCanvasToWindow(value));
```
volfied: Alongside checking for _udpatePosition in Scrubby mixin, you should also check for onScrub's existence here.

volfied: In fact, this entire function should bail at the beginning if `this.props.onScrub` is not defined. `getBoundingClientRect` is an expensive operation.

src/js/jsx/shared/Label.jsx
```diff
@@ -25,10 +25,47 @@ define(function (require, exports, module) {
...
+            onScrubStart: React.PropTypes.func
+        },
+
+        /**
+         * inform the parent that scrubbing has started
+         *
+         * @private
+         */
+        _updatePositionBegin: function () {
+            this.props.onScrubStart();
```
volfied: Ditto for existence checking.

volfied:
Got couple more comments. 

Also, can you hook up onScrubEnd as well, I see it being useful in some situations.

While testing, I saw that we weren't actually transforming the layer until I let go of the mouse, can you look into it as well?

Back to you @pineapplespatula

pineapplespatula:
I'm at a loss as to why we arent transforming the layer during the scrub action. I'll look into this more tomorrow, but my hunch is a core issue

pineapplespatula:
@jsbache diagnosed the drawing issue. This should be ready for another review

iwehrman:
CI checks are failing. This could be a false positive though. Does it check out for you?

iwehrman:
+1. The point of commit messages is to be useful to future humans, which is why we encourage rebasing commits: the end result of most PRs should a single commit with a message that explains the overall change.

pineapplespatula:
@volfied  what hasn't been addressed?

volfied:
Huh, Github must have served me a cached page, I still saw the old code.

PR #3214, opened 2015-11-06
Color themes!

iwehrman:
With this PR, we load one of four color themes to match the current Photoshop UI color stop. Your choices, from lightest to darkest are "original", "light", "medium" or "dark". It works in both debug and compiled modes. For the former, we figure out the color stop and then set a LESS global variable before calling the render function. For the latter, we compile the LESS files with four different values for the same `stop` global variable, and then load only the appropriate CSS file at load time.

Note that there is one pretty major SuperHack included, in which the paths used by the `@import` statements in the LESS files have to be rewritten to point to the "style/" subdirectory. It sure seems like this is something that should be possible to fix with some options to the LESS render method, for example, [`rootpath` or `relativeUrls`](http://lesscss.org/usage/#using-less-in-the-browser-options), but I'll be darned if I can get them to work. @designedbyscience: do you have any ideas? If there is _no_ solution to this, then as a last resort we can just do this search-and-replace once in our LESS files, but that definitely doesn't _feel_ like the right solution.

Later on we'll make it possible to switch themes from within Design Space, but this is a start. There is also some opportunity for additional cleanup in the different color theme files, like factoring out the duplicate helper functions, which I'm sure @placegraphichere will be excited about doing a little later. 

Requires at least Mac pgdev.867.

dark
medium
light
original


iwehrman:
Rebased and ready 2 rock.

iwehrman:
Presumably the super hack above could be resolved by less/less.js#1403. Alternatively, once we move to always using CSS that is pre-compiled by a watcher, we can use the `--include-path` option to get rid of this replacement hack.

volfied:
Nice, I will keep an eye out on the stops when I'm working on less/css part of webpack.

PR #3215, opened 2015-11-06
Move adapter requires to object requires

volfied:
Instead of directly referring to adapter sources, we expect spaces-adapter to export a single object that we require parts of.

Requires https://github.com/adobe-photoshop/spaces-adapter/pull/163

This is an intermediate step that'll make it a lot easier to work on webpack efforts in parallel.


iwehrman:
Everybody rebase!

PR #3219, opened 2015-11-10
Reduce plugin-compatibility check to a warning

iwehrman:
Details in adobe-photoshop/spaces-adapter#164. These two PRs should be merged at the same time.


mcilroyc:
too nice

PR #3222, opened 2015-11-10
Factor out colors-shared from color theme files

iwehrman:
This is a very simple factoring of the shared color styles functions into their own file, `colors-shared.less`, instead of being duplicated in each of the color theme files. It's just a copy-paste job.

Addresses: concerns.


shaoshing:
Wonderful!

PR #3225, opened 2015-11-10
Startup Columns Match Layout Preferences

placegraphichere:
Use the app preferences for columns to render the correct column setup
that will be displayed. Avoids the lurching UI drawing once the full app is
loaded.

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -421,14 +421,53 @@ define(function (require, exports, module) {
                         <IconBar />
                     </div>
                 );
-            } else {
+            } else if (this.state.propertiesVisible && this.state.layersLibrariesVisible) {
                 return (
+                    //  Show 2 columns and toolbar
+                    <div className="panel-set__container">
+                        <div ref="panelSet"
+                             className="panel-set">
+                            <PanelColumn visible={true} />
+                            <PanelColumn visible={true} />
+                        </div>
+                        <IconBar />
+                    </div>
+                );
+            } else if (this.state.propertiesVisible ||
+                       this.state.layersLibrariesVisible &&
+                       !this.props.singleColumnModeEnabled) {
+                return (
+                    // Just show one column and toolbar
                     <div className="panel-set__container">
                         <div ref="panelSet"
                              className="panel-set">
                             <PanelColumn visible={true} />
                         </div>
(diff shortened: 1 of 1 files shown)


iwehrman:
This is great! But, there seems to be a bug with my current pref state when launching into single-column mode, in which two columns + the icon bar are shown before the single column without an icon bar. Here's a gif and the prefs deets:

img

image

src/js/jsx/PanelSet.jsx
```diff
@@ -421,16 +421,55 @@ define(function (require, exports, module) {
...
+                    </div>
+                );
+            } else if (this.props.singleColumnModeEnabled) {
+                // Just show single column
+                return (
+                    <div className="panel-set__container panel-set__container__small-screen">
+                        <div ref="panelSet"
+                             className="panel-set">
+                            <PanelColumn visible={true}>
+                            </PanelColumn>
```
shaoshing: maybe consider closing the component in the open tag, like the others?

placegraphichere:
Placing this as "review only", as the issue below needs some discussion.

At the point when we draw the “fake” UI on initial load we do not know if there are any documents. This is an issue for the no-doc state, since we now will show the 2 columns as the fake UI. I’ve been searching for a simple thing to tell me, but so far, something first needs to initialize before it will log that any documents exist. I need some assistance or at least a cursory review on this.

---

Here's some logs of documents we know about when initially drawing the fake UI, note, there are no `activeDocuments` and the `documentIDs` are empty:
screen shot 2015-11-11 at 4 15 48 pm

---

And then after everything is loaded, `activeDocuments` is not null and `documentIDs` now have content.
screen shot 2015-11-11 at 4 16 01 pm


Perhaps before we add more stuff onto something that isn't yet capable of gathering all the info it needs, I could grab a hand on sorting this out, or just shelve this for now. 

Other than that, fixed @iwehrman's and @shaoshing's comments.

iwehrman:
It takes a little while to figure out whether there are any open documents, so your choices are a) render something before then, even though it might be incorrect; or b) wait to render anything (i.e., either any sort of UI, or just the panel) until we know. Depending on what you think is preferable, we can take different paths to accomplish it.

placegraphichere:
Just gonna say nope on this, it's either broken one way or the other.

PR #3226, opened 2015-11-11
Added global shortcuts to the search options

DivyaPrabhakar:
With this PR, global shortcuts are added to the options that are pulled up when the search bar is prompted. 

1.) commands.js now provides the search provider for both Global Shortcuts and Menu Commands. One thing I need input on is whether we want to combine global shortcuts and menu commands under one header "Commands" in the UI, or keep them under separate labels. For now, they are separate.  

2.) Whenever we call .addShortcut, an extra parameter is being added for a name. This allows us to provide a name for the shortcut when it shows up in the UI. Currently, this is way we will distinguish between which global shortcuts will be listed as options in the search bar. For example, pressing "esc" is the global shortcut to dismiss a dialog, but we do not want this to appear in the search options under Global Shortcuts.

### DIFF
--- a/src/js/actions/search/commands.js
+++ b/src/js/actions/search/commands.js
@@ -33,7 +33,8 @@ define(function (require, exports) {
 
     var events = require("js/events");
 
-    var menuActions = require("js/actions/menu");
+    var menuActions = require("js/actions/menu"),
+        shortcutActions = require("js/actions/shortcuts");
 
     /**
      * Get a localized label for the full path of the given menu entry ID
@@ -65,13 +66,13 @@ define(function (require, exports) {
     };
     
     /**
-     * Get a shortcut as a string for the given menu entry shortcut object
+     * Get a shortcut as a string for themen menu entry shortcut object
      *
      * @private
      * @param {object} fullShortcut
      * @return {string}
      */
-    var _getShortcut = function (fullShortcut) {
+    var _getMenuCommandString = function (fullShortcut) {
         var modifierBits = fullShortcut.modifiers,
             keyChar = fullShortcut.keyChar,
             keyCode = fullShortcut.keyCode,
@@ -135,7 +136,7 @@ define(function (require, exports) {
                         shortcut = "";
                     
                     if (currItem.shortcut) {
-                        shortcut = _getShortcut(currItem.shortcut);
+                        shortcut = _getMenuCommandString(currItem.shortcut);
                     }
 
                     if (ancestry) {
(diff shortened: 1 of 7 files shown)


src/js/actions/search/commands.js
```diff
@@ -161,7 +162,85 @@ define(function (require, exports) {
...
+
+        if (modifierBits.shift) {
+            shortcut += "\u21E7";
+        }
+
+        if (key) {
+            shortcut += key.toString().toUpperCase();
+        }
+
+        return " " + shortcut + "\u00a0\u00a0\u00a0\u00a0";
```
shaoshing: Just curious about the purpose of the four space characters here :D

DivyaPrabhakar: @shaoshing Alison's original code had 4 spaces, so I am assuming she talked about it with @placegraphichere and they agreed upon this for the way the shortcut renders in the Search Bar UI.

shaoshing: thanks for explaining.

src/js/actions/search/commands.js
```diff
@@ -171,12 +250,23 @@ define(function (require, exports) {
...
+
+    /**
+     * Find SVG class for global shortcuts
+     * If this needs to vary based on the item, use category list as parameter 
+     * (see getSVGCallback type in search store)
+     * 
+     * @return {string}
+    */
+    var _getSVGClassGlobalShortcut = function () {
+        return "global-shortcuts";
```
shaoshing: Did you forget to commit the icon file?

src/js/actions/shortcuts.js
```diff
@@ -158,6 +161,31 @@ define(function (require, exports) {
...
+     * Executes a given shortcut given an id of some sort
+     * 
+     * @param {string} id of shortcut clicked
+     * @return {Promise}
+     */
+    var executeShortcut = function (id) {
+        var shortcutStore = this.flux.store("shortcut"),
+            shortcuts = shortcutStore.getState().shortcuts;
+
+        if (!shortcuts) {
```
shaoshing: Is this checking necessary even though the default value in the shortcut store is an array?

src/js/actions/tools.js
```diff
@@ -878,7 +879,8 @@ define(function (require, exports) {
                 specs.push({
                     key: utilShortcuts.GLOBAL.TOOLS.SHAPE,
                     modifiers: {},
-                    fn: activateTool
+                    fn: activateTool,
+                    name: "Other " + tool.name.toLowerCase() + " shortcut"
```
shaoshing: This is a hidden shortcut and is duplicate with Rectangle tool. I am wondering if we should just hide it from the search as well.

src/js/actions/search/commands.js
DivyaPrabhakar: We are still deciding if Global shortcuts should be a separate section or not. If so, @placegraphichere would need to provide an icon.

src/js/actions/tools.js
DivyaPrabhakar: I wasn't sure about this but I agree that there should be one rectangle shortcut only. I'll make the shortcut "R" the primary one for now.

src/js/actions/search/commands.js
iwehrman: Maybe in the short-term you can just re-use the menu commands icon.

shaoshing:
Everything looks perfect!

iwehrman:
:tada:

PR #3227, opened 2015-11-11
History cleanup: fix resetIndex to only amend history

mcilroyc:
Updates `layers.resetIndex()` to only amend history.  This continues the effort started in a recent PR which did the same for resetLayers and resetBounds.  This required fixing the two instances that relied on resetIndex to *create* history state by explicitly creating a history state early. 

This adds a second flavor of the generic action to create history state called `newHistoryStateRogueSafe` which will amend a previous state IF it as create roguely (via PS initiated event, described as flow number 1 [here](https://github.com/adobe-photoshop/spaces-design/wiki/History-Subsystem)

Two other things...
- Convert "editArtboardEvent" handler to use new flux action
- Remove 'move' handler from type tool because it is redundant to the handler in actions/transform

Testing note: this affects (hopefully with no functional change) the following:

1. placing layers (from lib, etc)
1. reordering layers
1. moving text layers while in type tool (using CMD drag)
1. transforming artboards

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -1620,11 +1620,9 @@ define(function (require, exports) {
      * though possibly out of order w.r.t. Photoshop's model.
      *
      * @param {Document=} document Document for which layers should be reordered, if undefined, use current document
-     * @param {boolean=} suppressHistory Optional. If true, emit an event that does NOT add a new history state
-     * @param {boolean=} amendHistory Optional. If true, update the current state (requires suppressHistory)
      * @return {Promise} Resolves to the new ordered IDs of layers as well as what layers should be selected
      */
-    var resetIndex = function (document, suppressHistory, amendHistory) {
+    var resetIndex = function (document) {
         if (typeof document === "undefined") {
             document = this.flux.store("application").getCurrentDocument();
 
@@ -1641,15 +1639,7 @@ define(function (require, exports) {
                     });
             })
             .then(function (payload) {
-                if (suppressHistory) {
-                    if (amendHistory) {
-                        return this.dispatchAsync(events.document.history.amendment.REORDER_LAYERS, payload);
-                    } else {
-                        return this.dispatchAsync(events.document.REORDER_LAYERS, payload);
-                    }
-                } else {
-                    return this.dispatchAsync(events.document.history.optimistic.REORDER_LAYERS, payload);
-                }
+                this.dispatch(events.document.history.amendment.REORDER_LAYERS, payload);
             })
             .then(function () {
                 // get the document with latest selected layers, after layer reordering.
@@ -1694,10 +1684,13 @@ define(function (require, exports) {
 
         var numberRef = (typeof target === "number"),
             targetRef = numberRef ? layerLib.referenceBy.index(target) : layerLib.referenceBy[target],
-            reorderObj = layerLib.reorder(layerRef, targetRef),
-            reorderPromise = descriptor.playObject(reorderObj);
+            reorderObj = layerLib.reorder(layerRef, targetRef);
       
-        return reorderPromise.bind(this)
+        return this.transfer(historyActions.newHistoryState, document.id, strings.ACTIONS.REORDER_LAYERS)
+            .bind(this)
+            .then(function () {
+                return descriptor.playObject(reorderObj);
+            })
             .catch(function (err) {
                 // Ignore reorder errors if target was a string (called by one of the bring/send methods below)
                 // since PS fails if we try to move a layer past the doc bounds
@@ -1707,9 +1700,7 @@ define(function (require, exports) {
                 }
             })
             .then(function () {
-                // TODO this resetIndex causes a history state, eventually want to move this to a more explicit
-                // history state creation action
-                return this.transfer(resetIndex, document, false, false);
+                return this.transfer(resetIndex, document);
             })
             .then(function () {
                 // The selected layers may have changed after the reorder.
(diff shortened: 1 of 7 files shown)


src/js/actions/transform.js
```diff
@@ -1136,6 +1136,40 @@ define(function (require, exports) {
...
+
+        return this.transfer("history.newHistoryStateRogueSafe", nextDoc.id)
+            .bind(this)
+            .then(function () {
+                return this.transfer("layers.resetBounds", nextDoc, nextDoc.layers.allSelected);
+            })
+            .then(function () {
+                nextDoc = appStore.getCurrentDocument();
+                return this.transfer("layers.resetSelection", nextDoc);
+            }).then(function () {
```
iwehrman: I know you just moved this, but can you fix the alignment of this `.then`?

iwehrman: Also, since I'm staring at this, does anyone know if it's really necessary to call all three of these actions in a row sequentially? I almost wonder if it would just be faster here to call `updateDocument`...

iwehrman: It's very confusing above that we call `resetBounds` on the selected layers, but THEN call `resetSelection`...

src/js/actions/transform.js
```diff
@@ -1146,21 +1180,11 @@ define(function (require, exports) {
...
-            return this.flux.actions.layers.resetBounds(nextDoc, nextDoc.layers.allSelected)
-                .bind(this)
-                .then(function () {
-                    nextDoc = appStore.getCurrentDocument();
-                    this.flux.actions.layers.resetSelection(nextDoc);
-                }).then(function () {
-                    nextDoc = appStore.getCurrentDocument();
-                    this.flux.actions.layers.resetIndex(nextDoc, true);
-                });
+            this.flux.actions.transform.handleTransformArtboard();
```
iwehrman: Don't you want to return this promise instead of `Promise.resolve`?

src/js/actions/layers.js
```diff
@@ -1694,10 +1684,13 @@ define(function (require, exports) {
 
         var numberRef = (typeof target === "number"),
             targetRef = numberRef ? layerLib.referenceBy.index(target) : layerLib.referenceBy[target],
-            reorderObj = layerLib.reorder(layerRef, targetRef),
-            reorderPromise = descriptor.playObject(reorderObj);
+            reorderObj = layerLib.reorder(layerRef, targetRef);
       
-        return reorderPromise.bind(this)
+        return this.transfer(historyActions.newHistoryState, document.id, strings.ACTIONS.REORDER_LAYERS)
```
iwehrman: Just to confirm: it's important that this run before we play the action below? Is there any opportunity for more concurrency?

src/nls/root/strings.json
```diff
@@ -39,6 +39,7 @@
         "SET_TYPE_TRACKING": "Set Type Tracking",
         "SET_TYPE_ALIGNMENT": "Set Type Alignment",
         "UNGROUP_LAYERS": "Ungroup layers",
+        "REORDER_LAYERS": "Reorder Layers",
```
iwehrman: I just realized that I don't understand what these names are used for, since it doesn't seem like they're making their way into Photoshop?

iwehrman:
A few questions, but generally looks good. This is all looking a better and better, methinks.

src/js/actions/transform.js
mcilroyc: I fixed the indenting of `.then()`.  I stared at this block for awhile too, and made a quick TODO comment in the function header because it seems heavy-handed (and in a weird order! as you said) but I was hoping not to address it as part of this PR which is otherwise safe-ish.

iwehrman: :white_check_mark:

src/nls/root/strings.json
mcilroyc: I've been trying to keep meaningful history state "names" when possible ... in preparation for some day displaying them to the user?  Or at least for troubleshooting.

src/js/actions/transform.js
mcilroyc: I'm not sure at this point, but I'll change it to return the action's promise.

src/js/actions/layers.js
mcilroyc: The problem here is that we ARE still getting historyState events from PS even if we are initiating the change.  I like to prevent, when possible, a race condition in how the history state gets incremented.  By waiting for this new history state to be created in our model *before* playing the action in PS, we insure that the resultant historyState event from PS will be a no-op.  Maybe we should talk about this in more detail, but that's my story for now.

I could be persuaded that it is OK to use the "amend rogue" flow even for these optimistic style actions.

mcilroyc:
I made two changes, and otherwise whined about how I don't want to make any further structural changes right now.  

Perhaps the most important question is that of concurrency of 'optimistic' actions... but rest assured that once I get further along in getting things consistent across all the actions, that is something I'm interested in fixing up right (and hopefully providing some scaffolding to make it easier for everyone)

src/js/actions/layers.js
iwehrman: No, this makes sense. There may some structural opportunities for additional concurrency, but your explanation of why `newHistoryState` should precede everything seems sound. This confirms my understanding. It doesn't immediately seem worthwhile to have a looser kind of history event which can come before or after Photoshop's event, in the sense that it probably wouldn't buy us much, and it would make it harder for us to convince ourselves of its correctness.

src/nls/root/strings.json
iwehrman: This is interesting. We're currently attempting to maintain a 1:1 correspondence with PS history states, right? Those have their own strings and definitely are displayed in the standard PS UI. But, conceivably, we could loosen the correspondence so that we have more history states than PS... For example, I wonder how configuring exports works now w.r.t. history states? Maybe you can explain this to me offline...

iwehrman:
:hamburger: :fries:

PR #3228, opened 2015-11-11
Run grunt tasks concurrently

iwehrman:
This changes the `test` and `requirejs` grunt tasks to execute their subtasks concurrently. Since the subtasks just spawn grunt child processes, this effectively runs the subtasks in parallel. 

On an i7, this speeds up `grunt test` from ~15s to 9s, and `grunt requirejs` (which is called by `grunt compile` and `grunt build`) from 140s to 45s.

This requires you to run `npm install`.


pineapplespatula:
this looks great :trophy:

PR #3229, opened 2015-11-11
refactor Actions to use an object for thier properties

pineapplespatula:
This did touch an awful lot, with most of it being boiler plate, 

fluxcontroller should be revisited now that we have objects, but i made the most trivial changes for now

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -253,50 +253,43 @@ define(function (require, exports, module) {
             
             if (!dependencies) {
                 dependencies = new Set([action]);
-                if (actionObject && actionObject.transfers) {
-                    actionObject.transfers.forEach(function (dependency, index) {
-                        // Translate action pathnames to unsynchronized action functions
-                        if (typeof dependency === "string") {
-                            dependency = this._actionsByName.get(dependency);
-                        }
-
-                        // Validate transfer declarations
-                        if (!dependency) {
-                            var actionName = this._actionNames.get(action);
-                            throw new Error("Transfer declaration " + index + " of " + actionName + " is invalid.");
-                        }
-                        
-                        // Resolve child dependencies before proceeding
-                        _resolveDependencies(dependency, stack).forEach(dependencies.add, dependencies);
-                    }, this);
-                }
-
-                actionDependencies.set(action, dependencies);
-
-                var reads,
-                    writes;
-
                 if (actionObject) {
-                    reads = actionObject.reads || locks.ALL_LOCKS;
-                    writes = actionObject.writes || locks.ALL_LOCKS;
-                } else {
-                    reads = locks.ALL_LOCKS;
-                    writes = locks.ALL_LOCKS;
-                }
+                    if (actionObject.transfers) {
+                        actionObject.transfers.forEach(function (dependency, index) {
+                            // Translate action pathnames to unsynchronized action functions
+                            if (typeof dependency === "string") {
+                                dependency = this._actionsByName.get(dependency);
+                            }
 
-                // Calculate transitive lock sets based on the action's dependencies
-                dependencies.forEach(function (dependency) {
-                    if (dependency.action) {
-                        reads = reads.concat(dependency.action.reads || locks.ALL_LOCKS);
-                        writes = writes.concat(dependency.action.writes || locks.ALL_LOCKS);
-                    } else {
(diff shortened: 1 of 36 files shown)


src/js/fluxcontroller.js
```diff
@@ -272,13 +273,26 @@ define(function (require, exports, module) {
...
 
-                var reads = action.reads || locks.ALL_LOCKS,
-                    writes = action.writes || locks.ALL_LOCKS;
+                var reads,
+                    writes;
+
+                if (actionObject) {
+                    reads = actionObject.reads || locks.ALL_LOCKS;
+                    writes = actionObject.writes || locks.ALL_LOCKS;
+                } else {
```
iwehrman: Shouldn't all actions have a `.action` property? Why is this else branch needed?

src/js/fluxcontroller.js
```diff
@@ -272,13 +273,26 @@ define(function (require, exports, module) {
...
+                }
 
                 // Calculate transitive lock sets based on the action's dependencies
                 dependencies.forEach(function (dependency) {
-                    reads = reads.concat(dependency.reads || locks.ALL_LOCKS);
-                    writes = writes.concat(dependency.writes || locks.ALL_LOCKS);
+                    if (dependency.action) {
+                        reads = reads.concat(dependency.action.reads || locks.ALL_LOCKS);
+                        writes = writes.concat(dependency.action.writes || locks.ALL_LOCKS);
+                    } else {
```
iwehrman: Ditto here.

src/js/fluxcontroller.js
iwehrman: If there is a good reason for this, it would be less redundant to do something like `var actionObject = action.action || {}, reads = actionObject.reads || locks.ALL_LOCKS`, etc.

src/js/fluxcontroller.js
```diff
@@ -561,9 +577,10 @@ define(function (require, exports, module) {
      */
     FluxController.prototype._applyAction = function (action, actionReceiver, params, parentActionName) {
         var flux = this.flux,
-            lockUI = action.lockUI,
-            post = action.post,
-            modal = action.modal || false,
+            actionObject = action.action,
```
iwehrman: Also, if (as discussed above) it _is_ possible for `action.action` to be falsey, then this could cause an error below.

iwehrman:
A few small quibbles about the fluxcontroller changes. Otherwise, this looks good. If anybody else has any concerns about this change, speak now!

src/js/fluxcontroller.js
pineapplespatula: we need to add some additional filtering if we want all of the "actions" that reach here to have an action property. I'll look into adding it.

pineapplespatula:
@iwehrman I didnt adress "_getLayersForDocument" since its not being transfered to, ( but maybe it should be?)

src/js/fluxcontroller.js
```diff
@@ -374,13 +384,14 @@ define(function (require, exports, module) {
      * @return {ActionReceiver}
      */
     FluxController.prototype._makeActionReceiver = function (proto, action, actionName) {
-        if (!action.writes) {
+        var actionObject = action.action;
+        if (actionObject && !actionObject.writes) {
```
iwehrman: This should probably be `!actionObject || !actionObject.writes`.

src/js/fluxcontroller.js
iwehrman: I'm still not clear about whether this is safe.

src/js/fluxcontroller.js
```diff
@@ -248,41 +248,48 @@ define(function (require, exports, module) {
...
-                            throw new Error("Transfer declaration " + index + " of " + actionName + " is invalid.");
-                        }
-                        
-                        // Resolve child dependencies before proceeding
-                        _resolveDependencies(dependency, stack).forEach(dependencies.add, dependencies);
-                    }, this);
-                }
-
-                actionDependencies.set(action, dependencies);
+                if (actionObject) {
```
iwehrman: Can we just require that all actions have a `.action` property?

src/js/fluxcontroller.js
pineapplespatula: i'm not sure we're at a point where we have properly filtered out all the of the actions from the broader "actions". i think we'll need to wait to make that change

iwehrman:
OK!

PR #3230, opened 2015-11-12
Convert remaining history-changing events to the new 'unifiedHistory' handling

mcilroyc:
This is almost entirely a lift-n-shift refactoring, except for two guides-related bug fixes that I could not resist.  This finishes up the effort to move away from using a nested event structure to determine how history state was effected by a given event.  Instead, a history property on the event payload is used.

Guide bug fixes:

- Guide and SmartGuide visibility should not be loaded from history cache during undo/redo
- After navigating history (undo/redo) we must resetGuidePolicies


src/js/actions/history.js
```diff
@@ -104,7 +104,8 @@ define(function (require, exports) {
             .then(function () {
                 var currentDocument = this.flux.store("application").getCurrentDocument();
                 
-                return this.transfer(layerActions.resetLayerVisibility, currentDocument);
+                return Promise.join(this.transfer(layerActions.resetLayerVisibility, currentDocument),
+                    this.transfer("guides.resetGuidePolicies"));
```
mcilroyc: Here go one of the bug fixes, and line 177 below

src/js/models/document.js
```diff
@@ -172,7 +172,9 @@ define(function (require, exports, module) {
     Document.prototype.overlayVisibility = function (document) {
         var nextLayers = this.layers.overlayVisibility(document.layers);
 
-        return this.set("layers", nextLayers);
+        return this.set("layers", nextLayers)
+            .set("guidesVisible", this.guidesVisible)
+            .set("smartGuidesVisible", this.smartGuidesVisible);
```
mcilroyc: This is the other bug fix: retain guide/smartguide visibility from current state when navigating to a cached history state

src/js/actions/transform.js
```diff
@@ -573,7 +585,8 @@ define(function (require, exports) {
 
             payload.size = newSize;
             
-            dispatchPromise = this.dispatchAsync(events.document.history.optimistic.RESIZE_DOCUMENT, payload);
+            dispatchPromise = this.dispatchAsync(events.document.history.RESIZE_DOCUMENT, payload);
+            // TODO is this missing historyStateInfo?
```
iwehrman: @DivyaPrabhakar @volfied: do either of you know anything about this?

mcilroyc: I'm interested to know, but I'm also now going back through these comments to unspool them myself.  This was primarily a note-to-self.

iwehrman:
Looks great. I just want to get some input on that `TODO` from @DivyaPrabhakar or @volfied before proceeding, leading either to a :+1: or a bug report.

src/js/actions/transform.js
DivyaPrabhakar: Nope, that wasn't from me.

iwehrman:
Nevermind, @mcilroyc is gonna solve that TODO.

Issue #3224, opened 2015-11-10
Select: Leaf layer highlighting doesn't work with modifier key


chadrolfs:
new doc, create a bunch of shapes and nest them in groups
deselect (nothing selected)
mouse over the shapes with Select tool (result top level group highlights)
hold down cmd/ctrl and mouse over shapes

Result: no highlighting at all

iwehrman:
Looks good, works good.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
YAY!  A bug fix to verify.

chadrolfs:
Here is a case where it doesn't work, where a selection bounding box includes unselected layers that you want to add to the selection (holding Cmd/Ctrl + Shift).  

![select-overlap](https://cloud.githubusercontent.com/assets/8442631/11155783/a45bfcf4-89fc-11e5-955d-a25eaa77a387.gif)

iwehrman:
FBNC 2 @chadrolfs again :notes:

chadrolfs:
2nd case was originally #3195 but now 'tis fixed, yay!

PR #3232, opened 2015-11-12
Fix leaf highlighting

volfied:
Our latest changes to mouse over highlighting did not consider leaf layers... to fix it, I've added layerstructure.selectableLeaves which is used twice in SuperselectOverlay. #3224

Also fixed a related issue, where due to priority being higher, `policy.beforeStartup` would run after `tools.beforeStartup` and overwrite the active tool's mouse mode, which broke highlighting until user switched tools.


iwehrman:
Looks good, works good.

PR #3236, opened 2015-11-12
Add transfer to getLayerIDsForDocumentID in two places

mcilroyc:
This fixes a regression in artboard creation, and layer ungrouping, both that seemed to have cropped up today.

Remove post verification of getLayerIDsForDocumentID (I think that was a copy paste error?)

two other other minor fixes in promise chains

cc: @pineapplespatula


iwehrman:
LGTM.

Issue #3221, opened 2015-11-10
On-canvas UI is too big


chadrolfs:
Here is what it looks like in 16.1 shipping html
![screen shot 2015-11-10 at 2 38 42 pm](https://cloud.githubusercontent.com/assets/8442631/11078164/00f2e6a8-87b9-11e5-9d02-49f59f567252.png)

Here is what it looks like on current master:
![screen shot 2015-11-10 at 2 38 22 pm](https://cloud.githubusercontent.com/assets/8442631/11078168/09a4217c-87b9-11e5-85bd-1e74c668abe1.png)

iwehrman:
It's not specific to the artboard adders. All on-canvas UI is too big:
![image](https://cloud.githubusercontent.com/assets/1075154/11102122/8e63a57a-886f-11e5-8e7c-cc80ff9f907a.png)

chadrolfs:
woah, it sure is!

iwehrman:
For some reason `window.getComputedStyle(window.document.documentElement).fontSize` is returning "16px" when we first set the root size. It returns the correct value ("10px") just a little later, but we aren't catching that. I'm not sure why this changed, but I suspect it has something to do with #3214. It seems like this shouldn't happen as long as we wait for styles to load before loading starting the controller. Investigating...

mcilroyc:
FBNC to @chadrolfs  ... note that I could not repro this issue, so I'm taking ian's word that this addresses it.

PR #3237, opened 2015-11-13
Fix initial on-canvas UI size

iwehrman:
This PR delays instantiating the flux controller (and hence initializing the stores) until `main.startup` is called. This addresses a bug in which the UI store would have incorrect initial state (`rootSize`) due to inspecting the DOM before styles had finished loading.

Addresses #3221.


mcilroyc:
LGTM.  

FWIW, I can not repro the related issue (based on that slim info in that issue), but nothing breaks that I can see!

PR #3239, opened 2015-11-13
History cleanup part 3

mcilroyc:
Internally within the history store, delay the creation/merging of HistoryState from props so that 'name' can be properly merged (if supplied)

Minor re-org of the following three layer actions so that they are more history-safe, and they supply a meaningful history state name: rename, lock, createArtboard;

Removed dead code: layers.addVectorMask

Removed 'finalize history' event, since it is no longer needed in the _layerTransformHandler


mcilroyc:
For those playing at home, I just rebased and simplified the changes a bit.  I'm no longer adding historyStateInfo in a few spots that I thought it was necessary (but turned out not to be).  So this PR is back to a stable spot.

src/js/actions/layers.js
```diff
@@ -1563,28 +1563,29 @@ define(function (require, exports) {
      * @returns {Promise}
      */
     var setLocking = function (document, layer, locked) {
-        var payload = {
+        if (layer.isBackground) {
+            return _unlockBackgroundLayer.call(this, document, layer);
```
iwehrman: Note to @pineapplespatula: here's another opportunity for a private-action conversion.

src/js/actions/layers.js
```diff
@@ -2413,49 +2414,6 @@ define(function (require, exports) {
     };
 
     /**
-     * create a vector mask on the selected layer
-     *
-     * @return {Promise}
-     */
-    var addVectorMask = function () {
```
iwehrman: @pineapplespatula: is it OK to remove this? Or are you planning to use it sometime soon?

iwehrman:
Looks OK to me. I'll wait for @pineapplespatula to chime in about the dead `addVectorMask` action.

src/js/actions/layers.js
pineapplespatula: go ahead and remove this

PR #3240, opened 2015-11-13
Simple layer index validation as a post condition of undo/redo

mcilroyc:
@iwehrman (or anybody): should we take this opportunity to factor out the _verify* functions into their own file so they can be used across actions?  Or is this "hack" OK for now (exposing the function directly)?


iwehrman:
The hack is perfectly fine for now. @pineapplespatula is working on cleaning up some of the controller infrastructure that relates to action visibility (i.e., "private" or "protected" actions), and may wish to consider this use case as well. Which is to say, maybe these verification things should just be actions as well; perhaps private ones.

PR #3216, opened 2015-11-06
Refactor Draggable/Droppable with performance improvements

shaoshing:
* Improved initial drag performance (#3181)
* Improved overall performance when dragging over droppable items
* Make the Droppabel / Draggable components easier to use and read.

### Test

Before: https://www.dropbox.com/s/eja1idsmxg6z1i4/dnd-before.mov?dl=0
After: https://www.dropbox.com/s/u6mz642ar10ueuv/dnd-after.mov?dl=0
Here is a document (I stole from @iwehrman long ago) contains 1000+ layers which makes it perfect for testing/comparing the administration: 

Here are some drag-and-drop scenarios:

* layer reordering: layer -> layer 
* create graphic from layer: layer -> library panel
* create layer from graphic: graphic -> canvas

PR #3245, opened 2015-11-13
Layer Panel performance improvements

shaoshing:
**This PR is based on #3216, so it will show unrelated commits; I will rebase once #3216 is merged. To view the actual changes, please see commit 6e5745a **

Previously we have a flat structure in the layer panel: it has no parent / children hierarchy in the DOM. This structure has two problems: whenever its document model changed, LayerPanel has to 1) check each LayerFace for rendering, and 2) enforces a relatively complicated `shouldComponentUpdate` and `render` implementations in LayerFace. In this PR, I made the change to convert the LayerFace list into a tree structure to avoid massive rendering due to parent layer change, and implemented a light-weight `layerDiff` function (< 25ms) for less SCU calls.

##### better performance for layer selection / expanding / toggling visibility

Before, when a layer is selected / expanded / hidden, all the LayerFace have to be checked and then re-render descendants to reflect the parent layer; this is slow, especially when the changed layer is the root (for example, collapsing an art board). With the introduced tree structure, we can simply update the changed LayerFace (changed layers are the diff of the current and next layer tree) and let the browser update the children via CSS selector (e.g. ```.layer__selected .children {background-color: blue}```). 

###### Layer selection before / after
layer-selection-before
layer-selection-after
(larger time span indicates the number of children updated)

###### Layer expansion before / after
layer-expansion-before
(larger time span indicates the number of children updated)

layer-expansion-after

##### Other attempts I tried

* manage the selection / expand / visibility state in each LayerFace component. I successfully implemented the individual selection state, but not seeing a lot of improvement. It might also become complicated when it comes to history / group selection / drag and drop
* improve Position / Size panel performance. I see a total of 60ms spend on these two panels when layer selection changed, but there are not much room for improvement as the update seems necessary.

##### Profiling tools

For people who are interested in performance, this is a snippet (https://gist.github.com/shaoshing/fd37e4bd9cf2ede17cb6) that I used which will mark the `render` and `shouldComponentUpdate` calls in the Timeline (will provide a much better view of what is running during the scripting time) and also log their run times. Might be useful to have some interesting findings.

screen shot 2015-11-12 at 12 33 03 pm
screen shot 2015-11-12 at 12 34 31 pm

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -56,6 +56,8 @@ define(function (require, exports, module) {
          * @type {boolean}
          */
         _isDragEventTarget: false,
+        
+        _isExpanded: false,
 
         getInitialState: function () {
             return {
@@ -67,7 +69,12 @@ define(function (require, exports, module) {
         },
         
         shouldComponentUpdate: function (nextProps, nextState) {
-            if (nextProps.isLayerChanged) {
+            var depth = nextProps.depth,
+                layer = nextProps.layer,
+                isLayerChanged = !!nextProps.changedLayerIDPaths[depth] && 
+                    nextProps.changedLayerIDPaths[depth].has(layer.id);
+
+            if (isLayerChanged) {
                 return true;
             }
             
@@ -489,15 +496,15 @@ define(function (require, exports, module) {
                 isDragging = this.state.isDragging,
                 isDropTarget = this.state.isDropTarget,
                 dropPosition = this.state.dropPosition,
-                isGroup = this.props.children;
+                hasChildren = this.props.childNodes;
             
             var layerClasses = classnames({
                 "layer": true,
                 "layer__select": layer.selected,
-                "layer-group": isGroup,
-                "layer-group__selected": isGroup && layer.selected,
-                "layer-group__collapsed": isGroup && !layer.expanded,
-                "layer-group__not-visible": isGroup && !layer.visible,
+                "layer-group": hasChildren,
+                "layer-group__selected": hasChildren && layer.selected,
+                "layer-group__collapsed": hasChildren && !layer.expanded,
+                "layer-group__not-visible": hasChildren && !layer.visible,
                 "layer-group__drag-target": isDragging
             });
 
@@ -550,6 +557,24 @@ define(function (require, exports, module) {
                 if (layer.textWarningLevel === 2) {
                     iconClassModifier = "face__kind__alert";
                 }
+            }
+
+            var layerList;
+
+            if (this.props.childNodes && (layer.expanded || this._isExpanded)) {
+                this._isExpanded = true;
+                
+                var LayersList = require("jsx!./LayersList");
+
+                layerList = (
+                    <LayersList
+                        disabled={this.props.disabled}
+                        document={this.props.document}
+                        rootLayer={layer}
+                        layerNodes={this.props.childNodes}
+                        depth={this.props.depth + 1}
+                        changedLayerIDPaths={this.props.changedLayerIDPaths}/>
+                );
             }
             
             return (
@@ -607,7 +632,7 @@ define(function (require, exports, module) {
                                     onClick={this._handleLockToggle}>
                                 </ToggleButton>
                             </div>
-                            {this.props.children}
+                            {layerList}
                         </li>
                     </Droppable>
                 </Draggable>
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -52,15 +52,24 @@ define(function (require, exports, module) {
         
         /**
          * Indicates whether the component is the initial target of a drag event.
-         
+         *
+         * @private
          * @type {boolean}
          */
         _isDragEventTarget: false,
         
+        /**
+         * True if the layer is or was expanded. Used to avoid
+         * rendering its child faces until they are first visible.
+         *
+         * @private
+         * @type {boolean}
+         */
         _isExpanded: false,
 
         getInitialState: function () {
             return {
+                layer: this._getDocument().layers.byID(this.props.layerID),
                 isDropTarget: false,
                 dropPosition: null,
                 isDragging: false,
@@ -68,11 +77,20 @@ define(function (require, exports, module) {
             };
         },
         
+        componentWillMount: function () {
+            this.getFlux().stores.document.addLayerFaceListener(this.props.documentID, this.props.layerID,
+                this._handleLayerFaceChange);
+        },
+        
+        componentWillUnmount: function () {
+            this.getFlux().stores.document.removeLayerFaceListener(this.props.documentID, this.props.layerID,
+                this._handleLayerFaceChange);
+        },
+
         shouldComponentUpdate: function (nextProps, nextState) {
             var depth = nextProps.depth,
-                layer = nextProps.layer,
-                isLayerChanged = !!nextProps.changedLayerIDPaths[depth] && 
-                    nextProps.changedLayerIDPaths[depth].has(layer.id);
+                isLayerChanged = !!nextProps.changedLayerIDPaths[depth] &&
+                    nextProps.changedLayerIDPaths[depth].has(nextProps.layerID);
 
             if (isLayerChanged) {
                 return true;
@@ -81,7 +99,8 @@ define(function (require, exports, module) {
             if (this.state.isDragging !== nextState.isDragging ||
                 this.state.dragStyle !== nextState.dragStyle ||
                 this.state.dropPosition !== nextState.dropPosition ||
-                this.state.isDropTarget !== nextState.isDropTarget) {
+                this.state.isDropTarget !== nextState.isDropTarget ||
+                !Immutable.is(this.state.layer, nextState.layer)) {
                 return true;
             }
             
@@ -95,7 +114,7 @@ define(function (require, exports, module) {
          * @param {SyntheticEvent} event
          */
         _handleIconClick: function (event) {
-            var layer = this.props.layer;
+            var layer = this.state.layer;
             if (!layer.isGroup) {
                 return;
             }
@@ -111,10 +130,8 @@ define(function (require, exports, module) {
 
             // The clicked layer may an have out-of-date document models due to
             // the aggressive SCU method in LayersPanel.
-            var documentID = this.props.document.id,
-                documentStore = this.getFlux().store("document"),
-                currentDocument = documentStore.getDocument(documentID),
-                currentLayer = currentDocument.layers.byID(this.props.layer.id);
+            var currentDocument = this._getDocument(),
+                currentLayer = currentDocument.layers.byID(this.state.layer.id);
 
             this.getFlux().actions.layers.setGroupExpansion(currentDocument, currentLayer,
                 !currentLayer.expanded, descendants);
@@ -130,11 +147,16 @@ define(function (require, exports, module) {
         _handleLayerNameChange: function (event, newName) {
             if (newName.length !== 0) {
                 if (newName !== this.props.layer.name) {
-                    this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+                    this.getFlux().actions.layers.rename(this._getDocument, this.props.layer, newName);
                 }
             } else {
                 this.refs.layerName.setValue(this.props.layer.name);
             }
+        },
+        
+        /** @ignore */
+        _getDocument: function () {
+            return this.getFlux().stores.document.getDocument(this.props.documentID);
         },
 
         /**
@@ -148,6 +170,13 @@ define(function (require, exports, module) {
             // TODO: Skip to next layer on the tree
             event.stopPropagation();
         },
+        
+        /** @ignore */
+        _handleLayerFaceChange: function (nextLayer) {
+            this.setState({
+                layer: nextLayer
+            });
+        },
 
         /**
          * Grabs the correct modifier by processing event modifier keys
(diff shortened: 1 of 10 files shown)


iwehrman:
I'll wait to try to read this until #3216 is merged and you've rebased this against master, but in my quick tests so far the performance improvements are tremendous. Group expand/collapse in particular is dramatically improved; like, by an order of magnitude. Bravo.

It's too bad that our action performance measurements don't currently take React rendering time into account. I'd love to change that, so that the numbers that are dumped the console when an action is played account both for the time spent in the queue and waiting for PS, but also the time spent re-rendering the UI. Maybe we could eventually do that if we wrote our own `StoreWatchMixin` to somehow return promises that only resolve once the corresponding `setState` callback has been applied? Just thinkin' crazy here.

Your snippet is cool, too! Did you also try using the built-in React performance-measuring tools? https://facebook.github.io/react/docs/perf.html It provides roughly similar information, though a) at a slightly less granular level, and b) disconnected from the timeline. I suspect that, as a team, we are less comfortable analyzing the timeline than we should be. Maybe sometime in the near future you could give us an in-depth tour of what you've learned by doing this? I'm interested at least :grin:

iwehrman:
This is now ready for rebase against master.

iwehrman:
@shaoshing: should this be assigned to someone other than you? Just wondering!

shaoshing:
@iwehrman yes I've rebased with master and you should be able to view the actual changes now. Sorry for the delay.

Regarding the React performance tool, I haven't tried it yet but it looks useful; I shall give it a try next time. It is a great idea to keep track of the component performance, like what we have for the actions, but I also noticed a downgrade in performance when I log every `render` and `shouldComponentUpdate` methods. I am not sure the cause but it may be a better idea by just tracking those components with large scope.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -87,35 +83,39 @@ define(function (require, exports, module) {
...
          * @type {Set.<number>}
          */
-        _mountedLayerIDs: null,
+        _expandedLayerIDs: new Set(),
+        
+        /**
+         * Set of layer IDs that are changed in the last document model update. This is updated when LayerPanel 
+         * receives a new "document" prop. See "LayerPanel#_layerDiff" for details.
+         */
+        _changedLayerIDs: new Set(),
```
iwehrman: I believe that these properties end up in the prototype, so you shouldn't initialize it to a mutable value. Instead, you should initialize it in `componentWillMount` (or similar) so that each panel instance has it's own mutable set instance.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -87,35 +83,39 @@ define(function (require, exports, module) {
...
-         * rendering faces until they are first visible. Initialized when the
-         * panel is mounted.
+         * Set of layer IDs that is or was expanded. Used to avoid
+         * rendering child faces until they are first visible.
          *
          * @private
          * @type {Set.<number>}
          */
-        _mountedLayerIDs: null,
+        _expandedLayerIDs: new Set(),
```
iwehrman: Ditto here.

src/js/jsx/sections/layers/LayersPanel.jsx
iwehrman: (I could be wrong if, e.g., React is internally doing some kind of ES6 class magic. )

iwehrman:
Well, after reading this, I admit that I'm not _quite_ sure I understand what's going on here. On the one hand, I see that you've reflected the layer structure into the component structure. This seems reasonable enough, but I wouldn't expect this to provide a huge benefit on it's own across differently structured PSDs. And I can't actually convince myself that this is any faster when selecting individual layers in a big flat document compared to master. But you're also computing answers up front to the  `shouldComponentUpdate` methods of each face, and in cases where I _do_ see a big speedup (like when collapsing a large group), it must be from that. It's not immediately obvious to me though why this structural change is necessary though. The downside of it is definitely that you seem to be partially re-creating the React virtual diffing algorithm right in the `LayersPanel`. It's true that `shouldComponentUpdate` is complicated on master, but I don't quite understand why it's slow. I guess the derived properties that are necessary for the various constant-time checks, when computed individually, are more expensive than if similar checks were computed hierarchically? This is my guess, at least, but like I said I don't completely get it. Can you help me understand better?

iwehrman:
An alternative approach could be as follows:

1. Change the layers panel so that, when the set of selected layers change, `shouldComponentUpdate` returns false.
2. Augment the document store to emit a new kind of event when the set of selected layers changes. Include with this selection-changed event the set of newly selected and newly deselected layers.
3. Add listeners for this selection-changed event to layer faces. If the face is in one of the two aforementioned sets, it updates itself.

This would require a different implementation based on whether the layers panel renders represents faces as a tree or a list. If it's a tree, then the sets above could be "precise" (i.e., not expanded to include descendents and ancestors), descendants and ancestors are updated with pure CSS. The downside to this is that if you a group is changed then we have to call sCU on all its children. If, on the other hand, the panel is represented as a flat list, then the sets of selection added/removed layers needs to include all its ancestors, and they all have to be re-rendered. But, all and only the necessary layers are re-rendered. 

As an aside, I still don't completely understand the trade-offs between a flat and hierarchical structure here.

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -124,43 +124,129 @@ define(function (require, exports, module) {
                 return true;
             }
 
-            if (this.props.visible !== nextProps.visible) {
+            if (this.props.visible !== nextProps.visible ||
+                this.props.active !== nextProps.active) {
```
mcilroyc: nit: couldn't this be combined with the block above?

shaoshing:
Hi @iwehrman  @mcilroyc , thanks for reviewing. I think a flat structure has bad performance when it comes to long-nested layer list, as any update will 1) trigger all SCU, and 2) require a large number of re-renders when the target layer has many children. One way to solve issue 1 is the store solution: let LayerFace listen to a store and the store emits event to the changed components. I believe this will have notable improvements on leaf layers, since all they require is updating classes. However, issue 2 remains unsolved. One example is selecting an art board: in a flat structure, we still have to update all children. That’s how I come to the solution with hierarchical structure + LayerPanel diff. On one handle, the diff algorithm helps keep minimum SCU calls; on the other hand, the hierarchical structure allows us to use CSS for style updates to eliminate the massive children updates.

The Tradeoff is that we lost some React’s “purity” by doing part of it job in LayerPanel (the diff part), according to Cory, and there is no major improvement in layer selection: the total script time is around 200ms when selecting leaf layer (v.s. 250ms before). One thing worth to mention is that within the 200ms run time, around 100ms is spent on actions and other renderings (transform, appearance, etc.), which means even if we can manage to improve the LayerPanel’s performance to less than 50ms, we may still experiencing lagging in a large document. I am not suggesting we should give up trying, though, just sharing what I found so that we can have a better view of the problem.

screen shot 2015-11-23 at 11 34 17 am
(master branch, A is the start of a layer click event, B is the start of LayerPanel rendering. Scripting time is longer due to my tracking code)

screen shot 2015-11-23 at 11 39 56 am
(this branch, A is the start of a layer click event, B is the start of LayerPanel rendering. Scripting time is longer due to my tracking code)

@iwehrman we discussed the store solution before, and I did experiment it but gave up as it did not solve issue 2 with the flat structure. However with the hierarchical structure, things may become easier, and I am giving it another try now. I will post update later today or tomorrow morning.

shaoshing:
Hi @iwehrman @mcilroyc, I’ve completed my second attempt on improving the layer panel performance. The second version is based on my initial work, and here is the summarized road map of this PR:

##### Version 1
Version 1 converted the layer list into hierarchical structure, and calculate diff of current and next document in LayerPanel to control which LayerFace should be updated, to minimize the number of SUC called. version 1 brings the following improvements:

- faster expand/collapse layer group
- faster select/deselect layer group
- faster toggling visibility of layer group

##### Version 2 (current version)
in this version, since the diff tells us which layers are updated, or whether the layer tree is changed (when layer re-order/creation/deletion happened), we can move the diff algorithm to the document store, and have the LayerFace listen to the store. This allow the store to directly notify the changed layer to perform a update. Version 2 has the following improvements:

1. Faster layer selection response; the time it takes from clicking to highlighting the clicked layer has been reduced from 120ms to 40ms on average. However, this doesn’t mean the total runtime is 80ms less. The improved respond time is from the advantage that document store emits `layer-face-change` and `layer-tree-change` events before emitting a general `change` event, so that LayerFace can update without waiting (my test shows that Transform and Size panel can take 100ms to re-render in a large document).
2. only actions that affect the LayerPanel will trigger its UI re-render. (e.g. if you are just changing fill color, the diff is not performed and thus no `layer-face-change` or `layer-tree-change` event is emitted)
3. things improved in version 1 also get better due to 1.

Sadly these improvements introduce extra complexity in our app. For example, In the document store, we need to perform precise layer diff to have the LayerPanel refresh correctly (I've tested most of the actions that invalidate the LayerPanel. you can see below for a list of things to test). Each action handler in the document store has to indicate whether the change should trigger a layer diff (via the `performLayerDiff` option in `_setDocument`). In the React side, LayerFace can no longer rely on its props for latest layer and document model, since re-rendering in the LayerPanel is blocked. The correct way is passing in IDs, and let the LayerFace fetches the models from the store with the IDs. However, we may have a higher chance of encountering errors due to out-of-date data in these components. Due to these, I am not entirely confident to say this is the right direction for the problem. Maybe you can review my code, test the Vermilion document on my branch, and let me know your thoughts.

##### Things I tested (please let me know if anything is missing):

1. select/deselect layer (including multiple selections)
2. expand/collapse layer group / art board
3. toggle visibility/lock
4. re-order layer, layer group, art board
5. create/delete layer/group/art board
6. trying to undo 3 ~ 5

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -617,10 +591,29 @@ define(function (require, exports, module) {
                 }
             }
 
+            var layerList;
+
+            if (hasChildren && (layer.expanded || this._isExpanded)) {
+                this._isExpanded = true;
+                
+                var LayersList = require("jsx!./LayersList");
```
shaoshing: `LayerList` is also requiring `LayerFace`, so we need to put the require statement in the function scope, otherwise `LayersList` will be an empty object.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -154,7 +144,15 @@ define(function (require, exports, module) {
...
-            this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+            this.getFlux().actions.layers.rename(this._getDocument(), this.state.layer, newName);
+        },
+        
+        /**
+         * @private
+         * @return {Document}
+         */
+        _getDocument: function () {
+            return this.getFlux().stores.document.getDocument(this.props.documentID);
```
shaoshing: LayerFace uses this helper function to fetch the latest document whenever needed, instead of reading from its props. (actually the props do not include the document and layer models any more, they are replaced by their IDs)

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -65,53 +76,34 @@ define(function (require, exports, module) {
...
+                this._handleLayerFaceChange);
+        },
+
         shouldComponentUpdate: function (nextProps, nextState) {
-            // Drag states
+            var depth = nextProps.depth,
+                isLayerChanged = !!nextProps.changedLayerPaths[depth] &&
+                    nextProps.changedLayerPaths[depth].has(nextProps.layerID);
+
+            if (isLayerChanged) {
```
shaoshing: This part respond to `layer-tree-change` event initiated by root LayerList. It checks whether its layer is within the changed layer path (for example, when the layer's children is deleted / create / re-ordered). Otherwise it will skip its rendering (and its children)

mcilroyc:
My gut reaction is that this "version 2" is a bridge too far.  The extra complexity doesn't seem worth it.

Dispatch from the "can't react just be better?" department ... is there any chance that *some* improvements might be gained with the imminent React upgrade?

shaoshing:
We have one down vote for version 2 from Cory, so @iwehrman can you review and let us know your thoughts about version 2. If we decide to discard version 2, we can always revert to version 1 as version 2 is additional commits on top of 1.

BTW, this branch is rebased with master to run with webpack.

src/js/models/documentindex.js
```diff
@@ -137,7 +137,7 @@ define(function (require, exports, module) {
             hasNextActiveDocumentID = typeof nextActiveDocumentID === "number";
 
         if (hasOpenDocuments !== hasNextActiveDocumentID) {
-            throw new Error("There must be a next active document ID iff there are open documents");
+            throw new Error("There must be a next active document ID if there are open documents");
```
iwehrman: https://en.wikipedia.org/wiki/If_and_only_if

shaoshing: ha, that's interesting.

iwehrman:
For those playing along at home, @shaoshing and I talked about an alternate approach that may yield bigger speed-ups in some cases with less complexity. The current branch does provide a solid speedup, but I think I agree with @mcilroyc insofar as the perceived performance difference may not be worth the increased code complexity. I think increasing the code complexity is definitely a tradeoff worth considering, but we also want to make sure that, if we're going to load up our technical debt credit cards,  we get something really great in return.

I suggest that we do not delete this branch because it's certainly possible that we'll return to this approach later, but also that we can probably close this PR for now.

shaoshing:
Sounds good. Thanks for the feedback @mcilroyc @iwehrman . I will give this performance issue another try and post it in a separate branch.

PR #3247, opened 2015-11-13
Consolidate all localization to a util helper

volfied:
- Change all i18n!nls/... requires to use js/util/nls
 - All translated strings are now quote wrapped and are passed to nls.localize method

Another side effect of the webpack efforts, centralizing our localization code will help me replace it easier later on.

Usage is pretty similar, instead of requiring the dictionary with `i18n!` prefix, we now use `js/util/nls.localize` function to grab the localized string/tree at any point. 

It touches A LOT of files.. but it should be fairly clean.

### DIFF
--- a/src/js/util/nls.js
+++ b/src/js/util/nls.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
+ * Copyright (c) 2015 Adobe Systems Incorporated. All rights reserved.
  *  
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"), 
@@ -24,13 +24,19 @@
 define(function (require, exports) {
     "use strict";
 
-    var log = require("js/util/log");
+    var objectUtil = require("js/util/object");
 
-    var dictionaries = {
+    /**
+     * All the dictionaries we use for our strings
+     *
+     * @private
+     * @type {Object}
+     */
+    var _dictionaries = {
         strings: require("i18n!nls/strings"),
-        menus: require("i18n!nls/menu"),
-        macShortcuts: require("i18n!nls/shortcuts-mac"),
-        winShortcuts: require("i18n!nls/shortcuts-win")
+        menu: require("i18n!nls/menu"),
+        "shortcuts-mac": require("i18n!nls/shortcuts-mac"),
+        "shortcuts-win": require("i18n!nls/shortcuts-win")
     };
         
     /**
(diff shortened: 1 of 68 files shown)


src/js/util/nls.js
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
iwehrman: Should be 2015.

src/js/util/nls.js
```diff
@@ -0,0 +1,61 @@
...
+        macShortcuts: require("i18n!nls/shortcuts-mac"),
+        winShortcuts: require("i18n!nls/shortcuts-win")
+    };
+        
+    /**
+     * Thin wrapper around i18n so it's centralized
+     * which will help us replace it easily later on
+     *
+     * @param {string} key of the format [file].[dot-separated-key]
+     *
```
iwehrman: I think you're losing the battle for a newline between `@param` and `@return` :)

src/js/util/nls.js
```diff
@@ -0,0 +1,61 @@
...
+     */
+    var localize = function (key) {
+        var value = dictionaries;
+
+        key.split(".").some(function (part) {
+            if (value.hasOwnProperty(part)) {
+                value = value[part];
+            } else {
+                value = null;
+                log.warn("Translation not found for: " + key);
```
iwehrman: Why warn here? Shouldn't this throw?

iwehrman: Which is to say: I can't see how returning null here is better than throwing.

src/js/util/nls.js
```diff
@@ -0,0 +1,61 @@
...
+define(function (require, exports) {
+    "use strict";
+
+    var log = require("js/util/log");
+
+    var dictionaries = {
+        strings: require("i18n!nls/strings"),
+        menus: require("i18n!nls/menu"),
+        macShortcuts: require("i18n!nls/shortcuts-mac"),
+        winShortcuts: require("i18n!nls/shortcuts-win")
```
iwehrman: It would make me feel better (I think it would be less confusing) if the keys exactly matched the file names.

src/js/util/nls.js
```diff
@@ -0,0 +1,61 @@
...
+     * which will help us replace it easily later on
+     *
+     * @param {string} key of the format [file].[dot-separated-key]
+     *
+     * @return {string|Object} Translated, or fallen-back string, or the sub-tree for the ID
+     */
+    var localize = function (key) {
+        var value = dictionaries;
+
+        key.split(".").some(function (part) {
```
iwehrman: It's too bad we can't use `object.getPath` here?

iwehrman: I think we could if there were two parameters: one for the dictionary and one for the path into the dictionary...

iwehrman:
Some grips for @volfied, but no qualms with the general approach and I think we can merge soon.

src/js/util/nls.js
volfied: Let's do this separately for ALL the files later.

volfied:
@iwehrman Thank you for the review, back to you.

iwehrman:
Everybody get ready to rebase!

PR #3248, opened 2015-11-13
Add shift+cmd/ctrl policy to inner selection rectangle

volfied:
Addresses #3224... because we need more policies


iwehrman:
MOAR POLICIES.

PR #3249, opened 2015-11-14
first attempts at cleaning up fluxController

pineapplespatula:
this PR adds the ability to flag actions as private
it also adds a module tag to each action in flux controller, and filters out functions that are not actions from being called.

This may cause widespread side effects depending on how consitantly we implemented and used actions previously so I'm marking it as review only for now

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -309,7 +309,8 @@ define(function (require, exports) {
     };
     initInactiveDocuments.action = {
         reads: [locks.PS_DOC],
-        writes: [locks.JS_DOC]
+        writes: [locks.JS_DOC],
+        private: true
     };
 
     /**
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -77,11 +77,6 @@ define(function (require, exports, module) {
     var MAX_RETRY_WINDOW = 6400;
 
     /**
-    *@type {Object}
-    */
-    var _allActions = {};
-
-    /**
      * Priority order comparator for action modules.
      *
      * @private
@@ -90,8 +85,8 @@ define(function (require, exports, module) {
      * @return {number}
      */
     var _actionModuleComparator = function (moduleName1, moduleName2) {
-        var module1 = _allActions[moduleName1],
-            module2 = _allActions[moduleName2],
+        var module1 = actionIndex[moduleName1],
+            module2 = actionIndex[moduleName2],
             priority1 = module1._priority || 0,
             priority2 = module2._priority || 0;
 
@@ -114,7 +109,7 @@ define(function (require, exports, module) {
         this._synchronizedActions = new Map();
         this._idleTasks = new Set();
 
-        var actions = this._synchronizeAllModules(_allActions),
+        var actions = this._synchronizeAllModules(this._allActions),
             stores = storeIndex.create(),
             allStores = _.merge(stores, testStores || {});
 
@@ -145,6 +140,22 @@ define(function (require, exports, module) {
      * @type {ActionQueue} 
      */
     FluxController.prototype._actionQueue = null;
+
+    /**
+     * local store of all the available properly defined flux actions
+     * 
+     * @private 
+     * @type {Object}
+     */
+    FluxController.prototype._allActions = null;
+
+    /**
+     * Map from action functions to modules
+     *
+     * @private
+     * @type {Map.<function, string>}
+     */
+    FluxController.prototype._actionModules = null;
 
     /**
      * Map from unsynchronized action functions to pathnames.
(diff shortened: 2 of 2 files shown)


src/js/fluxcontroller.js
```diff
@@ -77,6 +77,11 @@ define(function (require, exports, module) {
     var MAX_RETRY_WINDOW = 6400;
 
     /**
+    *@type {Object}
```
iwehrman: Broken comment formatting, plus this variable should be part of the object state. And private.

src/js/fluxcontroller.js
```diff
@@ -446,6 +472,15 @@ define(function (require, exports, module) {
                         throw new Error(message);
                     }
     
+                    if (nextAction.private) {
+                        if (nextAction.module !== action.module) {
+                            var privateMessage = "Invalid transfer from " + actionName + " to " + nextActionName +
+                                    ". " + nextActionName + " is declaed private and is in a different module from" +
```
iwehrman: Typo.

src/js/fluxcontroller.js
```diff
@@ -588,6 +623,9 @@ define(function (require, exports, module) {
             actionTitle = "sub-action " + actionName + " of action " + parentActionName;
         } else {
             actionTitle = "action " + actionName;
+            if (action.private) {
+                throw new Error("Action " + actionName + "is private and cannot be exectued");
```
iwehrman: Typo.

src/js/fluxcontroller.js
```diff
@@ -202,6 +208,26 @@ define(function (require, exports, module) {
...
+     * @private
+     */
+    FluxController.prototype._initAllActions = function () {
+        Object.keys(actionIndex).forEach(function (actionModuleName) {
+            var actionModule = actionIndex[actionModuleName],
+                moduleActions = {};
+            Object.keys(actionModule)
+                .forEach(function (actionName) {
+                    if (actionModule[actionName].action) {
+                        actionModule[actionName].module = actionModuleName;
```
iwehrman: I don't think we should modify the `.action` object. Instead, we should just keep this state in the `FluxController` instances, the same way we keep the locks and the name/function mappings.

src/js/actions/documents.js
```diff
@@ -304,7 +304,8 @@ define(function (require, exports) {
     };
     initInactiveDocuments.action = {
         reads: [locks.PS_DOC],
-        writes: [locks.JS_DOC]
+        writes: [locks.JS_DOC],
+        private: true
```
iwehrman: :white_check_mark:

iwehrman:
Nice start!

pineapplespatula:
@iwehrman  :repeat_one:

src/js/fluxcontroller.js
```diff
@@ -202,6 +219,32 @@ define(function (require, exports, module) {
     });
 
     /**
+     * Initialize all action functions 
+     *
+     * @private
+     */
+    FluxController.prototype._initAllActions = function () {
+        this._allActions = {};
```
iwehrman: This should be a `Map`.

src/js/fluxcontroller.js
```diff
@@ -202,6 +219,32 @@ define(function (require, exports, module) {
...
+        this._actionModules = new Map();
```
iwehrman: I think this should be called `_actionModuleNames`? It maps actions to action module names, right? The JSDoc above should be clarified too.

iwehrman:
Is this supposed to be review-only, @pineapplespatula?

pineapplespatula:
since its such a broad change i would want a QE pass before it gets merged. if there is a better flag for that let me know

iwehrman:
Review-only is great, but you should assign it to @chadrolfs or @ktaki. Also, you might as well mark everything else as private now that should be private?

pineapplespatula:
@ktaki  can you give this branch a once over to see if i broke anything unexpected

ktaki:
@pineapplespatula Do you have any suggestion how your changes be tested? Any particular areas of concerns?

iwehrman:
@ktaki @pineapplespatula: I don't think this would cause any subtle bugs. If it seems to be working then it's probably working. I suggest you annotate the other private actions and we get on with it.

ktaki:
@pineapplespatula: You told me to hold testing while you were making some more changes. Let me know if you've done with it.

pineapplespatula:
Closing this PR

iwehrman:
Out of curiosity, why did you decide to close this instead of finishing it up?

pineapplespatula:
There aren't really any truly "private" actions, but there are a number of actions that should be "protected" however the amount of cleanup around the actions to stop exporting these functions which are not actions seemed a larger scope than i wanted to bite off for something that doesn't directly add to our goals

PR #3250, opened 2015-11-16
Move all CSS for different color stops into 1 file

designedbyscience:
Instead of loading different CSS files for different stops, now all component CSS is scoped under a class on the body tag, which is applied at load based on the PS color stop.

### DIFF
--- a/src/index-build.html
+++ b/src/index-build.html
@@ -25,7 +25,7 @@
 <head>
     <meta charset="utf-8">
     <title>Photoshop Design Space (optimized build)</title>
-    <link rel="stylesheet/less" type="text/css" href="style/main.less" />
+    <link rel="stylesheet/less" type="text/css" href="style/main.css" />
     <script src="js/require.js"></script>
     <script>
         var locale = window.navigator.language,
@@ -55,7 +55,7 @@
             }
         });
 
-        require(["js/init"]);
+        require(["js/init-build"]);
     </script>
 </head>
 <body>
--- a/src/index-build.html
+++ b/src/index-build.html
@@ -25,7 +25,7 @@
 <head>
     <meta charset="utf-8">
     <title>Photoshop Design Space (optimized build)</title>
-    <link rel="stylesheet/less" type="text/css" href="style/main.css" />
+    <link rel="stylesheet" type="text/css" href="style/main.css" />
     <script src="js/require.js"></script>
     <script>
         var locale = window.navigator.language,
--- a/src/index-build.html
+++ b/src/index-build.html
@@ -25,6 +25,7 @@
 <head>
     <meta charset="utf-8">
     <title>Photoshop Design Space (optimized build)</title>
+    <link rel="stylesheet/less" type="text/css" href="style/main.less" />
     <script src="js/require.js"></script>
     <script>
         var locale = window.navigator.language,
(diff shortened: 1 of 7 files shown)


src/index-build.html
```diff
@@ -54,7 +55,7 @@
             }
         });
 
-        require(["js/init-build"]);
+        require(["js/init"]);
```
volfied: Do we need to fix the `paths` in require.config?

volfied:
Thank you for making the less/CSS changes @designedbyscience, the PR needs a bit more work to get it working correctly for compile builds. But since you graciously offered your help with less, I'll take care of the rest.

designedbyscience:
Ah sorry, didn't even look at the config...

volfied:
@iwehrman Can you review our changes? I've gotten it to work under release as well. 

re@init-debug.js > init.js: These will be cleaned up nicer with webpack changes.

iwehrman:
This currently has failing CI checks.

iwehrman:
```
    Running "jshint:all" (jshint) task
       src/js/actions/layers.js
       1068 |            actionName = strings.ACTIONS.RENAME_LAYER,
                                      ^ 'strings' is not defined.
       1571 |            actionName = locked ? strings.ACTIONS.LOCK_LAYER : strings.ACTIONS.UNLOCK_LAYER,
                                               ^ 'strings' is not defined.
       1571 |            actionName = locked ? strings.ACTIONS.LOCK_LAYER : strings.ACTIONS.UNLOCK_LAYER,
                                                                            ^ 'strings' is not defined.
       2238 |                        name: strings.ACTIONS.CREATE_ARTBOARD
                                           ^ 'strings' is not defined.
```

volfied:
https://github.com/adobe-photoshop/spaces-design/blob/designedbyscience/the-one-less-file/src/js/actions/layers.js#L1068

volfied:
Travis is tripping.

src/index-build.html
```diff
@@ -25,6 +25,7 @@
 <head>
     <meta charset="utf-8">
     <title>Photoshop Design Space (optimized build)</title>
+    <link rel="stylesheet/less" type="text/css" href="style/main.css" />
```
iwehrman: Is the `rel=stylesheet/less` part correct here?

volfied:
@iwehrman Good catch, I had it changed locally on the built file but forgot it for the source file.

src/js/init.js
```diff
@@ -29,11 +29,6 @@ define(function (require) {
     var ui = require("./util/ui"),
```
iwehrman: I find this renaming to be more confusing than it was before, but I guess it's all going away soon...

src/style/main.less
```diff
@@ -281,3 +151,32 @@ input {
   html { font-size: @base-11; }
 
 }
+
+/* Setup Color Stops */
+.light {
```
iwehrman: I think all these class names need to be prefixed. Something like `.color-stop__light`, `color-stop__original`, etc.

iwehrman:
Back to @volfied / @designedbyscience for the class-name prefixing. Looks good otherwise! Thanks for fixing my superhack!

iwehrman:
Let's give it a whirl!

Issue #3256, opened 2015-11-17
Error opening Search dialog


ktaki:
With the Pg-Dev Mac build #886, clicking on search icon produces errors.

Transfer from search.toggleSearchBar to dialog.openDialog failed: Error: Translation not found for: menu.WINDOW.OPEN_DOCUMENT

![screen shot 2015-11-17 at 10 40 30 am](https://cloud.githubusercontent.com/assets/2029597/11221182/8d793cf8-8d18-11e5-9b26-db23bf70507f.png)

iwehrman:
After the fix for this in #3261, there are now a bunch of string keys that just map N -> N and I can't figure out why. This seems like useless cruft. Back to @pineapplespatula to clean it up.

volfied:
Back to @iwehrman to close

PR #3261, opened 2015-11-17
add some missing strings

pineapplespatula:
fix for issue #3256


iwehrman:
What is the point of these strings? Why are we mapping N -> N in the localization dictionary? This doesn't make sense. Now the loc team is "translating" them. I think this was probably the wrong fix for this issue?

PR #3262, opened 2015-11-17
Remove a hack around photoshop placeEvent during multiple object placement

mcilroyc:
This is related to watson 4074678 which is fixed as of playground-dev-mac 885.  When dragging two files from the OS into photoshop, previously only one of the subsequent `placeEvent` events would contain the layer ID.  We had a hack in place to fall back on updateDocument in this case. Now the layerID should be included by photoshop, so this PR removes the hack.

Note that in debug mode with post condition validation enabled, placing multiple objects will fail validation of the first N-1 events.  This is a false positive that can be ignored.


volfied:
Brilliant.

PR #3263, opened 2015-11-17
Webpack!

volfied:
Well, it's here... It touches a lot of files, but majority of those are ditching the prefixes and cleaning up src/nls.

You can read a guide of changes made at https://github.com/adobe-photoshop/spaces-design/wiki/Webpack-Dev-Guide

I've commented on any non-config / clean up change below.

---

Tested and approved by:
- [x] @shaoshing 
- [x] @mcilroyc 
- [x] @pineapplespatula 
- [x] @designedbyscience
- [x] @DivyaPrabhakar 
- [x] @placegraphichere 
- [x] @chadrolfs
- [x] @ktaki 
- [x] @iwehrman


.bowerrc
```diff
@@ -0,0 +1,3 @@
+{
+	"directory": "./bower_components"
```
volfied: For some reason, bower install, for me, installs everything at `src/bower_components`. This helps cement it at the right place, and we can remove this file once we move away from bower.

.gitignore
```diff
@@ -3,5 +3,5 @@ build/
 *.~
 bower_components
 node_modules
-src/font/*.otf
+src/vendor
```
volfied: Stale ignore updated

manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-    "start_url": "./src/index.html",
+    "start_url": "./build/index.html",
```
volfied: This is important, as we need to load the right index.html file.

src/index.html
```diff
@@ -25,23 +25,15 @@
...
-    <script src="js/config.js"></script>
-    <script type="text/javascript">
-        require.config({
-            config: {
-                i18n: {
-                    locale: window.navigator.language
-                }
-            }
-        });
+      var locale = window.navigator.language,
```
volfied: PG_DEBUG is passed in by webpack at compile time.

src/js/fluxcontroller.js
```diff
@@ -610,7 +610,7 @@ define(function (require, exports, module) {
             .bind(this)
             .then(function () {
                 var actionPromise = action.apply(actionReceiver, params);
-                if (!(actionPromise instanceof Promise)) {
+                if (!(actionPromise.then)) {
```
volfied: This is not easily fixable with a polyfill, but `bluebird.Promise` is confused by browser's `Promise`, so it's better to check for presence of `then` function (and confirm it's a thenable)

src/js/init.js
```diff
@@ -29,6 +29,8 @@ define(function (require) {
     var ui = require("./util/ui"),
         main = require("./main");
 
+    require("../style/main.less");
```
volfied: This allows webpack to parse less files and bundle them in.

src/style/main.less
```diff
@@ -52,35 +52,35 @@ input {
 
 @font-face {
   font-family: 'Adobe Clean';
-  src: url('file://shared/font/AdobeClean-Regular.otf');
+  src: url('~file://shared/font/AdobeClean-Regular.otf');
```
volfied: `~` allows less loader to parse these requires as absolute paths and leave them untouched.

src/js/util/global.js
```diff
@@ -26,11 +26,12 @@ define(function (require, exports) {
 
     /**
      * Indicates whether the application is running in debug mode.
+     * __PG_DEBUG__ is defined through webpack
      * 
      * @const
      * @type {boolean} 
      */
-    var DEBUG = !!window.__PG_DEBUG__;
+    var DEBUG = !!__PG_DEBUG__; // jshint ignore:line
```
iwehrman: You should be able to declare this as a JSHint global instead of just ignoring the whole line. http://jshint.com/docs/#inline-configuration

shaoshing:
giphy

shaoshing:
Excellent work! This is an important step for us toward ES6. I also tried the `--watch` (BTW `grunt compile --dev --watch ` does not work for me, I have to switch positons of the two params to get the live compiling feature working) option to get a feel of the compile performance and noticed that changing less files does not trigger a rebuild. I think that's probably because less files are built in a separate grunt task. Do we have any plan to add the less loader so that we don't have to manually rebuild the less files (I cannot find it in your TODOs)?

webpack.config.js
```diff
@@ -0,0 +1,131 @@
...
+require("es6-promise").polyfill(); // Required for css loading
+
+// If we call webpack with --dev, we only compile English language,
+// and create sourceMaps
+var devMode = process.argv.some(function (value) {
+    return value === "--dev";
+});
+
+// TODO: Similar to gruntfile, find a way to make this dynamic
+var languages = devMode ? ["en"] : ["en", "de", "fr", "ja"];
```
shaoshing: maybe store the `ALL_LOCALES` in system env and then we can access it from `process.env.ALL_LOCALES`

volfied: Grunt automatically grabs the locale list by folder names, so I need to figure out a way to pass it in to webpack config as well.

volfied:
@shaoshing I've made changes so webpack picks up locales from the Grunt process. I've also added a safety check to make sure we don't use webpack CLI to build, since we need Grunt for resource files.

volfied:
@iwehrman Pushed changes so `grunt dev` and `grunt compile` are two separate tasks.

`grunt dev` builds sourcemaps, only English, and keeps watchers alive
`grunt compile` builds all languages, uglifies them, and no sourcemaps

iwehrman:
Previously we had `grunt compile` distinct from `grunt build`, where the latter was equal to `grunt compile` + `grunt build`. Can we have that back? I don't see any reason to change that.

designedbyscience:
Works. AMD, we hardly knew ye

Gruntfile.js
```diff
@@ -26,36 +26,21 @@
...
-                useStrict: true,
-                config: {
-                    i18n: {
-                        locale: locale
-                    }
-                }
-            }
-        };
-    };
+    if (grunt.cli.tasks[0] === "dev") {
```
volfied: There is a bug here where if you type anything but `grunt dev`, webpack tries to build all locales. I'm looking into it. For now, avoid using anything but `grunt compile` and `grunt dev`

volfied:
@pineapplespatula , @ktaki, @iwehrman  and @placegraphichere, what will it take to get your guys' checks on the list in the top comment?

volfied:
With 157854b, this change will no longer require us to copy www-shared into src/vendor. Hooray!

volfied:
With 735fa91, dev builds will break our dependencies into their own chunk, so incremental rebuilds will only rebuild our own source files. This takes down incremental rebuild from 5 seconds to 2 seconds.

placegraphichere:
Pretty sure I don't fully understand all that I'll need to change in my day to day activities. But I know it will be a few things. Can we run through those things at some point? Otherwise, hooray!

iwehrman:
The longer we wait, the better this PR gets :grimacing:

pineapplespatula:
Can i debug remotely with files?

iwehrman:
@pineapplespatula No. Due to a CEF bug, you have to use the in-proc dev tools to use source maps. See here: https://bitbucket.org/chromiumembedded/cef/issues/1635/source-maps-broken-in-remote-devtools

pineapplespatula:
Is In-proc debugging ready for us to do full time development with it? Otherwise this seems pretty blocked to me.

pineapplespatula:
With @jsbache recent changes to in-proc debugging I think this is unblocked .(there are still issues, but they no longer blocking this PR in my opinion)

volfied:
8719345 Exposes some of our modules for snippets. I've duplicated the snippet page with essential debugging snippets for webpack usage: https://github.com/adobe-photoshop/spaces-design/wiki/Dev-Tool-Snippets-(for-Webpack)

volfied:
I just realized that every time I rebase, the commit numbers change. Oops

.gitignore
iwehrman: I think this is no longer necessary?

Gruntfile.js
```diff
@@ -67,7 +52,8 @@ module.exports = function (grunt) {
                 "src/**/*.js",
                 "src/**/*.jsx",
                 "test/**/*.js",
-                "test/**/*.jsx"
+                "test/**/*.jsx",
+                "!src/vendor/**/*"
```
iwehrman: I think this is also stale.

Gruntfile.js
```diff
@@ -76,7 +62,8 @@ module.exports = function (grunt) {
                     "*.js",
                     "*.json",
                     "src/**/*.js",
-                    "src/**/*.jsx"
+                    "src/**/*.jsx",
+                    "!src/vendor/**/*"
```
iwehrman: Ditto.

Gruntfile.js
```diff
@@ -106,7 +93,8 @@ module.exports = function (grunt) {
             src: [
                 "*.json",
                 "src/**/*.json",
-                "test/**/*.json"
+                "test/**/*.json",
+                "!src/vendor/**/*"
```
iwehrman: And so forth.

src/js/actions/libraries.js
```diff
@@ -1239,59 +1247,69 @@ define(function (require, exports) {
...
-                            // updateType: info (rename), representation (update content)
-                            headlights.logEvent("libraries", "element",
-                                ["update", elementType, properties.updateType].join("-"));
-                            break;
+        apiLoadPromise.bind(this)
+            .then(function () {
+                /* global ccLibraries */
+                // There is now a ccLibraries object in this scope
+                // So we emit that to the store
+                this.dispatch(events.libraries.LIBRARIES_API_LOADED, ccLibraries);
```
iwehrman: I don't really understand the point of putting this in the store...

src/js/actions/libraries.js
```diff
@@ -1239,59 +1247,69 @@ define(function (require, exports) {
...
-                            break;
-                        case "deleteElement":
-                            headlights.logEvent("libraries", "element", ["delete", elementType].join("-"));
-                            break;
-                        case "updateElement":
-                            // updateType: info (rename), representation (update content)
-                            headlights.logEvent("libraries", "element",
-                                ["update", elementType, properties.updateType].join("-"));
-                            break;
+        apiLoadPromise.bind(this)
```
iwehrman: We usually align all this methods on their own new line.

src/js/actions/libraries.js
iwehrman: Oh, I see that it's accessed in other modules. I guess this makes sense, then.

src/js/main.js
```diff
@@ -162,6 +162,16 @@ define(function (require, exports) {
...
         _spaces._debug.enableDebugContextMenu(true, function () {});
+
+        /* global __PG_DEBUG__ */
+        if (__PG_DEBUG__) {
+            // Expose these for snippet usage, only available in debug builds
+            window.__ADAPTER_DESCRIPTOR__ = adapter.ps.descriptor;
+            window.__ADAPTER_OS__ = adapter.os;
+            window.__ADAPTER_UI__ = adapter.ps.ui;
+            window.__FLUX_CONTROLLER__ = _controller;
+            window.__LOG_UTIL__ = log;
```
iwehrman: Can we also expose `js/util/performance`? Also, do we need different variables for all the adapter objects? Can we just expose `adapter`?

src/js/stores/library.js
```diff
@@ -48,6 +47,11 @@ define(function (require, exports, module) {
     var LibraryStore = Fluxxor.createStore({
 
         /**
+         * @type {Object}
```
iwehrman: Should be `?object` I think. (Lower-case, at the very least.)

src/js/stores/library.js
```diff
@@ -545,6 +565,15 @@ define(function (require, exports, module) {
          */
         generateTempFilename: function (element) {
             return element.name.replace(_TRIM_FILE_NAME_REGEXP, "") + "@" + element.id + ".psd";
+        },
+
+        /**
+         * Returns the loaded libraries API
+         *
+         * @return {Object} CCLibraries API global object
```
iwehrman: Should be `object`.

src/js/util/global.js
```diff
@@ -20,17 +20,18 @@
  * DEALINGS IN THE SOFTWARE.
  * 
  */
-
+/* global __PG_DEBUG__ */
```
iwehrman: Can we put this into the JSCS or JSHint config instead of muddying the files?

iwehrman:
@volfied: should we consider forcing the edit menu actions to be enabled when in debug mode to work around the bug @dtristram mentioned with the in-proc dev tools?

iwehrman:
Looking good so far. Back to @volfied with a few questions and nits.

iwehrman:
Also, I think we should have a grunt task which starts the watchers without going through all the linting steps. In fact, you might just want to remove the linting steps from the existing grunt task.

iwehrman:
Also, I still think that grunt task names should be verbs, and `grunt dev` is vague. If it were me, I would call it `grunt watch`.

iwehrman:
Another suggestion: since you're splitting the external dependencies out of the main bundle, you could omit the source maps for those. That might speed things up.

src/index-debug.html
```diff
@@ -24,38 +24,17 @@
...
-    <title>Photoshop Design Space (optimized build)</title>
-    <link rel="stylesheet" type="text/css" href="style/main.css" />
-    <script src="js/require.js"></script>
-    <script>
-        var locale = window.navigator.language,
-            parts = locale.match(/\w+/),
-            language = parts.length > 0 && parts[0];
+    <title>Photoshop Design Space (development build)</title>
+    <link rel="stylesheet" type="text/css" href="style.css" />
+    <script type ="text/javascript" src="./externalDeps.js"></script>
```
iwehrman: This is bad because it's loading one bundle after another, which is slower than just loading everything at once. Look at this profile run:
image
It takes over 900ms before we start executing our own JavaScript. This seems like it must be a performance regression compared to using the sources compiled by r.js...

iwehrman: Also, I don't understand the motivation behind splitting out the external dependencies into a separate bundle.

volfied: I'll see what I can do in terms of this. My logic was that this is only on dev builds, and it saves more than a second on re-compile time.

iwehrman: Well, it does seem like there is a minor performance regression here, but I guess it isn't as bad as I first thought, since it seems like with the r.js-compiled sources it also takes 850ms or so before we start running our code:
image

iwehrman:
@volfied: would you mind reporting some basic startup benchmark numbers comparing webpack to the r.js-compiled builds?

volfied:
@iwehrman Not sure if I got this right, but here are the r.js results:

[](http://cl.ly/image/370h463a3g3E)
[](http://cl.ly/image/1C2U0A0m1y04)

And here are the webpack results:

[](http://cl.ly/image/0b2Y0k193W3C)
[](http://cl.ly/image/3L3t432y0p1m)

volfied:
Latest changes:

 - Instead of `grunt dev`, now it's `grunt debug`, and it doesn't lint your code. That's up to the dev to run `grunt` to make sure their code is clean.
 - compiled nls files get put in build/nls. This helps with git workflow as files in src/nls were troublesome and required special cases in config files

volfied:
@iwehrman @ktaki, please let me know if there are any more issues you'd like to see fixed before this lands. This has been one fun plane to fly

ktaki:
Checked the checkbox above.

iwehrman:
I think we're basically good to go here. I'd like to merge this after we update Perforce on Friday. If you have any other concerns, speak now!

volfied:
Rebased, again. Also added descriptions to Grunt tasks.

iwehrman:
Everybody hold on to your butts.

Issue #1802, opened 2015-07-22
Simplify layer-kind checks


shaoshing:
We should add a helper function to Layer model for checking layer kind. This will avoid [redundant code](https://github.com/adobe-photoshop/spaces-design/search?utf8=%E2%9C%93&q=layerKinds) like `layer.kind !== layer.layerKinds.GROUPEND`. Instead, it may be improved by using `layer.is("GROUPEND")`

iwehrman:
I support both of these (in principle, at least), but I suggest slicing off one the two into a separate issue since they're completely independent.

iwehrman:
I factored out this unrelated request into #1935:

> It is unlikely to have multiple strokes and fills in the Layer model, so its `fills` and `strokes`
> properties should not be List type.

shaoshing:
Fixed by #3273

PR #3273, opened 2015-11-18
Refactor Layer kinds

shaoshing:
##### Unified Layer kind enumerable 

Before we used three different ways of accessing the layer kinds list:

* Layer.layerKinds
* layer.layerKinds (Layer instance)
* layerLib.layerKinds (direct access from the adapter)

Now the unified way is

* Layer.KINDS

##### Use string value for `Layer#kind`, instead of number.

For example, `layer.kind` will return `GROUP`, instead of `7`. Here are some benefits

* easier for debugging
* easier to use meaningful names in CSS / icon files and translations

##### Added new properties for layer kind checking (addressed #1802)

Layer has the following new properties:

* isText (replaced Layer#isTextLayer())
* isVector (replaced Layer#isVector())
* isSmartObject (replaced Layer#isSmartObject())
* isPixel
* isAdjustment
* isGroup
* isGroupEnd
* is3D
* isVideo


src/js/models/layer.js
```diff
@@ -469,6 +559,26 @@ define(function (require, exports, module) {
...
     };
+    
+    /**
+     * Return the capitalized layer kind name. For example: SMARTOBJECT -> SmartObject
+     *
+     * @private
+     * @param {LayerKind} layerKind
+     * @return {string}
+     */
+    var _capitalizeLayerKindName = function (layerKind) {
```
volfied: This function should be called camelCaseLayerKindName, as you're camelCasing it.

src/js/models/layer.js
```diff
@@ -469,6 +559,26 @@ define(function (require, exports, module) {
...
+        switch (layerKind) {
+            case Layer.KINDS.SMARTOBJECT:
+                return "SmartObject";
+            case Layer.KINDS.GROUPEND:
+                return "GroupEnd";
+            case Layer.KINDS.SOLIDCOLOR:
+                return "SolidColor";
+            default:
+                return _.capitalize(layerKind.toLowerCase());
```
volfied: https://lodash.com/docs#camelCase

src/js/models/layer.js
```diff
@@ -484,32 +594,37 @@ define(function (require, exports, module) {
...
-            proportionalScaling: false,
-            isArtboard: !!isArtboard,
-            bounds: isArtboard ? new Bounds(boundsDescriptor) : null,
-            isLinked: false,
-            vectorMaskEnabled: false,
-            vectorMaskEmpty: true,
-            textWarningLevel: 0,
-            initialized: true
-        });
+        var Bounds = require("./bounds"),
```
volfied: Please move this require to the top of the file.

src/js/models/layer.js
```diff
@@ -537,34 +652,38 @@ define(function (require, exports, module) {
...
-            vectorMaskEnabled: layerDescriptor.vectorMaskEnabled,
-            exportEnabled: exportEnabled,
-            vectorMaskEmpty: layerDescriptor.vectorMaskEmpty,
-            textWarningLevel: layerDescriptor.textWarningLevel,
-            isLinked: _extractIsLinked(layerDescriptor),
-            initialized: initialized || selected
-            // if not explicitly marked as initialized, then it is initialized iff it is selected
-        };
+        var Bounds = require("./bounds"),
+            Fill = require("./fill"),
```
volfied: Ditto

src/js/models/layer.js
volfied: I'll give you the benefit of the doubt, is there a reason these are in function scope?

src/js/models/layer.js
```diff
@@ -711,6 +808,6 @@ define(function (require, exports, module) {
 
         return nextLayer;
     };
-
+
```
volfied: :scream:

volfied:
Looks like it requires rebase. I also have some questions/comments.

Great job with the clean up though, it's so much clearer now.

src/js/models/layer.js
shaoshing: We actually need capitalized (GroupEnd) instead of camelCased (groupEnd) string here, because it is concatenated with "is" later.

src/js/models/layer.js
shaoshing: Sorry, I forgot to leave a comment on this. This is for resolving the circular dependencies between Layer and Bounds (also Fill). If it is put on top, `Bounds` will be an empty object as it is also requiring `Layer`. This is one of the recommended ways of dealing the situation (http://requirejs.org/docs/api.html#circular). 

Another way that will allow us putting the require statement on top, is exposing the model via the `export` object, just like what we have for actions. For example:

```
// bounds.js
var Layer = require("layer");
...
exports.Model = Bounds.

// layer.js
var Bounds = require("bounds");  // Bounds refers to the exports object, so change in the bounds.js also reflects in here.

function createLayer() {
   ...
   var bounds = new Bounds.Model();
}
```

NOTE: ES6 package has the `export default ...` feature that will solve this nice, but until we are ready for it, we may have keep using the require inside function.

shaoshing: I am wonder if this will break webpack?

volfied:
https://www.youtube.com/watch?v=iwFKSl-3dPc

src/js/models/layer.js
volfied: https://www.youtube.com/watch?v=iwFKSl-3dPc

DivyaPrabhakar: @volfied LOL

shaoshing:
@volfied your comments and questions have been addressed. I also merged your branch to fix the jsdoc errors (thanks for the PR). Now back to you.

src/js/models/layer.js
shaoshing: All right sir!

volfied:
Let's do this.

Issue #3231, opened 2015-11-12
Guide positions should be loaded lazily


iwehrman:
We currently load guide position information for each document even when guides are disabled. Instead, we should load this information lazily when guides are first shown.

iwehrman:
Fixed by #3274.

PR #3274, opened 2015-11-18
Load guides lazily if they're not initially visible

iwehrman:
This PR delays loading guide information until the guides are first made visible. This is a minor performance optimization. Addresses #3231.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -191,16 +191,22 @@ define(function (require, exports) {
      * Get an array of guide descriptors for the given document descriptor.
      *
      * @private
-     * @param {object} doc Document descriptor
+     * @param {object|number} docSpec Document descriptor or document ID
      * @return {Promise.<Array.<object>>}
      */
-    var _getGuidesForDocument = function (doc) {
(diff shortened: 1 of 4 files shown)


src/js/actions/documents.js
```diff
@@ -328,15 +334,19 @@ define(function (require, exports) {
                     return;
                 }
 
-                var currentRef = documentLib.referenceBy.current;
-                return _getDocumentByRef(currentRef)
+                var currentRef = documentLib.referenceBy.current,
+                    documentPromise = _getDocumentByRef(currentRef),
+                    deselectPromise = descriptor.playObject(selectionLib.deselectAll());
```
volfied: Any reason why you moved this outside?

volfied:
Looks good as well, but curious about reasoning behind a code move @iwehrman

src/js/actions/documents.js
iwehrman: Because it is safe to do so AFAICT, and moving it up could prevent delays. If we wait and then submit that command it's worse than submitting it early to PS for parallel execution.

iwehrman: But, yes, it's unrelated.

PR #3275, opened 2015-11-18
Fix JSDoc errors due to jsdoc update

volfied:
:sparkles:


iwehrman:
Weird that JSDoc wasn't flagging these to begin with.

PR #3283, opened 2015-11-19
Prevent numerical menu labels from being localized.

pineapplespatula:
fix for issue #3256


DivyaPrabhakar:
- This is failing CI tests
- I checked out this branch and Menu Commands have no search results now

pineapplespatula:
this is rebased now. I dont see any difference between master and this branch in functionality.

volfied:
Couple weeks/months down the line, one of us is going to look at this commit message and say WTF, and will have to read the code change to see what changed. It would help everyone if the commit message was a bit more descriptive of what it accomplished:

> Prevent numerical menu labels from being localized.

src/js/actions/search/commands.js
```diff
@@ -50,18 +50,20 @@ define(function (require, exports) {
             menuTree;
 
         parts.forEach(function (part) {
-            nlsPath = nlsPath + "." + part;
-            menuTree = nls.localize(nlsPath);
+            if (!_.isNumber(parseInt(part, 10))) {
```
volfied: There is no `else` statement here, do we not need to append the number itself to the resultPath at the end?

pineapplespatula: I'm not sure, it doesn't seem like it should matter but also it wont hurt

iwehrman:
In whose court is the proverbial ball here? Looks like the last commit is 10 days old...

PR #3285, opened 2015-11-19
Suppress layer panel scroll-into-view on click

iwehrman:
This is a minor performance optimization. It eliminates one [forced synchronous layout](https://developers.google.com/web/tools/chrome-devtools/profile/rendering-tools/analyze-runtime#how-to-identify-layout-bottlenecks) when clicking in the layers panel by suppressing the scroll-into-view behavior, which is unnecessary in that case. The Chrome timeline seems to indicate that this saves a reliable 5ms of layout on each click. This isn't really noticeable, but maybe it's educational? 

Before (some spent in layout):
image

After (no time spent in layout):
image

### DIFF
--- a/src/js/jsx/sections/layers/LayersPanel.jsx
+++ b/src/js/jsx/sections/layers/LayersPanel.jsx
@@ -105,9 +105,32 @@ define(function (require, exports, module) {
          */
         _lastScrolledTo: Immutable.List(),
 
+        /**
+         * Used to suppress scrolling into view when the selection is changed
+         * by clicking directly on a layer face.
+         *
+         * @private
+         * @type {boolean}
+         */
+        _suppressNextScrollTo: false,
+
         componentWillMount: function () {
             this._setTooltipThrottled = synchronization.throttle(os.setTooltip, os, 500);
             this._mountedLayerIDs = new Set();
+        },
+
+        /**
+         * A capture-phase click handler for the container. Used to suppress
+         * scrolling into view when clicking on the layers panel directly.
+         *
+         * @private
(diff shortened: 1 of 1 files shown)


src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -105,13 +105,44 @@ define(function (require, exports, module) {
...
+            if (event.target !== containerNode) {
+                this._suppressNextScrollTo = true;
+            }
+        },
+
         componentDidMount: function () {
             this._scrollToSelection(this.props.document.layers);
+
+            var containerNode = React.findDOMNode(this.refs.container);
+            containerNode.addEventListener("click", this._handleContainerClickCapture, true);
```
designedbyscience: Why can't do you do this in the JSX?

iwehrman: The `LayerFace` click handler calls `event.stopPropagation()`, so this has to be attached at the capture phase instead of the bubbling phase. I'm not aware of a way to attach capture-phase handlers in JSX. Is there one?

designedbyscience: Sí: https://facebook.github.io/react/docs/events.html#supported-events

iwehrman: Well sonofagun! There are probably a few other places where I've done this as well in my ignorance... Anyway, I'll fix it up, thanks!

iwehrman:
Commit amended to use React's built-in capture-phase event handler support.

PR #3288, opened 2015-11-19
beforeStartup/afterStartup cleanup

iwehrman:
What it says! The UI is perfectly usable before these are registered, so it doesn't seem important to delay rendering it until they're registered. Seems to shave 50-100ms off the time-to-UI.

===
Actually, this turned into a more general `beforeStartup`-to-`afterStartup` cleanup, as well as a cleanup of `tools.initTool` and `selectTool`. According to my measurements, this reduces the time to a fully ready to UI by ~700ms.

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -34,7 +34,6 @@ define(function (require, exports) {
         documentLib = require("adapter").lib.document,
         adapterOS = require("adapter").os,
         adapterUI = require("adapter").ps.ui,
-        adapterPS = require("adapter").ps,
         UI = require("adapter").ps.ui,
         OS = require("adapter").os,
         vectorMaskLib = require("adapter").lib.vectorMask;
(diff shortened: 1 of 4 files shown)


src/js/actions/tools.js
```diff
@@ -420,58 +419,61 @@ define(function (require, exports) {
     var selectTool = function (nextTool) {
```
iwehrman: The logic is the same, but I reduced unnecessary serialization here.

src/js/actions/tools.js
```diff
@@ -486,8 +488,7 @@ define(function (require, exports) {
             "toolRectangle.select",
             "toolSampler.select", "toolSampler.deselect",
             "toolType.select", "toolType.deselect"
-        ],
-        modal: true
```
iwehrman: Also, this used to be marked modal-safe, but also started by calling `endModalToolState`. But that's already the default behavior, so I removed both.

src/js/actions/tools.js
```diff
@@ -512,14 +513,6 @@ define(function (require, exports) {
                 }
                 
                 return this.transfer(selectTool, tool);
-            })
-            .catch(function (err) {
```
iwehrman: There's no point in trying to recover if we fail to select an inferred tool above.

src/js/actions/ui.js
```diff
@@ -654,25 +654,18 @@ define(function (require, exports) {
         // Enable target path suppression
         var pathPromise = adapterUI.setSuppressTargetPaths(false);
 
-        // Add additional shortcut CMD=, so that CMD+ and CMD= both work for zoom in.
```
iwehrman: Moved to `afterStartup`.

src/js/actions/tools.js
```diff
@@ -948,16 +958,13 @@ define(function (require, exports) {
             name: "Super Select Vector"
         });
 
-        var shortcutsPromise = this.transfer(shortcuts.addShortcuts, shortcutSpecs),
```
iwehrman: Moved to `afterStartup`.

volfied:
I see nothing wrong, nice separation.

Issue #3290, opened 2015-11-20
Export metadata should be loaded lazily


iwehrman:
Currently helper functions like `_getLayersForDocument` load "extension" metadata for _all_ layers. Instead, it should only load this metadata for selected layers, and `initializeLayers` (or something similar) should be used to initialize this data when the layer is first selected.

CC @mcilroyc

PR #3292, opened 2015-11-20
Load layer metadata information lazily 

iwehrman:
1. Load layer metadata lazily, so that when initializing a document metadata is only requested for selected layers. As with other lazy properties, metadata is loaded on-demand by `initializeLayers`/`resetLayers`. Addresses #3290. 
2. Respect the `lazy` parameter in `resetLayers` when choosing whether or not to load layer metadata when resetting layers.
3. Correctly reset/update layer export information as a result of `resetLayers`. Previously, updated export information that resulted from `resetLayers` was ignored.

This provides a modest document load-time speedup. For Vermilion, it seems to save a small few hundred milliseconds.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -135,6 +135,14 @@ define(function (require, exports) {
         .concat(_optionalLayerProperties);
 
     /**
+     * Namespace for extension metadata.
+     * 
+     * @const
+     * @type {string}
+     */
+    var METADATA_NAMESPACE = global.EXTENSION_DATA_NAMESPACE;
+
+    /**
      * Get layer descriptors for the given layer references. Only the
      * properties listed in the arrays above will be included for performance
      * reasons. NOTE: All layer references must reference the same document.
@@ -151,15 +159,20 @@ define(function (require, exports) {
             optionalPropertiesPromise = descriptor.batchMultiGetProperties(references, optionalProperties,
                 { continueOnError: true });
 
-        var nameSpace = global.EXTENSION_DATA_NAMESPACE,
-            extensionPlayObjects = references.map(function (ref) {
-                return layerLib.getExtensionData(ref[0], ref[1], nameSpace);
-            }),
+        var extensionPromise;
+        if (lazy) {
+            extensionPromise = Promise.resolve();
+        } else {
+            var extensionPlayObjects = references.map(function (ref) {
+                    return layerLib.getExtensionData(ref[0], ref[1], METADATA_NAMESPACE);
+                });
+
             extensionPromise = descriptor.batchPlayObjects(extensionPlayObjects)
                 .map(function (extensionData) {
-                    var extensionDataRoot = extensionData[nameSpace];
+                    var extensionDataRoot = extensionData[METADATA_NAMESPACE];
                     return (extensionDataRoot && extensionDataRoot.exportsMetadata) || {};
                 });
+        }
 
         return Promise.join(layerPropertiesPromise, optionalPropertiesPromise, extensionPromise,
             function (required, optional, extension) {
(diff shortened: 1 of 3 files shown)


src/js/actions/layers.js
```diff
@@ -206,50 +219,47 @@ define(function (require, exports) {
...
-        // And it is safe to ignore ALL bg layers because we don't use extension data on them
-        var nameSpace = global.EXTENSION_DATA_NAMESPACE,
-            numberOfLayers = doc.hasBackgroundLayer ? doc.numberOfLayers + 1 : doc.numberOfLayers,
-            indexRange = _.range(1, startIndex + numberOfLayers),
-            extensionPlayObjects = indexRange.map(function (i) {
-                var layerRef = layerLib.referenceBy.index(i);
-                return layerLib.getExtensionData(docRef, layerRef, nameSpace);
-            }),
+        var extensionPromise;
+        if (doc.hasBackgroundLayer && doc.numberOfLayers === 0 && targetLayers.length === 1) {
```
iwehrman: I special-cased this differently than you did previously, @mcilroyc. The reason is that it's fairly tricky to merge the lazy and metadata properties subset into the full properties set. I think this is actually a good, fast way to do it, but if something looks off to you please let me know.

src/js/actions/layers.js
```diff
@@ -206,50 +219,47 @@ define(function (require, exports) {
...
-                return lazyPropertiesPromise.each(function (lazyProperties) {
-                    if (!lazyProperties) {
-                        // A background will not have a layer ID
-                        return;
-                    }
-                    
-                    var lazyLayerID = lazyProperties.layerID;
 
-                    _.merge(propertiesByID[lazyLayerID], lazyProperties);
+                return extensionPromise.then(function (allData) {
```
iwehrman: See "tricky merging" above.

src/js/models/documentexports.js
```diff
@@ -67,6 +67,44 @@ define(function (require, exports, module) {
     }));
 
     /**
+     * Generate List of ExportAssets based on an array of simple props objects
```
iwehrman: The following two functions were factored directly out of `fromDescriptors` below.

src/js/stores/export.js
```diff
@@ -181,6 +183,23 @@ define(function (require, exports, module) {
                 documentExports = DocumentExports.fromDescriptors(payload);
 
             this._documentExportsMap = this._documentExportsMap.set(documentID, documentExports);
+            this.emit("change");
```
iwehrman: This is an unrelated bugfix. @mcilroyc, you may wish to confirm that adding this change event doesn't have any ill effects.

src/js/actions/layers.js
```diff
@@ -151,15 +159,20 @@ define(function (require, exports) {
             optionalPropertiesPromise = descriptor.batchMultiGetProperties(references, optionalProperties,
                 { continueOnError: true });
 
-        var nameSpace = global.EXTENSION_DATA_NAMESPACE,
-            extensionPlayObjects = references.map(function (ref) {
-                return layerLib.getExtensionData(ref[0], ref[1], nameSpace);
-            }),
+        var extensionPromise;
+        if (lazy) {
+            extensionPromise = Promise.resolve();
```
iwehrman: To test this case, note that all text layers (including non-selected ones) are reset when the font list is first initialized, and that happens when a text layer is first selected.

src/js/actions/layers.js
```diff
@@ -206,50 +219,47 @@ define(function (require, exports) {
...
-                    return (extensionDataRoot && extensionDataRoot.exportsMetadata) || {};
-                })
-                .then(function (extensionDataArray) {
-                    // add an empty object associated with the background layer (see note above)
-                    if (startIndex === 0) {
-                        extensionDataArray.unshift({});
-                    }
-                    return extensionDataArray;
-                });
+                    .map(function (extensionData) {
```
mcilroyc: over-indented, methinks

src/js/actions/layers.js
mcilroyc: gah

src/js/actions/layers.js
iwehrman: You got me. Commit amended.

mcilroyc:
looks good

PR #3293, opened 2015-11-22
Mask Mode App Ui Design Work

placegraphichere:
- Reorg the document bar buttons with Mask first
- Highlight for mask mode more pronounced
- Additional work to follow
- string loc thing updated


pineapplespatula:
LGTM

PR #3294, opened 2015-11-23
Tap for newHistoryState in libraries.applyLayerStyle

volfied:
We needed the path coming from `representation.getContentPath` to be passed to the next handler. So now we tap to create a new history state instead of `then`ing to it.


volfied:
@mcilroyc I think this was your recent history changes that caused this. Is the `tap` safe here? Do we need the result from the transfer?

mcilroyc:
my bad.  that's safe.

PR #3295, opened 2015-11-23
Async load CCLibraries API using scriptjs

volfied:
This solves the problem of having to copy www-shared/libs into src/vendor/libs by using async loader https://github.com/ded/script.js/

We now load it at libraries.beforeStartup asynchronously, and pass it to the library store so it's available through runtime without requiring again.

Another solution for this would be to use scriptjs at init.js and keep a global object, but this seems cleaner.

This is a PR into architecture/webpack, but I wanted to get some eyes on it separately. @iwehrman and @shaoshing in particular.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,8 @@
     "mathjs": "~2.2.0",
     "react": "~0.13.3",
     "spaces-adapter": "https://github.com/adobe-photoshop/spaces-adapter.git#^0.27.0",
-    "tinycolor": "~1.2.1"
+    "tinycolor": "~1.2.1",
+    "scriptjs": "~2.5.8"
   },
   "devDependencies": {
     "qunit": "~1.19.0",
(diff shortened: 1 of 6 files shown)


iwehrman:
I suggest creating the script-loading promise once at module-initialization time, and then `.then`ing it in `beforeStartup`, `afterStartup`, etc. The reason is that we don't actually want to load the script multiple times (i.e., every time that there's an error and we call `beforeStartup` again).

volfied:
@iwehrman I've added `onInitialize` method as a possible lifetime method for actions, and implemented it in libraries.js

Thank you

volfied:
@iwehrman After our offline discussion, I've moved apiLoad to the outermost scope instead of mucking with another lifetime method.

volfied:
Since merging this into a waiting to merge branch will be a PITA later on while rebasing, with @iwehrman 's OK, I'm closing this PR and pushing these commits to `architecture/webpack` manually.

PR #3298, opened 2015-11-24
Color Picker and Color Swatch UI improvements

placegraphichere:
Color Picker sliders now housed entirely inside slider
- looks better, feels better, I think it is better
- Gradients have been adjusted to compensate for slider circle being inside the bar
- Opacity has been moved and tweaked to read inline with the other values and save space

Several Design Tweaks to Color swatch and Panels

- Swatch sizes are based around 8 grid (3.2rem instead of 3.4)
- Transparency background tiling improved (2x2 instead of off size)
- Alignment of colors, stroke, effects now go full-column so eyeball aligns with other left side icons (trash, download for example in other panels)
- Small change to box-shadow on the color swatch rounded corners to reduce white edges
- removed any 100th decimal point for rem sizes for reduced 1/2px possibilities

An incremental improvement to the UI 

---

new:
screen shot 2015-11-23 at 6 20 44 pm

---

Old
screen shot 2015-11-23 at 6 21 12 pm

### DIFF
--- a/src/js/jsx/sections/style/ColorOverlayList.jsx
+++ b/src/js/jsx/sections/style/ColorOverlayList.jsx
@@ -174,7 +174,7 @@ define(function (require, exports, module) {
 
             return (
                 <div className={colorOverlayClasses}>
-                    <div className="formline formline__no-padding">
+                    <div className="formline formline__no-padding formline__space-between">
                         <div className="control-group control-group__vertical">
                             <ColorInput
                                 id={colorInputID}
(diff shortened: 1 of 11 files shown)


mcilroyc:
looks good to my eye.  @placegraphichere you may want to rebase so that this branch doesn't throw a bunch of warnings (due to the bluebird upgrade, I think)

mcilroyc:
meh, I'll just merge it!

Issue #3276, opened 2015-11-18
Error on Clear Guides


ktaki:
Create a new doc, add a guide, and run View->Clear Guides.

Action guides.clearGuides failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -1 Photoshop message: error: -1 message: Could not complete the Clear All Guides command because of a program error.

![screen shot 2015-11-18 at 3 01 35 pm](https://cloud.githubusercontent.com/assets/2029597/11257297/85ddc4c6-8e05-11e5-9cfd-cfdfb6adf92b.png)

mcilroyc:
related(?): I get an error when *creating* the first guide on a new document.  Almost certainly related to #3274   - I'll investigate further.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3299, opened 2015-11-24
Handle case where first guide is added to a new document

mcilroyc:
This addresses #3276.  A new document will not have "guides" property (due to recent lazy loading change), so this handles that case when setting the first guide.  @iwehrman please review to make sure this is inline with how we should recover from lazy loading of guides.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -532,13 +532,16 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var allocateDocument = function (documentID) {
-        var updatePromise = this.transfer(updateDocument, documentID),
-            selectedDocumentPromise = _getSelectedDocumentID();
+        return _getSelectedDocumentID()
+            .bind(this)
+            .tap(function (currentDocumentID) {
+                // if this document is current, call updateDocument with an undefined documentID argument
+                // so that it will interpreted as an update of the current document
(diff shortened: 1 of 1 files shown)


iwehrman:
It's unclear why this is happening before guides are initialized. That is, I don't understand why we're running code to clear or set guides when guides are disabled. Or, are we not doing that, and guide initialization is broken in a different way?

mcilroyc:
Guides are not initialized for new documents.  `allocateDocument` calls `updateDocument` which handles non-current documents as though they have guides visibility false
https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/actions/documents.js#L416

iwehrman:
To re-state our offline conversation for posterity: I think that is the bug that should be fixed. I intentionally did not add this check in my original PR because I assumed that it should not be necessary given the invariant that guides are initialized IFF guides have been set visible at least once.

mcilroyc:
Take two,  @iwehrman.  reverted the first attempt in favor of a change to allocateDocument which forces the document to be updated in "current" mode.  This properly initializes guides.

iwehrman:
It's a bummer that this probably adds a small delay, but probably only very small. At some point we might also want to make the parameter conventions for `updateDocument` a little less weird. Anywho, looks good.

PR #3300, opened 2015-11-24
Listen to tool store as well for border policy renewal, and short circuit tools.select on same tool

volfied:
This is a bug I found while working on some other bug. On `tools.select`, we uninstall the active border policies. But we never re-install them with the new tool.

At first I tried listening to the tool store for change events, but this bug happens when you press V *while* in select tool already, so there is no difference in tool store that we can pick up on.

And I think it's reasonable for `superselectTool.select` to transfer to resetBorderPolicies. I know refactoring of all resetBorderPolicies into a listener was made to speed up our actions, but this is one place where it is necessary to call it directly, and it doesn't block anything else.

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -419,14 +419,10 @@ define(function (require, exports) {
         var toolStore = this.flux.store("tool"),
             currentTool = toolStore.getCurrentTool();
 
-        // If same tool is being selected, make this a no-op
-        if (currentTool === nextTool) {
-            return Promise.resolve();
-        }
-
         // Remove the border policy if it's been set at some point
+        // But not if we're resetting the same tool
         var removeTransformPolicyPromise;
-        if (_currentTransformPolicyID) {
+        if (_currentTransformPolicyID && currentTool !== nextTool) {
             var policyID = _currentTransformPolicyID;
             _currentTransformPolicyID = null;
             removeTransformPolicyPromise = this.transfer(policy.removePointerPolicies, policyID, true);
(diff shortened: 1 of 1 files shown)


iwehrman:
Would it be better to just short-circuit `select` if we're switching to the tool that's already activated? That way we could just listen for tool store change events, and we wouldn't need this one-off call to `resetBorderPolicies`?

pineapplespatula:
I'm with Ian on this, I think a better fix would be to address why we are always uninstalling the active border policies in tools.select. and in the where the tool does not change, we should not do that.

volfied:
Fair enough, I reverted last commit and instead pushed changes so we short circuit on tools.select when it's the same tool. We also listen to tool store change events now. @iwehrman

pineapplespatula:
I like this better, but the title no longer matches the PR, can you update that?

mcilroyc:
Assigning to @pineapplespatula to push the button

src/js/actions/tools.js
```diff
@@ -417,7 +417,13 @@ define(function (require, exports) {
      * @return {Promise} Resolves to tool change
      */
     var selectTool = function (nextTool) {
-        var toolStore = this.flux.store("tool");
+        var toolStore = this.flux.store("tool"),
+            currentTool = toolStore.getCurrentTool();
+
+        // If same tool is being selected, make this a no-op
```
pineapplespatula: I'm not sure that making the whole select function a no-op is what we want, but I want to test masks with this a bit before i press the button.

pineapplespatula: The super select tool in the mask mode relies on the select action to reset the behavior in different scenarios. Can you make this instead of being a no-op, only to not uninstall border policies

volfied:
@pineapplespatula Thanks for checking for bugs, I've made the change.

volfied:
This is now rebased and ready to merge, @pineapplespatula

volfied:
@pineapplespatula is something blocking this?

pineapplespatula:
running grunt debug on this branch results in 

>> Local Npm module "grunt-contrib-requirejs" not found. Is it installed?
Warning: Task "debug" not found. Use --force to continue

volfied:
@pineapplespatula Try again, please.

pineapplespatula:
LGTM

PR #3302, opened 2015-11-24
Upgrade to bluebird 3.x

iwehrman:
This is a backward-incompatible upgrade from Bluebird 2.x to 3.x. Details here: http://bluebirdjs.com/docs/new-in-bluebird-3.html

This should be reviewed and merged with adobe-photoshop/spaces-adapter#169.

Note that Bluebird 3 contains many new warnings, most of which seem useful and relevant. However, the source of these warnings is currently very hard to pin down (see petkaantonov/bluebird#861), so I've disabled the warnings for now. When the warnings are a little easier to deal with, we should of course address their causes and enable the warnings. For now I've only made the minimal required changes because I'd like to use some of the cancellation features of version 3 in a separate branch.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,7 +24,7 @@
     "tests"
   ],
   "dependencies": {
-    "bluebird": "~2.9.34",
+    "bluebird": "~3.0.5",
     "classnames": "~2.1.3",
     "d3": "~3.5.6",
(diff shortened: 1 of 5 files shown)


volfied:
Sweet.

Issue #3303, opened 2015-11-24
Panning on canvas disconnects superselect overlay from layers


pineapplespatula:
If you select the superselect tool and while hovering over layers you pan the canvas you can easily create a state where you see things like 
![screen shot 2015-11-24 at 3 48 41 pm](https://cloud.githubusercontent.com/assets/1008531/11384380/12e92460-92c4-11e5-8481-728b7485bc57.png)

pineapplespatula:
@volfied I'm having trouble reproducing this consistently since it seems like there are other issues with overlays in master.

volfied:
Highlighting does not work currently due to us not receiving `OS.externalMouseMove` events. However, you can confirm that UI updates w.r.t panning by turning on Policy Frames in Help menu. 

Can you confirm that in master, with Help>Show Policy Frames the policies don't update as you pan?

pineapplespatula:
border policies update correctly on drag in master

chadrolfs:
related or dupe of #2754?

iwehrman:
This is CNR for me. Can you verify @chadrolfs?

ktaki:
There is a good chance you can reproduce a separation of the policy frame.

Open a new artboard doc. Select the artboard. Option-drag it to make copies of the artboard several times continuously and quickly. (You might be able to do it with option-dragging groups.)

![screen shot 2016-01-22 at 8 56 52 am](https://cloud.githubusercontent.com/assets/2029597/12524577/b69b7410-c112-11e5-8e84-30566d99bff4.png)

chadrolfs:
@ktaki I CNR on both Mac and Win.  Can you try and let me know if you can still repro?

ktaki:
CNR with mac build 997. I will keep this open and come back later.

ktaki:
Oh, well. I just got it. This time, Moving "Artboard 1 copy" using superselect tool, and manipulated W H X Y by scrubbing.

![screen shot 2016-01-22 at 4 41 37 pm](https://cloud.githubusercontent.com/assets/2029597/12526887/7e56048e-c127-11e5-92b5-6ce14ff76ea9.png)

ktaki:
I reproduced once more today. I think this is definitely reproducible. Scrubby issue might be related to the  issue #3591. Try using scrubby to change W and H, and then, drag and move the artboard.

chadrolfs:
Coordinate separation seen if you undo (see #3726 )

[...]

Issue #866, opened 2015-06-12
On-canvas transform controls: highlight alignment seems slightly off


DesignSpaceBot:
![transformcontrolsgap](http://f.cl.ly/items/233B0Y1o2h3D0x37232M/image-removed.png)

-- @chadrolfs at [2015-03-20T19:36:53Z](https://github.com/adobe-photoshop/spaces-design-archive/issues/866)

PR #3304, opened 2015-11-24
First iteration of HUD for vector mask tool

pineapplespatula:
Create a HUD for the vector mask tool 
When you select the super select tool and enter mask mode on a layer without a mask. the hud pops up to help you create one. 
screen shot 2015-11-24 at 3 55 46 pm
allowing you to easily create a circular vector mask, 
screen shot 2015-11-24 at 3 56 16 pm

create a vector mask over the whole rect, or to drop into the pen tool directly. 

Seth also added some very nice overlay drawings on mouse over. 

@ktaki or @chadrolfs should take a look at this one before it is merged. currently there is a behavior very similar to issue #3303 where the overlays look incorrect if you zoom or pan the canvas after entering the HUD


ktaki:
Looks good. It might be better if the size of the mask is slightly smaller than the image boundary if entire image is visible on canvas. If the image boundary is outside of visible area on canvas (fills entire canvas), the mask could be created in smaller than the current visible area. Otherwise, it is hard to see the mask created by clicking the shapes in the HUD.

src/js/actions/mask.js
```diff
@@ -142,6 +150,98 @@ define(function (require, exports) {
             menuActions.native]
     };
 
+    /**
+     * Create an circular mask the size of the currently selected layer
```
volfied: I was gonna say the comment has a typo, then I realized it doesn't even describe the function correctly.

src/js/actions/mask.js
```diff
@@ -142,6 +150,98 @@ define(function (require, exports) {
...
+     *
+     * @return {Promise}
+     */
+    var applyRectangle = function () {
```
volfied: I don't think I like the action names. They don't exactly describe what's going on. Maybe `makeRectangleMask` and `makeEllipseMask` would be better? What do you think?

src/js/actions/mask.js
```diff
@@ -142,6 +150,98 @@ define(function (require, exports) {
...
+        }
+
+        var currentLayers = currentDocument.layers.selected;
+
+        if (currentLayers.isEmpty()) {
+            return Promise.resolve();
+        }
+
+        var currentLayer = currentLayers.first(),
+            bounds = boundsLib.bounds(currentLayer.bounds),
```
volfied: If this is going to be applicable to not just single layers, but layer groups as well, you probably want to use `currentDocument.layers.childBounds(currentLayer)` here, otherwise it's ok the way it is.

src/js/actions/mask.js
```diff
@@ -142,6 +150,98 @@ define(function (require, exports) {
...
+                historyStateInfo: {
+                    name: nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"),
+                    target: documentLib.referenceBy.id(currentDocument.id),
+                    coalesce: true,
+                    suppressHistoryStateNotification: true
+                }
+            };
+
+        return descriptor.batchPlayObjects([vectorMaskLib.makeBoundsWorkPath(bounds),
+            vectorMaskLib.makeVectorMaskFromWorkPath(),
```
volfied: Can you indent these two lines so they're obviously part of the call one line higher, and not in the same indent with the `.bind` and `.then`. The code flow here makes it hard to read.

src/js/actions/mask.js
pineapplespatula: since they are in the mask action adding a Mask to the end of the function name seems redundant. I dont think there is much difference between Apply and Make.

src/js/actions/mask.js
```diff
@@ -142,6 +150,98 @@ define(function (require, exports) {
...
+                historyStateInfo: {
+                    name: nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"),
+                    target: documentLib.referenceBy.id(currentDocument.id),
+                    coalesce: true,
+                    suppressHistoryStateNotification: true
+                }
+            };
+            
+        return descriptor.batchPlayObjects([vectorMaskLib.makeCircularBoundsWorkPath(bounds),
+            vectorMaskLib.makeVectorMaskFromWorkPath(),
```
volfied: Ditto

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -210,11 +220,24 @@ define(function (require, exports, module) {
             svg.selectAll(".superselect-bounds").remove();
             svg.selectAll(".superselect-marquee").remove();
             svg.selectAll(".artboard-adder").remove();
+            svg.selectAll(".vector-mask-hud").remove();
 
             if (!currentDocument || this.state.modalState) {
                 return null;
             }
 
+            this._hudGroup = svg.insert("g", ".hud-group")
```
volfied: Maybe this can be in the if statement below. Otherwise we're creating an unnecessary "vector-mask-hud" `<g>` every time we redraw.

volfied: I'm also not fan of the name, considering it's for vectorMask HUD only.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -210,11 +220,24 @@ define(function (require, exports, module) {
...
+                .classed("vector-mask-hud", true);
+            
+            var currentLayer = currentDocument.layers.selected.first();
+                
+            if (currentLayer && currentLayer.vectorMaskEmpty) {
```
volfied: These ifs can be &&'ed together.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -327,7 +350,77 @@ define(function (require, exports, module) {
 
             this.updateMouseOverHighlights();
         },
+        
+        /**
+         * Draws either superselectable or leaf layer bounding boxes
```
volfied: Stale JSDOC

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -80,6 +80,13 @@ define(function (require, exports, module) {
         _scrimGroup: null,
 
         /**
+         * Owner group for the HUD elements
```
volfied: If this is vector mask only, better name and a description here please.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -327,7 +350,77 @@ define(function (require, exports, module) {
...
+        
+        /**
+         * Draws either superselectable or leaf layer bounding boxes
+         * 
+         * @param {SVGElement} svg SVG HTML element to draw in
+         * @param {LayerTree} layerTree layerTree of the current document
+         */
+        drawMaskPreview: function (svg, layerTree, type) {
+            var scale = this._scale,
+                boundsAB = layerTree.topAncestor(layerTree.selected.first()).isArtboard ?
```
volfied: Please assign `layerTree.topAncestor(layerTree.selected.first())` to a variable so it's not duplicated in two lines.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -327,7 +350,77 @@ define(function (require, exports, module) {
 
             this.updateMouseOverHighlights();
         },
+        
+        /**
+         * Draws either superselectable or leaf layer bounding boxes
+         * 
+         * @param {SVGElement} svg SVG HTML element to draw in
+         * @param {LayerTree} layerTree layerTree of the current document
+         */
```
volfied: Missing param here... On that note, are you sure you're `npm updated`, because jscs and jshint should complain about this.

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -327,7 +350,77 @@ define(function (require, exports, module) {
...
+            
+            svg.selectAll("g#maskPreview").remove();
+            
+            if (type) {
+                svg.insert("g", ":first-child")
+                    .attr("id", "maskPreview")
+                    .classed("superselect-bounds", true)
+                    .attr("transform", this.props.transformString);
 
+                // Skip empty bounds
```
volfied: Why even create the <g> above if you're going to short circuit here?

volfied:
Code itself on micro level looks good. But there are a lot of typos and jsdoc errors. 
 
The vector mask HUD code should be in it's own overlay.jsx file. We're bundling superselectOverlay with vectorMask work and it's going to make it a big pain down the road to separate them (and trust me, we will have to). Can you break that into it's own file?

pineapplespatula:
I disagree about separating vector mask and superselect overlay. Fundamentally the vector mask HUD is the proper behavior for using the vector mask mode while you have the superselect tool active. It is super select behavior, and therefor should stay in superselect overlay

pineapplespatula:
@volfied :repeat_one:

volfied:
This is about modularity and future proofing our code. Mixing the two will make bug fixes harder (e.g. "Which state are we in, oh we are calling an entirely different action"), and in a future Space where we don't have vector mask overlays, will require us to separate them.

pineapplespatula:
There is a flicker when the hud first comes up. I'm not sure where that is coming from ( maybe the transition between superselect and superselect vector tools?)

pineapplespatula:
refactoring this seems to have made it terribly buggy. I'm reassigning this to me until its more stable

pineapplespatula:
back to @volfied

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+        StoreWatchMixin = Fluxxor.StoreWatchMixin,
+        d3 = require("d3");
+
+    var UI = require("adapter").ps.ui;
+
+    var system = require("js/util/system"),
+        keyUtil = require("js/util/key");
+
+    var SuperselectVectorMaskOverlay = React.createClass({
+        mixins: [FluxMixin, StoreWatchMixin("tool", "document", "application", "ui", "modifier")],
```
volfied: Not all of these stores are pertinent to this overlay, can you clean out the ones that are not used. (Which might explain the blinking)

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+         */
+        drawOverlay: function () {
+            if (!this.isMounted()) {
+                return;
+            }
+
+            var currentDocument = this.state.document,
+                svg = d3.select(React.findDOMNode(this));
+
+            svg.selectAll(".superselect-bounds").remove();
```
volfied: This also might explain the blinking

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+            if (svg.selectAll(".vector-mask-hud") && svg.selectAll(".vector-mask-hud").remove) {
```
volfied: Why do you have to check for existence of `remove` function here?

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+                svg.selectAll(".vector-mask-hud").remove();
+            }
+
+            if (!currentDocument || this.state.modalState) {
+                return null;
```
volfied: You don't have to return null, just return is enough.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+         *
+         * @private
+         */
+        _drawVectorMaskHUDObjects: function () {
+            var uiStore = this.getFlux().store("ui"),
+                cloakRect = uiStore.getCloakRect();
+            
+            var layerTree = this.state.document.layers;
+    
+            var fluxActions = this.getFlux().actions,
```
volfied: you should put flux in a variable instead of calling `this.getFlux()` repeatedly.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+        },
+
+        componentDidUpdate: function (prevProps, prevState) {
+            if (this.state.vectorMaskMode !== prevState.vectorMaskMode) {
+                this.drawOverlay();
+            }
+        },
+
+        /**
+         * Calls all helper functions to draw super-select overlay
```
volfied: Stale JSDOC

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+                }
+
+                svg.insert("g", ":first-child")
+                    .attr("id", "maskPreview")
+                    .classed("superselect-bounds", true)
+                    .attr("transform", this.props.transformString);
+
+                // HACK: For some reason Photoshop's bounds seem to be shifted by ~1px to the
+                // bottom-right. See https://github.com/adobe-photoshop/spaces-design/issues/866
+                var offset = system.isMac ? 0 : scale;
```
volfied: This hack was very specific to bounds highlighting / on canvas transform overlays... Does it apply to masks? Can you turn it off and see if it causes #866 to appear?

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+         * Checks to see if the given position is under
+         * any always_propagate policies, where mouse events should go to PS
+         * and we shouldn't highlight layers
+         *
+         * @param {number} x
+         * @param {number} y
+         *
+         * @return {boolean} True if the pointer is under an always propagate policy
+         */
+        _positionUnderAlwaysPropagatePolicy: function (x, y) {
```
volfied: This function is unused in this file.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,346 @@
...
+                    this.drawMaskPreview(this._vectorMaskHudGroup, layerTree, "ellipse");
+                    d3.event.stopPropagation();
+                }.bind(this))
+                .on("mouseout", function () {
+                    // Apply a temp mask on hover
+                    this.drawMaskPreview(this._vectorMaskHudGroup, layerTree, false);
+                    d3.event.stopPropagation();
+                }.bind(this))
+                .on("click", function () {
+                    // Apply the color to selected layers
```
volfied: Stale comments

src/js/models/tools/superselect/vector.js
```diff
@@ -107,6 +109,8 @@ define(function (require, exports, module) {
             flux.actions.tools.select(toolStore.getToolByID("newSelect"));
         }
     };
+    
+    SuperSelectVectorTool.prototype.toolOverlays = [SuperselectOverlay, SuperselectVectorMaskOverlay];
```
volfied: Does the direct selection tool for editing vectors require highlighting?

src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -214,6 +216,11 @@ define(function (require, exports, module) {
             if (!currentDocument || this.state.modalState) {
                 return null;
             }
+            var currentLayer = currentDocument.layers.selected.first();
```
volfied: Newline here to separate if block from var

src/js/jsx/Scrim.jsx
```diff
@@ -251,15 +251,17 @@ define(function (require, exports, module) {
                 return null;
             }
 
-            if (!tool || !tool.toolOverlay || this.state.isAssetDragOver) {
+            if (!tool || !tool.toolOverlays || this.state.isAssetDragOver) {
                 return null;
             }
             
-            var ToolOverlay = tool.toolOverlay;
+            var ToolOverlays = tool.toolOverlays;
```
volfied: the variable should be in camelCase

src/js/jsx/Scrim.jsx
```diff
@@ -251,15 +251,17 @@ define(function (require, exports, module) {
...
 
-            return (
-                <ToolOverlay transformString={transformString} ref="toolOverlay"/>
-            );
+            return ToolOverlays.map(function (ToolOverlay) {
+                return (
+                    <ToolOverlay transformString={transformString} ref="toolOverlay"/>
```
volfied: You're adding the same `ref` to multiple components here.

volfied: Looks like you can remove the ref completely, as it's not in use.

volfied:
Reviewed

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
pineapplespatula: the blinking went away, i think We should leave this since we use that class at times when the HUD is active

src/js/models/tools/superselect/vector.js
pineapplespatula: yes. Otherwise there is a large flicker.

src/js/fluxcontroller.js
```diff
@@ -596,7 +596,7 @@ define(function (require, exports, module) {
         }
 
         var modalPromise;
-        if (!modal && flux.store("tool").getModalToolState()) {
+        if (!modal) {
```
volfied: What's the reason for this removal?

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
volfied: It is not the vector mask overlay's job to remove some other component's UI. 

Superselect Overlay should already clear out it's own svg elements during it's own `drawOverlay`, in which you eventually short circuit and stop drawing anything else.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,279 @@
...
+        drawOverlay: function () {
+            if (!this.isMounted()) {
+                return;
+            }
+
+            var currentDocument = this.state.document,
+                svg = d3.select(React.findDOMNode(this));
+
+            svg.selectAll(".superselect-bounds").remove();
+            if (svg.selectAll(".vector-mask-hud")) {
```
volfied: Not sure if you need to check for this, I think d3 is able to safely exit its functions if selector returns an empty array.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,279 @@
...
+            
+            var layerTree = this.state.document.layers;
+    
+            var fluxActions = flux.actions,
+                toolStore = flux.store("tool"),
+                remToPx = flux.store("ui").remToPx,
+                numIcons = 3;
+
+            var sampleSize = remToPx(2.4),
+                iconSize = Math.round(sampleSize * 1),
```
volfied: `* 1` ?

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -0,0 +1,279 @@
...
+                .attr("width", iconSize)
+                .attr("height", iconSize)
+                .on("mouseover", function () {
+                    // Apply a temp mask on hover
+                    this.drawMaskPreview(this._vectorMaskHudGroup, layerTree, "ellipse");
+                    d3.event.stopPropagation();
+                }.bind(this))
+                .on("mouseout", function () {
+                    // Remove the temp mask on hover
+                    this.drawMaskPreview(this._vectorMaskHudGroup, layerTree, false);
```
volfied: The JSDoc of this function states it accepts strings, if `false` is going to be a used value, please state that as well.

src/js/models/tools/superselect/vector.js
volfied: So when user is in direct edit mode of a vector (no masks), they're going to see bounds of other layers highlight? If there is a flicker, there must be a bug somewhere in the code unrelated to this.

volfied:
Reviewed @pineapplespatula . 

We can't get rid of the check in fluxcontroller. That's going to cause a big slow down in the app due to an unnecessary call to PS that fails in a lot of functions/transfers.

src/js/models/tools/superselect/vector.js
pineapplespatula: no they aren't, because the super select overlay doesn't draw in that case.

pineapplespatula:
@volfied :repeat:

src/js/models/tools/superselect/vector.js
volfied: If it doesn't draw, does it need to be included in the overlays for this tool?

volfied:
One more thing, @pineapplespatula. Right now, superselectoverlay is being drawn while we're in vector edit mode. It's not highlighting because we're not listening to the mouse events (due to policies), but the invisible rectangles are still in there. Please address this issue.

src/js/models/tools/superselect.js
```diff
@@ -37,6 +37,7 @@ define(function (require, exports, module) {
         system = require("js/util/system"),
         shortcuts = require("js/util/shortcuts"),
         SuperselectOverlay = require("jsx!js/jsx/tools/SuperselectOverlay"),
+        SuperselectVectorMaskOverlay = require("jsx!js/jsx/tools/SuperselectVectorMaskOverlay"),
```
volfied: With webpack merge, you need to get rid of the `jsx!` here during rebase.

src/js/models/tools/superselect/vector.js
```diff
@@ -31,6 +31,8 @@ define(function (require, exports, module) {
     var Tool = require("js/models/tool"),
         system = require("js/util/system"),
         EventPolicy = require("js/models/eventpolicy"),
+        SuperselectOverlay = require("jsx!js/jsx/tools/SuperselectOverlay"),
```
volfied: Ditto

pineapplespatula:
there are still some bugs around modal state, but i think they should be in a different PR

volfied:
@pineapplespatula What are the bugs? Also, I don't think "the HUD comes up much more slowly..." is a good commit description, especially since it does not describe anything the commit itself is doing.

If you rebase to a single commit, and log the issues that you're aware of, I'm good to merge this.

pineapplespatula:
i logged issues #3340 and #3378

volfied:
:+1:

Issue #3268, opened 2015-11-18
Errors on Next Document and Previous Document


ktaki:
Open two docs. Execute Window->Next Document and Window->Previous Document

Action documents.selectNextDocument failed: TypeError: Cannot read property 'get' of undefined
Action documents.selectPreviousDocument failed: TypeError: Cannot read property 'get' of undefined

![screen shot 2015-11-18 at 11 20 17 am](https://cloud.githubusercontent.com/assets/2029597/11251635/07f91ac4-8de7-11e5-8b0f-e595155af706.png)
![screen shot 2015-11-18 at 11 20 59 am](https://cloud.githubusercontent.com/assets/2029597/11251641/0c45b290-8de7-11e5-87fb-6f41967be56c.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3308, opened 2015-11-25
Point to the right internal variable in documentindex

volfied:
When we broke it apart, we must have missed a private variable name, causing the errors. 

Addresses #3268


iwehrman:
Whoopsie daisy.

Issue #3311, opened 2015-11-29
Error activating pen tool second time


iwehrman:
1. Reload Design Space with the pen tool acitve.
2. Activate to any other tool. 
3. Activate the pen tool.

Result:
```
Transfer from toolPen.select to shortcuts.addShortcut failed: Shortcut already exists: penBackspace(anonymous function) @ fluxcontroller.js:491tryCatcher @ bluebird.js:4912Promise._settlePromiseFromHandler @ bluebird.js:2997Promise._settlePromise @ bluebird.js:3054Promise._settlePromise0 @ bluebird.js:3099Promise._settlePromises @ bluebird.js:3178(anonymous function) @ bluebird.js:162setTimeout (async)schedule @ bluebird.js:133Async.settlePromises @ bluebird.js:161Promise._reject @ bluebird.js:3144Promise._settlePromise @ bluebird.js:3069Promise._settlePromise0 @ bluebird.js:3099Promise._settlePromises @ bluebird.js:3178(anonymous function) @ bluebird.js:162setTimeout (async)schedule @ bluebird.js:133Async.settlePromises @ bluebird.js:161Promise._reject @ bluebird.js:3144Promise._settlePromise @ bluebird.js:3069Promise._settlePromise0 @ bluebird.js:3099Promise._settlePromises @ bluebird.js:3178(anonymous function) @ bluebird.js:162setTimeout (async)schedule @ bluebird.js:133Async.settlePromises @ bluebird.js:161Promise._reject @ bluebird.js:3144Promise._settlePromise @ bluebird.js:3069Promise._settlePromise0 @ bluebird.js:3099Promise._settlePromises @ bluebird.js:3178(anonymous function) @ bluebird.js:162setTimeout (async)schedule @ bluebird.js:133Async.settlePromises @ bluebird.js:161Promise._reject @ bluebird.js:3144Promise._resolveCallback @ bluebird.js:2942Promise._settlePromiseFromHandler @ bluebird.js:3010Promise._settlePromise @ bluebird.js:3054Promise._settlePromiseCtx @ bluebird.js:3091(anonymous function) @ bluebird.js:152

Transfer from tools.select to toolPen.select failed: Error: Job canceled before execution
    at file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:154:33
    at Array.forEach (native)
[...]

PR #3314, opened 2015-11-29
Fix REMOVE_SHORTCUT handler

iwehrman:
This fixes an awful off-by-one bug in the shortcut store's `REMOVE_SHORTCUT` handler that was preventing the most recently installed shortcut from ever being uninstalled. 

Addresses #3311.


mcilroyc:
yikes.  Also seems like now we could do better than using `some()` to find the index, not that it really matters.

Issue #3313, opened 2015-11-29
Tool activation is slow


iwehrman:
The time to update the toolbar when switching tools is terrible. The `tools.selectAction` action seems to take at least ~400ms, and we don't dispatch to the stores to update the React components until the very end of that. Consequently, the total time to switch tools is probably something like 700ms, which is abysmal.

iwehrman:
Fixed by #3315.

PR #3315, opened 2015-11-29
Faster tool activation

iwehrman:
This PR splits the tool-change dispatch into two parts:

1. an early, optimistic part that causes the view to re-render; and
2. a later part that contains the installed policy IDs.

This significantly improves perceived tool-activation time. Addresses #3313.

### DIFF
--- a/src/js/stores/tool.js
+++ b/src/js/stores/tool.js
@@ -112,7 +112,8 @@ define(function (require, exports, module) {
         initialize: function () {
             this.bindActions(
                 events.RESET, this._handleReset,
-                events.tool.SELECT_TOOL, this._handleSelectTool,
+                events.tool.SELECT_TOOL_START, this._handleSelectTool,
+                events.tool.SELECT_TOOL_END, this._handleSelectTool,
                 events.tool.MODAL_STATE_CHANGE, this._handleModalStateChange,
                 events.tool.VECTOR_MASK_MODE_CHANGE, this._handleVectorMaskModeChange,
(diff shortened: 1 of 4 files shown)


src/js/stores/tool.js
```diff
@@ -202,15 +203,21 @@ define(function (require, exports, module) {
...
          */
         _handleSelectTool: function (payload) {
             var tool = payload.tool;
 
             this._previousTool = this._currentTool;
             this._currentTool = tool;
-            this._currentKeyboardPolicyID = payload.keyboardPolicyListID;
-            this._currentPointerPolicyID = payload.pointerPolicyListID;
+
+            if (payload.keyboardPolicyListID) {
```
mcilroyc: during the handling of the "start" event, does it make sense to clear the current*policy vars?  Otherwise it seems like there is a period of inconsistency?  Actually, as I type this, I realize that this inconsistency properly reflects the actual (hopefully brief) inconsistent state that we're in.  So nevermind.

mcilroyc:
I guess this worries me a little bit that, especially on slow machines, there's a potential for the user being falsely assured of the tool change (especially using KBSC) and then having canvas interactions interpreted in confusing ways.  But I can't convince myself this is an actual problem.  And this sure FEELS faster!

iwehrman:
The only good solution to this problem (which exists before and after this PR) would be to lock the UI during the `tools.select` action.

mcilroyc:
I'll also just say this - even though it may not really make sense - but maybe an animation to reflect the "you clicked this, but hang tight" nature of it.  Like a flash/pulse kinda thing.

PR #3316, opened 2015-11-30
Menu bug-fix grab bag

iwehrman:
This PR contains various menu fixes and cleanups. In particular: 

1. The maps (`rootMap` and `submenuMap`) are now derived properties, which reduces complexity and at least a few bugs by obviating their maintenance.
2. Many instances of mutable object references have been purged in favor strict immutability. In particular, there is now an immutable `MenuShortcut` model, instances of which may be referenced by `MenuItem` instances.
3. The `checked` property of `MenuItem` is now properly a boolean. Previously its range also variously, erroneously included 0, 1, "on" and "off".
4. Descriptor export has been tidied up so that unnecessary cruft is not submitted to the adapter when installing new menu definitions.
5. The open-document `MenuItem` instances no longer refer directly to document models; instead only their IDs are referenced.

As a result, state changes that should not affect the state of the menu now properly result in structurally equal `MenuBar` instances, which results in significantly fewer menu-installation adapter calls. This may help avoid a bug in which the menu is corrupted if it is open while installing a new menu because we'll do that less frequently.

---

As @tomwmcrae pointed out, many of these bugs are caused by Immutable record "type" violations. It sure would be great if there were a way to do dynamic (not static) type checks when in debug mode to ensure that our Immutable record constructors were being applied to models with valid types at instantiation time. There is some discussion of a hypothetical feature along those lines here: facebook/immutable-js#546.

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -28,11 +28,13 @@ define(function (require, exports, module) {
         Immutable = require("immutable");
 
     var MenuItem = require("./menuitem"),
+        MenuShortcut = require("./menushortcut"),
         keyutil = require("js/util/key"),
         global = require("js/util/global"),
         pathUtil = require("js/util/path"),
-        system = require("js/util/system");
-    
+        system = require("js/util/system"),
+        object = require("js/util/object");
+
     /**
      * A model for the menu bar application currently shows
      *
@@ -65,15 +67,22 @@ define(function (require, exports, module) {
          *
          * @type {Immutable.Map.<string, object>}
          */
-        actions: null,
+        actions: null
+    });
 
+    Object.defineProperties(MenuBar.prototype, object.cachedGetSpecs({
         /**
          * Map from ID to root menus
          *
          * @type {Immutable.Map.<string, MenuItem>}
          */
-        rootMap: null
-    });
+        rootMap: function () {
+            return Immutable.Map(this.roots
+                .map(function (entry) {
+                    return [entry.itemID, entry];
+                }));
+        }
+    }));
 
     /**
      * Process the raw action descriptor into enablement rules
@@ -113,7 +122,7 @@ define(function (require, exports, module) {
                 if (descriptor.hasOwnProperty("$payload")) {
                     action.$payload = descriptor.$payload;
                 }
-                actionMap.set(id, action);
+                actionMap.set(id, Immutable.Map(action));
             } else {
                 if (prop !== "$enable-rule") {
                     _processMenuActions(descriptor, actionMap, enablerMap, id);
@@ -379,12 +388,9 @@ define(function (require, exports, module) {
         _artboardFromTemplates(menuObj, menuActionsObj, templates);
 
         var menuID = menuObj.id,
-            rootMap = new Map(),
             // Process each root submenu into roots
             roots = Immutable.List(menuObj.menu.map(function (rawMenu) {
-                var rootItem = MenuItem.fromDescriptor(rawMenu);
-                rootMap.set(rootItem.id, rootItem);
-                return rootItem;
+                return MenuItem.fromDescriptor(rawMenu);
             })),
             actions = new Map(),
             enablers = new Map();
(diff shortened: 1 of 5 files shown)


src/js/models/menubar.js
```diff
@@ -606,20 +602,20 @@ define(function (require, exports, module) {
                     "label": label,
                     "command": id,
                     "enabled": !appIsModal,
-                    "checked": Immutable.is(document, currentDocument) ? "on" : "off",
-                    "shortcut": (index < 9) ? {
+                    "checked": Immutable.is(document.id, currentDocument.id),
```
volfied: If we're comparing two numbers, do we need immutable.is here?

src/js/models/menuitem.js
```diff
@@ -86,21 +82,43 @@ define(function (require, exports, module) {
         commandKind: null,
 
         /**
-         * @type {{modifiers: number=, keyCode: number}}
+         * @type {?Shortcut}
```
volfied: `MenuShortcut` here.

src/js/models/menuitem.js
```diff
@@ -86,21 +82,43 @@ define(function (require, exports, module) {
...
+
+            return Immutable.Map(this.submenu
+                .toSeq()
+                .filterNot(function (entry) {
+                    return entry.type === "separator";
+                })
+                .map(function (entry) {
+                    return [entry.itemID, entry];
+                })
+                .toList());
```
volfied: Is `toList` necessary here? Immutable.Map ctor accepts iterables.

src/js/models/menuitem.js
```diff
@@ -283,16 +320,22 @@ define(function (require, exports, module) {
      * @return {MenuItem}
      */
     MenuItem.prototype.updateSubmenuProps = function (submenuID, props) {
-        var menuItem = this.submenuMap.get(submenuID),
-            menuIndex = this.submenu.indexOf(menuItem);
+        var menuIndex = this.submenu.findIndex(function (menuItem) {
```
volfied: Why not use submenuMap here? It's a cached property and would be much faster.

volfied:
Got couple questions @iwehrman

src/js/models/menuitem.js
mcilroyc: But does Map constructor accept a sequence?  Mainly I'm responding to this comment so that I'll start getting notifications of future comments.

src/js/models/menuitem.js
iwehrman: I was trying to avoid building it unnecessarily, but... now that I say that out loud, it doesn't really make any sense. Will fix.

src/js/models/menubar.js
iwehrman: Derp.

iwehrman:
Issues addressed/obviated, I think. I changed `submenuMap` to map from IDs to indices, so that it could support the two different uses cases in constant-ish time.

src/js/models/menuitem.js
```diff
@@ -86,21 +82,58 @@ define(function (require, exports, module) {
...
+                return null;
+            }
+
+            return Immutable.Map(submenu.reduce(function (map, entry, index) {
+                if (entry.type === "separator") {
+                    return map;
+                } else {
+                    return map.set(entry.itemID, index);
+                }
+            }, new Map()));
```
volfied: Hey @mcilroyc, please correct me if I'm wrong, but I remember you having to change some code like this in the past because it was really slow. Am I remembering wrong?

volfied:
Thanks for the changes @iwehrman, I just want to check with @mcilroyc and you on my last comment about creating the map through a reduce, otherwise looks good. 

So, @mcilroyc, please hit the merge button if my worries are misplaced and things look good to you.

src/js/models/menuitem.js
mcilroyc: I do remember suggesting that a good pattern is to create an Array of [K, V] and using that to create an Immutable.Map after the reduction.  However, I think this method is fine too.  What we *don't* want is to use reduce which creates an Immutable.Map and then `set` it within the loop.

PR #3317, opened 2015-11-30
Cleanup of TextInput definition and usage

designedbyscience:
1. Removed property expansion in rendering the inputs
2. `singleClick` and `live` seemingly accomplished the same goal. I renamed the property `doubleClickToEdit`, which is more informative than either of the other two, and set the default to `false`, since we only use this property in LayerFace and Library Assets
3. `TextInput` does not use or need the `type` property. Removed
4. `editable` and `disabled` also seemed to accomplish the same goal. Also, confusing with the `live` property. Condensed everything down to `disabled`
5. `preventHorizontalScrolling` is used by elements with `doubleClickToEdit={true}`. I set this to default to `true`, as it doesn't harm anything and then allows us to not have 2 dependent properties that must be set in the usage of TextInput (not preventing horizontal scrolling is the exception)
6. Fixed a typo in a CSS class name

I'm going to update the specs, but high level, here are places where TextInput is used:
* Export panel, suffix field
* Layers panel, layer name
* Libraries panel, creating a new library name input and asset names
* Color panel, color string input
* Datalists, all of them, including Search Bar

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -653,10 +653,9 @@ define(function (require, exports, module) {
                                         title={layer.name + tooltipPadding}
                                         className="face__name"
                                         ref="layerName"
-                                        type="text"
+                                        doubleClickToEdit={true}
                                         value={layer.name}
-                                        editable={!this.props.disabled && nameEditable}
-                                        preventHorizontalScrolling={true}
+                                        disabled={this.props.disabled || !nameEditable}
                                         onKeyDown={this._skipToNextLayerName}
                                         onChange={this._handleLayerNameChange}>
                                     </TextInput>
--- a/src/js/jsx/shared/TextInput.jsx
+++ b/src/js/jsx/shared/TextInput.jsx
@@ -33,19 +33,6 @@ define(function (require, exports, module) {
 
     var Focusable = require("../mixin/Focusable");
 
-    var _typeToClass = {
-        simple: "column-4",
-        percent: "column-3",
-        degree: "column-3",
-        color: "column-11",
-        shadow: "column-9",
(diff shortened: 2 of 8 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -653,10 +653,9 @@ define(function (require, exports, module) {
                                         title={layer.name + tooltipPadding}
                                         className="face__name"
                                         ref="layerName"
-                                        type="text"
+                                        doubleClickToEdit={true}
                                         value={layer.name}
-                                        editable={!this.props.disabled && nameEditable}
```
shaoshing: Maybe this should be `this.props.disabled || !nameEditable`.

shaoshing:
Hey @designedbyscience this looks awesome. Love the reduced number of params and the renames, and your detailed summary. I still have one question, then I am ready to merge it. :+1:

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -653,10 +653,9 @@ define(function (require, exports, module) {
...
-                                        preventHorizontalScrolling={true}
+                                        disabled={this.props.disabled && !nameEditable}
```
shaoshing: Maybe this should be this.props.disabled || !nameEditable.

src/js/jsx/shared/TextInput.jsx
```diff
@@ -33,19 +33,6 @@ define(function (require, exports, module) {
 
     var Focusable = require("../mixin/Focusable");
 
-    var _typeToClass = {
```
mcilroyc: bye felicia

designedbyscience:
Great, changed, thanks @shaoshing

Issue #3259, opened 2015-11-17
Opening a LLSO has incorrect coordinates and highlight location


chadrolfs:
new doc
drag a LLSO onto canvas (note it auto-commits)
hit Enter (opens LLSO)

Result: artboard adders and highlighting coordinates are off

chadrolfs:
![openllso-badcoords](https://cloud.githubusercontent.com/assets/8442631/11222175/cd7d7814-8d1d-11e5-8c6c-1f1d6b3a6cbc.gif)

shaoshing:
Might related to #3251 because I also see this in regular document. Will check again once #3251 is resolved.

shaoshing:
![screen shot 2015-11-30 at 3 13 09 pm](https://cloud.githubusercontent.com/assets/118264/11487559/e8bc45ae-9774-11e5-8f58-5080df7be8f0.png)

shaoshing:
Fixed by #3320

Issue #3312, opened 2015-11-29
Marquee selection is broken


iwehrman:
1. Create a new document.
2. Click and drag within the document canvas.

Expected: a marquee selection rectangle. 
Result: nothing.

The `select.drag` action is at least being called.

chadrolfs:
related to #3158

shaoshing:
Fixed by #3320.

Issue #3251, opened 2015-11-16
On-canvas highlighting broken


chadrolfs:
create a new doc, select the artboard
Result: adders do not show on-canvas

create some rects and mouse over
result: no blue highlighting

chadrolfs:
Show Policy Frames doesn't work either

volfied:
Something changed and we no longer get OS.externalMouseMove events, I'm looking at it.

volfied:
https://github.com/adobe-photoshop/spaces-design/issues/3278 is a dupe of this.

volfied:
https://jira.corp.adobe.com/browse/PS-306

chadrolfs:
related to #3312

shaoshing:
Fixed by #3320

PR #3320, opened 2015-11-30
Change Scrim SCU to skip isDropTarget when false

volfied:
This must have slipped through the cracks during the drag&drop work. Scrim SCU never re-updated itself due to fact that both isDropTarget and nextProps.dropTarget being false. Now it will return false there if and only if both are true.

Addresses #3312, #3251, and bunch of other dupes.


shaoshing:
Nice! and it also fixed #3259

Issue #3318, opened 2015-11-30
Cannot create new layers


ktaki:
With the latest pg-Dev Mac build #906 and master branch (#3315, feec60e), new layers cannot be created with shape tools, pen tool, or type tool. The pg-Dev Mac build #906 itself does not have the issue.

ktaki:
This does not happen with the Mac build #903. Could this be a Core issue caused by tristam's submissions?

iwehrman:
Is there a warning/error/stack trace in the console?

ktaki:
No errors are shown in the console.

iwehrman:
Seems to work again in pgdev 908.

ktaki:
Well, not with my machine. So, you cannot reproduce this? I will try with other machine.

chadrolfs:
yeah, i can repro this bug with pgdev.910 with master @ feec60e

ktaki:
@chadrolfs: Good to hear you can repro. Windows seems working fine. Have you tried on Win?

chadrolfs:
start with Select tool, reset DS, hit r to switch to Rect tool, try to draw a shape.

if you reset DS and you already have a shape tool active, you can create shapes.

ktaki:
Thanks for clarifying the steps. Windows has the same issue.

volfied:
Can repro on Mac with Chad's steps.

volfied:
We have a foreign PROP_TO_BROWSER leftMouseDown policy whenever we start with superselect tool, it doesn't go away, and breaks things. This is the cause, I'm working on it now.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3321, opened 2015-11-30
Don't ignore 0 policy IDs during tool swap

volfied:
Addresses #3318 

Tool store was not saving the initial superselect tool's "send to browser" policy, since the ID was 0, so it was sticking and preventing other tools that relied on sending mouse events to Photoshop.


iwehrman:
Sorry, I just pushed a little fix for this right to master before I noticed your PR:
https://github.com/adobe-photoshop/spaces-design/commit/383075a3a61f22f2099744e1e8e3e7b30b5f9b57 In any case, fixed.

Issue #3154, opened 2015-10-28
Undo after changing transform fields with negative values doesn't update correctly


chadrolfs:
new doc
create 2 different sized Rects and select them both
enter W transform field, enter a negative number (-50), hit enter
Result: min value for each rect is set (0.1) but W field shows mixed state ( -- )
Repeat last step for H field
Result: now that field shows what you'd expect 0.1
cmd/ctrl + z undo (repeat)
Result: canvas updates but field remains 0.1
undo eventually results in 
`Transfer from history.decrementHistory to layers.resetLayerVisibility failed: Error: Photoshop returned an error when playing command number: 3 Photoshop code: 24946 Photoshop message: error: 24946 message: base element not found`

chadrolfs:
![undo-xform](https://cloud.githubusercontent.com/assets/8442631/10798775/73666956-7d66-11e5-82b5-7fbeba2253de.gif)

chadrolfs:
related to #3151

chadrolfs:
![screen shot 2015-10-28 at 11 28 26 am](https://cloud.githubusercontent.com/assets/8442631/10798842/d1fe4556-7d66-11e5-95a4-4969963884af.png)

mcilroyc:
Ok.  good news and bad news.  
bad: this is actually two separate bugs
good: I can narrow down the worse of the two bugs, and they can be treated separately, I think.  That is, the undo related bug has no relation to whether or not the number input shows a mixed state after entering a negative value:

repro:
start in the same way @chadrolfs described.  new doc, two rects, select both.  then...
- click into `W` field, input -50, hit *enter*
- (while still focused in input) cmd-Z
- Note in the console that it immediately enqueues both setSize and undo actions.  This is the root of the error that will follow if you undo again

My hunch is that, because this happens with W but not H, it may have to do with `_handleWidthKeyDown` in Size.jsx (unsubstantiated)

The separate bug with mixed states has to do with rounding, but I haven't investigated further
![rounding bounds](http://cl.ly/image/2k032m15053Q/Image%202015-10-28%20at%202.36.24%20PM.png)

I will say that I think these are both deferrable bugs given the edge behavior required to trigger them.

cc @iwehrman since you pinged me on slack about this.

iwehrman:
Also note that the controller reset seems to handle this failure as gracefully as possible.

chadrolfs:
Simple steps to hit error:
cmd/ctrl + n
[...]

PR #3322, opened 2015-11-30
undo/redo menu commands use preserveFocus

mcilroyc:
For the undo and redo menu commands, supply the "preserveFocus" switch in the payload.  This seems like a better experience, and addresses #3154.  I can't find any downside to this.

Better experience: For example, change the Width of a rectangle using the number input, then undo.  Retaining focus in the input seems natural

The bug fix: Previously when changing to a negative value of width (which clamps to 0.1), undoing it would cause a setSize AND an undo to be performed in rapid succession.

### DIFF
--- a/src/static/menu-actions.json
+++ b/src/static/menu-actions.json
@@ -144,10 +144,16 @@
         },
         "UNDO": {
             "$action": "edit.undo",
+            "$payload": {
+                "preserveFocus": true
+            },
             "$enable-rule": "have-document,earlier-history"
         },
         "REDO": {
             "$action": "edit.redo",
+            "$payload": {
+                "preserveFocus": true
+            },
             "$enable-rule": "have-document,later-history"
         },
         "SELECT_ALL": {


mcilroyc:
@chadrolfs or @ktaki  could you please try this branch?  This is a minor change from a code perspective, but has moderately significant impact to functionality.  When doing an undo/redo with an active input, we *no longer* "blur" the input prior to executing the undo/redo.  I think this makes sense, but would like your opinion(s).

ktaki:
The error reported in the issue #3154 does not show up on the branch. The negative value shows up as "--" is still there, but this is a separate issue. After negative value was entered and corrected to "0.1" and hit return to remove focus, you still need to undo twice to get the original value back. This, I think, is a minor glitch we can defer. Overall, the undo changes do not cause new issues. @chadrolfs How do you say?

chadrolfs:
as soon as I get through my assigned issues, i'll take a look at this.

chadrolfs:
Yeah, seems to resolve the bug.  In both cases, I'm seeing mixed results, sometimes it resolves to: 0.1, other times it resolves to: -- for both W and H.  I'm confused.  And then if I undo, undo, undo, then redo, redo, redo, redo, I end up with something really unexpected:

undo-wh

chadrolfs:
it looks like marquee select isn't working either.

mcilroyc:
@chadrolfs @ktaki thanks for checking.  I think the issue of "mixed" results should be treated separately as the underlying cause is separate.  Probably as a separate bug, or maybe under the umbrella of #3153.  If we're not seeing the gross error during undo, then #3154 seems fixed.  That's my opinion.

@chadrolfs marquee select seems to work for me (I'm on an older build dev mac 896).  Can you be more specific about how to repro?  Or can you confirm that the issue you are seeing is NOT in 383075a (which is the point on master that this branched from?

mcilroyc:
@chadrolfs I've rebased this branch to master (per our slack convo).  Also note that the commit I referenced in my previous comment may been incorrect, but that point is now moot.

chadrolfs:
things working great after rebase, thanks!

ktaki:
I think this is okay to merge now.

Issue #3212, opened 2015-11-05
2nd resize of expanded CC Lib asset fails


chadrolfs:
cmd/ctrl + n
load CC Lib: Test - OK to accidentally nuke
alt/opt + drag asset onto canvas (GearSale)
Resize it with Shift key to constrain, move and resize again

Result: 2nd attempt fails

chadrolfs:
![cc-lib-2nd-resizefail](https://cloud.githubusercontent.com/assets/8442631/10984951/285c9454-83d2-11e5-8ad0-9b2629bb936c.gif)

chadrolfs:
![screen shot 2015-11-05 at 3 34 28 pm](https://cloud.githubusercontent.com/assets/8442631/10985051/cca20986-83d2-11e5-9469-3f79d647d6bf.png)

chadrolfs:
![screen shot 2015-11-05 at 3 36 12 pm](https://cloud.githubusercontent.com/assets/8442631/10985082/098e32c0-83d3-11e5-8adb-7dad63fd560a.png)

volfied:
As the last screenshot shows, this is related to the bounds not updating correctly. Make sure to keep Help > Show Policy Frames enabled to see whether bounds are correctly updating as you resize the expanded layers.

volfied:
FBNC to @chadrolfs

PR #3323, opened 2015-11-30
Ignore adjustment layers when calculating layer bounds

shaoshing:
Fix for issue #3212 

Adjustment layers add effects to the layers below them and do not have a visual area, so their bounds should be ignored when calculating bounds of selected layers. Othwerise it will result in incorrect bounds (which is larger than the actual bounds) and break the transform feature.


volfied:
ok!

PR #3326, opened 2015-12-01
Idle initialization of the font list and unselected layers

iwehrman:
This PR adds infrastructure for idle task execution:

1. Action receivers now have a `whenIdle` method, which enqueues the given action once the queue and JavaScript engine are idle.
2. Unselected layers of activated documents are now initialized when idle instead of at first-selection time. (Note that this does not yet initialize not-yet-activated documents.) The action `layers.initializeLayersBackground` calls `initializeLayers` on 50 uninitialized layers for a given document, and then calls itself again when idle if there are still uninitialized layers.
3. The font list is initialized when idle instead of first-text-layer-selection time.

The scenarios in which this makes a noticeable performance impact are not terribly common (selecting a big range of layers that have never been selected previously), but I thought it would be good to get feedback on the approach before going much further.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -105,6 +105,8 @@ define(function (require, exports, module) {
         this._actionQueue = new AsyncDependencyQueue(CORES);
         this._initActionNames();
         this._initActionLocks();
+        this._synchronizedActions = new Map();
+        this._idleTasks = new Set();
 
         var actions = this._synchronizeAllModules(actionIndex),
             stores = storeIndex.create(),
@@ -179,12 +181,28 @@ define(function (require, exports, module) {
     FluxController.prototype._actionReceivers = null;
 
     /**
+     * Map from unsynchronized to synchronized actions.
+     *
+     * @private
+     * @type {Map.<Action, ActionReceiver>}
+     */
+    FluxController.prototype._synchronizedActions = null;
+
+    /**
      * Indicates whether or not the UI is currently locked.
      *
      * @private
      * @type {boolean}
      */
     FluxController.prototype._uiLocked = false;
+
+    /**
+     * The set of pending idle task promises.
+     *
+     * @private
+     * @type {Set.<Promise>}
+     */
+    FluxController.prototype._idleTasks = null;
 
     Object.defineProperties(FluxController.prototype, {
         "flux": {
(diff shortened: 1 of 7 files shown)


src/js/fluxcontroller.js
```diff
@@ -105,6 +105,8 @@ define(function (require, exports, module) {
         this._actionQueue = new AsyncDependencyQueue(CORES);
         this._initActionNames();
         this._initActionLocks();
+        this._sychronizedActions = new Map();
```
volfied: This is not declared in `FluxController.prototype` or anywhere, and is lacking JSDoc.

src/js/fluxcontroller.js
```diff
@@ -511,6 +522,122 @@ define(function (require, exports, module) {
...
+                    }
+
+                    if (!nextAction || (typeof nextAction !== "function")) {
+                        throw new Error("Exec passed an undefined action");
+                    }
+
+                    var params = Array.prototype.slice.call(arguments, 1),
+                        synchronizedAction = self._sychronizedActions.get(nextAction);
+
+                    return synchronizedAction.apply(null, params);
```
volfied: Shouldn't this be passing the `actionReceiver` as the binder?

src/js/fluxcontroller.js
```diff
@@ -511,6 +522,122 @@ define(function (require, exports, module) {
...
+                                    // Clean up the queue-activation handler
+                                    actionQueue.off("active", handleActive);
+                                    timer = null;
+                                    idle = null;
+
+                                    // Enqueue the task
+                                    this.enqueue.apply(this, params)
+                                        .then(resolve, reject);
+                                }.bind(this));
+                            }.bind(this), 3000);
```
volfied: 3 seconds seems a bit high, is there a specific reason you chose it?

iwehrman: Actually I was going to turn that down to 1 second, but I guess I forgot. There's no particular reason for that choice either though.

src/js/fluxcontroller.js
iwehrman: The synchronized action is already bound, so binding it again is a no-op.

src/js/fluxcontroller.js
iwehrman: Yes, good catch, I'll fix that.

iwehrman:
Commit amended to reduce queue timer to 1s, and also to fix the `_synchronizedActions` declarations (and to fix a consistent typo in that variable name).

volfied:
:+1: This is a nice utility to have

Issue #3307, opened 2015-11-25
Undo re-order of uninitialized layers failed


shaoshing:
When you re-order an uninitialized layer and hit undo, the LayerPanel is not updated until you click on the item to trigger an update. Please see video for repro steps:

https://www.dropbox.com/s/p7qbxxv4td99fzy/layer-reorder-undo.mov?dl=0

iwehrman:
Over to @chadrolfs to verify.

PR #3330, opened 2015-12-01
Loading document from history should initialize selected layers

mcilroyc:
This addresses #3307.  

It is possible for the current selection state (after undo/redo) to contain layers which were not initialized at the time the history state was captured.

Consider a newly loaded document with three shapes: 1, 2, & 3, with  Layer 1 selected.  Layers 2 and 3 are not fully initialized.

- Move Layer 2 below layer 3
- Layer 2 is now selected
- Undo
- Layer 2 is still selected, but must now be explicitly initialized

Please review @iwehrman.

### DIFF
--- a/src/js/actions/history.js
+++ b/src/js/actions/history.js
@@ -81,7 +81,8 @@ define(function (require, exports) {
     /**
      * Helper function to load a state from history store based on a count offset
      * and then to fix it up with the latest selection state, visibility state, and
-     * then reset border policies
+     * then reset border policies.  Selected layers are also initialized in case the current
+     * selection contains layers that were not initialized at the time of the previous history state
      *
      * @private
      * @param {number} documentID
@@ -92,7 +93,7 @@ define(function (require, exports) {
         return descriptor.getProperty(documentLib.referenceBy.id(documentID), "targetLayers")
             .bind(this)
             .catch(function () {
-                // no targetLayers property means no document is open
+                // no targetLayers property means no layer selected
                 return [];
             })
             .then(function (targetLayers) {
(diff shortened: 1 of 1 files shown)


src/js/actions/history.js
```diff
@@ -104,10 +105,13 @@ define(function (require, exports) {
                 });
             })
             .then(function () {
-                var currentDocument = this.flux.store("application").getCurrentDocument();
+                var document = this.flux.store("application").getCurrentDocument(),
```
iwehrman: Above in the `catch` clause it seems to imply that it's a possibility that no document will currently be open. I realize it's comparing apples to oranges somewhat, but is that something we should test for here? Should this bail out early if `getCurrentDocument()` returns null? Or is there some reason why you trust that that can never happen, despite the comment above?

iwehrman:
Just the one question. Seems reasonable. Once @pineapplespatula's controller changes go in, we should convert these private helper functions into private actions so that we can use `transfer` and make the locks easier to understand.

src/js/actions/history.js
mcilroyc: That catch clause (or, at least, the comment therein) makes absolutely no sense to me.  Assuming it was me that wrote it ... maybe I meant "no targetLayers property means no layer is selected"?.

Oh, looks like it was probably copied from actions/layers.js : 458.

I'll fix the comment ... or experiment with removing the `catch` completely ... which seems less safe, but 'tis the season.

mcilroyc: I confirmed that the catch clause is necessary because when no layers are selected, the `getProperty` throws an error.  I updated the comment.  I will admit to having stared at that comment before....and then blissfully ignoring it.  So, thanks.

mcilroyc:
back to you @iwehrman.  Comment fixed.  I agree that it sucks the way the helper functions are structured (2 deep!) and will gladly use the new private actions to fix it when the time comes.

iwehrman:
OK!

Issue #3329, opened 2015-12-01
Startup failure with pen tool and no doc


iwehrman:
1. From standard PS, select the pen tool
2. Switch to Design Space.

Result: complete failure.
```
Transfer from tools.select to toolPen.select failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Deselect” is not currently available.
    at eval (eval at undefined, <anonymous>:1:1)
    at eval (native)
From previous event:
    at Descriptor._batchPlayImmediate (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:394:14)
    at Descriptor.batchPlay (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:526:25)
    at Descriptor.batchPlayObjects (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:551:21)
    at Descriptor.playObject (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/bower_components/spaces-adapter/src/ps/descriptor.js:374:21)
    at Object.select (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/tool/pen.js:93:43)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:612:44)
From previous event:
    at FluxController._applyAction (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:611:14)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:474:37)
    at AsyncDependencyQueue._processNext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:291:27)
[...]

PR #3334, opened 2015-12-02
check for missing docuement in pen selection

pineapplespatula:
fix for issue #3329

### DIFF
--- a/src/js/actions/tool/pen.js
+++ b/src/js/actions/tool/pen.js
@@ -90,7 +90,13 @@ define(function (require, exports) {
         if (vectorMode) {
             selectVectorMask = descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
         } else {
-            selectVectorMask = descriptor.playObject(vectorMaskLib.dropPathSelection());
+            var applicationStore = this.flux.store("application"),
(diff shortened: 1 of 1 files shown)


src/js/actions/tool/pen.js
```diff
@@ -90,7 +90,11 @@ define(function (require, exports) {
         if (vectorMode) {
             selectVectorMask = descriptor.playObject(vectorMaskLib.activateVectorMaskEditing());
         } else {
-            selectVectorMask = descriptor.playObject(vectorMaskLib.dropPathSelection());
+            var applicationStore = this.flux.store("application"),
+                currentDocument = applicationStore.getCurrentDocument();
+            if (currentDocument) {
+                selectVectorMask = descriptor.playObject(vectorMaskLib.dropPathSelection());
```
iwehrman: It may well be perfectly fine to not initialize this variable if there is no current document w.r.t. `Promise.join`, but... we are usually careful to do this elsewhere. It seems like we should be consistent.

pineapplespatula: good catch. fixed

mcilroyc:
LGTM

Issue #3324, opened 2015-12-01
Datalist: typing the first character does not filter the dropdown list


shaoshing:
![datalist](https://cloud.githubusercontent.com/assets/118264/11489172/b661ec6e-9781-11e5-8ed5-2fcbb8b0c4d7.gif)

designedbyscience:
Confirmed fixed

PR #3335, opened 2015-12-02
Make dropdown list starts filtering for the first input character

shaoshing:
Fix for issue #3324 

`setState` is asynchronous, and it is not safe to read the state immediately after writting to it. In this case, `this.state.editing` (in line 162 ) may remian unchanged until the last `setState` get processed. Also, I don't think we need to check `this.state.editing` there as it is always set to true in this function, so I simply remove the checker to fix the issue.


mcilroyc:
This looks good to me, but assigning to @designedbyscience to double check

designedbyscience:
:+1:

Issue #3238, opened 2015-11-13
Sampling locked layer has unexpected results


chadrolfs:
Create two shape layers (1 and 2) with different Fill colors
Lock layer 1 and have it selected/targeted
click on the Sampler tool
mouse over layer 2 and click

Result: Locked layer 1 assumes fill color of layer 2.  If you do the opposite, select/target layer 2 (unlocked) and attempt to sample the locked layer 1, it won't.

Expected: If you have a locked layer targeted/selected, Sampling should not work.

volfied:
FBNC to @chadrolfs

chadrolfs:
hmm, this doesn't look fixed.  pgdev.915 master @ 62e89f2

chadrolfs:
still works oppositely (as stated in steps above)

shaoshing:
Hi @chadrolfs I am on dev 919 and it is working correctly. Can you give it a second try?
![screen shot 2015-12-07 at 9 47 18 am](https://cloud.githubusercontent.com/assets/118264/11634952/6eb1087c-9cc8-11e5-87ee-aa2f213d8e1e.png)
![screen shot 2015-12-07 at 9 47 26 am](https://cloud.githubusercontent.com/assets/118264/11634953/6eb44c80-9cc8-11e5-9e55-37bf93798571.png)

chadrolfs:
I have the locked layer selected and I sample an unlocked layer and it samples when it shouldn't.

![locked-sample](https://cloud.githubusercontent.com/assets/8442631/11671890/347742ae-9dc0-11e5-976c-2be1de410d5f.gif)

volfied:
FBNC to @chadrolfs

chadrolfs:
almost!
locked layer selected > sample anything fails correctly YAY!
unlocked layer selected > sample anything (including locked layer) doesn't sample locked layers. :(

PR #3336, opened 2015-12-02
Fix for sampling locked layer.

shaoshing:
Fix for issue #3238 

`layerTree.leaves` ignores locked layers, which causes sampling locked layer failed (no sampler HUD displayed). I change the way of retrieving the top hit layer by getting the first id from `topHitLayerID`, which is the top most hit layer id.

### DIFF
--- a/src/js/actions/sampler.js
+++ b/src/js/actions/sampler.js
@@ -414,12 +414,17 @@ define(function (require, exports) {
         return uiUtil.hitTestLayers(doc.id, coords.x, coords.y)
             .bind(this)
             .then(function (hitLayerIDs) {
-                var hitLayerMap = new Set(hitLayerIDs.toJS()),
-                    clickedLeafLayers = layerTree.leaves.filter(function (layer) {
-                        return hitLayerMap.has(layer.id);
-                    }),
-                    topLayer = clickedLeafLayers.last(),
+                var topLayer,
                     selected = layerTree.selected;
+                
+                hitLayerIDs.forEach(function (layerID) {
+                    var layer = layerTree.byID(layerID);
+                    
+                    if (!layer.isArtboard) {
+                        topLayer = layer;
+                        return false;
+                    }
+                });
 
                 // If we're trying to sample the only selected layer, surely it's a no-op
                 if (!topLayer || selected.size === 1 && selected.first() === topLayer) {


src/js/actions/sampler.js
```diff
@@ -409,17 +409,15 @@ define(function (require, exports) {
         var uiStore = this.flux.store("ui"),
             docStore = this.flux.store("document"),
             coords = uiStore.transformWindowToCanvas(x, y),
-            layerTree = doc.layers;
+            layers = doc.layers;
```
mcilroyc: I'm not saying I love the `layerTree` var name, but just pointing out that I think it has been used in the past to differentiate the "layer structure" from a simple list of layers.

mcilroyc:
Assigning to @volfied.  I know that ordering is crucial to this, and to my eye this new method seems problematic.

src/js/actions/sampler.js
shaoshing: Thanks for the tip. when I see `layerTree` I am always expecting a tree structure and kind of confusing. maybe the name `layerstructure` is better in this case? anyway I will change the name to maintain consistency.

volfied:
This doesn't seem right... in old method we're grabbing leaf layers and the last one after filtering them, but in your new code you're grabbing the top one from hitLayerIDs.

We want to make sure we're not trying to sample artboards/groups etc. 

@shaoshing Can you, with absolute certainty, prove that these two codes return the same layer in the same document (barring the locked problem you're trying to solve)?

src/js/actions/sampler.js
```diff
@@ -414,12 +414,17 @@ define(function (require, exports) {
...
             .then(function (hitLayerIDs) {
-                var hitLayerMap = new Set(hitLayerIDs.toJS()),
-                    clickedLeafLayers = layerTree.leaves.filter(function (layer) {
-                        return hitLayerMap.has(layer.id);
-                    }),
-                    topLayer = clickedLeafLayers.last(),
+                var topLayer,
                     selected = layerTree.selected;
+                
+                hitLayerIDs.forEach(function (layerID) {
```
shaoshing: The first id in `hitLayerIDs` is the top layer id.

src/js/actions/sampler.js
```diff
@@ -414,12 +414,17 @@ define(function (require, exports) {
...
+                    var layer = layerTree.byID(layerID);
+                    
+                    if (!layer.isArtboard) {
```
shaoshing: `hitLayerIDs` already excludes group layers, so here we just need to make sure the layer is not an art board.

src/js/actions/sampler.js
```diff
@@ -414,12 +414,17 @@ define(function (require, exports) {
...
+                        topLayer = layer;
+                        return false;
```
shaoshing: stop the forEach loop here

shaoshing:
Hi @volfied I've updated the code to handle group and art board propertly. Please give it a second look.

volfied:
Alrighty then!

Issue #3328, opened 2015-12-01
Opening file where artboard is selected by default blinks adders (resetBorderPolicies x2)


chadrolfs:
new doc
drag a LLSO (that is an artboard assets selected by default) onto canvas (note it auto-commits)
hit Enter (opens LLSO)

Result: blink of adders (reset border policy twice)

![llso-openedit-blink](https://cloud.githubusercontent.com/assets/8442631/11514007/37ddad1a-982b-11e5-9314-1e37640918ef.gif)

chadrolfs:
![screen shot 2015-12-01 at 1 02 54 pm](https://cloud.githubusercontent.com/assets/8442631/11514142/d20f8ca0-982b-11e5-9041-43c9d79115a4.png)

volfied:
Well, I am completely baffled by this. Problem is, for a brief second, uiStore disables overlays, and this causes Scrim to re-render and render `null` for the superselectOverlay. Then the overlaysEnabled turns on again, and Scrim re-renders the superselectOverlay correctly, causing the blink.

I thought this was a draw debouncing issue, or fix could be provided there, but because React is unmounting the component, our drawOverlay doesn't really do much. The parent component is being taken our from under us. 

@iwehrman, ui store's `overlaysEnabled`, do you think we can debounce the emit("change") for that so that if the `overlaysEnabled` switches to false and quickly switches back to true, it doesn't emit a change event? How terrible would that be?

iwehrman:
FBNC to @chadrolfs. 

@volfied: Yes, I think we could try that. I did try that before, but then ran into _other_ issues. By the time I gave up I'd unraveled half of the sweater.

PR #3337, opened 2015-12-02
Correctly check for inequality in transform matrices during policy reset

volfied:
```
> [0,1] === [0,1]
< false
```

So we should use `_.isEqual` instead to check equality.

Doesn't really address #3328, but that one is a lot hairier than it looks.


iwehrman:
Less bad!

Issue #3338, opened 2015-12-02
StoreWatchMixin broken after controller reset


iwehrman:
After the controller resets due to an error, React components that registered for changes via the `StoreWatchMixin` no longer receive change events. To test:

1.` Help > Test Action Failure`
2. Zoom in or out and observe that the `Zoom` component in the toolbar does not change.

volfied:
This only happens with UI store, as far as I can tell.. hmmm

iwehrman:
Fixed by #3343.

PR #3343, opened 2015-12-03
Set the correct border policy change handler, SCOPES

volfied:
This is the victim of bad indentation and bad scope.

`tools.beforeStartup` was setting a local `_borderPolicyChangeHandler` variable, instead of the file scoped one. So during onReset, we would call `removeListener` on various stores with an `undefined` handler, which would cause the store to drop ALL the listeners, turning the app unresponsive after a reset.

Addresses #3338


iwehrman:
LGTM.

PR #3348, opened 2015-12-03
Color theme switching

iwehrman:
This PR makes it possible to change the UI color stop (theme) from Design Space via `Window > Color Theme`. Changing the color stop updates the UI theme, the titlebar color and also the artboard backdrop color. Notes:

1. Currently there is a Photoshop bug that prevents the artboard backdrop color from being updated until the canvas is panned or otherwise invalidated, but @jsbache is looking into that.
2. This moves the `color-stop__XYZ` class from `body` to the `Main` React element, which is now in charge of deciding what stop to render.
3. The stop is still fetched at application start time so that the first render has the correct theme, but it's also fetched in `ui.beforeStartup` so that the correct value can make it into the store. This is slightly redundant, but tolerable I think.
4. I made some unrelated whitespace cleanups in main.less since I was in there anyway.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -41,7 +41,7 @@
     "requirejs": "~2.1.20",
     "requirejs-i18n": "~2.0.6",
     "requirejs-text": "~2.0.14",
-    "spaces-adapter": "https://github.com/adobe-photoshop/spaces-adapter.git#^0.28.0",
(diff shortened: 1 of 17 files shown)


iwehrman:
The aforementioned PS bug: https://jira.corp.adobe.com/browse/PS-452

volfied:
Requires latest and greatest, but works!

jsbache:
Fixed in team sandbox with CL: 1054893. Please verify

Issue #3149, opened 2015-10-27
Alt/Opt + drag artboard to duplicate after zoom out shifts overlay highlight/bounds


chadrolfs:
cmd/ctrl + n
draw a couple rects
zoom out a few times (cmd/ctrl + -)
with Select tool click + drag on AB title with Alt/Opt key down to duplicate
marquee select

Result: overlay highlight/bounds is shifted.

chadrolfs:
![highlight-afterzoom](https://cloud.githubusercontent.com/assets/8442631/10772171/ed52339a-7cb1-11e5-8b26-8834fd7f5184.gif)

volfied:
Can't repro currently.

chadrolfs:
blocked due to highlighting being broken

volfied:
@chadrolfs Tried couple times and I can't repro this. If you can, can you add more precise steps?

chadrolfs:
Odd, I can repro on 2 different machines

chadrolfs:
close all open doc, reset DS, then start from the top.  There aren't really any additional steps to what's above.

mcilroyc:
please validate, @chadrolfs

chadrolfs:
validated

PR #3353, opened 2015-12-03
Update transform after a duplicate drag

volfied:
Not entirely sure on the root cause of bug #3149, but from running my snippets I could see that it wasn't the bounds that were the culprit, but the transformation. We get `move` and `moveToArtboard` notifications, and we must be updating our transform before Photoshop finished settling the transform matrix.

I know this is vague, but updating the transform after the duplicate solves the problem.

Addresses #3149

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1302,7 +1302,11 @@ define(function (require, exports) {
                 // a somewhat more optimistic copy routine, instead of addLayers
                 // which doesn't know that the layers being added are copies of
                 // existing layers.
-                return this.flux.actions.layers.addLayers(currentDoc, toIDs, true, false);
+                return this.flux.actions.layers.addLayers(currentDoc, toIDs, true, false)
+                    .bind(this)
+                    .then(function () {
+                        return this.flux.actions.ui.updateTransform();
+                    });
             } else {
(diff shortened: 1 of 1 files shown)


volfied:
Btw, another solution is to use the new `whenIdle` API to enqueue `updateTransform` on drag. But that causes a second delay and looks bad.

iwehrman:
There are a lot of problems with this whole function. Note that the way debounce is used here is completely broken. Also, it seems like this should be converted into an action unless there's a good reason not to.

mcilroyc:
@iwehrman what do you mean by "completely broken"?  If this is related to my comment 'debouncing is critical even for just one "move" ...', then I should point out that we could fix that another way ... IF we determined that we don't need the standard debounce functionality (ie, if we are sure we don't need to handle a series of quick 'move' events.

iwehrman:
For every event, we create a new debounced function and then call it and then it goes out of scope. That is equivalent to just calling the function after a fixed delay, I believe. A debounced function can only really debounce if you call it more than once. So, I assume that was not the intent of this code?

mcilroyc:
Created an issue to track further improvements.  https://github.com/adobe-photoshop/spaces-design/issues/3391

Issue #3357, opened 2015-12-04
Artboards get clipped outside original area after transform


ktaki:
I do not know how to describe this, but please follow the steps to reproduce. Feel free to change the title.
 
New AB doc
Add a rectangle
Set W to 300
tab
Set H to 300
tab
Set X to 0
Hit return
Cmd-z
V to select superselect tool
Cmd-A
Grab the artboard title to move artboard

![screen shot 2015-12-03 at 4 02 40 pm](https://cloud.githubusercontent.com/assets/2029597/11578115/5c1c5688-99d7-11e5-9990-d096fff2d382.png)

volfied:
Happens on Core as well (as in, you can switch to core, and it stays that way. We should let the artboards team know of this issue.

ktaki:
Yes, it stays that way when switched to standard Ps, but I cannot make it happen in standard Ps. Note that this does not happen with 16.1 GM DS. I can contact cjb.

ktaki:
Verified the fix.

PR #3359, opened 2015-12-04
Play setSize and swap in a transaction with autoExpand dance included

volfied:
We were playing the `setAutoExpand` calls outside the `playLayerActions` calls, so on an undo, we would get back to the state where we turned auto expand off.

Now we use the handy transactions to make sure the autoExpand dance is included in the history state.

Addresses #3357

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -484,14 +484,6 @@ define(function (require, exports) {
                 
         var dispatchPromise = this.dispatchAsync(events.document.history.REPOSITION_LAYERS, payload);
 
-        // Make sure to show this action as one history state
-        var options = {
-            historyStateInfo: {
-                name: nls.localize("strings.ACTIONS.SWAP_LAYERS"),
-                target: documentRef
-            }
-        };
-
         var autoExpandEnabled = false;
 
         var hasArtboard = layers.some(function (layer) {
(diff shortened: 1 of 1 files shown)


src/js/actions/transform.js
```diff
@@ -504,6 +504,11 @@ define(function (require, exports) {
             headlights.logEvent("edit", "layers", "swap-non-artboard");
         }
 
+        var transaction = descriptor.beginTransaction(options),
```
mcilroyc: style nit: it seems a little awkward how `options` is dangling a dozen or so lines above.

src/js/actions/transform.js
```diff
@@ -533,6 +538,9 @@ define(function (require, exports) {
             })
             .then(function () {
                 return this.transfer(layerActions.resetIndex, document);
+            })
+            .then(function () {
```
mcilroyc: My gut says that the transaction should be ended prior to calling resetIndex ... but I don't actually know that it matters.

volfied: Good catch, it might not matter but it's probably safer to end it in the correct place for future proofing.

volfied:
Thanks @mcilroyc !

mcilroyc:
alrighty

src/js/actions/transform.js
```diff
@@ -532,6 +535,9 @@ define(function (require, exports) {
                 }
             })
             .then(function () {
+                descriptor.endTransaction(transaction);
```
iwehrman: Doesn't this return a promise that you need to wait for???

Issue #3258, opened 2015-11-17
Errors on closing a doc while in text editing mode


ktaki:
Tested with pg-dev Mac build 886.

Create a text layer. While in text editing mode, Cmd-W to close the doc. Click on Do not save.

Action layers.resetLayers failed: Error: unknown exception type

![screen shot 2015-11-17 at 11 04 25 am](https://cloud.githubusercontent.com/assets/2029597/11221784/c23bd27c-8d1b-11e5-8a87-f5958e55c58e.png)
![screen shot 2015-11-17 at 11 04 41 am](https://cloud.githubusercontent.com/assets/2029597/11221787/ca45ad80-8d1b-11e5-8f51-c8ebfa87d311.png)

mcilroyc:
fixed by ref'd PR.  back to @ktaki

ktaki:
Verified the fix.

PR #3361, opened 2015-12-04
Short circuit resetLayers if passed in doc no longer exists

volfied:
Addresses #3258, PS would diligently send us the `make` event for the text layer under edit as we're closing the document, and we would enqueue the `resetLayers` action. And since the document was gone, things would go on the fritz. Now we check to see if the document model still exists, similar to what handleHistoryState does.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -481,6 +481,13 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
+        var storedDocument = this.flux.store("document").getDocument(document.id);
+
+        if (!storedDocument) {
+            log.debug("Ignoring this resetLayers call, the document does not exist");
+            return Promise.resolve();
+        }
+
         var docRef = documentLib.referenceBy.id(document.id),
             layersPromise;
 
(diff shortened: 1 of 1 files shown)


src/js/actions/layers.js
```diff
@@ -482,6 +482,13 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
+        var upToDateDocument = this.flux.store("document").getDocument(document.id);
```
mcilroyc: It sure would be nice to just change the first argument of `resetLayers` to be documentID.  Another day I guess.  Something about `upToDateDocument` rubs me the wrong way, as a variable name.

mcilroyc: Splitting hairs further ... really we probably should take this a step further and validate that document.id === appStore.currentDocumentID right?

iwehrman: In theory we allow resetting layers in the non-active document?

iwehrman: @mcilroyc @volfied: I agree that there is a problem with our pattern of passing in an immutable model in actions. We should talk at the upcoming team meeting about the best possible replacement pattern.

mcilroyc: Well, I'll make one last-ditch effort to suggest that we add a "document exists" check to the document store, but I won't stand in the way of this PR because it seems sufficient.

volfied:
@mcilroyc I've changed the variable name. As far as changing our action variable pattern goes, I'd rather do those all at once after we decide on what to do.

PR #3362, opened 2015-12-04
Move from bower to npm, upgrade React to 0.14.0

volfied:
Next step in the build system, so we have only one dependency manager. 

I've also upgraded React to 0.14.0 while at it. React is pretty good at warning the dev if you use a deprecated API, but key changes are:
 - `findDOMNode` is part of `react-dom`.
 - `this.refs.[bla]` will return what used to be `this.refs.[bla].getDOMNode()`
 - React addons are all need to be required separately, in form of `react-addons-addon-name`

Next steps will be to clean up the `resolve` section in webpack config so we require these libraries by their real names. And ditch the AMD format in our files (big change)


iwehrman:
Remind me why @shaoshing's custom JSDoc plugin is no longer needed?

package.json
```diff
@@ -38,20 +53,35 @@
...
+    "immutable": "^3.7.5",
+    "lodash": "^3.10.1",
+    "loglevel": "^1.4.0",
+    "mathjs": "^2.4.2",
+    "react": "^0.14.2",
+    "react-addons-pure-render-mixin": "^0.14.2",
+    "react-dom": "^0.14.2",
+    "scriptjs": "^2.5.8",
+    "spaces-adapter": "adobe-photoshop/spaces-adapter.git",
+    "tinycolor2": "^1.1.2",
```
iwehrman: Can we alias this to `tinycolor` so that we don't have to change all the require calls?

src/index-debug.html
```diff
@@ -38,6 +38,6 @@
     </script>
 </head>
 <body>
-    <div class="main"></div>
+    <div class="main" id="app"></div>
```
iwehrman: What purpose do these ids serve?

package.json
volfied: I wanted to avoid resolving, personal preference.

src/js/main.js
```diff
@@ -121,7 +123,7 @@ define(function (require, exports) {
             });
 
         var renderPromise = new Promise(function (resolve) {
-            React.render(new Main(props), window.document.body, function () {
+            ReactDOM.render(new Main(props), window.document.getElementById("app"), function () {
```
iwehrman: Hmm, I guess this is why? But, you should still just use the class instead of adding a new ID.

iwehrman: Actually, I don't really understand this change either. Please explain!

volfied:
@iwehrman @shaoshing's custom JSDoc plugin added a special case for all the React functions we didn't add a JSDoc to. I've instead added an option to ignore expression functions when checking for JSDoc. I wanted to try to avoid using any custom libraries, since they eventually go stale.

package.json
iwehrman: Not sure what you mean by "avoid resolving". I meant aliasing it in the webpack config.

iwehrman:
I seem to recall that ignoring expressions also ignores a lot of things that we currently check. So, making this change doesn't cause extra warnings, but wouldn't it miss some JSDoc that we already have? I could be remembering incorrectly...

volfied:
We are also running grunt-jsdoc and that complains with bad jsdoc as well. We might be dropping some of the jsdocs we miss, but in terms of upgradability I think it's better to keep what the project provides.

Ideally, @shaoshing would make a PR to jscs-jsdoc and get some Github fame, and we would use his settings once more.

iwehrman:
`.bowerrc` should be removed, and the reference to `bower_components` should be removed from .gitignore.

iwehrman:
Anyway, this seems fine. I pushed a commit that removed that file and that line. So, let's just do it.

PR #3367, opened 2015-12-07
Small NumberInput cleanup, consistent property naming

designedbyscience:
NumberInput usage was already pretty clean across Design Space. Similar to TextInput, I cleaned up some old properties, removed the property expansion that had existed, added a few notes for the different properties.

Additionally, moved everything to use `placeholder` as the property name for placeholder text. This is the HTML property name, NumberInput and TextInput had slightly varying names.


shaoshing:
Looks good to me!

PR #3368, opened 2015-12-07
Datalist refactoring

shaoshing:
This PR made the following improvements:

* added `onHighlightedChange` callback to distinguish with `onChange`. The first one is fired when the current highlighted option is changed (mouse over or using arrow key), and the latter is fired when the user confirm the selection via mouse click / return / enter.
* Removed `live` `autoSelect ` prop, as you can use the `onHighlightedChange` instead. Also clean up some internal state.
* added `onClose` (you can search `Datalist~onClose` for details). With this we should be able to revet to initial selection if user close the list with selecting the highlighted option for Type/Blend mode, but things get more complicated when it comes to MIXED state, so I did not introduce the change in this PR yet.
* Better jsdoc for callbacks.
* Bug fix: disabled Tab key in the search dialog. otherwise the search input will lose focus while the dialog is still open, which does not look right.

@designedbyscience can you please review? Thank you.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -403,7 +403,8 @@ define(function (require, exports, module) {
         /**
          * Perform action based on ID
          *
-         * @param {string} id
+         * @private
+         * @type {Datalist~onInputChange}
          */
         _handleChange: function (id) {
             if (id === null) {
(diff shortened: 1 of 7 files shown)


shaoshing:
@designedbyscience I've addressed your commends, which I don't know why they are not here any more.

designedbyscience:
Yes, that is strange...

Anyways, looks good. I like the changes

Issue #3366, opened 2015-12-06
Webpack watcher aborts when compilation fails


iwehrman:
The webpack watcher currently aborts whenever compilation fails. This happens all the time during development, and it's quite frustrating to have to constantly re-start the watcher. It would be a lot cooler if it did not do this.

PR #3369, opened 2015-12-07
Keep webpack watcher alive on compilation errors

volfied:
Simple fix for #3366, should make developing a bit faster.


iwehrman:
Thanks!

Issue #3331, opened 2015-12-02
Export Presets on 2x artboards need to adjust export sizes accordingly


placegraphichere:
- create artboard, say for an iPhone6
- create artwork 
- add iOS preset exports on the artwork
- artwork will think it's at 1x, but should be at 2x since the templates are sized at 2x

- expected: artwork should respect the artboard "x" size and scale up/down accordingly for each asset

-User feedback:
https://twitter.com/manfredalbrecht/status/672005189285584896

I think I know what we can do here, I may try and fix.

chadrolfs:
https://watsonexp.corp.adobe.com/#bug=4096099 for 16.1.1 BRC

tomwmcrae:
How will a user understand how the "0.33" and "0.67" additions will help them?

chadrolfs:
via @placegraphichere: 
Designing at 3x for say the iPad Pro, assets that need to be scaled to a 1x and 2x  size would need to use .33x or .67x.
for the iPhone 6, you are designing at 2x with the template we provide. So going to iPadPro needs to be 1.5x. For 1x its 0.5x.
there is a concept of “designed at” that is hard to articulate to users, since there is no real standard design size when using a pixel based artboard
.33x and .67x are for iPad pro. So if you design at 3x, the 1x assets need to be 0.33, 2x assets need the 0.67

ktaki:
0.33 and 0.67 have been added to Ps 16.1.2. Are there any further tasks for this issue?

iwehrman:
The additional export sizes were only a very partial workaround for this bug. It is no longer fix in progress, so I've removed that label.

PR #3370, opened 2015-12-07
Add 0.33 and 0.67 export scales [master]

iwehrman:
This is a tiny workaround for #3331. This does not fully address the issue. This pull request is against master; there will be a different PR against the `ps-16.1` branch.


mcilroyc:
just confirming... we aren't adding these to either of the "preset" groups?

PR #3371, opened 2015-12-07
Add 0.33 and 0.67 export scales [ps-16.1]

iwehrman:
This is the companion to #3370 which updates the `ps-16.1` branch. Please test this against the shipping version of Photoshop 16.1.


iwehrman:
Note that these CI checks are failing due to the fact that a recently updated patch version of the JSDoc grunt task is returning new warnings. I believe the warnings are harmless. This bug is captured in #3372.

mcilroyc:
scales are free

PR #3373, opened 2015-12-07
Notify webpack build status via system notifications.

shaoshing:
* You will receive a system notification everytime when webpack rebuilds. This might have better experience than reloading Design Space without knowing whether your change is integrated. 
* Syntax error will not abort `grunt debug`. It will send system notification instead.

screen shot 2015-12-07 at 11 06 31 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,8 @@
     "uglifyjs": "^2.4.10",
     "url-loader": "^0.5.6",
     "webpack": "^1.12.3",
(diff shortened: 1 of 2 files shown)


iwehrman:
Needs a rebase!

shaoshing:
Rebased!
nzg9ojt

volfied:
Isn't that handy?

PR #3374, opened 2015-12-07
Factor panel actions/store/lock out of ui actions/store/lock

iwehrman:
This PR factors a panel action module, store and lock out from the existing UI action module, store and lock. This reduces contention for the UI lock somewhat at startup time, and also arguably also makes this  code a little easier to understand. In some cases it wasn't clear what state/action should go where, but the rule of thumb I tried to use was: if it has to do with the _window_ put it in the UI store, and if it has to do with a _panel_ put it in the panel store. (We should probably rename the UI store to something more appropriate, but I don't currently have a suggestion, so I left it alone.) This seems to shave ~100ms off startup time on my MBP due to reduced lock contention.

### DIFF
--- a/src/js/actions/ui.js
+++ b/src/js/actions/ui.js
@@ -228,13 +228,13 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var setOverlayCloaking = function () {
-        var uiStore = this.flux.store("ui"),
-            cloakRect = uiStore.getCloakRect();
+        var panelStore = this.flux.store("panel"),
+            cloakRect = panelStore.getCloakRect();
 
         return adapterUI.setOverlayCloaking(cloakRect, ["scroll"], "afterPaint");
     };
     setOverlayCloaking.action = {
-        reads: [locks.JS_UI],
+        reads: [locks.JS_PANEL],
         writes: [locks.PS_APP],
         modal: true
     };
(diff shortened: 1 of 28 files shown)


volfied:
Feels like marquee related information should be in UI store. Or maybe I don't understand the logic enough yet.

iwehrman:
The hope was to move more of the app/feature-specific stuff out of the UI store. So, I disagree!

src/js/actions/ui.js
```diff
@@ -290,22 +290,22 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var updatePanelSizes = function (sizes) {
-        return this.dispatchAsync(events.ui.PANELS_RESIZED, sizes)
+        var transformPromise = this.transfer(updateTransform),
+            dispatchPromise = this.dispatchAsync(events.panel.PANELS_RESIZED, sizes)
             .bind(this)
```
volfied: This needs to be indented.

volfied:
You forgot some `events.ui.TOGGLE_OVERLAYS` (at least one in transform.js#beforeStartup)

volfied:
I pushed a commit myself fixing this, since I'm likely to merge it after giving it one more look.

volfied:
We can probably sit and argue about the dividing line all day, but since this helps startup speed and doesn't hurt anything, I'm merging it. And will make another PR for more lock cleanup.

PR #3379, opened 2015-12-07
Initialize debug snippet hook-ups correctly

volfied:
`__FLUX_CONTROLLER__` was `undefined` before, because it wasn't initialized at assignment time.

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -116,6 +116,14 @@ define(function (require, exports) {
             flux: _controller.flux,
             initialColorStop: stop
         };
+        
+        if (__PG_DEBUG__) {
+            // Expose these for snippet usage, only available in debug builds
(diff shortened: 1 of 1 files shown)


src/js/main.js
```diff
@@ -116,6 +116,14 @@ define(function (require, exports) {
             flux: _controller.flux,
             initialColorStop: stop
         };
+        
+        if (__PG_DEBUG__) {
+            // Expose these for snippet usage, only available in debug builds
+            window.__PS_ADAPTER__ = adapter;
+            window.__FLUX_CONTROLLER__ = _controller;
+            window.__LOG_UTIL__ = log;
+            window.__PERF_UTIL = performanceUtil;
```
iwehrman: Also, the underscores are inconsistent...

Issue #3271, opened 2015-11-18
Selecting a layer group may result in collapsing the group.


shaoshing:
This occurs on some layers.

![select-layer](https://cloud.githubusercontent.com/assets/118264/11253313/92883898-8def-11e5-8378-9c0862b0f7c4.gif)

Reproduce
1. open `VermilionArtboards.psd`
2. expand art board `Desktop Large` while holding OPT key
3. scroll to `Fashion` group (it should be expanded)
4. select the group, and it is selected but also collapsed

shaoshing:
JIRA: https://jira.corp.adobe.com/browse/PS-439

shaoshing:
These groups appears to be expanded, but switching to classic shows that they are actually collapsed.

mcilroyc:
to @chadrolfs - note the required build of mac dev >= 932

PR #3381, opened 2015-12-07
Fix for expand/collapse all groups

shaoshing:
**This PR requires dev >= 932 and Adapter PR https://github.com/adobe-photoshop/spaces-adapter/pull/174**

Fix issue #3271 that not all descendant groups are expanded when option-click a group.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -817,11 +817,11 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer|Immutable.Iterable.<Layer>} layers
-     * @param {boolean} expand If true, expand the groups. Otherwise, collapse them.
-     * @param {boolean=} descendants Whether to expand all descendants of the given layers.
+     * @param {boolean} expand - If true, expand the groups. Otherwise, collapse them.
+     * @param {boolean=} recursive - Whether to expand/collapse all descendants groups of the given layers.
      * @return {Promise}
      */
-    var setGroupExpansion = function (document, layers, expand, descendants) {
+    var setGroupExpansion = function (document, layers, expand, recursive) {
         if (layers instanceof Layer) {
             layers = Immutable.List.of(layers);
         }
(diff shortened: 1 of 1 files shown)


src/js/actions/layers.js
```diff
@@ -888,14 +890,14 @@ define(function (require, exports) {
         }
 
         var documentRef = documentLib.referenceBy.id(document.id),
-            layerRefs = layers
+            layerRefs = targetLayers
```
shaoshing: the descriptor no longer needs all descendant layer group IDs in order to expand/collapse descendant groups recursively.

iwehrman:
Out of curiosity, does this provide any sort of performance improvement?

src/js/actions/layers.js
```diff
@@ -835,7 +835,9 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
-        if (descendants) {
+        var targetLayers = layers;
```
iwehrman: Can we just get rid of this extra variable?

iwehrman: Eh, I guess not...

shaoshing:
Possibily since the descriptor only needs to check the layer list once with recursive option (just my guess), but I doubt it is significant enought to have noticiable impact on the UI.

src/js/actions/layers.js
```diff
@@ -817,11 +817,11 @@ define(function (require, exports) {
      *
      * @param {Document} document
      * @param {Layer|Immutable.Iterable.<Layer>} layers
-     * @param {boolean} expand If true, expand the groups. Otherwise, collapse them.
-     * @param {boolean=} descendants Whether to expand all descendants of the given layers.
+     * @param {boolean} expand - If true, expand the groups. Otherwise, collapse them.
```
mcilroyc: nit: this is not an established pattern (yet?) of separating the param name from the description with a `-`.  Convince me we should?

mcilroyc:
jsdoc nit aside, this seems to work.  merging.

src/js/actions/layers.js
shaoshing: I don't think so, although I am seeing some jsdoc are using this pattern.

iwehrman: There should not be a dash separating the param name from the description.

PR #3387, opened 2015-12-07
Make source map works for both internal and external dev tools.

shaoshing:
By using the `inline-source-map` option https://webpack.github.io/docs/configuration.html#devtool
Tested on dev 926.


volfied:
Crazy, this used to crash CEF before we updated. Thank you for finding out that this now works!

:tada: :tada: :tada:

Issue #3197, opened 2015-11-04
Unable to align children within artboard


chadrolfs:
cmd/ctrl + n
draw some shapes and group them (within AB)
(in layers panel) select the group and the artboard layer (cmd/ctrl + click)

Result: align icons/cmds are disabled  (in classic PS, the are not)

volfied:
FBNC to @chadrolfs

chadrolfs:
fix for this may have caused/exposed this bug #3468

PR #3388, opened 2015-12-07
No nesting except ABs rule, used for align commands

volfied:
This PR enables Align operations for artboards, given all other rules still apply.

We were using `hasStrictSelectedAncestors` for both Align and Distribute buttons. However, artboards are alignable ancestors as far as Photoshop is concerned, so I've broken the disabled checks for the UI buttons into two separate functions, and use a custom checker for Align, while still using the available functions for Distribute.

Also introduced a new menu rule: `no-nesting-except-artboards`, where nesting is allowed in selected layers as long as it's the artboard in the selection.

This addresses #3197.

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -213,13 +213,18 @@ define(function (require, exports, module) {
                 (document !== null) &&
                 !document.unsupported &&
                 (document.layers !== null) &&
-                (document.layers.selectedNormalized.size === 2),
+                (document.layers.selected.size === 2),
             "layers-selected-2+":
                 (document !== null) &&
                 !document.unsupported &&
                 (document.layers !== null) &&
+                (document.layers.selected.size > 1),
+            "layers-selected-normalized-2+":
+                (document !== null) &&
+                !document.unsupported &&
+                (document.layers !== null) &&
                 (document.layers.selectedNormalized.size > 1),
-            "layers-selected-3+":
+            "layers-selected-normalized-3+":
                 (document !== null) &&
                 !document.unsupported &&
                 (document.layers !== null) &&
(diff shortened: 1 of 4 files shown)


src/js/models/menubar.js
```diff
@@ -219,7 +219,7 @@ define(function (require, exports, module) {
                 (document !== null) &&
                 !document.unsupported &&
                 (document.layers !== null) &&
-                (document.layers.selectedNormalized.size > 1),
+                (document.layers.selected.size > 1),
```
mcilroyc: why is the "layers-selected-2" rule (above) still hanging on with `selectedNormalized` ?

src/js/models/menubar.js
```diff
@@ -253,6 +253,16 @@ define(function (require, exports, module) {
                         return layer !== ancestor && document.layers.selected.contains(ancestor);
                     });
                 })),
+            "no-nesting-except-artboards":
+                (document !== null) &&
+                !document.unsupported &&
+                (document.layers !== null) &&
+                !(document.layers.selected.some(function (layer) {
+                    return document.layers.ancestors(layer).some(function (ancestor) {
```
mcilroyc: mild suggestion of putting this into `layerstructure` as a cached property ... since it seems slightly intensive.

mcilroyc:
@volfied one question, one suggestion (not demand).  Otherwise looks solid.

src/js/models/menubar.js
volfied: `layers-selected-2+` was only used by align commands. Normalizing gets rid of selected children of other selected layers, which is not what we want for align.

Since Align is the only client of `layers-selected-2+`, I went ahead and changed only that, but I think your confusion is justification enough that we may need a separate rule for align and leave all `layer-selected-#` rules with `selectedNormalized`

Will revisit.

src/js/models/menubar.js
volfied: We already had `hasSelectedAncestors`, but in this case I needed to ignore artboards in those selected ancestors. I was trying to avoid adding more cached properties like `hasSelectedAncestorsButNoArtboards`... but in this case I'm trading performance for code size, which is not a good trade.

Will revisit.

volfied:
@mcilroyc Re-visited this and cleared it up. Please review.

mcilroyc:
Just shy of a month old, not bad!
merging.

Issue #3350, opened 2015-12-03
X and Y values not shown if they are set to zero and ref point is center


ktaki:
Draw a shape. In the property panel, set the reference point to center. Enter 0 to X field and hit return. Repeat for Y. The both show empty fields.

![screen shot 2015-12-03 at 12 03 30 pm](https://cloud.githubusercontent.com/assets/2029597/11572440/2d0c4ea0-99b6-11e5-9ac2-d557a94df169.png)

ktaki:
Verified the fix.

PR #3389, opened 2015-12-08
Ref point center and X, Y set to zero now displays

DivyaPrabhakar:
References issue #3350

### DIFF
--- a/src/js/util/collection.js
+++ b/src/js/util/collection.js
@@ -81,7 +81,7 @@ define(function (require, exports) {
     var pluck = function (iterable, property, notSetValue) {
         return iterable.map(function (obj) {
             if (obj) {
-                if (obj.hasOwnProperty(property) || obj[property]) {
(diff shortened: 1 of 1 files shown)


mcilroyc:
This seems like a significant change to a commonly used utility (pluck).  I'm not sure this is safe.  Pardon me if you've already cleared this with @iwehrman or something along those lines.

mcilroyc:
I'm going to compare this with lodash's pluck, and then I'll stop griping and merge it

DivyaPrabhakar:
@mcilroyc I have not cleared it with anyone so the scrutiny is welcome

mcilroyc:
Ok, I now believe this is safe.  Part of my misunderstanding was the particular case that this handles.  An example is the "xCenter" property of `bounds` which is a cached derived property on the Bounds model.  It will not be returned by hasOwnProperty, but it ALSO can not be fetched with Immutable's `get()` function.  

It is still lame that we have to maintain our own version of `pluck` for immutable... because of little programming errors like this.

Issue #3382, opened 2015-12-07
Menu commands are broken in search


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/11641760/d9f4bcfc-9cee-11e5-8030-6b31534d881f.png)

CC @DivyaPrabhakar

mcilroyc:
looks fixed to me, but leaving with @iwehrman as one final check... and to make sure he didn't have some other concern with the underlying PR.

PR #3390, opened 2015-12-08
Menu Commands fixed in search

DivyaPrabhakar:
References issue #3382

### DIFF
--- a/src/js/actions/search/commands.js
+++ b/src/js/actions/search/commands.js
@@ -123,13 +123,11 @@ define(function (require, exports) {
     var _menuCommandSearchOptions = function () {
         var menuStore = this.flux.store("menu"),
             menu = menuStore.getApplicationMenu(),
-            menuMap = menu.rootMap,
             roots = menu.roots.reverse();
 
         var menuCommands = [];
         roots.forEach(function (root) {
-            var currRoot = menuMap.get(root.id),
-                nodes = [currRoot],
+            var nodes = [root],
                 currItem;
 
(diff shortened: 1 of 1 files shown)


src/js/actions/search/commands.js
```diff
@@ -123,13 +123,11 @@ define(function (require, exports) {
     var _menuCommandSearchOptions = function () {
         var menuStore = this.flux.store("menu"),
             menu = menuStore.getApplicationMenu(),
-            menuMap = menu.rootMap,
             roots = menu.roots.reverse();
 
         var menuCommands = [];
         roots.forEach(function (root) {
-            var currRoot = menuMap.get(root.id),
```
iwehrman: Do you know what the purpose of this line was previously? It looks superfluous to me, but I just want to confirm that you agree. (The same functionality is still available via `MenuBar.prototype.byID`, btw.)

mcilroyc: Looks to me like it was just a misunderstanding during the original implementation.... like maybe it seemed that it was necessary to get the root from the map, event though it wasn't.

DivyaPrabhakar: @iwehrman @mcilroyc I agree about it being a misunderstanding -- it is odd that it was working before and I had the same question of why this was here in the first place. But yes both of your observations are correct, it was just a small mistake in what reference was being inserted into the array.

mcilroyc:
Seems legit.  merging.

PR #3392, opened 2015-12-08
Clean up ui locks

volfied:
We had some JS_UI locks that were no longer valid, those are removed. In most cases, these locks are essentially redundant with the transfers, so it helps keep things cleaner.

Almost all type actions seemed unnecessary, but @pineapplespatula, am I missing something?

Also in edit.js, since we dispatch `panel.TOGGLE_OVERLAYS` now, changed the lock there, this was a dormant bug.

### DIFF
--- a/src/js/actions/edit.js
+++ b/src/js/actions/edit.js
@@ -422,7 +422,7 @@ define(function (require, exports) {
     };
     redo.action = {
         reads: [locks.JS_APP, locks.JS_DOC],
-        writes: [locks.JS_UI],
+        writes: [locks.JS_PANEL],
         transfers: [history.incrementHistory],
         post: [layers._verifyLayerIndex],
         modal: true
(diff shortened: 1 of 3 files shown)


pineapplespatula:
The UI does change during the update calls but it is via a dispatched command, I'm not sure if that necessitates a UI Lock (but regardless they should probably be on the update calls and not on the set calls.

otherwise this looks good. although someone else will have to speak for 
changing   `writes: [locks.JS_UI]` to `writes: [locks.JS_PANEL]`

pineapplespatula:
paging @mcilroyc or @iwehrman

src/js/actions/edit.js
```diff
@@ -422,7 +422,7 @@ define(function (require, exports) {
     };
     redo.action = {
         reads: [locks.JS_APP, locks.JS_DOC],
-        writes: [locks.JS_UI],
+        writes: [locks.JS_PANEL],
```
mcilroyc: this seems safe to me, @pineapplespatula ... since it is now consistent with `undo`.

PR #3399, opened 2015-12-09
Adding tag/version strings to adapter and generator-connection deps in package.json

mcilroyc:
This isn't as cool as it used to be with `bower`, which allowed semver strings to be supplied, but I think this is better than not supplying a version at all.  @volfied ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "classnames": "^2.2.0",
     "d3": "^3.5.9",
     "fluxxor": "^1.7.3",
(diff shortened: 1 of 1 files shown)


mcilroyc:
I just realized that the adapter version I specified here is actually the one required by the probably-ready-to-merge PR #3381

volfied:
That PR is still under review, but updating spaces-adapter should not have any effect on existing code, merging this.

PR #3401, opened 2015-12-09
Handle locked layers when using the sampler tool

shaoshing:
This PR handles locked layers in different sampler scenarios:

1. select a locked layer: sampler is disabled (no HUD and mouse click), no sampler overlay
2. select multiple locked layers: the same as 1
3. selections of locked and unlocked layers: sampler is enabled, but only applies to unlocked layers. locked layers are excluded from the sampler overlay.

Fixes issue #3238


volfied:
Alright!

Issue #3325, opened 2015-12-01
Datalist: wrong position when the dropdown list is above the input


shaoshing:
Dropdown list should update its position to bottom align with the input below, when its height changed.

![datalist-dropdown](https://cloud.githubusercontent.com/assets/118264/11511022/784d73d6-981b-11e5-87bf-f9984d0a9672.gif)

mcilroyc:
@ktaki could you please eyeball this?  Fixed with recent PR.

mcilroyc:
Note that the position issue is only visible (prior to this PR) after entering some filtering text on datalists that appear *above* the input.

ktaki:
It looks good now.
<img width="99" alt="screen shot 2015-12-09 at 4 31 26 pm" src="https://cloud.githubusercontent.com/assets/2029597/11703263/5cb32d78-9e92-11e5-9f50-696ac39c1798.png">

PR #3402, opened 2015-12-09
Fix wrong dialog position when it is placed above its target.

shaoshing:
This PR fixes issue #3325.

Before:

screen shot 2015-12-09 at 11 45 00 am

After:

screen shot 2015-12-09 at 11 45 13 am

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -110,6 +110,38 @@ define(function (require, exports, module) {
             this.getFlux().store("dialog").registerDialog(this.props.id, this._getDismissalPolicy());
         },
 
+        componentWillUnmount: function () {
+            if (this.state.open) {
+                this._removeListeners();
+                this.getFlux().actions.dialog.closeDialog(this.props.id);
+            }
+
+            this.getFlux().store("dialog").deregisterDialog(this.props.id);
+        },
+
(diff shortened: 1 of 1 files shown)


src/js/jsx/shared/Dialog.jsx
```diff
@@ -211,34 +211,40 @@ define(function (require, exports, module) {
...
-                            placedDialogTop = clientHeight -
+                            topPosition = clientHeight -
                                 dialogBounds.height - dialogMarginTop - dialogMarginBottom;
                         }
                     }
 
                     this.setState({
-                        "topPosition": placedDialogTop + "px",
+                        "topPosition": topPosition + "px",
+                        "bottomPosition": bottomPosition + "px",
```
mcilroyc: I can't find any user-facing issue from this... but it is possible that `topPosition` and `bottomPosition` to be accessed without being defined... resulting in the string "undefinedpx" being added to state.
ss

mcilroyc:
works fine AFAICT.  But I have one gripe that is just legitimate enough to push back to @shaoshing

shaoshing:
@mcilroyc Thanks for the careful review, that's a really good catch. I've fix the `undefinedpx` value. I also made some extra change to move some functions in `Dialog.jsx` to their proper positon. Back to you.

mcilroyc:
bueno

PR #3403, opened 2015-12-09
Factor actions.panel out of actions.ui

iwehrman:
I actually had this in my branch for #3374 but I forgot to update the PR... Anyway, this just splits `actions.panel` out of `actions.ui` in the obvious way.


volfied:
Looks good, but you forgot two actions. I just pushed a commit fixing those

Issue #3384, opened 2015-12-07
Lost spinner on Export


ktaki:
A spinner is no longer shown while exporting assets. This is the case for both quick export from Export panel and exporting from Export dialog.

mcilroyc:
@designedbyscience update here if think your work fixes this issue.

iwehrman:
FBNC to @ktaki.

ktaki:
The spinner does not show up in the Export panel on Mac. It's there for Export dialog on Mac and both Export panel and Export dialog on Windows.

<img width="289" alt="screen shot 2015-12-09 at 4 07 39 pm" src="https://cloud.githubusercontent.com/assets/2029597/11702925/99002310-9e8f-11e5-86f5-d3936c77f147.png">

ktaki:
Sorry, wrong statement in the previous comment. The spinner does NOT shown in the Export panel on Mac and Windows. It is shown in Export dialog on Mac and Windows. Mac and Windows versions are consistent.

mcilroyc:
Export panel fixed by second PR.  back to @ktaki again.  @designedbyscience is sorry : )

ktaki:
Verified the fix.

Issue #3217, opened 2015-11-06
Chromium 47 SVG animation deprecation warning


iwehrman:
In Chromium 47, we get the following warning triggered by react-with-addons.js:20843:
```SVG's SMIL animations (<animate>, <set>, etc.) are deprecated and will be removed. Please use CSS animations or Web animations instead.```
~~Are we actually using SVG animations anywhere? It's possible that this warning will go away if we upgrade to React 0.14.~~

designedbyscience:
I think we are for the loader

iwehrman:
Sure enough: https://github.com/adobe-photoshop/spaces-design/blob/2a5dd74d5d414563efae4b9b075e5beeddd24bd1/src/img/ico-loader.svg?short_path=d608a50#L5-L12

placegraphichere:
assigning to me

placegraphichere:
This issue is a lot of fun. Seems that I can easily embed CSS into the SVG that will do the same animation. Problem is we do the `<use>` thing in order to place svg. The embedded CSS animation gets cut when you do that for some reason. Example of that will work in preview/browser, but not in `<svg><use>` that we have in Design Space:

---

`<svg version="1.1" id="loader" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 width="100%" height="100%" viewBox="0 0 16 16" xml:space="preserve">
    <style type="text/css">    
        #c1 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: 0s;
            animation: opacityLoop 1s infinite ease-in-out;
            animation-delay: 0s;
        }
        #c2 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: .1666s;
        }
        #c3 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: .3333s;
        }
        #c4 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: .4999s;
        }
        #c5 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: .6666s;
        }
        #c6 {
            -webkit-animation: opacityLoop 1s infinite ease-in-out;
            -webkit-animation-delay: .8333s;
        }
        @-webkit-keyframes opacityLoop {
[...]

PR #3405, opened 2015-12-09
Replace SVG animated loader with a CSS animated one

designedbyscience:
SMIL SVG animations are being deprecated in Chrome in favor of CSS animations. Due to how we include SVGs (using the `<use>` tag), we cannot simply add CSS animations to the individual SVGs. I added some CSS in order to target the proper tag with the spinning loaded animation.

Since we are no longer pulling in the SVG in Main.jsx, I was able to remove that code as well as the url-loader, which @volfied says was only used for that.

Addresses #3384 and #3217

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "react-addons-perf": "^0.14.2",
     "react-tools": "^0.13.1",
     "uglifyjs": "^2.4.10",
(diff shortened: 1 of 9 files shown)


iwehrman:
It spins.

PR #3407, opened 2015-12-09
Convert the search registration functions to actions

iwehrman:
What it says!

@volfied @shaoshing This also fixes a potential bug in the CC Libraries implementation in which multiple collectionsLoaded handlers could be registered when resetting the controller. I fixed it by removing any stale handlers before re-registering in `libraries.afterStartup`.


volfied:
Looks good to me, @shaoshing, can you double check the `actions/libraries.js` changes since that's your domain.

shaoshing:
the overall changes look good, and the libraries part is safe I think. Ready to merge.

PR #3409, opened 2015-12-09
Factor actions.groups out of actions.layers

iwehrman:
The goal here is only to reduce the size of `actions.layers`. Also:

* moves the mask actions from `actions.layers` into `actions.mask`, and
* converts `layers.unlockBackgroundLayer` into action.


volfied:
You forgot `models/tools/superselect.js#297` `layers.deleteVectorMask`

volfied:
Otherwise looks good.

iwehrman:
Fixed + rebased. Back to @volfied.

Issue #3265, opened 2015-11-17
Layer names can be left blank


ktaki:
In Layers panel, double click a name to select entire name for edit, hit delete key, and then hit return key. The name is left blank. It should show the original name.

volfied:
FBNC to @ktaki

ktaki:
Verified the fix for both layer's name and library asset name.

PR #3410, opened 2015-12-09
Revert input to original value if the new layer/asset name is empty.

shaoshing:
This PR fixes issue that user can input empty layer name (issue #3265). 
I also found the same issue on libraries asset name, and it is fixed too.

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -155,7 +155,13 @@ define(function (require, exports, module) {
          * @param {string} newName 
          */
         _handleLayerNameChange: function (event, newName) {
-            this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+            if (newName.length !== 0) {
+                if (newName !== this.props.layer.name) {
+                    this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+                }
+            } else {
+                this.refs.layerName.setValue(this.props.layer.name);
+            }
         },
 
         /**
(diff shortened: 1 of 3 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -155,7 +155,13 @@ define(function (require, exports, module) {
          * @param {string} newName 
          */
         _handleLayerNameChange: function (event, newName) {
-            this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+            if (newName.length !== 0) {
+                if (newName !== this.props.layer.name) {
+                    this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
+                }
+            } else {
+                this.refs.layerName.setValue(this.props.layer.name);
```
volfied: Is this necessary? isn't the _handleLayerNameChange being called from the TextInput?

mcilroyc: This is handling the special case where TextInput *tries* to set it to empty string, and now LayerFace is saying *nah, set it back to what it was because we don't allow empty layer names*

volfied: I see, it strikes me as non-React to call setValue on another component, but TextInput isn't exactly the most friendly of React components :)

mcilroyc: for the sake of conversation... I know what you mean @volfied.  I wonder about an alternate approach:  If the onChange handler (`_handleLayerNameChange` in this case) determines that the changed text value is invalid (empty, in this case), then it could just force *itself* to re-render based on on its current props/state... which would in turn cause TextInput to reset back to a steady state.

shaoshing: Hey guys, I am creating a new PR which handles the empty value scenario better. will ping you there.

PR #3412, opened 2015-12-09
Add Menu command to create a vector mask from a shape layer

pineapplespatula:
requires https://github.com/adobe-photoshop/spaces-adapter/pull/175 
added new menu item
screen shot 2015-12-09 at 3 19 49 pm

which is only enabled if you select two layers, one which is a vector shape and the other is a layer that allows for vector masks to be applied 

screen shot 2015-12-09 at 3 21 38 pm
screen shot 2015-12-09 at 3 22 03 pm

### DIFF
--- a/src/js/actions/mask.js
+++ b/src/js/actions/mask.js
@@ -27,13 +27,17 @@ define(function (require, exports) {
     var Promise = require("bluebird"),
         Immutable = require("immutable");
 
+    var historyActions = require("./history"),
+        layerActions = require("./layers"),
+        locks = require("../locks");
+
     var descriptor = require("adapter").ps.descriptor,
         boundsLib = require("adapter").lib.bounds,
         vectorMaskLib = require("adapter").lib.vectorMask,
-        documentLib = require("adapter").lib.document;
+        documentLib = require("adapter").lib.document,
+        layerLib = require("adapter").lib.layer;
 
-    var locks = require("../locks"),
-        events = require("../events"),
+    var events = require("../events"),
         nls = require("js/util/nls"),
         layerActionsUtil = require("js/util/layeractions");
 
@@ -129,7 +133,7 @@ define(function (require, exports) {
         reads: [locks.JS_TOOL],
         writes: [locks.PS_DOC, locks.JS_DOC]
     };
-    
+
     /**
      * Handle escaping from vector mask for the selection-based vector mask tools.
      * Most tools can just call tools.changeVectorMaskMode directly.
@@ -336,6 +340,88 @@ define(function (require, exports) {
         transfers: ["tools.select"]
     };
 
+    /**
+     * if there are two selected layers create a vector mask on one of them from the path of the shape in the 
+     * other selected layer
+     *
+     * @return {Promise}
+     */
+    var createVectorMaskFromShape = function () {
+        var appStore = this.flux.store("application"),
+            currentDocument = appStore.getCurrentDocument();
+
+        if (!currentDocument) {
+            return Promise.resolve();
+        }
+
+        var currentLayers = currentDocument.layers.selected;
+
+        if (currentLayers.size !== 2) {
+            return Promise.resolve();
+        }
+
+        var shapeLayers = currentLayers.filter(function (layer) {
+                return layer.isVector;
+            }),
+            maskLayers = currentLayers.filter(function (layer) {
+                return currentDocument.layers.canSupportVectorMask(layer) &&
+                    !layer.vectorMaskEnabled;
+            });
+
+        if (shapeLayers.isEmpty() || maskLayers.isEmpty()) {
+            return Promise.resolve();
+        }
+       
+        var shapeLayer = shapeLayers.first(),
+            maskLayer = maskLayers.first(),
+            shapeLayerRef = layerLib.referenceBy.id(shapeLayer.id);
+
+        var options = {
+            paintOptions: {
+                immediateUpdate: true,
+                quality: "draft"
+            },
+            historyStateInfo: {
+                name: nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"),
+                target: documentLib.referenceBy.id(currentDocument.id)
+            }
+        },
+            playObjects = [layerLib.select(shapeLayerRef),
+                vectorMaskLib.createMaskFromShape(maskLayer.id)],
+            deleteShapeLayer = currentDocument.layers.all.some(function (layer) {
+                    return !layer.isGroupEnd && !layer.isGroup && layer.id !== shapeLayer.id;
+                });
+
+        if (deleteShapeLayer) {
+            playObjects.push(layerLib.delete(shapeLayerRef));
+        }
+        
+        var playPromise = descriptor.batchPlayObjects(playObjects, options),
+            historyPromise = this.transfer(historyActions.newHistoryState, currentDocument.id,
+                nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"));
+
+        return Promise.join(historyPromise, playPromise)
+            .bind(this)
+            .then(function () {
+                return this.transfer(layerActions.resetLayers, currentDocument, maskLayer);
+            })
+            .then(function () {
+                if (deleteShapeLayer) {
+                    var payload = {
+                        documentID: currentDocument.id,
+                        layerIDs: Immutable.List.of(shapeLayer.id)
+                    };
+                    
+                    this.dispatch(events.document.history.DELETE_LAYERS, payload);
+                }
+            });
+    };
+    createVectorMaskFromShape.action = {
+        reads: [locks.JS_APP, locks.JS_DOC],
+        writes: [locks.PS_DOC],
+        transfers: [layerActions.resetLayers, layerActions.removeLayers, historyActions.newHistoryState]
+    };
+
     exports.editVectorMask = editVectorMask;
     exports.selectVectorMask = selectVectorMask;
     exports.deleteVectorMask = deleteVectorMask;
@@ -343,4 +429,5 @@ define(function (require, exports) {
     exports.handleEscapeVectorMask = handleEscapeVectorMask;
     exports.applyEllipse = applyEllipse;
     exports.applyRectangle = applyRectangle;
+    exports.createVectorMaskFromShape = createVectorMaskFromShape;
 });
--- a/src/nls/en/shortcuts-mac.json
+++ b/src/nls/en/shortcuts-mac.json
@@ -32,7 +32,8 @@
         },
         "LAYER": {
             "CREATE_ARTBOARD": "b",
-            "DUPLICATE": "j"
+            "DUPLICATE": "j",
+            "NEW_VECTOR_MASK_FROM_SHAPE": "m"
         },
         "ARRANGE": {
             "FLIP_HORIZONTAL": "/",
--- a/src/static/menu-win.json
+++ b/src/static/menu-win.json
@@ -320,6 +320,18 @@
                 {
                     "id": "NEW_ARTBOARD_FROM_TEMPLATE",
                     "submenu": []
+                },
+                {
+                    "separator": true
+                },
+                {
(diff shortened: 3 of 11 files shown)


pineapplespatula:
@ktaki  or @chadrolfs  if you have some spare cycles to bang on this i'm sure there are edge cases i haven't tested yet

chadrolfs:
loading I see these errors (correct adapter branch is npm linked):
screen shot 2015-12-11 at 9 52 03 am
screen shot 2015-12-11 at 9 52 12 am

pineapplespatula:
did you rerun grunt build after syncing the branch. (JSON doesnt recompile with the watcher)

chadrolfs:
no, had grunt debug running.

ktaki:
My result here. Initialization was successful, but got the following error when executed the command. Any suggestions?

screen shot 2015-12-11 at 12 22 33 pm

pineapplespatula:
did you re link (with npm ) my adapter branch jecohen/vectorMaskFromShape?

ktaki:
Can you suggest the commands? What I did was re-issued 'npm install' and 'npm update'.

pineapplespatula:
in the adapter branch `npm link`
and then in the design space branch `npm link spaces-adapter`

ktaki:
Are those commands needed every time switching spaces-adapter to other branch? Anyway, after done that, the result was the same error. 

@chadrolfs Are you still getting errors on initialization phase?

chadrolfs:
yeah, i blew away node modules in adapter, reinstalled npm and relinked.  still getting promise warnings.  Check with @volfied  or rebase against master.  I recall he made a change regarding those warnings.

pineapplespatula:
@chadrolfs  I rebased. give it a shot

chadrolfs:
hmmm, i'm still seeing the warnings.  I also tried blowing away and re-installing/linking node modules in both repos.

pineapplespatula:
The adapter branch has been merged, so this should have become an easier branch to test. Are you seeing the warnings while using the master branch of the adapter?

chadrolfs:
Ahhhh, that helped.  Not seeing those warnings now.  Thanks.

chadrolfs:
On problem I ran into: Undo updates the canvas but not the layers panel (you have to undo 2x)

iwehrman:
Sounds like this should still be assigned to @pineapplespatula, unless you think @chadrolfs' latest bug report is an unrelated issue.

pineapplespatula:
The issue chad found has been fixed in the latest commit

chadrolfs:
Just pulled the latest and seeing this on load:
`Action menu.beforeStartup failed: Error: Menu entry NEW_VECTOR_MASK_FROM_SHAPE has a shorcut with an unlocalized key.`

chadrolfs:
screen shot 2015-12-17 at 3 53 02 pm

chadrolfs:
pgdev.948, branch @ c297f87

ktaki:
My error reported earlier (not a function) has been resolved.

pineapplespatula:
@chadrolfs did you restart grunt compile after switching branches?

ktaki:
Open https://photoshop.slack.com/files/ktaki/F0GUF4QTH/twolayerssidebyside.psd.
Select bottom two layers. Cmd+m.
Select top two layers. Cmd+m.

screen shot 2015-12-17 at 5 08 13 pm

If you do first top two layers, then the bottom two layers, you may see no errors on console but the following on canvas and layers panel.

screen shot 2015-12-17 at 5 12 58 pm

screen shot 2015-12-17 at 5 12 50 pm

This file may produce other errors when playing with undo and redo on Cmd+m.

ktaki:
What should happen when a masked layer and a shape layer are selected? Currently menu is enabled but does nothing.

screen shot 2015-12-18 at 1 52 05 pm copy

iwehrman:
I suggest naming this menu item "New Vector Mask From Layer" or possibly "Create Vector Mask From Layer". CC @placegraphichere

placegraphichere:
"Create Vector Mask from Layer" sounds about right. That kind of language is used for Clipping Masks "Create Clipping Mask" when a proper layer selection is made in standard. Using "New" would probably not be right. 

The only other idea would be "Apply Layer as Vector Mask". But I think we're splitting hairs on language.

pineapplespatula:
@placegraphichere  with regards to @ktaki's question about the behavior when there already is a mask on the layer. Would you expect it to replace the vector mask, or for the vector mask to be disabled? 
(or some other option)

placegraphichere:
When multi-selection is a `Shape` and `Layer+VectorMask` my initial thought is that we `Group` the `Layer+VectorMask` and apply the `Shape` as a vector mask to the new `Group`. In this way we maintain the original vectorMask and apply a mask as was asked for. Continuing to allow new vector masks seems like a positive for people. Restricting based on layer type seems like it needs explanation, and creates possible confusion.

Another option would be to `Intersect Shape Areas` boolean the `Shape` layer to the existing `vectorMask`. But results there may be problematic. 

The final option would be to disable the option. But that's kind of weird, since I may not have noticed the layer mask I have selected is the root issue. 
- - -
I think in any case we may also need a `Remove Vector Mask` option in the file menu to accompany this.

pineapplespatula:
I had to go with disabling the option for now. I created a separate issue about this problem, but it probably requires core work to solve.

pineapplespatula:
this now requires https://github.com/adobe-photoshop/spaces-adapter/pull/180

iwehrman:
Needs a rebase.

src/js/actions/mask.js
```diff
@@ -330,6 +334,88 @@ define(function (require, exports) {
...
+
+        var options = {
+            paintOptions: {
+                immediateUpdate: true,
+                quality: "draft"
+            },
+            historyStateInfo: {
+                name: nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"),
+                target: documentLib.referenceBy.id(currentDocument.id),
+                coalesce: false,
```
iwehrman: Why don't you want to coalesce the events here? On the face of it, it seems reasonable to do so.

iwehrman: Either way, can you confirm that undo/redo works correctly here?

src/js/actions/mask.js
```diff
@@ -330,6 +334,88 @@ define(function (require, exports) {
...
+            }
+        },
+            playObjects = [layerLib.select(shapeLayerRef),
+                vectorMaskLib.createMaskFromShape(maskLayer.id)],
+            notShapeLayer = collection.difference(currentDocument.layers.all,
+                Immutable.List.of(shapeLayer)),
+            notShapeLayerContainsLeaf = notShapeLayer.some(function (layer) {
+                    return !layer.isGroupEnd && !layer.isGroup;
+                }),
+            deleteShapeLayer = !notShapeLayer.isEmpty() && notShapeLayerContainsLeaf;
```
iwehrman: It seems like this could just as easily be computed without creating an intermediate data structure with `collection.difference` that contains all but one layer by just adding one more conjunct to to the `.some` call above.

src/nls/en/shortcuts-mac.json
```diff
@@ -32,7 +32,8 @@
         },
         "LAYER": {
             "CREATE_ARTBOARD": "b",
-            "DUPLICATE": "j"
+            "DUPLICATE": "j",
+            "NEW_VECTOR_MASK_FROM_SHAPE": "m"
```
iwehrman: Should there be a shortcut on Windows too?

src/nls/fr/shortcuts-mac.json
```diff
@@ -32,7 +32,8 @@
         },
         "LAYER": {
             "CREATE_ARTBOARD": "b",
-            "DUPLICATE": "j"
+            "DUPLICATE": "j",
+            "NEW_VECTOR_MASK_FROM_SHAPE": "m"
```
iwehrman: I think it is best to _not_ add shortcuts to the non-English dictionaries. If a key is missing, it will fall back to English automatically, and editing these files directly might be confusing for the loc team.

src/static/menu-win.json
```diff
@@ -320,6 +320,12 @@
                 {
                     "id": "NEW_ARTBOARD_FROM_TEMPLATE",
                     "submenu": []
+                },
+                {
+                    "separator": true
+                },
+                {
+                    "id":"NEW_VECTOR_MASK_FROM_SHAPE"
```
iwehrman: Again, just want to confirm that you do not want a keyboard shortcut for this on Windows. If that is the case, why not?

iwehrman:
Back to @pineapplespatula. Code's looking good!

src/js/actions/mask.js
pineapplespatula: that is why it is false, when it was true i got history errors.

src/nls/en/shortcuts-mac.json
pineapplespatula: cmd M on windows is reserved for minimize. I'm happy to have one, but when i talked originally with @placegraphichere  we decided on the route of not giving windows a shortcut for now

src/js/actions/mask.js
iwehrman: How many times do you have to undo to get back to the original state?

src/nls/en/shortcuts-mac.json
placegraphichere: I'm not sure what to do here without a WIN machine to try, but as far as I can tell, Standard just uses Control+M for other things than minimize.

screen shot 2016-01-11 at 1 21 29 pm

https://helpx.adobe.com/photoshop/using/default-keyboard-shortcuts.html

iwehrman: So then let's go with Ctrl-M unless, for some reason, that doesn't work in Design Space. If it doesn't, then that could conceivably be a plugin bug, since apparently it does work in standard PS.

src/js/actions/mask.js
pineapplespatula: just once

iwehrman: Paging @mcilroyc to shed some light on this.

mcilroyc: I think the `removeLayers` transfer around line 416 is throwing a wrench in this.  That action wants to create its own history state.  I suggest omitting that call, and simply dispatching a `DELETE_LAYERS` event to remove the offending layer(s), and that will naturally amend history state (which I think is what you want)

src/nls/en/shortcuts-mac.json
pineapplespatula: ok added a win shortcut

pineapplespatula:
@iwehrman :repeat:

src/js/actions/mask.js
```diff
@@ -336,6 +340,88 @@ define(function (require, exports) {
...
+                return this.transfer(layerActions.resetLayers, currentDocument, maskLayer);
+            })
+            .then(function () {
+                if (deleteShapeLayer) {
+                    var payload = {
+                        documentID: currentDocument.id,
+                        layerIDs: Immutable.List.of(shapeLayer.id)
+                    };
+                    
+                    return this.dispatch(events.document.history.DELETE_LAYERS, payload);
```
mcilroyc: don't `return` this synchronous dispatch.

iwehrman:
Looks good. Bombs away!

src/js/actions/mask.js
```diff
@@ -336,6 +340,88 @@ define(function (require, exports) {
...
+                    };
+                    
+                    this.dispatch(events.document.history.DELETE_LAYERS, payload);
+                }
+            });
+    };
+    createVectorMaskFromShape.action = {
+        reads: [locks.JS_APP, locks.JS_DOC],
+        writes: [locks.PS_DOC],
+        transfers: [layerActions.resetLayers, layerActions.removeLayers, historyActions.newHistoryState]
```
mcilroyc: note for the future @pineapplespatula - no longer need to list the transfer to `layerActions .removeLayers`

PR #3413, opened 2015-12-09
Return the full promise in libraries.beforeStartup

volfied:
Found this out while triaging a bug. BeforeStartup should return only when it's fully finished.


mcilroyc:
seems legit

Issue #3408, opened 2015-12-09
Search menu command description localized strings are broken


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/11700395/63a59de6-9e7e-11e5-8fab-a4a2f616d23d.png)

iwehrman:
CC @pineapplespatula @DivyaPrabhakar

iwehrman:
Fixed by #3415.

PR #3415, opened 2015-12-10
Localized search strings fixed 

DivyaPrabhakar:
The biggest change here is that I replaced parseInt() with our own parseNumber(). The if statement was not executing; after this the strings are accessed and rendered correctly. References #3408

### DIFF
--- a/src/js/actions/search/commands.js
+++ b/src/js/actions/search/commands.js
@@ -27,6 +27,7 @@ define(function (require, exports) {
     var _ = require("lodash"),
         Immutable = require("immutable"),
         keyUtil = require("js/util/key"),
+        mathUtil = require("js/util/math"),
         system = require("js/util/system"),
         nls = require("js/util/nls");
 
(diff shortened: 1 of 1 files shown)


src/js/actions/search/commands.js
```diff
@@ -60,12 +61,10 @@ define(function (require, exports) {
                 }
 
                 if (menuTree.$MENU) {
-                    resultPath += menuTree.$MENU + ">";
+                    resultPath += menuTree.$MENU + " > ";
```
iwehrman: I realize this isn't your doing, but: do you know where _else_ this separator is used? It seems like something that should be defined once. Actually, even better if this separator is localized so that it could be different in different languages. (Who knows if `>` makes sense in German or Japanese?)

DivyaPrabhakar: Currently, this is the only place that the `>` separator is used, so in the short term I don't think there is anything to be done here. There are multiple places in search that use separators, albeit of different types, e.g. "/" between layer groups when searching for layers, ":" between Library name and type, "    " after shortcuts.

iwehrman: Thanks for the research! We might want to chat about unifying all these separators with @placegraphichere, but it doesn't seem like a major problem right this moment.

Issue #3404, opened 2015-12-09
Library Asset subtitles can push buttons out of view


designedbyscience:
![untitled-1___66_7___rectangle_1__rgb_8____](https://cloud.githubusercontent.com/assets/22372/11697096/fed5f408-9e85-11e5-80dd-de36871f5eae.png)

iwehrman:
@placegraphichere: is this still FIP?

shaoshing:
This is fixed by #3583. Over to @designedbyscience to verify.

iwehrman:
This looks fixed to me!

PR #3416, opened 2015-12-10
Library description pushes buttons off screen Fix for #3404

placegraphichere:
Limit width of title area DIV to a max-width so the subtitle span won't push the button off screen

### DIFF
--- a/src/style/sections/libraries/libraries-section.less
+++ b/src/style/sections/libraries/libraries-section.less
@@ -262,9 +262,13 @@
             background-color: @list-select-color;
             padding-right: 0;
             .libraries__asset__title {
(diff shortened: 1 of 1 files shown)


designedbyscience:
I think there should be a way to accomplish this goal only using flex box properties, inside of specifying dimensions of things

placegraphichere:
yeah, however the `ul` that holds the title and 3 icons items `column-14 button-radio libraries__asset__buttons` won't respect an assigned width or max width. So the content just keeps flowing. Open to other ideas for sure

iwehrman:
This will be subsumed by the PR that updates the implementation of `TextInput` for better performance.

PR #3419, opened 2015-12-10
Add a timeout warning to cc-lib-api load promise

volfied:
Hopefully we'll figure out the root of this issue and remove this timeout check, but for now, adding this to get some repro cases from everyone.

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -1267,6 +1267,7 @@ define(function (require, exports) {
         descriptor.addListener("toolModalStateChanged", _toolModalStateChangedHandler);
 
         return apiLoadPromise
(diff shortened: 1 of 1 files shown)


src/js/actions/libraries.js
```diff
@@ -1267,6 +1267,7 @@ define(function (require, exports) {
         descriptor.addListener("toolModalStateChanged", _toolModalStateChangedHandler);
 
         return apiLoadPromise
+            .timeout(3000, "CC Libraries API load timeout, please don't restart and notify the chatroom!")
```
mcilroyc: please don't (restart and notify), or please don't restart, but DO notify.  I'm confused.  JUUUST kidding.

PR #3420, opened 2015-12-10
Better way to handle empty value in TextInput

shaoshing:
* Instead of handling empty value in the handler, I added new option `allowEmpty` which will ignore empty value and skip the `onChange` handler.
* `onChange` was not clear before: it could be onChange handler or onCommit handler (when `continuous` prop is true). I reduce its scope by adding a new handle `onInputChange`, which also helps remove the `continuous` prop. Please see doc for difference of two handlers.
* bug fix: disallow empty value in ColorPicker. Otherwise inputting empty value will set color to black.

@volfied @mcilroyc can you review since you both reviewed #3410 ?

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -155,13 +155,7 @@ define(function (require, exports, module) {
          * @param {string} newName 
          */
         _handleLayerNameChange: function (event, newName) {
-            if (newName.length !== 0) {
-                if (newName !== this.props.layer.name) {
-                    this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
-                }
-            } else {
-                this.refs.layerName.setValue(this.props.layer.name);
-            }
+            this.getFlux().actions.layers.rename(this.props.document, this.props.layer, newName);
         },
 
         /**
@@ -664,7 +658,8 @@ define(function (require, exports, module) {
                                         value={layer.name}
                                         disabled={this.props.disabled || !nameEditable}
                                         onKeyDown={this._skipToNextLayerName}
-                                        onChange={this._handleLayerNameChange}>
+                                        onChange={this._handleLayerNameChange}
+                                        allowEmpty={false}>
                                     </TextInput>
                                     {showHideButton}
                                 </span>
--- a/src/js/jsx/sections/libraries/assets/AssetSection.jsx
+++ b/src/js/jsx/sections/libraries/assets/AssetSection.jsx
@@ -91,13 +91,7 @@ define(function (require, exports, module) {
          * @param {string} newName
          */
         _handleRename: function (event, newName) {
-            if (newName.length !== 0) {
-                if (this.props.displayName !== newName) {
-                    this.getFlux().actions.libraries.renameAsset(this.props.element, newName);
-                }
-            } else {
-                this.refs.input.setValue(this.props.displayName);
-            }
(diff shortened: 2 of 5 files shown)


designedbyscience:
Could we figure out something better than having `handleChange`, `handleInputChange` and also `handleDOMChange`? Must be better naming...

mcilroyc:
Another high level comment: I feel like maybe the "allowEmpty" concept is too narrow.  Can we just find a good pattern for letting the parent's "onChange" handler do the right thing, or provide "validation" feedback so that future situations can be handled similarly without having to stuff more "pre validation" into Text Input.

mcilroyc:
... I suppose the counter-argument to my previous comment is that if "allowEmpty" is a common validation, why not provide this convenience as part of TextInput.  I suppose I retract!

shaoshing:
@designedbyscience I suggest naming `onChange` as `onCommit` as the first one is confusing sometime, especially for input, as the standard onChange event is fired when it has new input. 

@mcilroyc `allowEmpty` is just like the `min` and `max` validator in NumberInput, so it should be common enought to extract as option. I also agree with your suggestion of provding feedback to have flexibility for special scenarios, for example, the Datalist `onChange` handler https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/jsx/shared/Datalist.jsx#L76

designedbyscience:
This conversation also brings up some problems with React naming. In native DOM land, `onChange` is really `onCommit`. And the continuous event is `input`... but React recommends using their `onChange` instead of `onInput`: https://facebook.github.io/react/docs/forms.html. 

So, do we name things closer to the native DOM definitions or to the React ones?

designedbyscience:
(Oh, but making these things separate props seems like a good idea, I am only quibbling about naming)

shaoshing:
So below are the relations of different namings, which show we should rename our `onChange` to avoid confusion?
```
oninput (DOM) == onChange (React)
onChange (React) == onInputchange (Design Space)
onChange (React) != onChange (DS) <- confusion
```

mcilroyc:
If I'm understanding this correctly ... I think my recommendation is that we change `TextInput` component's currently-named `onChange`property to `onCommit`, which then allows `onChange` to clearly be the "continuous" version used by DataList.   

This may slightly buck the react naming pattern ... but I think we're steering clear of their main warning/guidance not to use directly the DOM's built-in `oninput` event handler.  I'm going to assign this to @designedbyscience and force him to be the decider :hammer: :hammer:

mcilroyc:
wait, @shaoshing do we even still need both `onInputChange` and `onDOMChange`  (as you currently have them)?  Aren't they both just used by DataList?

shaoshing:
@mcilroyc my recent change on the Datalist separated the `onChange` event to `onHighlightedChange` (live) and `onChange` (commit). just a reminder that they need a rename if we decide to go with the new naming.

For `TextInput`, we don't need `onInputChange` as it is replaced by `onChange`, and looks like `onDOMChange` is only used by `Datalist` and can be replaced too.

iwehrman:
Is this blocked or is it just waiting for review?

mcilroyc:
My original intention in assigning it to @designedbyscience was just to have him confirm that he agreed to the new naming that I was suggesting.  I will look at the newest commit now, and merge unless something offends me!

mcilroyc:
merging.  @shaoshing can you please attach a bug to this, if it exists, or create a new one to track the validation by @chadrolfs etc. since this touches a core component `TextInput`

Issue #3286, opened 2015-11-19
Scrubby mixin is handling all window mousemove events


iwehrman:
The Scrubby mixin is intercepting ALL window mousemove events because it is registering itself on the window at component mount time. Instead, it should only register this handler when you mousedown on the input's label and unregister the handler on mouseup.

volfied:
Grabbing from Jeffrey as I'm wiring up all the scrubbies.

mcilroyc:
satisfied @iwehrman ?

PR #3421, opened 2015-12-10
Scrubbiness for more

volfied:
- Refactored Scrubby sliders, addressing #3286 and cleaning the code a bit
 - Added scrubbiness to Width and Height labels, also coalescing the scrubs
 - `_getMoveLayerActions` now uses absolute positioning if coalescing is true, but translate otherwise. This change was done in order to not change existing behavior, but have scrub behavior be correct.

This is still a WIP, while I hook up other controls, so I will be updating the PR with more files as I go along, but the core functionality, and example Scrub behavior should be reviewed in case I made some error that'll carry on to other scrub implementations.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -98,19 +98,24 @@ define(function (require, exports) {
      *                                     parent artboard of the layer
      * @param {string} refPoint Two character string denoting corner/edge of the layer to set the position into
      * @param {Array.<{layer: Layer, x: number, y: number}>} moveResults payload for updating each layer's bounds
+     * @param {boolean=} translate Whether to use translate or set absolute position by given coordinates
+     *                             absolute position is default
      *
      * @return {Immutable.List<{layer: Layer, playObject: PlayObject}>}
      */
-    var _getMoveLayerActions = function (document, targetLayer, position, refPoint, moveResults) {
+    var _getMoveLayerActions = function (document, targetLayer, position, refPoint, moveResults, translate) {
         var overallBounds = position.relative ?
                 document.layers.relativeChildBounds(targetLayer) :
                 document.layers.childBounds(targetLayer),
             positionKeys = uiUtil.getPositionKeysByRefPoint(refPoint),
             deltaX = position.hasOwnProperty("x") ? position.x - overallBounds[positionKeys.x] : 0,
             deltaY = position.hasOwnProperty("y") ? position.y - overallBounds[positionKeys.y] : 0,
+            newX = translate ? deltaX : (overallBounds.left + deltaX),
+            newY = translate ? deltaY : (overallBounds.top + deltaY),
             documentRef = documentLib.referenceBy.id(document.id),
             movingLayers = document.layers.descendants(targetLayer),
-            layerRef = [documentRef, layerLib.referenceBy.id(targetLayer.id)];
+            layerRef = [documentRef, layerLib.referenceBy.id(targetLayer.id)],
+            descriptorFn = translate ? layerLib.translate : layerLib.setPosition;
 
         moveResults = moveResults || [];
 
@@ -130,7 +135,7 @@ define(function (require, exports) {
 
         return Immutable.List.of({
             layer: targetLayer,
-            playObject: layerLib.translate(layerRef, deltaX, deltaY)
+            playObject: descriptorFn(layerRef, newX, newY)
         });
     };
          
@@ -404,44 +409,59 @@ define(function (require, exports) {
      * @param {boolean} position.relative If true, x and y will be relative to the owner artboard of layer
      * @param {string} refPoint Two character string denoting the active reference point so we know 
      *                 which corner of the layer to move
-     *
+     * @param {object} options Batch play options
+     * 
      * @return {Promise}
      */
-    var setPosition = function (document, layerSpec, position, refPoint) {
+    var setPosition = function (document, layerSpec, position, refPoint, options) {
+        options = _.merge({}, options);
         layerSpec = layerSpec.filterNot(function (layer) {
             return layer.isGroupEnd;
         });
 
         var payload = {
-                documentID: document.id,
-                positions: [],
-                history: {
-                    newState: true,
-                    name: nls.localize("strings.ACTIONS.SET_LAYER_POSITION")
-                }
-            },
-            options = {
-                paintOptions: _paintOptions,
-                historyStateInfo: {
-                    name: nls.localize("strings.ACTIONS.SET_LAYER_POSITION"),
-                    target: documentLib.referenceBy.id(document.id)
-                }
-            };
+            documentID: document.id,
+            positions: [],
+            coalesce: !!options.coalesce,
+            history: {
+                newState: true,
+                name: nls.localize("strings.ACTIONS.SET_LAYER_POSITION")
+            }
+        };
 
+        options = _.merge(options, {
+            paintOptions: _paintOptions,
+            historyStateInfo: {
+                name: nls.localize("strings.ACTIONS.SET_LAYER_POSITION"),
+                target: documentLib.referenceBy.id(document.id),
+                coalesce: !!options.coalesce,
+                suppressHistoryStateNotification: !!options.coalesce
+            }
+        });
+
+        // If coalescing, we use absolute setPosition function to avoid model mismatch
         var dispatchPromise = this.dispatchAsync(events.document.history.REPOSITION_LAYERS, payload),
             translateLayerActions = layerSpec.reduce(function (actions, layer) {
                 var layerActions = _getMoveLayerActions.call(this,
-                        document, layer, position, refPoint, payload.positions);
+                        document, layer, position, refPoint, payload.positions, !options.coalesce);
                 return actions.concat(layerActions);
             }, Immutable.List(), this);
 
-        var positionPromise = layerActionsUtil.playLayerActions(document, translateLayerActions, true, options)
+        var transaction = descriptor.beginTransaction(options),
+            actionOpts = _.merge(options, {
+                transaction: transaction
+            });
+
+        var positionPromise = layerActionsUtil.playLayerActions(document, translateLayerActions, true, actionOpts)
+            .then(function () {
+                return descriptor.endTransaction(transaction);
+            });
+
+        return Promise.join(dispatchPromise, positionPromise)
             .bind(this)
             .then(function () {
                 return this.transfer(layerActions.resetIndex, undefined);
             });
-
-        return Promise.join(dispatchPromise, positionPromise);
     };
     setPosition.action = {
         reads: [],
@@ -556,10 +576,12 @@ define(function (require, exports) {
      * @param {Layer|Immutable.Iterable.<Layer>} layerSpec Either a Layer reference or array of Layers
      * @param {{w: number=, h: number=}} size
      * @param {string} refPoint reference point, vertical position first then horizontal
+     * @param {object} options Batch play options
      *
      * @returns {Promise}
      */
-    var setSize = function (document, layerSpec, size, refPoint) {
+    var setSize = function (document, layerSpec, size, refPoint, options) {
+        options = _.merge({}, options);
         layerSpec = layerSpec.filterNot(function (layer) {
             return layer.isGroupEnd ||
                 document.layers.strictAncestors(layer)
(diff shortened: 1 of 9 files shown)


src/js/actions/transform.js
```diff
@@ -404,44 +408,59 @@ define(function (require, exports) {
...
+            documentID: document.id,
+            positions: [],
+            coalesce: !!options.coalesce,
+            history: {
+                newState: true,
+                name: nls.localize("strings.ACTIONS.SET_LAYER_POSITION")
+            }
+        };
+
+        options = _.merge(options, {
```
mcilroyc: This will come as no surprise to you, I'm sure ... but I don't like this.  I think it is confusing when action options get merged in to "play" options ... especially when it gets *assigned back to the same variable*

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -78,6 +139,64 @@ define(function (require, exports, module) {
...
+         * Update the current width of the selected layers based on change from initial scrub position
+         *
+         * @private
+         * @param {number} deltaX
+         */
+        _handleWScrub: function (deltaX) {
+            if (_.isNumber(this._initialScrubW)) {
+                var document = this.props.document,
+                    uiStore = this.getFlux().store("ui"),
+                    resolvedDelta = uiStore.zoomCanvasToWindow(deltaX),
```
mcilroyc: Can this `zoomCanvasToWindow` be generalized and stuff inside the scrubby mixin?  Seems like the individual components should have to be be concerned with it.

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -42,7 +43,36 @@ define(function (require, exports, module) {
     var Size = React.createClass({
         mixins: [FluxMixin],
 
-        shouldComponentUpdate: function (nextProps) {
+        /**
+         * Variables for storing inital scrub value to determine correct offset.
+         * @private 
+         * @type {number}
+         */
+        _initialScrubW: null,
```
mcilroyc: I'm not sure I see the value of having this exist as a private variable outside of `state`.  Isn't it just "the uniform value of the width"?  (If the widths are not uniform, then scrubbing is essentially disabled)

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -78,6 +139,64 @@ define(function (require, exports, module) {
...
+        },
+
+        /**
+         * Update the current width of the selected layers based on change from initial scrub position
+         *
+         * @private
+         * @param {number} deltaX
+         */
+        _handleWScrub: function (deltaX) {
+            if (_.isNumber(this._initialScrubW)) {
```
mcilroyc: If the widths are not uniform, then scrubbing is disabled.  It might be more clear/readable to make this determination at render time and just disable scrubbing (attach a no-op handler instead?).  Just a thought.

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -78,6 +139,64 @@ define(function (require, exports, module) {
...
+        _handleHScrub: function (deltaX) {
+            if (_.isNumber(this._initialScrubH)) {
+                var document = this.props.document,
+                    uiStore = this.getFlux().store("ui"),
+                    resolvedDelta = uiStore.zoomCanvasToWindow(deltaX),
+                    newHeight = Math.max(this._initialScrubH + resolvedDelta, this.refs.height.props.min);
+
+                if (newHeight > 0) {
+                    this.getFlux().actions.transform.setSizeThrottled(document, document.layers.selected,
+                        { h: newHeight }, this.props.referencePoint, { coalesce: true });
```
mcilroyc: The initial `setSize` needs to be non-coalesced.... I don't see that happening here, unless I'm missing it

mcilroyc:
I have a few questions about scrubby/state/coalesce in size.jsx.  Assigning back to @volfied since they may apply to the other components too.  In general I like the cWRP/SCU approach in size.jsx.

src/js/jsx/sections/transform/Size.jsx
volfied: I think it's contextual more than general. I wouldn't want the deltaX to be multiplied by any number on rotation scrub or radius scrub for example.

src/js/jsx/sections/transform/Size.jsx
volfied: Scrubby mixin will send the distance since the initial mousedown, so as we call different actions for each scrubby, model gets updated, and the state changes. This allows us to keep scrubbing continuous. I am not a fan of it either, but consdiering we throttle the action calls, and state changes, this is the most straightforward way. I'm open to suggestions.

src/js/jsx/sections/transform/Size.jsx
volfied: Yes, I agree. As I move things to state, instead of computing state on render, this will be easier to do.

src/js/jsx/sections/transform/Size.jsx
volfied: Is that a requirement? As far as I can tell, things work fine without it.

volfied:
@mcilroyc I am working on the other files to make changes similar to Size.jsx, but I answered all but one of your questions with non-task items (where I explain why I made the decision I made).

I will re-assign it once the changes are done, and keep the rest of my scrubby hook-up work in a separate PR (I'd like to keep this PR this size)

src/js/jsx/sections/transform/Size.jsx
volfied: I've decided against this, `uniformValue` is a costly operation, and we want to avoid calling it in every component update. Instead, calling it once on scrub begin and checking for nullity (we don't need `_.isNumber` as `uniformValue` returns null if it's not uniform) during scrubs is the fastest way. I don't think it adds much dirt to code.

src/js/actions/transform.js
volfied: For the sake of keeping it same across the code, can we delegate this to its own issue and track it there?

volfied:
@mcilroyc This is now ready for review, I've addressed your concerns only in replies and not much in code change :)

pineapplespatula:
@volfied  can you refactor your commits into one more descriptive commit

src/js/actions/transform.js
```diff
@@ -98,19 +98,23 @@ define(function (require, exports) {
      *                                     parent artboard of the layer
      * @param {string} refPoint Two character string denoting corner/edge of the layer to set the position into
      * @param {Array.<{layer: Layer, x: number, y: number}>} moveResults payload for updating each layer's bounds
+     * @param {boolean} translate Whether to use translate descriptor or not
```
pineapplespatula: add that setPosition is the alternative to the translate descriptor

src/js/actions/transform.js
pineapplespatula: I'm with cory on this it seems like we're overloading this options variable and some parts should get broken out. It seems like we only care about options.coalesce in this function. So I'm not sure why we're passing in the entire options object instead of just the coalesce field

src/js/actions/transform.js
```diff
@@ -556,10 +575,12 @@ define(function (require, exports) {
      * @param {Layer|Immutable.Iterable.<Layer>} layerSpec Either a Layer reference or array of Layers
      * @param {{w: number=, h: number=}} size
      * @param {string} refPoint reference point, vertical position first then horizontal
+     * @param {object} options Batch play options
      *
      * @returns {Promise}
      */
-    var setSize = function (document, layerSpec, size, refPoint) {
+    var setSize = function (document, layerSpec, size, refPoint, options) {
```
pineapplespatula: ditto from above, pass in coalesce instead of options.

volfied:
@pineapplespatula I'll rebase them once I clean up the code, but I'm not sure how I can make commits more descriptive at this point.

src/js/actions/transform.js
volfied: While I agree the option object juggling can be done better, I think passing a single boolean for an option is a step backwards. If we decide to add another option to be passed in actions, then we'll have to revert back to this.

src/js/actions/transform.js
volfied: Please refer to https://github.com/adobe-photoshop/spaces-design/pull/3421#discussion_r47837901 for my reply.

volfied:
@mcilroyc This is now rebased, and I've made sure to include a non-coalesced action on scrub start so we have a history state to go back to. Please review, and I will start working on the rest of scrubbies on a different branch.

src/js/actions/transform.js
pineapplespatula: I think if there is only one option currently that wrapping it in an object is overkill and should be done when the second parameter is added. but if you want to make it an object for now i'm ok with a name change for one of the two option objects and a more explicit JSDoc for the parameter stating that coalesce is all we care about currently

src/js/jsx/sections/style/Radius.jsx
```diff
@@ -40,72 +40,130 @@ define(function (require, exports, module) {
...
 
-        shouldComponentUpdate: function (nextProps) {
+        shouldComponentUpdate: function (nextProps, nextState) {
+            return this.state !== nextState;
+        },
+
+        componentWillReceiveProps: function (nextProps) {
             var getRelevantProps = function (props) {
-                var layers = props.document.layers.selected;
+                    var layers = props.document.layers.selected.filter(function (layer) {
```
mcilroyc: It is unfortunate that we filter "selected layers with radii" here, and then may immediately do it again in `_calculateState`.  Adding this subset of layers to layerstructure might help, but may be overkill.  Mentioning it anyway, for kicks.

src/js/jsx/sections/style/Radius.jsx
```diff
@@ -40,72 +40,130 @@ define(function (require, exports, module) {
...
+            this.setState({
+                layers: layers,
+                scalars: scalars,
+                maxRadius: maxRadius,
+                maxRadiusInput: Math.floor(maxRadius)
+            });
+        },
+
+        _handleScrubStart: function () {
+            this._initialScrubRadius = collection.uniformValue(this.state.scalars);
```
mcilroyc: I feel very strongly that _initialScrubRadius should be a property of the state.  `uniformValue` is not hard to compute, and it is fundamentally a stateful property of the component.

src/js/jsx/sections/style/Radius.jsx
volfied: I have a different plan for all these plucks we do to compare things. Working on it on the side.

src/js/actions/transform.js
pineapplespatula: at least i think you should add {boolean} option:coalesce to the JSDoc

src/js/actions/transform.js
```diff
@@ -1154,13 +1192,7 @@ define(function (require, exports) {
...
                 .toArray(),
             rotateObj = layerLib.rotate(layerRef, angle),
-            options = {
-                historyStateInfo: {
-                    name: nls.localize("strings.ACTIONS.ROTATE_LAYERS"),
-                    target: documentLib.referenceBy.id(document.id)
-                }
-            },
-            historyPromise = this.transfer(historyActions.newHistoryState, document.id,
+            historyPromise = coalesce ? Promise.resolve() : this.transfer(historyActions.newHistoryState, document.id,
```
mcilroyc: cool.  I had considered an alternate approach of providing a coalesce parameter to `newHistoryState` which would then pass that along to the event emitted to the history store... but I see no functional advantage to that approach, nor much of a "readability" advantage.

src/js/jsx/sections/style/Radius.jsx
```diff
@@ -44,68 +44,137 @@ define(function (require, exports, module) {
...
+            this.setState({
+                layers: layers,
+                scalars: scalars,
+                maxRadius: maxRadius,
+                maxRadiusInput: Math.floor(maxRadius)
+            });
+        },
+
+        _handleScrubStart: function () {
+            var currentRadius = collection.uniformValue(this.state.scalars);
```
mcilroyc: I still am slightly uneasy about computing the uniform value inside this scrubStart handler, and using the existence of `state.scrubRadius` to imply "scrub enabled" within the scrub handler ... but it's not a show-stopper.

src/js/jsx/sections/style/Radius.jsx
```diff
@@ -44,68 +44,137 @@ define(function (require, exports, module) {
...
+                this.setState({
+                    scrubRadius: currentRadius
+                });
+
+                this.startCoalescing();
+            }
+        },
+
+        _handleRadiusScrub: function (deltaX) {
+            if (this.state.scrubRadius === null) {
```
mcilroyc: this does double duty: if scrubRadius is null, it could be that there is non-uniform value... or simply that setState (which is async) has completed.  Nothing wrong here, just pointing it out (wrt to my prev comment)

src/js/jsx/sections/style/Radius.jsx
```diff
@@ -44,68 +44,137 @@ define(function (require, exports, module) {
...
         },
 
         render: function () {
-            var layers = this.props.document.layers.selected.filter(function (layer) {
-                    return layer.radii;
-                });
-
-            // If there is not at least one selected vector layer, don't render
-            if (layers.isEmpty()) {
+            if (!this.state) {
```
mcilroyc: Is this still necessary, now that you are calculating state during `componentWillMount`?  Maybe so, I just felt like maybe this was originally introduced because cWRP was not called for initial render...

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -88,8 +239,8 @@ define(function (require, exports, module) {
             var document = this.props.document,
                 flux = this.getFlux(),
                 referencePoint = flux.store("panel").getState().referencePoint;
-
-            flux.actions.transform
+
```
mcilroyc: :space_invader:

src/js/jsx/sections/transform/Rotate.jsx
```diff
@@ -50,18 +52,36 @@ define(function (require, exports, module) {
          */
         _lastAngle: null,
```
mcilroyc: This seems like it has been replaced by `state.value`.  It seems we are maintaining it for no reason now?

mcilroyc: I see the comment at line 168, but that doesn't seem to explain why we can't just use state.value

mcilroyc:
get scrubbed

Issue #3364, opened 2015-12-05
Search header text is hard to read in lighter themes


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/1A133P2Y3w3O3g2w331O/Screen%20Recording%202015-12-04%20at%2006.35%20PM.gif)

![image](https://cloud.githubusercontent.com/assets/1075154/11605497/f2f2de72-9ab5-11e5-8637-deff1f57903c.png)

iwehrman:
Fixed by #3423.

PR #3423, opened 2015-12-10
Fix search bar colors in all color stops

DivyaPrabhakar:
References #3364 (@placegraphichere @iwehrman this might be of most interest to you guys) 

The most important parts of the PR I need approval on are: 

a.) The spacing and layout of the search bar now -- the positions of the search icon, the list icons, the category headers, and the search inquiry as the user is typing. 

For each color stop, check the colors of these with respect to the entire UI in that color stop: 
b.) The search icon
c.) The category headers
d.) The list icons
e.) The text after the list icons
f.) The ancestry text after each option 

As is, we are able to use the same less variable names for each of the items above, no matter the color stop. If changes are to be made to specific color stops, we will just need to add specific selectors targeting that color stop. Note -- Rest assured the blurriness is not actually how the searchbar is rendered, the images are weird for some reason.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -36,6 +36,7 @@ define(function (require, exports, module) {
     var collection = require("js/util/collection"),
         nls = require("js/util/nls"),
         Datalist = require("js/jsx/shared/Datalist"),
+        SVGIcon = require("js/jsx/shared/SVGIcon"),
         headlights = require("js/util/headlights"),
         Button = require("js/jsx/shared/Button");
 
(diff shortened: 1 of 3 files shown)


iwehrman:
Something ain't right:
image

That's in the no-doc UI.

iwehrman:
The inner box is not centered in the outer box:
image

iwehrman:
On retina, I think the left-edge-blue-highlight line is so small that it's hard to see:
image

iwehrman:
Just a few additional suggestions. Looking good!

iwehrman:
The borders still have different widths :scream:
Left: 
image

Right: 
image

Also, there's still a gap separating the blue highlight from the edge of the list in the lighter themes:
image

DivyaPrabhakar:
@iwehrman 

I fixed the measurements to now be equal. 


As for the gap separation in the highlight, I am not seeing that on my machine.

iwehrman:
Looks good!

Issue #3187, opened 2015-11-03
Search: clear field icon (x) is not shown when you select a category


chadrolfs:
cmd/ctrl + f
arrow down to display datalist, click on Open Documents

Result: Open Docs category icon is shown in search field but (x) icon to clear field is not

alternately, if you type "Open" then click on Open Documents category, the icon shows in field as in steps above but now the (x) icon shows

chadrolfs:
![search-x](https://cloud.githubusercontent.com/assets/8442631/10918194/4a5dc484-8219-11e5-8fa3-2ab5af1c3e6c.gif)

DivyaPrabhakar:
@chadrolfs I am not able to repro this. I am using Mac build 936 and on master, here is what I see: 

![](https://www.dropbox.com/s/p5kkonpj2w86i4g/xsearchbar.gif?raw=1)

chadrolfs:
Instead of first hitting any alpha-numeric keys (letter "o") follow the step in the bug.  cmd/ctrl + f, then down arrow, click on Open Docs

chadrolfs:
I can still repro on pgdev.936 and master @ d44c3f9

DivyaPrabhakar:
@chadrolfs got it now, thanks.

PR #3427, opened 2015-12-11
Updates 'clear search'  icon whenever filters are changed

DivyaPrabhakar:
References #3187

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -192,11 +192,13 @@ define(function (require, exports, module) {
                 filterValues = _.drop(idArray);
 
             var updatedFilter = id ? _.uniq(this.state.filter.concat(filterValues)) : [],
-                filterIcon = id && this.getFlux().store("search").getSVGClass(updatedFilter);
+                filterIcon = id && this.getFlux().store("search").getSVGClass(updatedFilter),
+                nextHasInputValue = id !== null;
             
             this.setState({
                 filter: updatedFilter,
-                icon: filterIcon
+                icon: filterIcon,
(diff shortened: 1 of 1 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -194,6 +194,12 @@ define(function (require, exports, module) {
             var updatedFilter = id ? _.uniq(this.state.filter.concat(filterValues)) : [],
                 filterIcon = id && this.getFlux().store("search").getSVGClass(updatedFilter);
             
+            var nextHasInputValue = id !== null;
+            
+            if (nextHasInputValue !== this.state.hasInputValue) {
+                this.setState({ hasInputValue: nextHasInputValue });
```
volfied: Instead of calling setState here, can you add this key-value pair to the setState call that's made below?

DivyaPrabhakar:
@volfied I made the change but I feel the `if` statement is a bit redundant. Let me know what you think and also, can you explain to me why one call to `setState`, instead of two, is better in this case?

volfied:
Since this component does not have a `shouldComponentUpdate`, any calls you make to `setState` will cause a re-render. React usually tries to collect multiple `setState` calls together. In this case though, I'd like a single `setState` call for cleanliness. 

I would call `setState` with all three keys in both cases, and pass the same value. If you really care about the performance, you can implement a `shouldComponentUpdate` method and check the prevalent props and state variables to decide whether component should render or not. Youc an find examples of this in most of our React components.

DivyaPrabhakar:
@volfied Got it, thanks!

volfied:
:+1:

PR #3428, opened 2015-12-11
Enable LESS source map

shaoshing:
The new config will enable LESS source map for internal/remote dev tool.


volfied:
Nice catch!

Issue #3430, opened 2015-12-11
Mask Mode with multiple layers selected does not work


placegraphichere:
- select an image layer and shape layer
- enter Mask Mode
- HUD will show and you can preview the results for the image layer
- Results in path showing that is not actually a mask, unclear what it is
![screen shot 2015-12-11 at 2 28 48 pm](https://cloud.githubusercontent.com/assets/3229167/11754824/8ad5673e-a013-11e5-8ada-9b77e4701b03.png)
![screen shot 2015-12-11 at 2 29 22 pm](https://cloud.githubusercontent.com/assets/3229167/11754829/998c33de-a013-11e5-8584-559e1a8099c5.png)

mcilroyc:
@pineapplespatula after merging your PR, I get a new error.  (note that your unmerged branch seemed to work fine).

two selected layers: 
![ss](http://cl.ly/0x1E2k0j3e0m/Image%202016-01-07%20at%205.07.56%20PM.png)

console (press 'm' and click rectangle tool in HUD) http://cl.ly/3l2k3F0q1w1K

mcilroyc:
Ok.  false alarm on that.  But.. phase two of my concern:   I think the multi-layer-selected case is still weird if the "first" layer is the shape layer.  It works fine if the layers look as my screenshot above, but if they order is reversed, then hitting "m" has no effect.  I'll toss this to @ktaki to get his take on the completion of this bug.  If @pineapplespatula feels that my concern is addressed elsewhere, or otherwise invalid, speak up.

ktaki:
HUD is not shown, but you can draw a vector path with a shape tool. The result is a path drawn but does not create a mask as it was shown in the first description above. I think the vector mask icon and 'm' shortcut key should be disabled if multiple layers are selected.

ktaki:
Retested the latest master branch. A mask is created when a layer and a shape layer are selected and hit m or Cmd-m regardless of the order of layers selected. If more than two layers are selected, the menu and icon on the doc bar are dimmed. Closing this as fixed.

PR #3432, opened 2015-12-11
switch vector mask HUD actions to use Layer Actions Util to avoid iss…

pineapplespatula:
…ues when multiple layers are selected

this is a fix for issue #3430

### DIFF
--- a/src/js/actions/mask.js
+++ b/src/js/actions/mask.js
@@ -34,7 +34,8 @@ define(function (require, exports) {
 
     var locks = require("../locks"),
         events = require("../events"),
-        nls = require("js/util/nls");
+        nls = require("js/util/nls"),
+        layerActionsUtil = require("js/util/layeractions");
 
     var _CLEAR_PATH = 106;
 
@@ -129,7 +130,6 @@ define(function (require, exports) {
         writes: [locks.PS_DOC, locks.JS_DOC]
     };
     
-
     /**
      * Handle escaping from vector mask for the selection-based vector mask tools.
      * Most tools can just call tools.changeVectorMaskMode directly.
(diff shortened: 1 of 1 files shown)


src/js/actions/mask.js
```diff
@@ -269,11 +269,16 @@ define(function (require, exports) {
...
+                playObject: [
+                    vectorMaskLib.makeBoundsWorkPath(bounds),
+                    vectorMaskLib.makeVectorMaskFromWorkPath(),
+                    vectorMaskLib.deleteWorkPath()]
+            });
 
-        return descriptor.batchPlayObjects([vectorMaskLib.makeBoundsWorkPath(bounds),
-                vectorMaskLib.makeVectorMaskFromWorkPath(),
-                vectorMaskLib.deleteWorkPath()], options)
+        return layerActionsUtil.playLayerActions(currentDocument, createActions, true, options)
```
mcilroyc: I *think* you can get away with just using `playSimpleLayerActions` instead of having to prepare a List-Of-One for `playLayerActions`.

mcilroyc:
back to @pineapplespatula to investigate my wild claim.  Otherwise seems fine.

src/js/actions/mask.js
pineapplespatula: you still have to make a List-Of-One for the layer. But, it is reduces the file by 4 lines so i guess its a win.

src/js/actions/mask.js
```diff
@@ -269,11 +269,14 @@ define(function (require, exports) {
                     coalesce: true,
                     suppressHistoryStateNotification: true
                 }
-            };
-
-        return descriptor.batchPlayObjects([vectorMaskLib.makeBoundsWorkPath(bounds),
+            },
+            layerList = Immutable.List.of(currentLayer),
+            createActions= [
```
mcilroyc: :space_invader:  (see travis failure)

pineapplespatula:
@mcilroyc : sorry about that.

mcilroyc:
ok, merging this now.  I have concerns that this only half fixes the referenced issue ... but I'll take up that debate over there.

PR #3441, opened 2015-12-18
Verify selected layer bounds

iwehrman:
This PR adds a post-verification step for selected layer bounds to actions that modify the bounds. It also "cleans up" (and hopefully doesn't break?) the layer transform event handler.

With this on, I immediately notice a few warnings. Some of them are "false" in that there are actions racing to update the bounds, and the post-verification step sometimes runs before a later action fixes things up. I also see a number of what appear to be true (if perhaps not severe?) warnings when transforming layers via the transform panel due to rounding issues. In any case, I think we should add these warnings first and triage them later.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -446,7 +446,8 @@ define(function (require, exports) {
     setPosition.action = {
         reads: [],
         writes: [locks.PS_DOC, locks.JS_DOC],
-        transfers: [layerActions.resetIndex]
+        transfers: [layerActions.resetIndex],
+        post: [layerActions._verifySelectedBounds]
     };
 
     /**
@@ -546,7 +547,8 @@ define(function (require, exports) {
     swapLayers.action = {
         reads: [],
         writes: [locks.PS_DOC, locks.JS_DOC],
-        transfers: [layerActions.resetIndex]
+        transfers: [layerActions.resetIndex],
+        post: [layerActions._verifySelectedBounds]
     };
 
     /**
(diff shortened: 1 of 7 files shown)


src/js/actions/transform.js
```diff
@@ -1241,7 +1250,71 @@ define(function (require, exports) {
...
+
+        var appStore = this.flux.store("application"),
+            currentDoc = appStore.getCurrentDocument();
+
+        if (!currentDoc) {
+            return Promise.resolve();
+        }
+
+        // newDuplicateSheets move events should be processed immediately, not debounced
+        if (event.newDuplicateSheets) {
```
iwehrman: This used to be debounced, but the debouncing was broken in the sense that it was functionally identical to just delaying the code from executing by 200ms. I couldn't think of any reason why that delay was necessary either, so I just removed it. It's entirely possible that this breaks something, but at least it's less mysterious...

src/js/models/layerstructure.js
```diff
@@ -464,7 +464,7 @@ define(function (require, exports, module) {
         /**
          * The subset of Layer models that are all selected or are descendants of selected
          *
-         * @return {Immutable.List.<Layer>}
+         * @return {Immutable.OrderedSet.<Layer>}
```
iwehrman: Unrelated JSDoc fix.

src/js/actions/transform.js
volfied: We'll see!

volfied:
Nice, thanks for doing this.

Issue #3372, opened 2015-12-07
Shrinkwrap third-party dependencies


iwehrman:
We should start using the shrinkwrap feature of NPM to allow our tagged releases and archival branches to specify specific versions of our third-party dependencies. Recently we had trouble with an updated patch-version of the `jsdoc` grunt task returning new warnings on our archived release branch, which caused our CI tests to fail for PRs against that branch. A shrinkwrap file (updated at least for releases) would have averted that problem.

PR #3442, opened 2015-12-18
Add a shrinkwrap file

iwehrman:
This adds an npm-shrinkwrap.json file to the repository. This forces npm install to only choose specific versions of packages that are known to work in our repository. This will help us guard against packages that don't quite follow semver rules. More information here: https://docs.npmjs.com/cli/shrinkwrap

Addresses #3372.


iwehrman:
See also adobe-photoshop/spaces-adapter#178.

Issue #3438, opened 2015-12-17
Remove use of React isMounted 


iwehrman:
`isMounted` in React components is going to be deprecated soon:

https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html

I haven't found any uses of `isMounted` in our code base that seem egregious (i.e., that result in memory leaks due to failing to unregister listeners from unmounted components), but I guess we won't have a choice in the future.

volfied:
For most D3 components, this can be done by keeping a flag, which is the suggested way. 

For Select.jsx we can probably cancel the listener callback if we unmount.

And for `AssetPreviewImage` flag seems like the easier solution. Or we'll have to Promisify entire `_updateRenditionPath`and cancel that promise on `componentWillUnmount`

iwehrman:
Fixed by #3445.

PR #3445, opened 2015-12-18
Use cancellation and local flag to remove isMounted calls

volfied:
Addresses #3438, by removing isMounted calls across the board.

For D3 overlays, we rely on a flag, which is simple/stupid way.

For Select, we cancel the timeout callback using `window.clearTimeout`

For AssetPreviewImage, we cancel the promise, which is a no-op if it's already resolved.

### DIFF
--- a/src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
+++ b/src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
@@ -49,6 +49,15 @@ define(function (require, exports, module) {
          * @type {number}
          */
         _elementLastModified: 0,
+
+        /**
+         * Rendition update promise currently running for this asset preview
+         * will be cancelled if still around during unmount time
+         * This was added to prevent using `this.isMounted()`
+         *
+         * @type {Promise}
+         */
+        _renditionUpdate: null,
         
         getInitialState: function () {
             return {
@@ -63,6 +72,13 @@ define(function (require, exports, module) {
         
         componentWillUpdate: function () {
             this._updateRenditionPath();
+        },
+
+        componentWillUnmount: function () {
+            // This will prevent the callback that sets state from being called
+            if (this._renditionUpdate) {
+                this._renditionUpdate.cancel();
+            }
         },
         
         shouldComponentUpdate: function (nextProps, nextState) {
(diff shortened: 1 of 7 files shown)


src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
```diff
@@ -49,6 +49,15 @@ define(function (require, exports, module) {
...
         _elementLastModified: 0,
+
+        /**
+         * Rendition update promise currently running for this asset preview
+         * will be cancelled if still around during unmount time
+         * This was added to prevent using `this.isMounted()`
+         *
+         * @type {Promise}
+         */
+        _renditionUpdate: Promise.resolve(),
```
iwehrman: It's generally not kosher to define prototype properties using mutable values. I think you're being careful here not to modify the object directly on the prototype, but it would be safer to just initialize this null here and then in `componentWillUnmount` to see if the value is truthy before calling `cancel` on it.

src/js/jsx/shared/Select.jsx
```diff
@@ -143,6 +143,8 @@ define(function (require, exports, module) {
      */
     var Select = React.createClass({
 
+        _mountTimeoutID: null,
```
iwehrman: Needs JSDoc.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -59,6 +59,13 @@ define(function (require, exports, module) {
          */
         _scrimGroup: null,
 
+        /**
+         * Flag to keep track of mounted state
+         *
+         * @type {boolean}
+         */
+        _mounted: null,
```
iwehrman: I can't rationally explain why any of these `isMounted` calls would be necessary here to begin with, since all the calls happen either synchronously in the same frame as a React library call or in a window resize handler. Maybe we can just remove these?

src/js/jsx/tools/PolicyOverlay.jsx
```diff
@@ -64,16 +64,25 @@ define(function (require, exports, module) {
          */
         _drawDebounced: null,
 
+        /**
+         * Flag to keep track of mounted state
+         *
+         * @type {boolean}
+         */
+        _mounted: null,
```
iwehrman: This makes more sense because the draw calls are debounced, and we don't otherwise have a way to cancel a debounced call. Note that lodash's version of `debounce` _does_ provide a cancellation method. So, my guess is that the aforementioned `isMounted` checks were copied and pasted from here? Furthermore, if you ever want to get fancy, you could probably also do this by adding a cancel method to our async-debounced functions.

src/js/jsx/tools/SuperselectVectorMaskOverlay.jsx
```diff
@@ -41,6 +41,21 @@ define(function (require, exports, module) {
...
+         * @type {boolean}
+         */
+        _mounted: null,
+
+        componentDidMount: function () {
+            this._mounted = true;
+        },
+
+        componentWillUnmount: function () {
+            this._mounted = false;
```
iwehrman: My guess is that the mountedness checks can be removed from this file as well.

src/js/jsx/tools/GuidesOverlay.jsx
volfied: c/p artifact, wasn't needed, cleaning up.

volfied:
@iwehrman Thank you for the review. Since we didn't debounce our drawing functions using `synchronization.debounce`, I only used lodash's `debounce.cancel` function.

I've cleaned all `_isMounted` flags as well.

iwehrman:
Brilliant.

Issue #3446, opened 2015-12-19
Rotate input not reset when switching layers


mcilroyc:
This is probably a regression introduced by the recent scrubby PR.  I am working on a fix, so assigning to myself.  cc @volfied 

1. rotate a layer by inputing a number into the input
1. select a different layer
1. note that the rotate input retains the previous value

mcilroyc:
FBNC to @ktaki  (thanks for noticing this one)

ktaki:
Verified the fix.

PR #3447, opened 2015-12-19
Clean up of Rotate.jsx to remove _lastAngle in favor of react state.value

mcilroyc:
This is an attempt to get rid of `_lastAngle` to hopefully simplify the tracking of angle state in the Rotate component.  It seems to work, but am interested in @volfied's keen eye.  This also addresses #3446 which I noticed in master while comparing the two branches.

### DIFF
--- a/src/js/jsx/sections/transform/Rotate.jsx
+++ b/src/js/jsx/sections/transform/Rotate.jsx
@@ -46,14 +46,7 @@ define(function (require, exports, module) {
         },
 
         /**
-         * Last angle on input for relative rotation
-         *
-         * @type {number}
-         */
-        _lastAngle: null,
-
-        /**
-         * Value of the input as user scrubs this control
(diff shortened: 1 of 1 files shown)


src/js/jsx/sections/transform/Rotate.jsx
```diff
@@ -74,14 +67,21 @@ define(function (require, exports, module) {
...
-            // and we want to update them only when selection changes without re-render
+            // If selection has changed, reset the value back to zero
             if (differentSource) {
-                this._lastAngle = 0;
-                this._scrubAngle = 0;
+                this.setState({
+                    differentSource: true,
+                    value: 0
+                });
+            } else {
```
mcilroyc: I amended my original commit with this additional  `else` block - to avoid unnecessary renders

volfied:
Looks great, thank you!

PR #3448, opened 2015-12-19
Anticipate rogue history states when handling transform events.

iwehrman:
In #3441 I "cleaned up" the layer transform handler in a way that may, in some cases, cause history to not be recorded correctly. In particular, the transform events may be associated with rogue history events, and the handler should anticipate this. (Before #3441 we just inserted an awkward delay with the hope that any other events would already have been delivered and handled.) Without handling such out-of-order history events in some way, the later transfers to reset actions would then not amend history correctly.

I haven't actually been able to reproduce this problem, but that may just be because all the transform events I've been able to trigger so far have the correct ordering w.r.t. history states. Or maybe there is some other reason why this isn't actually a problem. In any case, I think this was the narrative that @mcilroyc explained to me, and so at least he should probably look into whether this makes sense!

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -1290,29 +1290,34 @@ define(function (require, exports) {
             if (event.trackerEndedWithoutBreakingHysteresis) {
                 return Promise.resolve();
             } else {
-                var textLayers = currentDoc.layers.allSelected.filter(function (layer) {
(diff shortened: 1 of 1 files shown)


mcilroyc:
This is as I would expect, and my initial tests show this is history-correct.  Our long national nightmare may soon be over.  This particular nightmare.

iwehrman:
:us: :us:

Issue #3452, opened 2015-12-21
Radius control is enabled for artboards


iwehrman:
The Radius control is enabled but does nothing when an artboard is selected. It should be disabled if none of the selected layers are shape layers.

iwehrman:
Fixed by #3456.

PR #3456, opened 2015-12-21
Disable Radius for non-radii layers

volfied:
Addresses #3452 

I've also noticed that we've kept `Vector.jsx` around without requiring it, so cleaned it up.


src/js/jsx/sections/style/Radius.jsx
```diff
@@ -77,7 +77,10 @@ define(function (require, exports, module) {
          * @param {object} props
          */
         _calculateState: function (props) {
-            var layers = props.document.layers.selected.filter(function (layer) {
+            var disabled = props.disabled || props.document.layers.selected.some(function (layer) {
```
iwehrman: I think you've got the wrong quantifier here. You want "all selected layers do not have radii" but this is "some selected layer does not have a radii." As long as the control is valid for some layer it should be enabled, and the invalid layers should be filtered out in the handler.

src/js/jsx/sections/style/Vector.jsx
```diff
@@ -1,65 +0,0 @@
...
-                    <Radius
-                        document={this.props.document}
-                        layers={vectorLayers} />
-                </div>
-            );
-        }
-    });
-
-    module.exports = Vector;
-});
```
iwehrman: :+1:

volfied:
@iwehrman I've made the changes. I've also noticed that we had to account for all layers in the selection to make sure we would re-render correctly.

However, as it stands, mixed appearances panel will always disable Radius when layer kinds are mixed, I will open a new issue for this as I feel like the entire Mixed Appearance panel requires work on this.

iwehrman:
LGTM.

Issue #3450, opened 2015-12-20
Overlays disabled after zooming


iwehrman:
1. Create a new artboard document.
2. Select the artboard. The overlay adders are visible.
3. Cmd-+ to zoom in.

Result: the overlays are disabled even after the operation is complete and remain so even after panning the canvas or modifying the document.

mcilroyc:
FBNC to @chadrolfs

PR #3457, opened 2015-12-21
Declarative overlay management

iwehrman:
This PR adds declarative management of overlay enablement by way of a new `hideOverlays` boolean action property. When actions like this are executing the overlays are disabled, and then they're re-enabled when all the `hideOverlay` actions are complete. This generally simplifies things (fewer explicit event dispatches) and allows for local reasoning ([as they say!](https://en.wikipedia.org/wiki/Separation_logic)) about the state of the overlays.

One exception to this simple handling is the "scroll" event handler. We use a debounced handler there, but we also want to hide the overlays immediately and not reveal them until after the debounced handler has run. I could imagine handling this case declaratively as well, but since it's currently a one-off I didn't bother.

Addresses #3450 and #2714.

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -623,8 +623,7 @@ define(function (require, exports) {
         var eventKind = adapterOS.eventKind.LEFT_MOUSE_DOWN,
             coordinates = [x, y],
             dragModifiers = keyUtil.modifiersToBits(modifiers),
-            diveIn = system.isMac ? modifiers.command : modifiers.control,
-            dontDeselect = modifiers.shift;
+            diveIn = system.isMac ? modifiers.command : modifiers.control;
 
         if (panning) {
             var dragEvent = {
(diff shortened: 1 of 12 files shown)


src/js/actions/superselect.js
```diff
@@ -636,43 +635,44 @@ define(function (require, exports) {
...
-                    })
-                    .catch(function () {}) // Move should silently fail if there are no selected layers
-                    .finally(function () {
-                        // Re show the transform controls, which will appear once we're out of the modal state
-                        return descriptor.playObject(
-                            toolLib.setToolOptions("moveTool", { "$Abbx": true }),
-                            { canExecuteWhileModal: true });
-                    });
-            }
+        var uiStore = this.flux.store("ui"),
```
iwehrman: There are actually no substantive differences here; I just flattened the control structure a little.

src/js/actions/ui.js
```diff
@@ -349,27 +349,51 @@ define(function (require, exports) {
...
-        var DEBOUNCE_DELAY = 200;
-
-        var setTransformDebounced = synchronization.debounce(function (event) {
-            if (event.transform) {
-                return this.flux.actions.ui.setTransform(event.transform);
-            }
-        }, this, DEBOUNCE_DELAY, false);
+        var scrolling = false,
+            updateTransformDebounced = synchronization.debounce(function () {
+                return this.flux.actions.ui.updateTransform()
```
iwehrman: I changed this from `setTransform` to `updateTransform` because at least some of the artboard overlay bugs were due to `set`ing an old transform right after a previous call to `updateTransform`. Currently `setTransform` (which just updates the model with a given transform, typically from an event handler) is dead code. I'm open to removing it, but I thought I'd check with someone else before I did that.

src/js/fluxcontroller.js
```diff
@@ -755,6 +760,14 @@ define(function (require, exports, module) {
                 return actionPromise;
             })
             .tap(function () {
+                if (hideOverlays) {
+                    actionReceiver.dispatch(events.panel.END_CANVAS_UPDATE);
+                }
+
+                if (lockUI && !uiWasLocked) {
```
iwehrman: Unrelated: I moved this into the previous `tap` since there wasn't any reason for an additional handling frame.

src/js/stores/ui.js
```diff
@@ -285,8 +285,6 @@ define(function (require, exports, module) {
             }
 
             this._zoom = payload.zoom;
-            this._overlaysEnabled = true;
```
iwehrman: Unrelated stale property cleanup.

iwehrman:
Also note: when the first action elects to disable overlays, it happens immediately. But when the last action to elect to disable overlays declares that it's state update is complete, re-enablement is delayed by 100ms before doing so to prevent flashing.

iwehrman:
Also also note: I looked through all the actions and this is my best guess about which ones should have `hideOverlays` enabled. It's entirely possible that I either missed some, or was overzealous. For example, you may yourself consider whether or not most of the `transform` actions should have `hideOverlays` enabled. I thought about it and decided no, but the correctness of that decision is perhaps not completely obvious...

src/js/actions/ui.js
mcilroyc: Seems OK to me, but leaving a note here for @volfied in case he wants to weigh in after the fact.

mcilroyc:
I love the approach.  I see no obvious mistakes in where you chose to include `hideOverlays`, and the associated bugs seem to be squashed.  Merging.

Issue #3458, opened 2015-12-22
Document activation is slow


iwehrman:
There are at least two problems:

1. ~~We currently don't start updating the UI until after Photoshop finishes changing the document and updating the canvas. Usually, we should be able to perform these operations in parallel.~~ Fixed by #3459.
2. For documents that are already initialized, we shouldn't need to change more than one or two CSS classes because each document's panel structure is maintained separately in the DOM. I suspect it takes longer than this because our `shouldComponentUpdate` functions aren't sharp enough to limit re-rendering to the top-most components.

iwehrman:
We currently spend ~180ms out of a total of ~360ms (in Flat_Style.psd) doing CSS style recalculation and layout. I think we should be able to restructure the CSS to avoid this.
![image](https://cloud.githubusercontent.com/assets/1075154/12128635/602ff446-b3b3-11e5-826c-cc778c92056c.png)

iwehrman:
Unsurprisingly, it seems that almost all of the style recalculation and layout is caused by the layers panel. Here's what document switching looks like when the layers panel is collapsed:
![image](https://cloud.githubusercontent.com/assets/1075154/12129368/05188398-b3b8-11e5-9e96-4b28b151add0.png)

iwehrman:
Also, as guessed by @designedbyscience, we pay the same exorbitant layout cost when expanding the layers panel:
![image](https://cloud.githubusercontent.com/assets/1075154/12129588/6491ac54-b3b9-11e5-9d36-ab31f99ee49e.png)

iwehrman:
Kicking this to @shaoshing since he's working on a PR for it now.

iwehrman:
@chadrolfs to use judgment about whether or not the problem is solved.

chadrolfs:
seems pretty zippy to me (at least on Mac).  Windows isn't bad but a tad slower than Mac.

PR #3459, opened 2015-12-22
Faster document activation

iwehrman:
~~This relies on #3457, which should be merged before this.~~

This improves perceived document switching time by exploiting additional concurrency in `documents.selectDocument`. In particular, if the document is already initialized, we dispatch the event which causes the UI to update (which is expensive) in parallel with the Photoshop call to select the document (which is also expensive). The former still explicitly follows the latter in case the document has not yet been initialized.

I've tried to generally follow the sequential constraints strongly implied by the existing mask-cleanup actions, but I'm not completely sure that the ordering (or lackthereof) is entirely correct. In particular, I don't really understand why, on master, the call to `deselectAll` follows the Photoshop selection change. I gather from that the it doesn't really matter whether this is called before or after the select command. @pineapplespatula: let me know if this makes sense.

I don't _think_ it's necessary to call `resetLinkedLayers` the first time a document is initialized; only every other time (ugh). @volfied: let me know if this makes sense.

I've removed the call to `history.queryCurrentHistory` because either a) the document is uninitialized, in which case the current history will be initialized by `updateDocument`, or b) the document is already initialized, in which case we don't need to init its history again. @mcilroyc: let me know if this makes sense.

I still think we can do better than this, especially for big documents. In theory the fast case shouldn't take any longer than it takes to update a few top-level CSS classes. But, for big documents, this seems to take longer, presumably because the application store's change event causes more than just the top-level components to re-render. 

Partially addresses #3458. (Just "partially" because of the aforementioned excessive React rendering.)

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -766,48 +766,50 @@ define(function (require, exports) {
         var flux = this.flux,
             documentID = typeof documentSpec === "number" ? documentSpec : documentSpec.id,
             document = flux.stores.document.getDocument(documentID),
-            documentRef = documentLib.referenceBy.id(documentID);
+            documentRef = documentLib.referenceBy.id(documentID),
+            uninitialized = !document.layers;
 
-        return this.transfer("panel.cloak")
+        var selectPromise = descriptor.playObject(documentLib.select(documentRef))
             .bind(this)
             .then(function () {
-                return descriptor.playObject(documentLib.select(documentRef));
-            })
-            .then(function () {
-                if (!document.layers) {
+                var updateTransformPromise = this.transfer(ui.updateTransform),
+                    initPromise;
+
+                if (uninitialized) {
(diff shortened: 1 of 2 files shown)


iwehrman:
Oh, and this also increases the timeout in `Guard` after which we call `resetCursor` from 100 to 500ms, which effectively delays changing the cursor so that, in practice, it shouldn't appear when switching to an already-initialized document.

iwehrman:
Rebased and ready 2 rock.

mcilroyc:
regarding the question of `history.queryCurrentHistory`: Your reasoning seems sound.

volfied:
`resetLinkedLayers` looks good to me, I'm ok to merge this.

volfied:
Assigning to @pineapplespatula for his input on vector mask changes.

pineapplespatula:
LGTM

Issue #3443, opened 2015-12-18
Unhandled error verifying history.revertCurrentDocument postcondition


iwehrman:
```
Verifying 1 postcondition for action history.revertCurrentDocument
fluxcontroller.js:842 Action history.revertCurrentDocument failed: TypeError: Cannot read property 'apply' of undefined
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:77262:41)
    at Array.map (native)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:77261:46)
```

The stack trace above refers to this line: 
https://github.com/adobe-photoshop/spaces-design/blob/228fad007a04c9524691675087d0fa0301284912/src/js/fluxcontroller.js#L766

The specified post is a function reference which turns out to be undefined at this point. I think the problem is a circular module reference among the history and layer actions. Probably the right thing to do is to convert the post conditions to be full-blown actions, and to then allow postconditions to be specified symbolically with strings as is possible elsewhere.

PR #3460, opened 2015-12-22
Verification post-conditions as actions

iwehrman:
This PR converts the verification postconditions to quasi-actions and allows them to be specified symbolically with pathnames instead of just via direct references. This also moves all of the verification actions into new modules underneath `actions/verify` in an attempt to tidy up `actions.documents` and `actions.layers`.

Above I referred to these as quasi-actions because the controller doesn't quite treat them the same way when executing them as action post-conditions. In particular, their locks are ignored and they're always executed in parallel under the reasonable assumption that they are side-effect free. At some point, as @mcilroyc and I have discussed, it would be nice to have both `pre` and `post` action lists which are treated a little more like action transfers, for consistency and uniformity. This is one step in that direction, but more remain.

Addresses #3443.

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -66,6 +66,8 @@ define(function (require, exports, module) {
         transform: require("./transform"),
         type: require("./type"),
         typetool: require("./typetool"),
-        ui: require("./ui")
+        ui: require("./ui"),
+        verifyDocuments: require("./verify/documents.js"),
+        verifyLayers: require("./verify/layers.js")
     };
 });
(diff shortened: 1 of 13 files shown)


src/js/actions/index.js
```diff
@@ -66,6 +66,8 @@ define(function (require, exports, module) {
         transform: require("./transform"),
         type: require("./type"),
         typetool: require("./typetool"),
-        ui: require("./ui")
+        ui: require("./ui"),
+        verifyDocuments: require("./verify/documents.js"),
```
volfied: Not needed in this PR, but it might be useful to have a single index file for verification methods that we can exclude in release builds.

volfied:
Needs rebase!

iwehrman:
Rebased.

Issue #3451, opened 2015-12-21
Radius slider is slow


iwehrman:
It's slow even when only the lone shape in a document is being transformed. I haven't looked into it yet, but I think the reason is that either the `Radius` or `Range` component is stateless. This means that it won't update it's position until it receives a change event from a flux store, which results in a great deal of latency while dragging.

iwehrman:
My initial guess wasn't quite right. My current guess is that the problem is two-fold:

1. We don't get mouse events from the adapter while playing Photoshop commands.
2. We re-render the entire UI at each update, which is terribly inefficient and in itself seems to take 60-80ms.

We can't do anything about the former. We can do something about the latter.

iwehrman:
The scripting time is has been reduced significantly. This is now blocked on the infrequency of mousemove events from Photoshop. CC @jsbache @dtristram

PR #3462, opened 2015-12-22
Experiment: direct radius updates

iwehrman:
This is an experimental change with the goal of speeding up the `Radius` component. With this change, the `DocumentStore` will no longer emit "change" events that result from `RADII_CHANGED` dispatches. Instead, a new, single-purpose "radiusChanged" event is emitted, and only the `Radius` component listens for these events. Consequently, `RADII_CHANGED` dispatches can be handled in ~5ms instead of ~40ms on master. The savings comes from not having to virtually re-render much higher level components without any effect on the DOM. This only works because no other components care about the details of border radii; future React components that care about radii changes would have to listen to this event in addition to "change" events. So, the cost is some extra complexity. Is this additional complexity worthwhile? I don't know; that's why I called this an experiment!

Also note that, incidentally, this _particular_ changes doesn't result in a radius slider that is significantly more responsive despite looking significantly better in the Chromium timeline. I believe this is because we're hitting the same Photoshop/Spaces architectural limitation as in the color picker: mouse events are not delivered while Photoshop is busy playing commands. In particular, commands that set change border radii.

Here's a rough idea of how the timeline changes.
Before:
image

After:
image

The profiles above are at roughly the same zoom level. Each of the two big spikes in the "before" run is from a document "change" handler, and the smaller, secondary spikes are from the menu and overlays. In the "after" run, the two "radiiChange" handlers are of the same magnitude as the secondary spikes and so they're hard to distinguish at a glance. You can also see roughly from the summaries in each run that there isn't too much practical difference in terms of redraw frequency: ~4fps :person_frowning: 

I'm hopeful that we can use a trick like this to reduce the time to redraw the layers panel on `SELECT_LAYERS_BY_ID` dispatches. That should not be similarly plagued with event routing issues. @shaoshing is working on this now.

Partially addresses #3451, but not completely due at least to the aforementioned well known Photoshop architectural issues.

### DIFF
--- a/src/js/jsx/sections/style/Radius.jsx
+++ b/src/js/jsx/sections/style/Radius.jsx
@@ -57,33 +57,53 @@ define(function (require, exports, module) {
                 nextRelevantProps = getRelevantProps(nextProps);
 
             if (!Immutable.is(relevantProps, nextRelevantProps)) {
-                this._calculateState(nextProps);
+                this._calculateState(nextProps.document.layers.selected, nextProps.disabled);
             }
         },
 
+        /**
+         * Handler for the document store's "radiiChange" event, which is emitted
+         * when just the radii of selected layers have changed.
+         *
+         * @param {Document} document
+         */
+        _handleStoreChange: function (document) {
+            if (document.id !== this.props.document.id) {
+                return;
+            }
+
+            this._calculateState(document.layers.selected, this.props.disabled);
+        },
+
         componentWillMount: function () {
-            this._calculateState(this.props);
+            this._calculateState(this.props.document.layers.selected, this.props.disabled);
+            this.getFlux().store("document").on("radiiChange", this._handleStoreChange);
+        },
+
+        componentWillUnmount: function () {
+            this.getFlux().store("document").off("radiiChange", this._handleStoreChange);
         },
 
         /**
-         * Given the props object, calculates and sets the state
-         * This is refactored out of componentWillReceiveProps, because 
-         * it doesn't run on initial render, causing a null state
+         * Given the props object, calculates and sets the state. This is factored
+         * out of componentWillReceiveProps, because it doesn't run on initial render,
+         * causing a null state.
          *
          * @private
-         * @param {object} props
+         * @param {Immutable.List.<Layer>} layers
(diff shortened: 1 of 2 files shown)


mcilroyc:
seems legit.  If we continue adding new specialized change event variations ... it might help to index them somewhere as constants.  Something to help us track down what events exist and might be useful.

Roisholto:
Please I will like to unsubscribe to this mail


Sent from Outlook Mobile






On Mon, Jan 4, 2016 at 9:31 AM -0800, "Cory McIlroy" <notifications@github.com> wrote:





seems legit.  If we continue adding new specialized change event variations ... it might help to index them somewhere as constants.  Something to help us track down what events exist and might be useful.

---

volfied:
@Roisholto I'm sorry you're getting unwanted e-mails. You must have subscribed to the pull request at some point. Please visit https://github.com/adobe-photoshop/spaces-design/pull/3462 and click the Unsubscribe button at the right side-bar to stop receiving emails.

volfied:
@iwehrman Do you want this to be reviewed for merge? Or is this experimental? If you'd like us to merge, can you rebase.

iwehrman:
I've rebased. I'm not 100% sure, but I think it's worth merging. It reduces render time from ~80ms to ~30ms. It doesn't noticeably improve the performance of the slider because the gaps between mouse events are >>> 80ms :person_frowning: It doesn't break the rules in a way that I think is risky, but it certainly does add some complexity. I dunno, I'm happy to have more opinions.

iwehrman:
I've also inquired with @jsbache and @dtristram about the fairly egregious ~200ms gaps between mousemove events. Because, of course, this change would become a lot more compelling if we were getting mousemove events more frequently!

mcilroyc:
I think it is OK to merge this - especially if it serves as a test-bed for future mousemove event rate improvements.

In the long run, I think there's some work to be done to make this pattern scale (ie, pick the right level of granularity of new change events).  "radiiChange" *might* be a little too granular(?)

volfied:
We can worry about granularity of the event name if we decide to implement this wider

volfied:
Roll the dice!

PR #3463, opened 2015-12-23
Statically eliminate debug code

iwehrman:
This PR replaces `global.debug` and its references with the `__PG_DEBUG__` global variable, and instructs uglify to define it as `false`. This makes it easier (read: possible) for uglify to statically elide dead debugging code blocks. This eliminates redundant runtime checks and, for fun, also reduces the size of the compiled bundles by ~2k. This only affects the non-debug compiled builds.


volfied:
Clean.

PR #3469, opened 2016-01-06
LayersPanel: hierarchical structure (part 1)

shaoshing:
This PR applies the following changes to the LayersPanel for better performance:

1. Convert to hierarchical layer structure so that we can make sure of browser rendering with CSS selectors, and thus eliminate the number of components needed to be updated when selecting/expanding a layer/group.

~~2. Stores layer's selected/expanded/visible attributes in LayerFace's internal state so that we can re-render (using `setState`) the component immediately without waiting for the flux cycle to be completed.~~

##### Result

~~* Since we skip the flux cycle, the duration for a layer to react to a select event has been reduced from `100ms-400ms` to `less than 10ms`. (time is measured between `LayerFace#handleLayerClick` and `LayerFace#componentDidUpdate` within the same instance)~~
* With the hierarchical structure, I also observed improvement when selecting/expanding/collapsing/hiding layer group (I did not take any measurement for them)

##### Known Issues

~~* Sometimes the previously selected layer (A) remains highlighted while the currently selected layer (B) is already selected. This is because A has to wait for the flux cycle and it takes longer than B. The solution is using an event emitter to notify A when B is selected.~~
* This is not the target of this PR, but I noticed that mouse over different layer / library asset appears to be lagging even we are not running any scripts.

@placegraphichere I am assigning to this you since you will work on the performance measurement, but anyone is welcomed to review and leave some comments :clap:

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -57,6 +57,15 @@ define(function (require, exports, module) {
          */
         _isDragEventTarget: false,
 
+        /**
+         * Used to suppress scrolling into view when the selection is changed
+         * by clicking directly on a layer face.
+         *
+         * @private
+         * @type {boolean}
+         */
+        _suppressNextScrollTo: false,
+
         getInitialState: function () {
             return {
                 isDropTarget: false,
@@ -65,7 +74,7 @@ define(function (require, exports, module) {
                 dragStyle: null
             };
         },
-        
+
         shouldComponentUpdate: function (nextProps, nextState) {
             // Drag states
             if (this.state.isDragging !== nextState.isDragging ||
@@ -84,35 +93,33 @@ define(function (require, exports, module) {
                 return true;
             }
 
-            var document = this.props.document,
-                nextDocument = nextProps.document;
-
-            // Depth changes
-            var currentDepth = document.layers.depth(this.props.layer),
-                nextDepth = nextDocument.layers.depth(nextProps.layer);
-
-            if (currentDepth !== nextDepth) {
-                return true;
+            return false;
+        },
+        
+        componentDidMount: function () {
+            // Scroll into view if the current layer is the first selected layers.
+            if (this.props.document.layers.selected.last() === this.props.layer) {
+                this._scrollIntoView(false);
+            }
+        },
+        
+        componentDidUpdate: function (prevProps) {
+            if (this._suppressNextScrollTo) {
+                this._suppressNextScrollTo = false;
+                return;
             }
 
-            // Deeper selection changes
-            var childOfSelection = document.layers.hasSelectedAncestor(this.props.layer);
-            if (childOfSelection || nextDocument.layers.hasSelectedAncestor(nextProps.layer)) {
-                if (!Immutable.is(document.layers.allSelected, nextDocument.layers.allSelected)) {
-                    return true;
+            this._scrollIntoView(prevProps.layer.selected);
+        },
+        
+        _scrollIntoView: function (prevSelected) {
+            if (this.props.layer.selected && !prevSelected) {
+                var node = ReactDOM.findDOMNode(this);
+
+                if (node) {
+                    node.scrollIntoViewIfNeeded();
                 }
             }
-
-            // Given that the face hasn't changed and no selected ancestor has changed, this
-            // component only needs to re-render when going from having a collapsed ancestor
-            // (i.e., being hidden) to not having one (i.e., becoming newly visible).
-            var hadCollapsedAncestor = document.layers.hasCollapsedAncestor(this.props.layer),
-                willHaveCollapsedAncestor = nextDocument.layers.hasCollapsedAncestor(nextProps.layer),
-                hadInvisibleAncestor = document.layers.hasInvisibleAncestor(this.props.layer),
-                willHaveInvisibleAncestor = nextDocument.layers.hasInvisibleAncestor(nextProps.layer);
-
-            return hadCollapsedAncestor !== willHaveCollapsedAncestor ||
-                   hadInvisibleAncestor !== willHaveInvisibleAncestor;
         },
 
         /**
@@ -179,6 +186,7 @@ define(function (require, exports, module) {
          */
         _handleLayerClick: function (event) {
             event.stopPropagation();
+            this._suppressNextScrollTo = true;
 
             var modifier = "select";
             if (event.shiftKey) {
@@ -229,13 +237,14 @@ define(function (require, exports, module) {
          * @param {SyntheticEvent} event
          */
         _handleLayerEdit: function (event) {
+            event.stopPropagation();
+
             var layer = this.props.layer;
             if (layer.locked || !layer.visible) {
                 return;
             }
 
             this.getFlux().actions.superselect.editLayer(this.props.document, this.props.layer);
-            event.stopPropagation();
         },
 
         /**
@@ -246,9 +255,10 @@ define(function (require, exports, module) {
          * @param {boolean} toggled Flag for the ToggleButton, true means locked
          */
         _handleLockToggle: function (event, toggled) {
+            event.stopPropagation();
+
             // Locked if toggled, visible if not
             this.getFlux().actions.layers.setLocking(this.props.document, this.props.layer, toggled);
-            event.stopPropagation();
         },
         
         /**
@@ -278,7 +288,7 @@ define(function (require, exports, module) {
 
             // Do not allow reordering to exceed the nesting limit.
             var doc = this.props.document,
-                targetDepth = doc.layers.depth(target);
+                targetDepth = this.props.depth;
 
             // Target depth is incremented if we're dropping INTO a group
             switch (dropPosition) {
@@ -293,7 +303,7 @@ define(function (require, exports, module) {
             default:
                 break;
             }
-
+            
             // When dragging artboards, the nesting limit is 0 because artboard
             // nesting is forbidden.
             var draggingArtboard = draggedLayers.some(function (layer) {
@@ -348,7 +358,7 @@ define(function (require, exports, module) {
          */
         _getDropPosition: function (dragPosition) {
             var layer = this.props.layer,
-                bounds = ReactDOM.findDOMNode(this).getBoundingClientRect(),
+                bounds = ReactDOM.findDOMNode(this.refs.face).getBoundingClientRect(),
                 dropPosition;
 
             if (layer.isGroup) {
@@ -411,10 +421,12 @@ define(function (require, exports, module) {
                 this._isDragEventTarget = false;
             }
             
-            this.setState({
-                isDragging: false,
-                dragStyle: null
-            });
+            if (this.isMounted()) {
+                this.setState({
+                    isDragging: false,
+                    dragStyle: null
+                });
+            }
         },
         
         /**
@@ -490,7 +502,7 @@ define(function (require, exports, module) {
                 dropPosition = isDropTarget ? this._getDropPosition(dragPosition) : null,
                 canDropLayer = isDropTarget && this._validCompatibleDropTarget(
                     this.props.layer, draggedLayers, dropPosition);
-            
+
             this.setState({
                 isDropTarget: canDropLayer,
                 dropPosition: dropPosition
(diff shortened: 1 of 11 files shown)


iwehrman:
Here's something that might be interesting. I looked at a layer expansion in the timeline and saw this:
image
Note that the pi chart shows ~60% of the time in paint, but I only saw a little blip off to the right... Until I scrolled down:
image
There it is! And it's caused by the `GuidesOverlay` update. I wonder if it would make any difference if we delayed this update to the next frame? I.e., instead of having `GuidesOverlay.componentDidUpdate` call `drawOverlay()` directly it instead did something like `requestAnimationFrame(function () { this.drawOverlay() }.bind(this));` Maybe that's not exactly what we want, but something like that? Or maybe it would be just as bad if we did that. I don't know. Just an observation!

iwehrman:
I guess what's happening is that the layout is being forced to update synchronously (instead of waiting for the end of the frame) by the call to `getBoundingClientRect` in `GuidesOverlay`. And this update is expensive because it requires a layout of all the new DOM nodes representing the layer faces. So perhaps this isn't a major problem after all given that there isn't _another_ expensive layout that follows this directly. I'm not quite sure yet...

designedbyscience:
By Layer Expansion, do you mean groups?

iwehrman:
Yes, sorry.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -637,10 +655,11 @@ define(function (require, exports, module) {
                         onDrop={this._handleDrop}
                         onDragTargetMove={this._handleDragTargetMove}
                         onDragTargetLeave={this._handleDragTargetLeave}>
-                        <li className={classnames(layerClasses)}>
+                        <div className={layerClasses}>
```
designedbyscience: Any particular reason for changing the type of tag?

designedbyscience:
For selection, we aren't reducing total time of the operation, just prioritizing the UI update before the action. 

How much work would it be to fix the first Known Issue? I was pretty easily able to click around and have it appear as if 2 items were selected for a few moments while the now-deselected layer was catching up.

designedbyscience:
When collapsing a group, SamplerOverlay updates on the next frame...why does that happen?

iwehrman:
@shaoshing: I think you can disregard my comment above. I'm still messing around with layout, but I don't think your PR significantly affects this. 

However, I wonder if it would make sense to split this into two separate PRs for easier evaluation: one that renders the layer faces hierarchically (and perhaps moving the scrolling code into `LayerFace` as well); and another that creates a fast path for selection changing. What do you think @designedbyscience?

iwehrman:
Basically, I see a decent speedup when expanding and collapsing groups, but I'm having a hard time convincing myself that the selection changes are a net gain. And I'm guessing that the expanding and collapsing change is due to the hierarchical layer face change, and so far it seems like we're mostly agreed that that is a good change. I'm still not sure if the experiment to create a fast path to update layer face selection state is good though, and splitting these PRs up might help me understand the relative merits of that change alone.

shaoshing:
@iwehrman your suggestion sounds good. I've removed the fast path code out of this branch, and will create a new one once this is merged. As far as I can tell, we still gain improvement by using this structure. Also, because the GuideOverlay/DocumentHeader are rendering before the LayerPanel, the forced-synchronous update is disappeared, but like you said, I think the impact is small as we don't see other large rendering after it. 

screen shot 2016-01-11 at 8 21 52 am

@designedbyscience `How much work would it be to fix the first Known Issue?` This requires all LayerFace to listen to the store for selection change event, and the store needs to diff the pre and current layers and notify the subscribers. These will not take longer than 2-3 hours. I will include this experiment in a separate branch. `When collapsing a group, SamplerOverlay updates on the next frame` Hmm, I don't see SamplerOverlay in the timeline. Do you mean SuperSelectOverlay?

iwehrman:
I noticed recently that `SuperselectOverlay` does not have any `shouldComponentUpdate` method, which results in a lot of redundant renders. Even if this isn't a performance problem per se, it definitely adds some unnecessary noise when trying to profile and debug. @volfied: is this something you could look into?

volfied:
@iwehrman https://github.com/adobe-photoshop/spaces-design/issues/3491 to keep track, I'll look into it.

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
iwehrman: The rule is: when you create a new file, use the current year in the copyright header. So, this should be (c) 2016. Also, I think it would be more consistent to name this file `LayerGroup`.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -408,10 +418,12 @@ define(function (require, exports, module) {
                 this._isDragEventTarget = false;
             }
             
-            this.setState({
-                isDragging: false,
-                dragStyle: null
-            });
+            if (this.isMounted()) {
```
iwehrman: Using `isMounted` is apparently considered an anti-pattern by the React team, and I think the method is slated for deprecation. @volfied recently went through the project and removed all such instances. He can give you some suggestions for the best way to deal with this.

iwehrman:
I can't convince myself that this is faster than master when it comes to expanding/collapsing groups. I think I observe a small (~20ms) speed up when selecting leaf layers. @shaoshing: are you seeing something different from that? As an example, I'm using Vermilion with all layers expanded. I'm selecting the `pad copy 4` layers at the top of the tree, and recursively expanding/collapsing `Desktop Large`. @volfied do you have ~15m to test this against master using the Chrome timeline to gather another performance data point?

iwehrman:
Performance observations from @placegraphichere would also be great!

Assigning to @volfied for that second opinion since @designedbyscience is out this week.

src/js/jsx/sections/layers/LayerFace.jsx
volfied: @shaoshing Please refer to https://github.com/adobe-photoshop/spaces-design/pull/3445 for ways I've accomplished this. 

At a cursory glance, I don't see the need for this check. `_getDropPosition` is called only by `_handleDragTargetMove`, and you call setState in there with no checks, which is part of the render callbacks.

volfied:
@shaoshing if you can fix the `isMounted` situation while this is on my plate, I will try to get to it before review today.

volfied:
I've compared action finish times on master vs this branch. Both for `layers.select` and `layers.setGroupExpansion`, master seems faster. Am I missing something ?

volfied:
I am not convinced by the performance gains on this, back to @shaoshing to fix the nits, and @iwehrman to merge it or not.

volfied:
I realize that this is part 1 of 2, so if part 2 actually makes gains based on this PR, this is going to be a win.

shaoshing:
This structure actually has several measurement scenarios. Below is the result of the scenarios I tested. The new structure saves us some time in scripting, but also increases time in rendering due to DOM complexity. 

### Direct selection on layer face (small improvement)
before
69bc7944-e478-4b7c-8247-5820730bfbf9
after 
7c879266-908a-4fb5-adbb-9dab005d28ab

### Artboard selection (better performance)
before
dd1a3f8c-73b7-4b56-99b3-ddcc856b2594
after
530c052a-213a-464e-829e-97489c376b14


### Expand artboard (better)
before
9109fa08-2de5-46c8-a66e-de63ad6963f2
after
e23a3340-7f10-4792-b2ac-026cb80ff4a8

### Collapse artboard (better)
before
b596233e-b829-4b3e-87ca-844b99cdf275

after
83d18412-4924-40fa-b2bb-cb16abe898f2

### Toggling Visibility (worsen)
Before
a5693273-b0c8-41ed-93a5-c51c9709b24a

After
7705ac77-e80f-4826-a353-08d69ed3fba9

shaoshing:
@iwehrman @volfied @designedbyscience  After more experiments on the DOM and styles, I found the reason for long rendering time is likely caused by the large number of `input element` in the LayerPanel. I made a quick change to `replace the input with span` and the performance boost is pretty impressive:

### Artboard Selection
Before
0d841574-b9fd-11e5-9249-a03b2063f9cf
After
screen shot 2016-01-14 at 11 53 51 am

### Expand an Artboard
Before
24cd33fa-b9fd-11e5-9faa-43e8e7a2cd95
After
screen shot 2016-01-14 at 11 55 47 am


### Toggling Artboard Visibility
Before
48308e6e-b9fd-11e5-9f02-f649c1be6369
After
screen shot 2016-01-14 at 11 56 37 am

iwehrman:
I've been doing some more careful benchmarking using some of @shaoshing's hints above to separate out the different use cases. The short version is that I'm fairly convinced that this branch yields decent gains across the board and without any significant individual performance regression. Furthermore, because this **offloads work from the JavaScript engine to the layout engine** I think that this branch has  better potential for future optimization of our styles and layout. In other words, all else being equal, it seems like a good idea to push our performance problems from scripting (where we've worked very hard for a long time to remove performance bottlenecks) to the style and layout where we haven't yet done significant optimization work. 

In a little more detail, I benchmarked this branch against master in a number of different ways, and repeated each class of measurement between 3 and 6 times. I also tracked both overall CPU time and the subset of scripting time specifically. (The idea being that _everything else besides scripting_ remains a potential target for style and layout optimization.) Here are the results:

test | master CPU | shao CPU | master JS | shao JS
------------- | ------------- | ----------- | ----------- | --------------
select leaf layer | 180 | **174** | 144 | **135**
select large artboard | 436 | **402** | 244 | **140**
collapse large artboard | 306 | **271** | **211** | 212
expand large artboard | 506 | **496** | 209 | **200**
toggle visibility of large artboard | 608 | **__408__** | 379 | **__104__**

@shaoshing's branch is the clear (if modest) winner both in overall CPU time and in just the scripting time. :tada:

I haven't yet looked over the code carefully, but I'll do that tomorrow so we can get this merged ASAP. Great work, @shaoshing!

For future reference: I'm looking at SHA 2ef700f71260c22d6530e0146afbca7ae5df5b07.

shaoshing:
@iwehrman thanks for the detailed evaluation! This is an excellent example/template for other performance PR as well. I am also looking forward to the potential improvement that we can achieve from the rendering side.

src/js/jsx/sections/layers/LayerFace.jsx
shaoshing: I am working on this now.

chadrolfs:
Wow, this is really awesome guys!  Nice work @shaoshing and thanks for the additional metrics @iwehrman

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -229,13 +237,14 @@ define(function (require, exports, module) {
          * @param {SyntheticEvent} event
          */
         _handleLayerEdit: function (event) {
+            event.stopPropagation();
```
iwehrman: Is there a particular reason why you moved these `stopPropagation` calls? Or does it just seem like a good pattern?

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+        mixins: [FluxMixin],
+
+        getDefaultProps: function () {
+            return {
+                depth: 0,
+                isRoot: false
+            };
+        },
+
+        render: function () {
```
iwehrman: I'm surprised that there isn't a `shouldComponentUpdate` method here! Was that an explicit choice? Is that because you don't want to bother with a conservative check for whether the child faces should update as well? 

Related question: did you consider a model in which `LayerGroup` has children instead of a prop for it's child faces?

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    var React = require("react"),
+        Fluxxor = require("fluxxor"),
+        FluxMixin = Fluxxor.FluxMixin(React),
+        classnames = require("classnames");
+
+    var DummyLayerFace = require("./DummyLayerFace"),
+        LayerFace = require("./LayerFace");
+
+    var LayersGroup = React.createClass({
```
iwehrman: I still think should be called `LayerGroup` singular :)

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+        mixins: [FluxMixin],
+
+        getDefaultProps: function () {
```
iwehrman: It would be good to add `propTypes` with documentation since this is a new file and we've got no excuse not to!

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -284,9 +149,12 @@ define(function (require, exports, module) {
...
                         className={containerClasses}
-                        onClick={this._handleContainerClick}
-                        onClickCapture={this._handleContainerClickCapture}>
-                        {childComponents}
+                        onClick={this._handleContainerClick}>
+                        <LayersGroup
+                            isRoot={true}
+                            disabled={this.props.disabled}
+                            document={doc}
+                            layerNodes={doc.layers.roots}/>
```
iwehrman: I suggest just naming this prop `layers`, just because it's simpler and easier to understand. Also, the parameter is actually a list of `Layer` models and not `LayerNode` models.

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+            return {
+                depth: 0,
```
iwehrman: In theory you can also use `LayerStructure.prototype.depth` to get the depth of a layer in constant time. It doesn't seem like a problem to duplicate it in the component props though, and it's probably slightly faster this way, so... anyway.

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+                    var className = classnames({
+                        "layer-group": layersGroup,
+                        "layer-group__selected": layersGroup && layer.selected,
+                        "layer-group__collapsed": layersGroup && !layer.expanded,
+                        "layer-group__not-visible": layersGroup && !layer.visible
+                    });
+
+                    results.push(
+                        <li key={layer.key} className={className}>
```
iwehrman: At some point, after this PR, it would be interesting to try getting rid of this extra structure of a `<li>` and then a `<LayerFace>` Instead, maybe the children layers could just be children of the `LayerFace` component?

src/js/jsx/sections/layers/LayersGroup.jsx
```diff
@@ -0,0 +1,103 @@
...
+                                childNodes={layerNode.children}/>
+                            {layersGroup}
+                        </li>
+                    );
+                }
+                
+                return results;
+            }.bind(this), []);
+
+            if (this.props.isRoot) {
```
iwehrman: I think this block only gets executed once per `LayerPanel` render, right? Maybe this one-time logic should happen in the parent component, in which rendering happens less frequently?

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -215,61 +124,17 @@ define(function (require, exports, module) {
         },
 
         render: function () {
-            var doc = this.props.document;
-
-            var layerComponents = doc.layers.allVisibleReversed
-                    .filter(function (layer) {
-                        // Do not render descendants of collapsed layers unless
```
iwehrman: I think this optimization got dropped. Was that a conscious choice or an accident?

src/js/jsx/sections/layers/LayersPanel.jsx
```diff
@@ -158,44 +105,6 @@ define(function (require, exports, module) {
         },
 
         /**
-         * Scrolls the layers panel to make (newly) selected layers visible.
-         *
-         * @param {LayerStructure} layerStructure
-         */
-        _scrollToSelection: function (layerStructure) {
```
iwehrman: I'm glad this logic was simplified. Do you know of any changes in behavior in the new implementation?

iwehrman:
OK, done with this pass. Looks good! Lots of questions, but mostly just for the sake of discussion to make sure most of the plausible stones have been turned over at least once. Back to @shaoshing!

src/js/jsx/sections/layers/LayerFace.jsx
shaoshing: yse, I am moving them to the top to have more unified coding style here.

src/js/jsx/sections/layers/LayersGroup.jsx
shaoshing: `LayerGroup` it is :D

src/js/jsx/sections/layers/LayersGroup.jsx
shaoshing: Good point. I included the solution in my previous PR (https://github.com/adobe-photoshop/spaces-design/commit/05c90a51684af68897a62a2efa4e9f017fed820f#diff-599bf64b66bfd695d31114df0d8d0c63R172), which is a layer diff algorithm in the `LayerPanel` or in the root `LayerGroup`, as we want to avoid repeating the diff in each nested `LayerGroup` when the deepest leaf changed. However, it adds extra complexity, so we decided to not do it.

src/js/jsx/sections/layers/LayersGroup.jsx
shaoshing: I also thought about this structure, but it will require all parent `LayerFace` to be re-rendered when a child layer changed :(

src/js/jsx/sections/layers/LayersGroup.jsx
shaoshing: Yes, only the root layer group has the dummy layer face. I am not sure exposing the `DummyLayerFace` to the `LayerPanel` is a good idea, as they seem unrelated, besides, this will require the `LayerGroup` to provide a way to allow appending layer face to the list, which will add extra complexity.

src/js/jsx/sections/layers/LayersPanel.jsx
shaoshing: Hmm, `doc.layers.roots` indeed is a list of `LayerNode` models. Prove: https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/models/layernode.js#L95-L103

src/js/jsx/sections/layers/LayersPanel.jsx
shaoshing: The behavior remains unchanged. The first newly selected LayerFace will scroll into view unless it is directly selected by user from the LayerPanel

src/js/jsx/sections/layers/LayersPanel.jsx
shaoshing: Another nice catch! I've added this optimization back to the new LayerGroup component.

shaoshing:
@iwehrman thanks again for the thorough evaluation and review. I've updated the PR to reflect your comments, and also replied your questions. Back to you.

src/js/jsx/sections/layers/LayersPanel.jsx
iwehrman: Doh, sure enough. Sorry!

src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -0,0 +1,137 @@
...
+                                document={this.props.document}
+                                layer={layer}
+                                childNodes={layerNode.children}/>
+                            {layerGroup}
+                        </li>
+                    );
+                }
+                
+                return results;
+            }.bind(this), []);
```
iwehrman: I believe `Immutable.Iterable.prototype.reduce` supports a third `receiver` argument. We prefer that, when available, over explicit calls to `.bind`.

src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -0,0 +1,137 @@
...
+         * @type {Boolean}
+         */
+        isRendered: false,
+        
+        propTypes: {
+            /**
+             * List of Layer nodes in the same layer group.
+             * @type {Immutable.List.<LayerNode>}
+             */
+            layerNodes: React.PropTypes.object.isRequired,
```
iwehrman: I think we can actually do a little better here by specifying `React.PropTypes.instanceOf(LayerNode).isRequired`, and similarly for `parentLayer` and `document` below.

iwehrman:
Besides, the additional `shouldComponentUpdate` comment in `LayerGroup`, just two more nits with the call to `.bind` and the stronger `propTypes`.

iwehrman:
Back to @shaoshing.

shaoshing:
@iwehrman I've added comment for the LayerGroup SCU function, and added validation for instance type. Please check it out.

iwehrman:
img

Issue #1766, opened 2015-07-20
Double-click group on open pixel area does nothing


placegraphichere:
- Create a group with empty pixel area between objects. 
- Double click the group in one of those open areas. 
- Nothing happens.

Expected: I would think I would be "entering" the group if it still maintains that I am selecting that group on double click. Having no response feels off.

![double-click-empty-area-in-group](https://cloud.githubusercontent.com/assets/3229167/8780904/b151f296-2ed1-11e5-8e85-ac8677dc0ecf.gif)

volfied:
If it enters the group, which layer do we select?

Another option is to deselect all.

placegraphichere:
enter group, select none?

volfied:
so leave the selection as is, which is current behavior?

volfied:
I don't think this is a bug. We don't have a concept of "entering" a group. Being in a group means having one or more of it's children selected. Double clicking an empty area in a group should be a no-op, as user is diving into nothing.

chadrolfs:
It should deselect all.

placegraphichere:
agree with @chadrolfs. It should deselect. Ai and Sketch have slightly different experiences with this, but both deselect. Double click the empty area in Sketch will deselect and attempt to select what is behind the group (deselects if nothing is there). In Ai, their selection model is more accurate, and a single click will select whatever is behind (or deselect if nothing is there).

chadrolfs:
fix for M3 if time permits and fix is low risk

iwehrman:
The current behavior is not flagrantly broken, and I do not think this is a low-risk change, so I'm deferring to M4.

mcilroyc:
Fixed by recent PR.  Seems like it works as was decided by the above discussion.   FBNC to @chadrolfs

chadrolfs:
Looks good on mac, checking win...

chadrolfs:
:tada:

PR #3472, opened 2016-01-07
Take hitTest into consideration when falling through in double click

volfied:
Addresses #1766 

Problem was, with first click on a double click, we would deselect, and then we would "fall through" which took group bounds into consideration, but not pixel areas that PS tells us.

So now, I make sure that layers we fall through into are actually covered by the click, which prevents the "reselect" of a group when double clicking an empty area within that group.


mcilroyc:
LGTM.  works different now.

Issue #3111, opened 2015-10-23
Sampling layer behind a masked group shows incorrect overlay


placegraphichere:
This one is kind of hard to describe, but the video shows it pretty clearly. 

Trying to sample a layer behind a layer with a large px area but is masked will show the overlay for the large pixel layer, but it will sample correctly. This is just showing the wrong overlay hinting. It works correctly in the end.

![mask-sampler-hittest](https://cloud.githubusercontent.com/assets/3229167/10705445/892ced8a-79a3-11e5-9a4e-919f9cd58639.gif)

iwehrman:
FBNC to @placegraphichere and/or @chadrolfs.

chadrolfs:
Looks good to me, closing this puppy.

PR #3473, opened 2016-01-07
Use hitTest for sampler overlay highlighting

volfied:
Adresses #3111 - we were using our own bounds calculations, which would consider masked layers in their full rectangle. Now we rely on hitTest, similar to what we do with superselect, so masked out areas don't highlight the layer.

### DIFF
--- a/src/js/jsx/tools/SamplerOverlay.jsx
+++ b/src/js/jsx/tools/SamplerOverlay.jsx
@@ -36,7 +36,9 @@ define(function (require, exports, module) {
 
     var system = require("js/util/system"),
         headlights = require("js/util/headlights"),
-        mathUtil = require("js/util/math");
+        mathUtil = require("js/util/math"),
+        uiUtil = require("js/util/ui"),
+        collection = require("js/util/collection");
 
     // Used for debouncing the overlay drawing
     var DEBOUNCE_DELAY = 200;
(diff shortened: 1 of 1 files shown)


iwehrman:
It worries me that the promise returned from hitTest isn't used to synchronize this function. It seems like we could be executing a lot of these update calls concurrently. Should this function return a promise and then be throttled?

src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -518,34 +520,49 @@ define(function (require, exports, module) {
...
-                    return;
-                }
+            uiUtil.hitTestLayers(this.state.document.id, canvasMouse.x, canvasMouse.y)
+                .bind(this)
+                .then(function (hitLayerIDs) {
+                    var selectableLayers = this.state.document.layers.selectable,
+                        selectableLayerIDs = collection.pluck(selectableLayers, "id"),
+                        considerIDs = collection.intersection(hitLayerIDs, selectableLayerIDs);
+
+                    return considerIDs.findLast(function (id) {
```
pineapplespatula: what happens in the case where there isn't a layer returned by hitTestLayers?

volfied:
@iwehrman we do the same thing in superselect highlighting as well. If we throttle/debounce this, then highlighting will appear broken to the user and look worse. So far it worked ok, and I'm hoping with the proposed hitTest improvements on the core side, things will work faster in the future.

@pineapplespatula when hitTest returns an empty array, intersection is also empty, and nothing gets highlighted.

mcilroyc:
nudging to @iwehrman because he had a concern that may be explained by barkin's comment.

iwehrman:
I don't love it, but OK.

Issue #3158, opened 2015-10-28
Marquee select doesn't work dragging from masked out area


chadrolfs:
Open CannoSelectLayers.psb
click and drag in empty space to marquee select

Result: no marquee

Workaround: click+drag from outside masked layer bounds or shift+click on layers or shift+drag

volfied:
@chadrolfs Is this fixed alongside all other highlighting/marquee fixes?

chadrolfs:
no, this is still an issue with current resources, pgdev.921 master @ dd265a7

volfied:
After #3474 is merged, please consider that #3475 has been created to keep track of marquee select issues on vector masked layers with polygon masks.

mcilroyc:
Addressed (partially?) by 3474.  passing to @chadrolfs for review.

PR #3474, opened 2016-01-07
Use hitTest to start marquee select

volfied:
Marquee select only starts if there is nothing under the pointer. With vector masked layers, we would consider masked out areas still within the layer, so now we hitTest to double check the click/drag location and start marquee on those areas.

Adresses #3158

### DIFF
--- a/src/js/actions/superselect.js
+++ b/src/js/actions/superselect.js
@@ -641,37 +641,43 @@ define(function (require, exports) {
         }
 
         var uiStore = this.flux.store("ui"),
-            canvasCoords = uiStore.transformWindowToCanvas(x, y),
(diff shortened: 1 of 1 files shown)


mcilroyc:
Seems fine.  I'll ask the obvious question: does the added hitTest give a noticeable lag?  (I don't notice it being any worse than before)

volfied:
It certainly adds some lag, but given how widely we use it and we don't have another choice to fix this issue, I think this is okay.

Marquee will start from where user clicked, so the lag will not affect the user's experience, only delay the marquee drawing a little bit.

mcilroyc:
> the lag will not affect the user's experience, only delay the marquee drawing a little bit
that seems like it *does* affect the user's experience! : )

anyway, seems like this is better than before.

Issue #3163, opened 2015-10-28
Vector Combine icons and commands execute on layers with masks


chadrolfs:
cmd/ctrl + n
drag a LLSO onto canvas (hit Enter to place) or create a Type layer
apply a mask to LLSO (m, r, drag a rect mask, m to exit)
drag a rect shape on canvas
select both layers
hit one of the vector Combine icons

Result: layers are merged/flattened/rasterized into a pixel layer.

chadrolfs:
it's certainly a workaround if you want to rasterize your type, heh.

PR #3476, opened 2016-01-07
disable combine operations between different types of path objects

pineapplespatula:
fix for issue #3163

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -586,6 +586,17 @@ define(function (require, exports, module) {
                 !curLayer.isVector &&
                 !curLayer.isBackground &&
                 !curLayer.locked;
+        },
+
+        /**
+         * determine if there are only Shape layers in the current selection
+         *
+         * @return {boolean} if the layer selection contains only vector shapes
+         */
+        "selectedLayersContainOnlyShapeLayers": function () {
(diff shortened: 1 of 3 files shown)


volfied:
While this is correct, I had to go in `layerstructure.js` to see how `selectedLayersCanHaveVectorMask` is important to disabling combine operations. I think you should instead add a layerstructure property to make sure all selected layers are shape layers. It's better to check for 1 type than n-1 types.

pineapplespatula:
I added a selectedLayersContainOnlyShapeLayers function which seems to be a little simpler

src/js/models/layerstructure.js
```diff
@@ -586,6 +586,17 @@ define(function (require, exports, module) {
...
                 !curLayer.locked;
+        },
+
+        /**
+         * determine if there are only Shape layers in the current selection
+         *
+         * @return {boolean} if the layer selection contains only vector shapes
+         */
+        "selectedLayersContainOnlyShapeLayers": function () {
+            return this.selected.all(function (layer) {
```
volfied: Can you use `.every()` here, even if .all works, it's not documented anywhere, and is not conventional. We use `every` everywhere else.

volfied:
Back to you @pineapplespatula

pineapplespatula:
@volfied  :repeat:

PR #3479, opened 2016-01-08
Integrate Adobe Stock search into search bar within DS

DivyaPrabhakar:
This PR adds the option to search Adobe Stock via the search bar. (JIRA PS-656) 
It also solves issues #3610, #3529, #3524. 

Testing: 

- The category "Adobe Stock" is now added to the search bar as a potential category the user can choose

- If the user selects the category "Adobe Stock", the user can: 
      > - Enter a string, press `enter` and adobestock.com will open to the results page of the user search 
      > - Enter no string, press `enter` and adobestock.com homepage will open 
      > - Enter a string, click the option `Search Adobe Stock...` and adobestock.com will open to the results page of the user search 
      > - Enter no string, click the option `Search Adobe Stock...` and adobestock.com homepage will open

- If the user selects any or no category within the search bar: 
      >  - If the user types in any string to the search bar and there are no results found, pressing `enter` should execute no action and nothing should happen to the state of the search bar.
      >  - If the user types in a string to the search bar and there are no results, and the user clicks the option `No Results Match`, nothing should happen to the state of the search bar. 
      >  - If the user presses `tab` to autocomplete and there is a a suggestion, the user inputted string should autocomplete. 
      >  - If the user presses `tab` to autocomplete and there is a a suggestion, the user inputted string should autocomplete. 
     >  - If the user presses `tab` to autocomplete and there is no suggestion, the user inputted string not change.
     > - If the user searches for a partial match of the option string, and then presses `enter`, the first option out of the list of options should select. Example: User types `documents`, the options list returns with "Open Documents", "Recent Documents", "documents-1353.psd", pressing enter should choose the option "Open Documents". 

- The category "Adobe Stock" is now added and after selecting it, entering a string within the input field and pressing "Enter" should open a new tab in the browser with the executed search and its results. 
- Searching Adobe Stock from DS will only happen when the "Adobe Stock" filter is applied. 
- When the "Adobe Stock" filter is applied, only the search bar is present, "No options for this result" should not be shown.
- From the browser, the user can choose an image and drag it directly into their document, save it to their local machine, or add it to their library.

### DIFF
--- a/src/js/actions/search.js
+++ b/src/js/actions/search.js
@@ -60,11 +60,16 @@ define(function (require, exports) {
         transfers: [dialog.openDialog, dialog.closeDialog]
     };
 
+    /**
+     * Registers the search payloads to the search store to load the search bar
+     *
+     * @return {Promise}
+     */
     var beforeStartup = function () {
         var searchStore = this.flux.store("search");
 
         searchStore.registerSearch(ID,
-            ["CURRENT_DOC", "RECENT_DOC", "ALL_LAYER", "MENU_COMMAND", "LIBRARY", "GLOBAL_SHORTCUT"]);
+            ["ADOBE_STOCK", "CURRENT_DOC", "RECENT_DOC", "ALL_LAYER", "MENU_COMMAND", "LIBRARY", "GLOBAL_SHORTCUT"]);
 
         return Promise.resolve();
     };
--- a/src/js/actions/search/stock.js
+++ b/src/js/actions/search/stock.js
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2016 Adobe Systems Incorporated. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+define(function (require, exports) {
+    "use strict";
+
+    var Immutable = require("immutable");
+
+    var events = require("js/events"),
+        locks = require("js/locks"),
+        adapter = require("adapter"),
+        nls = require("js/util/nls");
+
+    var SEARCH_STOCK_URL = "https://stock.adobe.com/search?k=",
+        ADOBE_STOCK_URL = "https://stock.adobe.com/";
+    
+    /**
+     * Executes Adobe Stock search by opening a URL with the search term
+     *
+     * @return {Promise}
+     */
+    var _confirmStockSearch = function (searchTerm) {
+        var searchURL;
+        
+        if (searchTerm) {
+            searchURL = encodeURI(SEARCH_STOCK_URL + searchTerm);
+        } else {
+            searchURL = encodeURI(ADOBE_STOCK_URL);
+        }
+
+        return adapter.openURLInDefaultBrowser(searchURL);
+    };
+
+    /**
+     * Find SVG class for Adobe Stock
+     * If this needs to vary based on the item, use category list as parameter 
+     * (see getSVGCallback type in search store)
+     * 
+     * @return {string}
+     */
+    var _getSVGClass = function () {
+        return "libraries-stock";
+    };
+
+    /**
+     * Make list for Stock options to choose from, currently empty
+     * 
+     * @private
+     * @return {Immutable.List.<object>}
+     */
+    var _adobeStockSearchOptions = function () {
+        return Immutable.List();
+    };
+
+    /**
+     * Creates an object with properties to populate the search bar 
+     * datalist when there are no options present. 
+     * 
+     * @private
+     * @return {object}
+     */
+    var _createNoOptionsDefault = function () {
+        return {
+            "noOptionsString": nls.localize("strings.SEARCH.NO_OPTIONS_STOCK"),
+            "noOptionsExecute": _confirmStockSearch,
+            "noOptionsType": "filter"
+        };
+    };
+
+    /**
+     * Register Adobe Stock payload information for search
+     *
+     * @return {Promise}
+     */
+    var registerAdobeStock = function () {
+        var adobeStockPayload = {
+            "type": "ADOBE_STOCK",
+            "getOptions": _adobeStockSearchOptions,
+            "filters": Immutable.List.of("ADOBE_STOCK"),
+            "handleExecute": _confirmStockSearch,
+            "shortenPaths": false,
+            "getSVGClass": _getSVGClass,
+            "noOptionsDefault": _createNoOptionsDefault
+        };
+
+        return this.dispatchAsync(events.search.REGISTER_SEARCH_PROVIDER, adobeStockPayload);
+    };
+    registerAdobeStock.action = {
+        reads: [],
+        writes: [locks.JS_SEARCH]
+    };
+
+    /**
+     * Send information about Adobe Stock to search store
+     *
+     * @return {Promise}
+     */
+    var afterStartup = function () {
+        return this.transfer("search.stock.registerAdobeStock");
+    };
+    afterStartup.action = {
+        reads: [],
+        writes: [],
+        transfers: ["search.stock.registerAdobeStock"]
+    };
+
+    exports.registerAdobeStock = registerAdobeStock;
+    exports.afterStartup = afterStartup;
+});
--- a/src/js/jsx/Search.jsx
+++ b/src/js/jsx/Search.jsx
@@ -53,14 +53,26 @@ define(function (require, exports, module) {
          * 
          * @private
          * @param {string} itemID ID of selected option
+         * @param {function=} noOptionsFn function to execute if there are no options
          */
-        _handleOption: function (itemID) {
+        _handleOption: function (itemID, noOptionsFn) {
+            var searchStore = this.getFlux().store("search");
             this._closeSearchBar()
                 .bind(this)
                 .then(function () {
-                    if (itemID) {
-                        var searchStore = this.getFlux().store("search");
-                        searchStore.handleExecute(itemID);
+                    // If there is a noOptionsFn, this indicates we want the default option,
+                    // that is present in the datalist when there are no results available,
+                    // to execute some behavior if selected. E.g. Selecting the filter
+                    // "Adobe Stock" defaults into "Search Adobe Stock..." and upon selection
+                    // we want it to execute a function to open the Stock homepage 
+                    if (noOptionsFn) {
+                        noOptionsFn(itemID);
+                    }
+                    // If there is noOptionsFn, this indicates the datalist has results available.
+                    else {
+                        if (itemID) {
+                            searchStore.handleExecute(itemID);
+                        }
                     }
                 });
         },
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -65,7 +65,6 @@ define(function (require, exports, module) {
                 searchStore = flux.store("search"),
                 searchState = searchStore.getState(this.props.searchID),
                 options = searchState.searchItems,
-
                 filterItems = options.filter(function (opt) {
                     return opt.type === "filter";
                 }),
@@ -104,7 +103,7 @@ define(function (require, exports, module) {
         getInitialState: function () {
             return {
                 // The currently applied filter
-                filter: [],
+                filters: [],
                 // SVG class for icon for the currently applied filter
                 icon: null,
                 hasInputValue: null
@@ -171,8 +170,8 @@ define(function (require, exports, module) {
                 this.refs.datalist.focus();
             }
 
-            if (prevState.filter !== this.state.filter && this.refs.datalist) {
-                this._updateDatalistInput(this.state.filter);
+            if (prevState.filters !== this.state.filters && this.refs.datalist) {
+                this._updateDatalistInput(this.state.filters);
                 // Force update because Datalist's state might not change at all
                 this.refs.datalist.forceUpdate();
             }
@@ -198,12 +197,12 @@ define(function (require, exports, module) {
             var idArray = id ? id.split("-") : [],
                 filterValues = _.drop(idArray);
 
-            var updatedFilter = id ? _.uniq(this.state.filter.concat(filterValues)) : [],
+            var updatedFilter = id ? _.uniq(this.state.filters.concat(filterValues)) : [],
                 filterIcon = id && this.getFlux().store("search").getSVGClass(updatedFilter),
                 nextHasInputValue = id !== null;
             
             this.setState({
-                filter: updatedFilter,
+                filters: updatedFilter,
                 icon: filterIcon,
                 hasInputValue: nextHasInputValue
             });
@@ -212,12 +211,11 @@ define(function (require, exports, module) {
         /**
          * Removes words from Datalist input that are contained in the ID
          *
-         * @param {Array.<string>} id ID of selected item
+         * @param {Array.<string>} idArray of selected items
          */
-        _updateDatalistInput: function (id) {
-            if (id) {
-                var currFilter = this.refs.datalist.getInputValue().split(" "),
-                    idString = _.map(id, function (idWord) {
+        _updateDatalistInput: function (idArray) {
+            if (idArray) {
+                var idString = _.map(idArray, function (idWord) {
                         var toRemove;
                         
                         try {
@@ -229,10 +227,25 @@ define(function (require, exports, module) {
 
                         return toRemove ? toRemove.replace(" ", "") : idWord;
                     }, this).join("").toLowerCase(),
+                    partialMatch,
+                    currFilter = this.refs.datalist.getInputValue().split(" "),
                     nextFilterMap = _.map(currFilter, function (word) {
-                        return idString.includes(word.toLowerCase()) ? "" : word;
+                        // This allows the datalist to populate with all or partial string matches 
+                        // to the user inputted string.
+                        if (idString.includes(word.toLowerCase())) {
+                            partialMatch = true;
+                            return "";
+                        } else {
+                            return word;
+                        }
                     }),
+                    nextFilter;
+                    
+                if (partialMatch) {
+                    nextFilter = "";
+                } else {
                     nextFilter = nextFilterMap.join(" ").trim();
+                }
 
                 this.refs.datalist.updateFilter(nextFilter);
             } else {
@@ -282,7 +295,7 @@ define(function (require, exports, module) {
                     if (option.type === "filter") {
                         // If it is the filter option for something that we already have filtered, don't
                         // show that filter option
-                        if (_.isEqual(this.state.filter, category)) {
+                        if (_.isEqual(this.state.filters, category)) {
                             return;
                         }
 
@@ -292,9 +305,9 @@ define(function (require, exports, module) {
                         }
                     }
 
-                    // All terms in this.state.filter must be in the option's category
-                    if (this.state.filter && this.state.filter.length > 0) {
-                        var dontUseTerm = _.some(this.state.filter, function (filterValue) {
+                    // All terms in this.state.filters must be in the option's category
+                    if (this.state.filters && this.state.filters.length > 0) {
+                        var dontUseTerm = _.some(this.state.filters, function (filterValue) {
                             return (!_.contains(category, filterValue));
                         });
 
@@ -417,14 +430,19 @@ define(function (require, exports, module) {
          *
          * @private
          * @type {Datalist~onChange}
+         * @param {string=} id of the selected option
+         * @param {string} searchTerm user input
+         * @return {{dontCloseDialog: boolean=}};
          */
-        _handleChange: function (id) {
-            if (id === null) {
+        _handleChange: function (id, searchTerm) {
+            if (!id) {
                 this.props.dismissDialog();
                 return;
             }
 
-            var filters = this.state.filter,
+            // These category filters, e.g. Choosing to search items only within the filter "Pixel Layers"
+            var filters = this.state.filters,
+                lastFilter = filters[filters.length - 1],
                 filtersString,
                 idSplit = id.split("-"),
                 type = idSplit[0],
@@ -437,8 +455,16 @@ define(function (require, exports, module) {
             } else {
                 filtersString = "filter-active";
             }
-            
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                // This payload will be used to dictate whether given the applied filters (0 or more), if 
+                // we want to take any action in the case the user input returns no search results
+                registeredSearchPayload = lastFilter ? searchStore.getPayloadFromFilter(lastFilter) : null;
+
             if (id !== PLACEHOLDER_ID) {
+                // This indicates there is a match to the user's input and they have selected an option
+                // from the datalist of options
                 if (_.contains(this.state.filterIDs, id)) {
                     this._updateFilter(id);
                     
@@ -447,6 +473,15 @@ define(function (require, exports, module) {
                 } else {
                     this.props.executeOption(id);
                     headlights.logEvent("search", filtersString, _.kebabCase(category));
+                }
+            } else {
+                // This indicates that the user has selected the default option when the datalist returns 
+                // no matching results to the user's input.
+                if (registeredSearchPayload && registeredSearchPayload.noOptionsDefault) {
+                    // Need to check if there is action to be taken when there are no options
+                    this.props.executeOption(searchTerm, registeredSearchPayload.noOptionsDefault().noOptionsExecute);
+                } else {
+                    return { dontCloseDialog: true };
                 }
             }
         },
@@ -462,20 +497,30 @@ define(function (require, exports, module) {
             var nextHasInputValue = value.length !== 0;
             
             if (nextHasInputValue !== this.state.hasInputValue) {
-                this.setState({ hasInputValue: nextHasInputValue });
+                this.setState({
+                    hasInputValue: nextHasInputValue
+                });
             }
         },
 
         /**
          * @private
          * @type {Datalist~onKeyDown}
+         * @param {SyntheticEvent} event
+         * @param {string} selectedID the id of the selected item in the datalist
+         * @param {Array.<Option>} optionsList of all the available filtered options
          */
-        _handleKeyDown: function (event, selectedID) {
+        _handleKeyDown: function (event, selectedID, optionsList) {
             switch (event.key) {
             case "Return":
+                var noOptionsExecute = optionsList ? optionsList.first().handleExecute : null;
+                if (!noOptionsExecute) {
+                    return { preventListDefault: true };
+                }
+                break;
             case "Enter":
-                if (selectedID === PLACEHOLDER_ID) {
-                    this._handleDialogClick(event);
+                if (selectedID === PLACEHOLDER_ID || !selectedID) {
+                    this._handleDialogClick(event, selectedID);
                 } else if (_.contains(this.state.filterIDs, selectedID)) {
                     this._updateFilter(selectedID);
 
@@ -493,7 +538,7 @@ define(function (require, exports, module) {
                 this.props.dismissDialog();
                 break;
             case "Backspace":
-                if (this.refs.datalist.cursorAtBeginning() && this.state.filter.length > 0) {
+                if (this.refs.datalist.cursorAtBeginning() && this.state.filters.length > 0) {
                     // Clear filter and icon
                     this._updateFilter(null);
                 }
@@ -510,7 +555,7 @@ define(function (require, exports, module) {
 
         render: function () {
             var searchStrings = nls.localize("strings.SEARCH"),
-                filters = this.state.filter,
+                filters = this.state.filters,
                 placeholderText,
                 clearInputBtn;
 
@@ -519,20 +564,28 @@ define(function (require, exports, module) {
                 lastFilter = filters[filters.length - 1];
 
             var searchPayload = searchStore.getPayloadFromFilter(lastFilter),
-                noOptionsString = searchPayload ? searchPayload.noOptionsString : null,
-                defaultNoOptions = nls.localize("strings.SEARCH.NO_OPTIONS");
+                // This returns a function that when called will return an object with the various properties
+                // for the case when that filter has no options associated with it
+                noOptions = searchPayload ? searchPayload.noOptionsDefault : null,
+                noOptionsObject = noOptions ? noOptions() : null,
+                noOptionsString = noOptionsObject ? noOptionsObject.noOptionsString : null,
+                noOptionsExecute = noOptionsObject ? noOptionsObject.noOptionsExecute : null,
+                noOptionsType = noOptionsObject ? noOptionsObject.noOptionsType : "placeholder",
+                defaultNoOptionsString = nls.localize("strings.SEARCH.NO_OPTIONS");
 
             var noMatchesOption = {
                 id: PLACEHOLDER_ID,
-                title: noOptionsString ? noOptionsString : defaultNoOptions,
-                type: "placeholder"
+                title: noOptionsString ? noOptionsString : defaultNoOptionsString,
+                titleType: noOptionsString ? "custom" : "default",
+                handleExecute: noOptionsExecute,
+                type: noOptionsType
             };
 
             // If we have applied a filter, change the placeholder text
             if (filters.length > 0) {
                 var categoryString = searchStrings.CATEGORIES[lastFilter],
                     category = categoryString ?
-                        categoryString.toLowerCase() : this.state.safeFilterNameMap[lastFilter];
+                        categoryString : this.state.safeFilterNameMap[lastFilter];
                 placeholderText = searchStrings.PLACEHOLDER_FILTER + category;
             } else if (!this.state.ready) {
                 placeholderText = searchStrings.PLACEHOLDER_INITIALIZING;
--- a/src/js/jsx/shared/Datalist.jsx
+++ b/src/js/jsx/shared/Datalist.jsx
@@ -305,7 +305,8 @@ define(function (require, exports, module) {
          */
         _handleInputKeyDown: function (event) {
             var dialog = this.refs.dialog,
-                select = this.refs.select;
+                select = this.refs.select,
+                selectedID;
 
             if (!select) {
                 switch (event.key) {
@@ -324,10 +325,25 @@ define(function (require, exports, module) {
                 }
             }
 
-            var cbOptions = this.props.onKeyDown(event, this.state.lastHighlightedID),
-                options = _.merge({ preventListDefault: false }, cbOptions);
+            // Finishes the autofill within the datalist
+            if (event.key === "Tab") {
+                selectedID = this.state.filter + this.refs.autocomplete.innerHTML;
+                this.setState({
+                    filter: selectedID
+                });
+            } else {
+                // The only time that the lastHighlightedID is undefined is when there is a partial string match 
+                // but the entire string does not match. Example: When user correctly spells the first word 
+                // of a search but incorrectly spells the second word of the search (i.e. Smart ojct). In that case
+                // take the first highlighted selection of the options list, skipping the header.
+                selectedID = this.state.lastHighlightedID ? this.state.lastHighlightedID : null;
+            }
 
-            if (!options.preventListDefault) {
+            var optionsList = select ? select.props.options : null,
+                cbOptions = this.props.onKeyDown(event, selectedID, optionsList),
+                preventList = _.merge({ preventListDefault: false }, cbOptions);
+
+            if (!preventList.preventListDefault) {
                 switch (event.key) {
                 case "ArrowUp":
                     select.selectPrev();
@@ -380,7 +396,8 @@ define(function (require, exports, module) {
          */
         _handleSelectClick: function (event, action) {
             var selectedID = action !== "apply" ? null : this.state.lastHighlightedID,
-                options = _.merge({ dontCloseDialog: false }, this.props.onChange(selectedID));
+                // The filter within datalist is the user input that in turn filters the datalist
+                options = _.merge({ dontCloseDialog: false }, this.props.onChange(selectedID, this.state.filter));
 
             if (!options.dontCloseDialog) {
                 var dialog = this.refs.dialog;
@@ -595,7 +612,18 @@ define(function (require, exports, module) {
                 });
             }
 
-            if (this.props.startFocused && this.refs.textInput) {
+            if (!this.refs.select) {
+                return;
+            }
+
+            var currentOptions = this.refs.select._getOptions(),
+                id;
+
+            if (currentOptions && currentOptions.size === 1) {
+                id = currentOptions.first().id;
+            }
+
+            if (this.props.startFocused && this.refs.textInput && id !== this.props.placeholderOption.id) {
                 this.focus();
             }
         },
@@ -695,23 +723,24 @@ define(function (require, exports, module) {
             }
 
             var dialog = searchableOptions && (
-                    <Dialog
-                        ref="dialog"
-                        id={"datalist-" + this.props.list + this._uniqkey}
-                        className={this.props.className}
-                        onOpen={this._handleDialogOpen}
-                        onClose={this._handleDialogClose}>
-                        <Select
-                            ref="select"
-                            options={searchableOptions}
-                            defaultSelected={selectedID}
-                            useAutofill={this.props.useAutofill}
-                            sorted={this.props.sorted}
-                            onChange={this._handleSelectChange}
-                            onClick={this._handleSelectClick}
-                            onClose={this._handleSelectClose} />
-                    </Dialog>
-                );
+                <Dialog
+                    ref="dialog"
+                    id={"datalist-" + this.props.list + this._uniqkey}
+                    className={this.props.className}
+                    onOpen={this._handleDialogOpen}
+                    onClose={this._handleDialogClose}>
+                    <Select
+                        ref="select"
+                        scrollToSearch={this.props.scrollToSearch}
+                        options={searchableOptions}
+                        defaultSelected={selectedID}
+                        useAutofill={this.props.useAutofill}
+                        sorted={this.props.sorted}
+                        onChange={this._handleSelectChange}
+                        onClick={this._handleSelectClick}
+                        onClose={this._handleSelectClose} />
+                </Dialog>
+            );
 
             var size = this.props.size,
                 svg = null;
--- a/src/js/stores/search.js
+++ b/src/js/stores/search.js
@@ -180,6 +180,7 @@ define(function (require, exports, module) {
          * Perform action for item
          *
          * @param {string} itemID Search item to handle
+         * @param {string=} lastFilter applied
          */
         handleExecute: function (itemID) {
             var idArray = itemID.split("-"),
@@ -235,7 +236,7 @@ define(function (require, exports, module) {
                 // Function to get properly named class for the svg icon
                 "getSVGClass": payload.getSVGClass,
                 // String to display if datalist does not have any options to present
-                "noOptionsString": payload.noOptionsString
+                "noOptionsDefault": payload.noOptionsDefault
             };
         },
 
@@ -271,38 +272,46 @@ define(function (require, exports, module) {
          */
         _getOptions: function (type) {
             var searchTypeInfo = this._registeredSearchTypes[type];
+            
             if (searchTypeInfo) {
                 var items = searchTypeInfo.getOptions();
-                
-                // Get shortest unique paths
-                var ancestors = collection.pluck(items, "pathInfo"),
-                    shortenedPaths = searchTypeInfo.shortenPaths ?
-                        pathUtil.getShortestUniquePaths(ancestors).toJS() : ancestors.toJS();
+                var itemMap;
 
-                var itemMap = items.map(function (item, index) {
-                    var newPathInfo = shortenedPaths[index] || "",
-                        name = item.name,
-                        style;
-                    // Don't show the path info if it is just the same as the item name 
-                    if (name === newPathInfo) {
-                        newPathInfo = "";
-                    }
+                if (items === null) {
+                    itemMap = [];
+                } else {
+                    // Get shortest unique paths
+                    var ancestors = collection.pluck(items, "pathInfo"),
+                        shortenedPaths = searchTypeInfo.shortenPaths ?
+                            pathUtil.getShortestUniquePaths(ancestors) : ancestors;
 
-                    if (name === "") {
-                        name = "Untitled";
-                        style = { "fontStyle": "italic" };
-                    }
+                    itemMap = items.map(function (item, index) {
+                        var newPathInfo = shortenedPaths.get(index) || "",
+                            name = item.name,
+                            untitledString = nls.localize("strings.SEARCH.UNTITLED"),
+                            style;
+                        
+                        // Don't show the path info if it is just the same as the item name 
+                        if (name === newPathInfo) {
+                            newPathInfo = "";
+                        }
 
-                    return {
-                        id: type + "-" + item.id.toString(),
-                        title: name,
-                        pathInfo: newPathInfo,
-                        svgType: item.iconID,
-                        category: item.category,
-                        style: style,
-                        type: "item"
-                    };
-                }.bind(this));
+                        if (name === "") {
+                            name = untitledString;
+                            style = { "fontStyle": "italic" };
+                        }
+
+                        return {
+                            id: type + "-" + item.id.toString(),
+                            title: name,
+                            pathInfo: newPathInfo,
+                            svgType: item.iconID,
+                            category: item.category,
+                            style: style,
+                            type: "item"
+                        };
+                    }, this);
+                }
 
                 // Label to separate groups of options
                 var headerLabel = {
--- a/src/nls/en/strings.json
+++ b/src/nls/en/strings.json
@@ -468,7 +468,9 @@
         "PLACEHOLDER": "Search, open & select",
         "PLACEHOLDER_INITIALIZING": "Initializing…",
         "PLACEHOLDER_FILTER": "Search ",
+        "NO_OPTIONS_STOCK": "Search Adobe Stock…",
         "NO_OPTIONS": "No results match your search",
+        "UNTITLED": "Untitled",
         "HEADERS": {
             "ALL_LAYER": "Layers",
             "CURRENT_LAYER": "Layers",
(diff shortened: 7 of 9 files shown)


iwehrman:
Let's wait to merge this until the "no results match" bug is fixed. I don't think we would want to ship this feature (even just in a sprint build) until that's resolved.

DivyaPrabhakar:
This PR has now been updated to reflect the bug fix, as per @iwehrman's comments. Ready for review.

src/js/actions/search.js
```diff
@@ -59,11 +61,30 @@ define(function (require, exports) {
         transfers: [dialog.openDialog, dialog.closeDialog]
     };
 
+    /**
+     * Executes Adobe Stock search by opening a URL with the search term
+     *
+     * @return {Promise}
+     */
+    var executeStockSearch = function (searchTerm) {
```
iwehrman: It seems like this should go in the `adobestock` module?

src/js/actions/search.js
```diff
@@ -73,6 +94,22 @@ define(function (require, exports) {
         transfers: []
     };
 
+    /**
+     * Send info about Adobe Stock to search store
+     *
+     * @return {Promise}
+     */
+    var afterStartup = function () {
+        return this.transfer("searchStock.registerAdobeStock");
```
iwehrman: And this too?

src/js/actions/search/adobestock.js
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
iwehrman: The rule is: whenever you make a new file, make sure the copyright notice uses the current year. So, this should be 2016 instead of 2014.

Aside: for brevity, I probably would have just named this `stock.js`, but if you really like `adobestock.js` then OK!

src/js/actions/search/adobestock.js
```diff
@@ -0,0 +1,86 @@
...
+        return "libraries-stock";
+    };
+
+    /**
+     * Make list for Stock options, currently empty
+     * 
+     * @private
+     * @return {Immutable.List.<object>}
+    */
+    var _adobeStockSearchOptions = function () {
```
iwehrman: I don't understand why this is a method. `Immutable.List()` always returns the same value!

iwehrman: Oh, I see what's happening now. But, I suggest just using an anonymous function literal in `registerAdobeStock` instead of a named, documented function as above. Also, it doesn't need to be bound with `.bind` because the body doesn't refer to `this`. (The only thing `.bind` does is set the value of `this`.)

src/js/jsx/search/SearchBar.jsx
```diff
@@ -45,7 +45,8 @@ define(function (require, exports, module) {
      * @const
      * @type {string}
      */
-    var PLACEHOLDER_ID = "NO_OPTIONS-placeholder";
+    var PLACEHOLDER_ID = "NO_OPTIONS-placeholder",
+        ADOBE_STOCK_FILTER = "ADOBE_STOCK";
```
iwehrman: This is OK I guess, but it's kind of a bummer that you are directly tying one particular search provider into the search bar. It would be nicer if different search providers could optionally register "default" handlers. The default handler, in this case, would execute the stock search. It would also take the place of the "no results found" message. If I haven't explained that well, let's talk more!

src/js/jsx/search/SearchBar.jsx
```diff
@@ -399,6 +400,13 @@ define(function (require, exports, module) {
         /** @ignore */
         _handleDialogClick: function (event) {
             this.refs.datalist.removeAutofillSuggestion();
+            // If string, it means no items match user inputted search query
+            var filters = this.state.filter,
+                value = event.target.value;
+
+            if (typeof value === "string" && filters.indexOf(ADOBE_STOCK_FILTER) !== -1) {
```
iwehrman: See above: it's lame that this click handler has stock-specific logic.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -416,10 +424,17 @@ define(function (require, exports, module) {
...
             var filters = this.state.filter,
                 filtersString,
-                idSplit = id.split("-"),
-                // Possible current types are Menu_Command, Current_Doc, Recent_Doc, All_Layers
-                type = idSplit[0],
-                category = "category-" + type;
+                idSplit,
+                type;
+
+            if (id) {
```
iwehrman: This is weird. Above, we check for `id === null`. I'm not even sure what type `id` is supposed to have here because it's undocumented but I guess it can be some other falsey value like `undefined`? In any case, this smells funny.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -428,6 +443,7 @@ define(function (require, exports, module) {
                 filtersString = "filter-active";
             }
             
+
```
iwehrman: Nope!

src/js/jsx/search/SearchBar.jsx
```diff
@@ -463,7 +479,7 @@ define(function (require, exports, module) {
             switch (event.key) {
                 case "Return":
                 case "Enter":
-                    if (selectedID === PLACEHOLDER_ID) {
+                    if (selectedID === PLACEHOLDER_ID || !selectedID) {
```
iwehrman: If this disjunct is necessary, you should explain why in a comment. Also, these parameters are undocumented. JSDoc that explains what it means if `selectedID` is falsey would be a good idea.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -515,7 +531,7 @@ define(function (require, exports, module) {
                 var lastFilter = filter[filter.length - 1],
                     categoryString = searchStrings.CATEGORIES[lastFilter],
                     category = categoryString ?
-                        categoryString.toLowerCase() : this.state.safeFilterNameMap[lastFilter];
+                        categoryString : this.state.safeFilterNameMap[lastFilter];
```
iwehrman: Was it not important before that `categoryString` be converted to lower case?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -543,6 +559,7 @@ define(function (require, exports, module) {
                         className="dialog-search-bar"
                         options={this.state.options}
                         startFocused={true}
+                        filters={filter}
```
iwehrman: It's weird that the singular word `filter` is being used to indicate a list of filters here?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -36,6 +36,8 @@ define(function (require, exports, module) {
         SVGIcon = require("js/jsx/shared/SVGIcon"),
         collection = require("js/util/collection");
 
+    var ADOBE_STOCK_FILTER = "ADOBE_STOCK";
```
iwehrman: Needs JSDoc.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -314,6 +316,8 @@ define(function (require, exports, module) {
                 case "Tab":
                     return;
                 case "Enter":
+                    this.props.onKeyDown(event);
```
iwehrman: Uhh, why is `Enter` being handled differently from `Return` here?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -702,7 +706,11 @@ define(function (require, exports, module) {
                 );
             }
 
-            var dialog = searchableOptions && (
+            var filters = this.props.filters;
+            var dialog;
+            
+            if (filters && filters.indexOf(ADOBE_STOCK_FILTER) === -1) {
```
iwehrman: It's definitely not acceptable for `Datalist` to have any knowledge of Adobe Stock. `Datalist` is a basic, shared component. This would be like Chrome adding Google-specific functionality to `<input type="text"/>`.

src/js/jsx/shared/Datalist.jsx
iwehrman: Nevermind, I just realized that this is in `Datalist`. Needs removal.

src/js/stores/search.js
```diff
@@ -271,36 +281,41 @@ define(function (require, exports, module) {
...
+                        var newPathInfo = shortenedPaths[index] || "",
+                            name = item.name,
+                            style;
+                        // Don't show the path info if it is just the same as the item name 
+                        if (name === newPathInfo) {
+                            newPathInfo = "";
+                        }
+
+                        if (name === "") {
+                            name = "Untitled";
```
iwehrman: I realize this isn't your code, but it sure seems weird that this string isn't localized. You might just want to file a separate bug for this?

src/js/stores/search.js
```diff
@@ -271,36 +281,41 @@ define(function (require, exports, module) {
...
+                    // Get shortest unique paths
+                    var ancestors = collection.pluck(items, "pathInfo"),
+                        shortenedPaths = searchTypeInfo.shortenPaths ?
+                            pathUtil.getShortestUniquePaths(ancestors).toJS() : ancestors.toJS();
+
+                    itemMap = items.map(function (item, index) {
+                        var newPathInfo = shortenedPaths[index] || "",
+                            name = item.name,
+                            style;
+                        // Don't show the path info if it is just the same as the item name
```
iwehrman: A newline should separate this declaration block from following statements.

src/js/actions/search.js
```diff
@@ -59,11 +61,30 @@ define(function (require, exports) {
         transfers: [dialog.openDialog, dialog.closeDialog]
     };
 
+    /**
+     * Executes Adobe Stock search by opening a URL with the search term
+     *
+     * @return {Promise}
+     */
+    var executeStockSearch = function (searchTerm) {
+        var searchURL = STOCK_URL + searchTerm;
```
iwehrman: You may need to URL-encode the search term: http://stackoverflow.com/a/3608791

iwehrman:
Back to @DivyaPrabhakar. Good start! I think it needs to be architected a little differently though so that the shared components don't have dependencies on Adobe Stock-specific code. I'm happy to chat about this offline!

src/js/actions/search.js
DivyaPrabhakar: I was also debating that. The reason I included it in `actions/search.js` is because for every other search module, their `execute` option was in an actions module pertaining to the executed task. For example, the `executeMenuCommand()` function is held within `actions/menu.js`. Since searching Adobe Stock didn't have a corresponding actions module, I included it in the `search` module. I know it was a bit of a stretch but it follows suit for the other search modules. Having said that, with the second opinion, I think moving it to `adobesstock.js` makes sense and I'll make the change.

src/js/actions/search.js
DivyaPrabhakar: Done.

src/js/actions/search/adobestock.js
DivyaPrabhakar: The reasons I kept this as a function were: 

1.) In the case that we expand on the functionality of Adobe Stock search in the future, we will have options that will populate through this function. I saw this as an intermediate state which returns an empty list now, that is not unique now, but later on we will end up changing and wanting this function back.

2.) For every registered search payload, payload.getOptions is a function and previously when I had the `getOptions` property be an empty list instead of a function call, it would run into errors at different points in the program when `getOptions()` was called instead of `getOptions`. We can go through and have it check for function or value, etc. but a.) only the stock module requires this and b.) like I said above, I anticipate changing the return value of the function so we would change any calls to `getOptions` in the future back to the original `getOptions()`

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: I think I understand but I want to clarify in person with you before I go ahead with this.

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: This conversion to lower case was inconsistent with how the strings in the search bar were being presented. 

In the dropdown, the options were capitalized.


Upon selection, the placeholder text would become lower case. 


I talked to Seth about this and he said removing the lower case upon selection makes sense.

src/js/jsx/shared/Datalist.jsx
DivyaPrabhakar: I am not sure, I didn't even notice that. It also happens in `SearchBar.jsx` _handleKeyDown(). Removing the `Return` case from both.

src/js/jsx/search/SearchBar.jsx
iwehrman: :white_check_mark:

src/js/jsx/shared/Datalist.jsx
```diff
@@ -314,7 +316,8 @@ define(function (require, exports, module) {
                 case "Tab":
                     return;
                 case "Enter":
-                case "Return":
```
iwehrman: My guess is that you'll want to remove this code once you get the `defaultOptions` working because this code path is only active when the `Select` component is _not_ being shown.

DivyaPrabhakar:
@iwehrman Back to you

iwehrman:
I opened the search UI and typed "s t o <enter>":
img

and this happened:
image

DivyaPrabhakar:
@iwehrman This error is actually happening on `master`. What is happening is that our string matching is currently only returning a successful result if the word entered matches the first word of the id of an option. For example: if I type in "sto", the list of results that are returned are "Adobe Stock". Then we go on to check if "sto" if positioned at the first index of the string "Adobe Stock". (see `datalist.jsx`, lines 556-568). Because it does not match, the autocomplete returns as `undefined` and then nothing gets stored as the state of the datalist `state` properites like `lastHighlightedID`. Later in the code, we use the value of `lastHighlightedID` to make decisions about executing datalist options on click. Because it is `undefined`,  we get the error that you see later when we are trying to split `id` that is `undefined`. 

If you type in "Adobe" and then select "Adobe Stock", then you will not run into any problem. We need to change our string matching -- should we do that in this PR?

iwehrman:
I get this warning in the console when I execute a search:
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you  called setState() on an unmounted component. This is a no-op. Please check the code for the TextInput component.

warning	@	warning.js:45
getInternalInstanceReadyForUpdate	@	ReactUpdateQueue.js:34
ReactUpdateQueue.enqueueSetState	@	ReactUpdateQueue.js:181
ReactComponent.setState	@	ReactComponent.js:65
(anonymous function)	@	TextInput.jsx:345
tryCatcher	@	bluebird.js:4978
Promise._settlePromiseFromHandler	@	bluebird.js:3042
Promise._settlePromise	@	bluebird.js:3099
Promise._settlePromise0	@	bluebird.js:3144
Promise._settlePromises	@	bluebird.js:3227
Promise._fulfill	@	bluebird.js:3168
Promise._settlePromise	@	bluebird.js:3112
Promise._settlePromise0	@	bluebird.js:3144
Promise._settlePromises	@	bluebird.js:3227
Promise._fulfill	@	bluebird.js:3168
(anonymous function)	@	bluebird.js:2482
```

iwehrman:
If I search for "100%" I get the following error:
```
Unhandled rejection Error: OpenURLInDefaltBrowser failed
    at eval (eval at undefined, <anonymous>:1:1)
    at eval (native)
From previous event:
    at Object.openURLInDefaultBrowser (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:415:23)
    at Object._confirmStockSearch (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:97398:25)
    at Fluxxor.createStore.handleExecute (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:74225:52)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:75579:34)
    at tryCatcher (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:5091:24)
    at Promise._settlePromiseFromHandler (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:3155:32)
    at Promise._settlePromise (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:3212:19)
    at Promise._settlePromise0 (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:3257:11)
    at Promise._settlePromises (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:3340:19)
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:275:26
    at MutationObserver.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/externalDeps.js:4389:14)
```

iwehrman:
The above error probably has to do with URI encoding. I left a comment about this earlier.

iwehrman:
The message shown by the default option is currently too engineery. We don't need to explicitly instruct people to press enter. I think it should just say "Search Adobe Stock for "<your current search query>". The `Datalist` component should be able to provide the current search query to the `SearchBar`. Also, it should be disabled until you actually type something in; pressing enter in that case should have no effect.  Alternatively, the default option message could be different depending on whether the user has typed anything yet or not.

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Executes Adobe Stock search by opening a URL with the search term
+     *
+     * @return {Promise}
+     */
+    var _confirmStockSearch = function (searchTerm) {
+        var searchURL = STOCK_URL + searchTerm;
+        return adapter.openURLInDefaultBrowser(searchURL);
+    };
+    _confirmStockSearch.action = {
```
iwehrman: It doesn't look like this is an action; it's just a plain function. So, this property isn't needed.

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
...
+
+    var STOCK_URL = search.SEARCH_STOCK_URL;
+    
+    /**
+     * Executes Adobe Stock search by opening a URL with the search term
+     *
+     * @return {Promise}
+     */
+    var _confirmStockSearch = function (searchTerm) {
+        var searchURL = STOCK_URL + searchTerm;
```
iwehrman: This is what should be URI-encoded. See previous comment.

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
...
+     * Creates an object with properties to populate the search bar 
+     * datalist when there are no options present. 
+     * 
+     * @private
+     * @return {object}
+    */
+    var _createNoOptionsDefault = function () {
+        return {
+            "noOptionsString": nls.localize("strings.SEARCH.NO_OPTIONS_STOCK"),
+            "noOptionsExecute": null
```
iwehrman: I'm confused. I thought the only use case for this execution function was for Adobe Stock? But, it's null here. What's the deal?

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Register Adobe Stock info for search
+     */
+    var registerAdobeStock = function () {
+        var adobeStockPayload = {
+            "type": "ADOBE_STOCK",
+            "getOptions": _adobeStockSearchOptions.bind(this),
+            "filters": Immutable.List.of("ADOBE_STOCK"),
+            "handleExecute": _confirmStockSearch.bind(this),
```
iwehrman: It looks like you're just using the standard execution function to handle default options. If that works then that's fine. Maybe there doesn't need to be a special execution function just for default options?

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
...
+        };
+    };
+
+    /**
+     * Register Adobe Stock info for search
+     */
+    var registerAdobeStock = function () {
+        var adobeStockPayload = {
+            "type": "ADOBE_STOCK",
+            "getOptions": _adobeStockSearchOptions.bind(this),
```
iwehrman: Again, this doesn't need to be bound because that function doesn't refer to `this`.

src/js/actions/search/stock.js
iwehrman: This function also doesn't refer to `this`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -400,6 +400,19 @@ define(function (require, exports, module) {
         /** @ignore */
         _handleDialogClick: function (event) {
             this.refs.datalist.removeAutofillSuggestion();
+            
+            // If string, it means no items match user inputted search query
+            var filters = this.state.filter,
```
iwehrman: I'm still confused about why this is called `filters` (plural).

src/js/jsx/search/SearchBar.jsx
```diff
@@ -400,6 +400,19 @@ define(function (require, exports, module) {
...
+                lastFilter = filters[filters.length - 1];
+                
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                payload = searchStore.getPayloadFromFilter(lastFilter),
+                noOptionsStock = nls.localize("strings.SEARCH.NO_OPTIONS_STOCK");
+
+            if (payload && payload.noOptions().noOptionsString === noOptionsStock) {
```
iwehrman: You've created an explicit dependency on Stock. There should be no explicit dependency on any provider. Consider that if this ever gets moved to standard PS we will have to shuffle the set of providers somewhat, and we can't predict how exactly. So, there should be no explicit dependencies in any generic infrastructure on specific implementations.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -407,20 +420,24 @@ define(function (require, exports, module) {
          * Perform action based on ID
          *
          * @private
-         * @type {Datalist~onInputChange}
```
iwehrman: I think @shaoshing wants that annotation there to indicate that this function is used as the `onInputChange` handler to the child `Datalist` instance.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -407,20 +420,24 @@ define(function (require, exports, module) {
...
+         * @param {string} id of the selected option
          */
         _handleChange: function (id) {
+            var filters = this.state.filter,
+                filtersString,
+                idSplit,
+                type;
+
             if (id === null) {
+                type = "";
```
iwehrman: If you're returning right after this, why bother initializing this variable?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -407,20 +420,24 @@ define(function (require, exports, module) {
...
                 this.props.dismissDialog();
                 return;
+            } else {
```
iwehrman: We prefer the following pattern:
```
if (badState) {
  cleanup();
  return;
}

// Normal case
...
```

Instead of:
```
if (badCase) {
  cleanup();
  return;
} else {
 // do stuff
}
```
We prefer the former because subjectively we find it easier to read by dismissing the exceptional cases first, and then proceeding with the main body at a lower indentation level.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -407,20 +420,24 @@ define(function (require, exports, module) {
...
+                idSplit = id.split("-");
```
iwehrman: For future reference, I really hate how we have two separate data items encoded in a single concatenated string. This is a terrible programming idiom. This doesn't need to be fixed now, but it should be fixed...

src/js/jsx/search/SearchBar.jsx
```diff
@@ -458,13 +475,16 @@ define(function (require, exports, module) {
 
         /**
          * @private
-         * @type {Datalist~onKeyDown}
```
iwehrman: @shaoshing wants this too, and so on for the existing `@type` annotations.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -407,20 +420,24 @@ define(function (require, exports, module) {
          * Perform action based on ID
          *
          * @private
-         * @type {Datalist~onInputChange}
+         * @param {string} id of the selected option
          */
         _handleChange: function (id) {
+            var filters = this.state.filter,
```
iwehrman: Also, all this is superfluous in case `id === null`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -458,13 +475,16 @@ define(function (require, exports, module) {
...
-         * @type {Datalist~onKeyDown}
+         * @param {SyntheticEvent} event
+         * @param {string} selectedID the id of the selected item in the datalist
          */
         _handleKeyDown: function (event, selectedID) {
             switch (event.key) {
-                case "Return":
                 case "Enter":
-                    if (selectedID === PLACEHOLDER_ID) {
+                    // The only time selectedID is falsey is when an Adobe Stock search
```
iwehrman: Presumably you are building infrastructure that is not specific to Adobe Stock, so this comment should not be specific to Adobe Stock.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -458,13 +475,16 @@ define(function (require, exports, module) {
 
         /**
          * @private
-         * @type {Datalist~onKeyDown}
+         * @param {SyntheticEvent} event
+         * @param {string} selectedID the id of the selected item in the datalist
          */
         _handleKeyDown: function (event, selectedID) {
             switch (event.key) {
-                case "Return":
```
iwehrman: Why remove this case?

src/js/jsx/shared/Datalist.jsx
iwehrman: I still think you want to remove this code for the reason above.

iwehrman: Also, the stock (or whatever default) result should work just like every other result. That means I should be able to click on it. But, I can't click on the stock result.

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: This is called `filters` because there can be more than one filter added at a time, i.e. "Layers" > "Text Layers"

iwehrman: OK, but then why is the state called `filter`?

iwehrman: The point is that the mismatch is confusing.

chadrolfs:
One issue I noticed on this branch that I couldn't repro on master was that sometimes the first character gets lost when typing in search after the filter is applied.
cmd + f
arrow down to Adobe Stock, hit enter to set the filter
type Ki, hit delete twice to delete Ki, then type it again.  I end up with i

DivyaPrabhakar:
@chadrolfs Now this shouldn't be happening, can you verify?

iwehrman:
Clicking the "Search Adobe Stock" result (once the filter is applied) doesn't work. For every other search result, this works.

iwehrman:
The search result (but NOT the input placeholder text) should have an ellipsis at the end to give a hint that you're about to open up a new window:
image

iwehrman:
I'm seeing this React warning whenever I perform a Stock search:
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the TextInput component.
```

DivyaPrabhakar:
@iwehrman Looking into the error now, the stock search filter with no user input should be fixed now. Opens to the homepage
@chadrolfs The error came back for me just FYI.

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,127 @@
...
+
+    var Immutable = require("immutable");
+
+    var events = require("js/events"),
+        locks = require("js/locks"),
+        adapter = require("adapter"),
+        nls = require("js/util/nls");
+
+    var SEARCH_STOCK_URL = "https://stock.adobe.com/search?k=",
+        ADOBE_STOCK_URL = "https://stock.adobe.com/at/";
```
iwehrman: This seems to be the Austrian home page for stock. I suggest just sending them to stock.adobe.com. But, this does raise the question of whether that's safe for all locales...

DivyaPrabhakar: Lol. I'll change that. For some reason Adobe Stock has defaulted to German and I cannot change any of the settings and when I do it reverts. I have cleared my cache, changed settings in Creative Cloud, etc, nothing works. So I didn't realize that was the not the default.

DivyaPrabhakar:
@iwehrman The stock URL to the universal homepage has been changed.

DivyaPrabhakar:
@iwehrman Finally back to you -- the layer selection problem I had a brief freak out over is fixed, the warnings are not occurring any more, and the no options states should all be working too.

iwehrman:
I'm still seeing this warning when I choose the Stock search option:
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you 
called setState() on an unmounted component. This is a no-op. Please check the code for the TextInput 
component.

warning @ warning.js:45
getInternalInstanceReadyForUpdate @ ReactUpdateQueue.js:34
ReactUpdateQueue.enqueueSetState @ ReactUpdateQueue.js:181
ReactComponent.setState @ ReactComponent.js:65
(anonymous function) @ TextInput.jsx:345
tryCatcher @ bluebird.js:4978
Promise._settlePromiseFromHandler @ bluebird.js:3042
Promise._settlePromise @ bluebird.js:3099
...
```

I also get this new error when I press Escape to dismiss the empty search UI without typing anything else in:
```
Uncaught TypeError: Cannot read property 'close' of undefined

React.createClass._handleInputKeyDown @ Datalist.jsx:334
React.createClass._handleKeyDown @ TextInput.jsx:313
ReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:71
executeDispatch @ EventPluginUtils.js:79
executeDispatchesInOrder @ EventPluginUtils.js:102
executeDispatchesAndRelease @ EventPluginHub.js:43
executeDispatchesAndReleaseTopLevel @ EventPluginHub.js:54
forEachAccumulated @ forEachAccumulated.js:23
EventPluginHub.processEventQueue @ EventPluginHub.js:259
runEventQueueInBatch @ ReactEventEmitterMixin.js:18
ReactEventEmitterMixin.handleTopLevel @ ReactEventEmitterMixin.js:34
handleTopLevelWithoutPath @ ReactEventListener.js:93
handleTopLevelImpl @ ReactEventListener.js:73
Mixin.perform @ Transaction.js:136
ReactDefaultBatchingStrategy.batchedUpdates @ ReactDefaultBatchingStrategy.js:62
batchedUpdates @ ReactUpdates.js:94
ReactEventListener.dispatchEvent @ ReactEventListener.js:204
```

iwehrman:
I suspect the latter error will be easy to track down, but the former may be more difficult. I'm happy to take a look with you if you'd like!

src/js/stores/search.js
```diff
@@ -271,38 +272,46 @@ define(function (require, exports, module) {
...
-                }.bind(this));
+                var itemMap;
+
+                if (items === null) {
+                    itemMap = [];
+                } else {
+                    // Get shortest unique paths
+                    var ancestors = collection.pluck(items, "pathInfo"),
+                        shortenedPaths = searchTypeInfo.shortenPaths ?
+                            pathUtil.getShortestUniquePaths(ancestors).toJS() : ancestors.toJS();
```
iwehrman: There's no need to call `toJS` here, which is slow. Instead of doing an array lookup below, you can not convert these immutable objects and call `.get(index)`. (I know you didn't create this bug, but you're in there now...)

src/js/jsx/search/SearchBar.jsx
```diff
@@ -458,20 +482,27 @@ define(function (require, exports, module) {
...
+         * @param {SyntheticEvent} event
+         * @param {string} selectedID the id of the selected item in the datalist
          */
         _handleKeyDown: function (event, selectedID) {
             switch (event.key) {
                 case "Return":
                 case "Enter":
-                    if (selectedID === PLACEHOLDER_ID) {
-                        this._handleDialogClick(event);
+                    // The only time selectedID is falsey is when an Adobe Stock search
```
iwehrman: This comment is surely out of date since you're not hard-coding anything about Adobe Stock in this component, right?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -429,12 +443,16 @@ define(function (require, exports, module) {
...
+            if (filters && filters.size === 0) {
                 filtersString = "filter-inactive";
             } else {
                 filtersString = "filter-active";
             }
-            
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                payload = searchStore.getPayloadFromFilter(lastFilter);
```
iwehrman: I guess this doesn't need to be changed here, but `payload` is not a very good word to use here because it's not descriptive at all. If I'm not familiar with this code, I'm scratching my head thinking "what the heck is a payload?". Ideally there would be a more descriptive term for whatever this function returns that makes it possible for me to keep reading without having to pause and go read that function to see roughly what the heck it does..

src/js/jsx/search/SearchBar.jsx
```diff
@@ -414,14 +426,16 @@ define(function (require, exports, module) {
...
          */
-        _handleChange: function (id) {
-            if (id === null) {
+        _handleChange: function (id, searchTerm) {
+            if (!id) {
                 this.props.dismissDialog();
                 return;
             }
 
             var filters = this.state.filter,
```
iwehrman: I think this was supposed to be renamed to `this.state.filters`?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -171,8 +170,8 @@ define(function (require, exports, module) {
                 this.refs.datalist.focus();
             }
 
-            if (prevState.filter !== this.state.filter && this.refs.datalist) {
-                this._updateDatalistInput(this.state.filter);
+            if (prevState.filter !== this.state.filters && this.refs.datalist) {
```
iwehrman: This should also be `.filters`.

src/js/actions/search/stock.js
```diff
@@ -0,0 +1,127 @@
...
+        return {
+            "noOptionsString": nls.localize("strings.SEARCH.NO_OPTIONS_STOCK"),
+            "noOptionsExecute": _confirmStockSearch,
+            "noOptionsType": "filter"
+        };
+    };
+
+    /**
+     * Register Adobe Stock info for search
+     */
```
iwehrman: Missing `@return {Promise}`.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -445,6 +463,12 @@ define(function (require, exports, module) {
                     this.props.executeOption(id);
                     headlights.logEvent("search", filtersString, _.kebabCase(category));
                 }
+            } else {
+                if (payload && payload.noOptionsDefault) {
+                    this.props.executeOption(searchTerm, payload.noOptionsDefault().noOptionsExecute);
+                } else {
+                    return { dontCloseDialog: true };
```
iwehrman: This function is also missing `@return {{dontCloseDialog: boolean}=}`.

iwehrman:
I'm still seeing the same `setState`-after-component-unmounted warning when I click the `Search Adobe Stock...` search result. It's still in the same place in `_beginEdit.`

iwehrman:
Back to @DivyaPrabhakar. Getting close!

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: I am assuming you mean a `@return {{dontCloseDialog: boolean}=}` at line 418 which has now been added but this comment won't be outdated because the change was not made to line 470.

DivyaPrabhakar:
@iwehrman and @ktaki: This PR has the fixes for the issues #3529 and #3524

src/js/jsx/search/SearchBar.jsx
```diff
@@ -415,14 +426,17 @@ define(function (require, exports, module) {
          *
          * @private
          * @type {Datalist~onInputChange}
+         * @param {string} id of the selected option
```
shaoshing: Missing the `searchTerm` JSDoc

src/js/jsx/search/SearchBar.jsx
```diff
@@ -230,9 +230,20 @@ define(function (require, exports, module) {
                         return toRemove ? toRemove.replace(" ", "") : idWord;
                     }, this).join("").toLowerCase(),
                     nextFilterMap = _.map(currFilter, function (word) {
-                        return idString.includes(word.toLowerCase()) ? "" : word;
+                        if (idString.includes(word.toLowerCase())) {
+                            partialMatch = true;
+                            // return "";
```
shaoshing: Should this line be removed or uncommented?

ktaki:
@DivyaPrabhakar: Somehow this can be reproduced with Windows platform only. Open the search bar. Type any single character. The error below happens.

searchbarerror

DivyaPrabhakar:
The current state of this branch is that it addresses all the bugs referenced in this PR as well as all the architecture concerns. @iwehrman -- sending to you to look over. 

Summary of the current situation: 
Search on `master` is currently broken -- keyup and keydown skips options in the datalist, pressing enter on a datalist option is broken. This is because of the change in event handling -- for every key press, on Windows we were getting `onkeydown` and `oninput` events; on Mac we were only getting `oninput` events. Therefore, we were seeing errors thrown on Mac that we were not seeing on Windows. That new event that is getting registered is causing the problems on datalist. If those get solved, the changes in this PR should ideally integrate perfectly within DS; only time will tell...

ktaki:
In the search dialog, select Adobe Stock. Enter a word to search. Hit return.

screen shot 2016-02-17 at 11 06 51 am

src/js/jsx/Search.jsx
```diff
@@ -53,14 +53,19 @@ define(function (require, exports, module) {
          * 
          * @private
          * @param {string} itemID ID of selected option
+         * @param {string=} noOptionsFn function to execute if there are no options
```
iwehrman: `noOptionsFn` does not look like a string to me!

src/js/jsx/search/SearchBar.jsx
```diff
@@ -216,7 +215,8 @@ define(function (require, exports, module) {
          */
         _updateDatalistInput: function (id) {
             if (id) {
-                var currFilter = this.refs.datalist.getInputValue().split(" "),
+                var partialMatch,
```
iwehrman: Style nit: in var declaration blocks, put the declarations with definitions first.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -230,9 +230,19 @@ define(function (require, exports, module) {
                         return toRemove ? toRemove.replace(" ", "") : idWord;
                     }, this).join("").toLowerCase(),
                     nextFilterMap = _.map(currFilter, function (word) {
-                        return idString.includes(word.toLowerCase()) ? "" : word;
+                        if (idString.includes(word.toLowerCase())) {
+                            partialMatch = true;
```
iwehrman: Map functions should always return a value. It's weird that this function only returns a value sometimes. If this is intentional, you should explicitly return `null` instead of having no return.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -216,7 +215,8 @@ define(function (require, exports, module) {
          */
         _updateDatalistInput: function (id) {
```
iwehrman: It's super, super confusing that this variable is called `id` when it is actually an array of `id`s.

src/js/jsx/search/SearchBar.jsx
iwehrman: (Or the empty string, or whatever the right explicit value should be.)

src/nls/en/strings.json
```diff
@@ -468,7 +468,9 @@
         "PLACEHOLDER": "Search, open & select",
         "PLACEHOLDER_INITIALIZING": "Initializing…",
         "PLACEHOLDER_FILTER": "Search ",
+        "NO_OPTIONS_STOCK": "Search Adobe Stock...",
```
iwehrman: This should be an ellipsis, not three periods.

src/js/stores/search.js
```diff
@@ -271,38 +272,46 @@ define(function (require, exports, module) {
...
+                } else {
+                    // Get shortest unique paths
+                    var ancestors = collection.pluck(items, "pathInfo"),
+                        shortenedPaths = searchTypeInfo.shortenPaths ?
+                                pathUtil.getShortestUniquePaths(ancestors) : ancestors;
+
+                    itemMap = items.map(function (item, index) {
+                        var newPathInfo = shortenedPaths.get(index) || "",
+                            name = item.name,
+                            style,
```
iwehrman: Style nit: reorder declarations.

src/js/stores/search.js
```diff
@@ -271,38 +272,46 @@ define(function (require, exports, module) {
...
+                        return {
+                            id: type + "-" + item.id.toString(),
+                            title: name,
+                            pathInfo: newPathInfo,
+                            svgType: item.iconID,
+                            category: item.category,
+                            style: style,
+                            type: "item"
+                        };
+                    }.bind(this));
```
iwehrman: For collections functions that accept a receiver argument, that is preferred over an explicit bind.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -701,25 +735,26 @@ define(function (require, exports, module) {
                     </div>
                 );
             }
-
+
```
iwehrman: Bad whitespace change.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -603,7 +626,18 @@ define(function (require, exports, module) {
...
-            if (this.props.startFocused && this.refs.textInput) {
+            var currentOptions,
+                id;
+
+            if (this.refs.select) {
+                currentOptions = this.refs.select._getOptions();
+            }
+            
+            if (currentOptions && currentOptions.size === 1) {
+                id = currentOptions.toJS()[0].id;
```
iwehrman: Calling `.toJS()` is very slow. It would be better to not do so if possible. It seems like you could just do `currentOptions.first().id` instead.

iwehrman: Also it seems like this whole if block could be moved inside the previous one?

iwehrman: Or, even better:
```javascript
if (!this.refs.select) {
  return;
}
```

src/js/jsx/shared/Datalist.jsx
```diff
@@ -323,10 +323,27 @@ define(function (require, exports, module) {
                 }
             }
 
-            var cbOptions = this.props.onKeyDown(event, this.state.lastHighlightedID),
-                options = _.merge({ preventListDefault: false }, cbOptions);
+            // All the potential options for the curret user input
```
iwehrman: Typo: `curret`

src/js/jsx/shared/Datalist.jsx
```diff
@@ -323,10 +323,27 @@ define(function (require, exports, module) {
                 }
             }
 
-            var cbOptions = this.props.onKeyDown(event, this.state.lastHighlightedID),
-                options = _.merge({ preventListDefault: false }, cbOptions);
+            // All the potential options for the curret user input 
+            var optionsList = select ? select.props.options : null,
```
iwehrman: It seems like this initialization could be delayed?

src/js/jsx/shared/Datalist.jsx
```diff
@@ -340,9 +357,15 @@ define(function (require, exports, module) {
                     break;
                 case "Enter":
                 case "Return":
-                    select.close(event, "apply");
-                    if (dialog && dialog.isOpen()) {
-                        dialog.toggle(event);
+                    var noOptionsExecute = optionsList ? optionsList.toJS()[0].handleExecute : null;
```
iwehrman: Again, if possible, please avoid calling `.toJS()`.

iwehrman: I don't think that `Datallist` should be doing anything that involves `handleExecute`, which is only relevant to `Search` AFAIK. Can you help me understand what this is used for so we can find a way to move this functionality out of `Datalist`?

src/js/stores/search.js
```diff
@@ -271,38 +272,46 @@ define(function (require, exports, module) {
...
-                }.bind(this));
+                var itemMap;
+
+                if (items === null) {
+                    itemMap = [];
+                } else {
+                    // Get shortest unique paths
+                    var ancestors = collection.pluck(items, "pathInfo"),
+                        shortenedPaths = searchTypeInfo.shortenPaths ?
+                                pathUtil.getShortestUniquePaths(ancestors) : ancestors;
```
iwehrman: Looks like you jumped a tab stop here.

iwehrman:
@DivyaPrabhakar: to fix the "Can only update a mounted component" problem:

1) Save the promise from `acquireFocus()` in object state, with something like:
```javascript
this._acquireFocusPromise = this.acquireFocus()
  .bind(this)
  .then(/* whatever */)
  .finally(function () {
    this._acquireFocusPromise = null;
  });
```

2) In `componentWillUnmount`, do something like: 
```javascript
if (this._acquireFocusPromise && /* status is "not yet resolved" */) {
  this._acquireFocusPromise.cancel();
}
```

If necessary, repeat for `releaseFocus`. 

CC @volfied: is this something that we could integrate into `Focusable`?

ktaki:
@DivyaPrabhakar: Tested the latest branch.

#3610 Autocompletion by tab key is working.
#3529 Still reproducible
#3524 TypeError is gone, but `Warning: setState(...):` is still happening.

The result of the first search (Limit search to...) is not listed in the dialog. For example, typing Layers and hit return does not list layers. Is this what you were referring to as "broken"?

iwehrman:
I think it's OK if #3529 is not fixed in this PR. The `setState` problem will be fixed in short order. The "result of first search" sounds like a blocker though.

volfied:
@iwehrman We can improve Focusable mixin to make sure the focus functions are only called while component is mounted. However, we wouldn't be able to have 1) where something happens in the then handler of the owner component, and then the promise pointer is set to null.

DivyaPrabhakar:
Updated the PR w.r.t. to the comments made. @ktaki #3529 should be fixed with these changes now. @iwehrman I removed the search specific checks from datalist that you asked for clarification about and moved it to searchbar.jsx (you can see what I am talking about on in searchbar.jsx, line 505-509 on this branch)

src/js/jsx/search/SearchBar.jsx
```diff
@@ -229,10 +227,23 @@ define(function (require, exports, module) {
 
                         return toRemove ? toRemove.replace(" ", "") : idWord;
                     }, this).join("").toLowerCase(),
+                    partialMatch,
```
iwehrman: Again, declarations without a definition should be moved to the end of var declaration blocks for stylistic consistency with the rest of the code.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -416,15 +427,19 @@ define(function (require, exports, module) {
...
-         * @type {Datalist~onChange}
+         * @type {Datalist~onInputChange}
+         * @param {string} id of the selected option
+         * @param {string} searchTerm user input
+         * @return {{dontCloseDialog: boolean=}};
          */
-        _handleChange: function (id) {
-            if (id === null) {
+        _handleChange: function (id, searchTerm) {
+            if (!id) {
```
iwehrman: Given that `id` is a non-optional parameter, this will only match if `id === ""`. Is that intentional?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -416,15 +427,19 @@ define(function (require, exports, module) {
...
                 this.props.dismissDialog();
                 return;
             }
 
-            var filters = this.state.filter,
+            var filters = this.state.filters,
+                lastFilter = filters ? filters[filters.length - 1] : id,
```
iwehrman: If filters is an empty array, this will throw.

iwehrman: Er, actually, it will just be undefined, because `filters.length - 1` will be -1, and `[-1] === undefined`. Anyway, Is this what you want?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -462,20 +487,30 @@ define(function (require, exports, module) {
...
+                });
             }
         },
 
         /**
          * @private
          * @type {Datalist~onKeyDown}
+         * @param {SyntheticEvent} event
+         * @param {string} selectedID the id of the selected item in the datalist
+         * @param {list=} optionsList
```
iwehrman: This is not a valid JSDoc type expression.

src/js/jsx/shared/Datalist.jsx
```diff
@@ -586,7 +601,7 @@ define(function (require, exports, module) {
 
         /**
          * Sets autofill suggestion to empty string and re-renders
-         *
+         *
```
iwehrman: Bad whitespace change.

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: This variable is used by the subsequent variable declarations, therefore must be present here. sorry for the ugly.

src/js/jsx/search/SearchBar.jsx
```diff
@@ -416,15 +429,19 @@ define(function (require, exports, module) {
          * Perform action based on ID
          *
          * @private
-         * @type {Datalist~onChange}
+         * @type {Datalist~onInputChange}
```
iwehrman: There is no longer an `onInputChange` prop to `Datalist`.

chadrolfs:
Tested this branch and all the bugs look fixed (tested Mac and Win).  Also, everything functions as expected and explained above.  From a user/functional standpoint, I think it's ready to merge.  @ktaki wanna have a look?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -416,15 +429,19 @@ define(function (require, exports, module) {
...
+         * @param {string} id of the selected option
+         * @param {string} searchTerm user input
+         * @return {{dontCloseDialog: boolean=}};
          */
-        _handleChange: function (id) {
-            if (id === null) {
+        _handleChange: function (id, searchTerm) {
+            if (id === null || !id) {
```
iwehrman: This is redundant, since if `id === null` then `!id` would suffice. Also the param type annotation says that string is required, so that would imply that it can't be null. Do you know if `id` is actually ever null?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -432,13 +449,21 @@ define(function (require, exports, module) {
                 category = "category-" + type;
 
             // Seeing if filters are active or not for analytics
-            if (filters.length === 0) {
+            if (filters && filters.length === 0) {
```
iwehrman: Is `filters` guaranteed to be an array? If so, the first conjunct is unnecessary. If not, you probably want `!filter || filters.length === 0` instead?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -432,13 +449,21 @@ define(function (require, exports, module) {
...
                 filtersString = "filter-inactive";
             } else {
                 filtersString = "filter-active";
             }
-            
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                // This payload will be used to dictate whether given the applied filter (0 or more), if 
+                // we want to take any action in the case the user input returns in no search results
```
iwehrman: I think you want `input returns no search results.` ?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -432,13 +449,21 @@ define(function (require, exports, module) {
...
+            if (filters && filters.length === 0) {
                 filtersString = "filter-inactive";
             } else {
                 filtersString = "filter-active";
             }
-            
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                // This payload will be used to dictate whether given the applied filter (0 or more), if
```
iwehrman: `applied filters` ?

ktaki:
Works much better :+1: Agree with Chad. Ready to merge. Will pound on it more on master.

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: This can be null (see: datalist.jsx `_handleSelectClick`), so I will change the parameter to optional and only have the `!id` check.

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: I confirmed that this `id` is in fact optional (there was a mistake in the JSDoc), therefore this check makes more sense now and I have kept it in.

iwehrman:
Let's do it!
img

Issue #3482, opened 2016-01-08
Expand-all-descendants only expands children


iwehrman:
1. Open VermilionArtboards.psd
2. Option-click the first artboard once or twice to expand all descendants.
3. Switch to standard PS.

Result: only children of the first artboard are expanded, and not all descendants.

iwehrman:
Also, `initializeBackgroundLayers` -> `resetLayers` can cause folder expansion state to change while you're using the app if you expand folder with subfolders that haven't yet been initialized.

iwehrman:
I think this is my own merge error. I think that somehow I lost @shaoshing's #3381 when I split `actions.groups` out of `actions.layers`.

iwehrman:
Fixed by #3483.

PR #3483, opened 2016-01-08
Fix recursive group expansion

iwehrman:
When I factored the `groups` actions out of `layers` actions, I think I lost @shaoshing's PR #3381 which makes correct use a `recursive` flag in the expand/collapse group action descriptor. 

Addresses #3482.


iwehrman:
@shaoshing Can you take a look at this? Should be pretty easy since it's an exact copy of your previous PR :grimacing:

shaoshing:
Looks great. also verified the fix.

Issue #3486, opened 2016-01-09
React is in development mode in compiled builds


iwehrman:
Even in compiled builds, the `Download the React DevTools for a better development experience` message appears in the console, which implies that React is running in development mode, doing things like gathering runtime statistics, checking propTypes, etc., that it should only be doing in debug builds. Among other places, there is discussion of this in facebook/react#3877. This is a regression from the recent require-webpack conversion.

volfied:
Thank you @iwehrman for catching and fixing this.

PR #3487, opened 2016-01-09
Define process.env.NODE_ENV for React

iwehrman:
This defines `process.env.NODE_ENV` to be `"development"` in debug builds and `"production" in compiled builds.` This is required for React to use a more optimized production mode in our compiled builds. 

Addresses #3486.


volfied:
Good catch!

PR #3489, opened 2016-01-09
Delay headlights logging until browser is idle

iwehrman:
This PR defers posting headlights events until the browser is idle, and removes the return value from `headlights.logEvent` to clarify that clients should not wait for any results. This is just a nit fix.


volfied:
Looks good to me!

PR #3490, opened 2016-01-11
Delay React render on drags by blocking dispatch

volfied:
On `superselect.drag`, we first run the `superselect.click` action to select whatever layer the user might be clicking on / wants to move. This calls `layers.select`, which dispatches optimistically to update our models and re-render everything on React side, before drag happens. This causes a visible delay on drags until our models/views update to send the drag operation to Photoshop to start dragging. (On average, 200 ms)

With this change, I save the selected layer IDs in the tool store, and instead of dispatching a select event, we just store those IDs, and wait for the move event to reach us before grabbing these stored IDs and emitting a select event for them, updating our models. This saves us on average 150 ms, bringing down the drag action length to 50 ms before Photoshop starts dragging.

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -821,10 +821,13 @@ define(function (require, exports) {
      * @param {string=} modifier Way of modifying the selection. Possible values
      *  are defined in `adapter/src/lib/layer.js` under `select.vals`.
      *  Default is similar to "select", but the pivot layer will always be cleared.
+     * @param {boolean?} quiet If set, will prevent dispatches as result of this select
+     *  Please use this sparingly, and making sure that you dispatch the correct events
+     *  so UI is updated at some point
      *
      * @returns {Promise}
      */
-    var select = function (document, layerSpec, modifier) {
+    var select = function (document, layerSpec, modifier, quiet) {
         if (layerSpec instanceof Layer) {
             layerSpec = Immutable.List.of(layerSpec);
         }
@@ -898,12 +901,24 @@ define(function (require, exports) {
             pivotID: nextPivot && nextPivot.id
         };
 
-        var dispatchPromise = this.dispatchAsync(events.document.SELECT_LAYERS_BY_ID, payload)
+        var dispatchPromise,
+            revealPromise;
+
+        // If quiet is set, we are doing a drag event, and don't want a UI update
+        // We dispatch the drag event with the selected layer IDs, so
+        // the actual dispatch can be done after a "move" event arrives from PS
+        if (quiet) {
+            dispatchPromise = this.dispatchAsync(events.tool.SELECT_TOOL_DRAG, payload);
+            revealPromise = Promise.resolve();
+        } else {
+            this.dispatchAsync(events.document.SELECT_LAYERS_BY_ID, payload)
                 .bind(this)
                 .then(function () {
                     return this.transfer(initializeLayers, document, nextSelected);
-                }),
+                });
             revealPromise = this.transfer(revealLayers, document, nextSelected);
+        }
+        
 
         var layerRef = (modifier === "deselect" ? layerSpec : nextSelected)
             .map(function (layer) {
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -818,16 +818,17 @@ define(function (require, exports) {
      * @param {Document} document Owner document
      * @param {Layer|Immutable.Iterable.<Layer>} layerSpec Either a single layer that
      *  the selection is based on, or a list of such layers.
-     * @param {string=} modifier Way of modifying the selection. Possible values
+     * @param {object=} options Object containins optional parameters
+     * @param {string=} options.modifier Way of modifying the selection. Possible values
      *  are defined in `adapter/src/lib/layer.js` under `select.vals`.
      *  Default is similar to "select", but the pivot layer will always be cleared.
-     * @param {boolean?} quiet If set, will prevent dispatches as result of this select
+     * @param {boolean=} options.quiet If set, will prevent dispatches as result of this select
      *  Please use this sparingly, and making sure that you dispatch the correct events
      *  so UI is updated at some point
      *
      * @returns {Promise}
      */
-    var select = function (document, layerSpec, modifier, quiet) {
+    var select = function (document, layerSpec, options) {
         if (layerSpec instanceof Layer) {
             layerSpec = Immutable.List.of(layerSpec);
         }
@@ -836,7 +837,9 @@ define(function (require, exports) {
             return Promise.resolve();
         }
 
-        var setPivot = layerSpec.size === 1,
+        var modifier = options ? options.modifier : "selectNoPivot",
+            quiet = options ? options.quiet : false,
+            setPivot = layerSpec.size === 1,
             selected = document.layers.selected,
             pivot = document.layers.pivot,
             nextSelected,
@@ -890,9 +893,12 @@ define(function (require, exports) {
                 });
             }
             break;
-        default:
+        case "selectNoPivot":
             nextSelected = layerSpec;
             nextPivot = null;
+            break;
+        default:
+            throw new Error("Layers.select is passed an invalid modifier");
         }
 
         var payload = {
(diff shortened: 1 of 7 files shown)


src/js/actions/layers.js
```diff
@@ -821,10 +821,13 @@ define(function (require, exports) {
      * @param {string=} modifier Way of modifying the selection. Possible values
      *  are defined in `adapter/src/lib/layer.js` under `select.vals`.
      *  Default is similar to "select", but the pivot layer will always be cleared.
+     * @param {boolean?} quiet If set, will prevent dispatches as result of this select
```
iwehrman: Confusingly, nullable types are written `?T` not `T?` whereas optional types are written `T=`. I think you actually want `boolean=` here.

src/js/actions/layers.js
```diff
@@ -821,10 +821,13 @@ define(function (require, exports) {
...
+     *  Please use this sparingly, and making sure that you dispatch the correct events
+     *  so UI is updated at some point
      *
      * @returns {Promise}
      */
-    var select = function (document, layerSpec, modifier) {
+    var select = function (document, layerSpec, modifier, quiet) {
```
iwehrman: Would now be a good time to convert `modifier` and `quiet` into a single options param? It's a little hard to understand what it means at the call site when you see `this.transfer("layers.select", layers, true, false)`.

src/js/actions/layers.js
```diff
@@ -898,12 +901,24 @@ define(function (require, exports) {
...
+            revealPromise;
+
+        // If quiet is set, we are doing a drag event, and don't want a UI update
+        // We dispatch the drag event with the selected layer IDs, so
+        // the actual dispatch can be done after a "move" event arrives from PS
+        if (quiet) {
+            dispatchPromise = this.dispatchAsync(events.tool.SELECT_TOOL_DRAG, payload);
+            revealPromise = Promise.resolve();
+        } else {
+            this.dispatchAsync(events.document.SELECT_LAYERS_BY_ID, payload)
```
iwehrman: Shouldn't this be assigned to `dispatchPromise`?

src/js/actions/layers.js
```diff
@@ -898,12 +901,24 @@ define(function (require, exports) {
...
                 .bind(this)
                 .then(function () {
                     return this.transfer(initializeLayers, document, nextSelected);
-                }),
+                });
             revealPromise = this.transfer(revealLayers, document, nextSelected);
+        }
+
```
iwehrman: Extra newline.

src/js/actions/superselect.js
```diff
@@ -671,6 +672,8 @@ define(function (require, exports) {
                             };
 
                             return adapterOS.postEvent(dragEvent);
+                        } else {
+                            return Promise.resolve();
```
iwehrman: Inside a `then` handler, this is superfluous.

src/js/actions/transform.js
```diff
@@ -1317,16 +1317,36 @@ define(function (require, exports) {
 
             return Promise.join(modelPromise, transformPromise);
         } else {
+            var dragUpdatePromise,
+                dragLayerIDs = this.flux.store("tool").getState().draggedLayerIDs;
```
iwehrman: Would be nicer stylistically to swap these.

src/js/actions/transform.js
```diff
@@ -1345,10 +1365,10 @@ define(function (require, exports) {
         }
     };
     handleTransformLayer.action = {
-        read: [locks.JS_APP, locks.JS_DOC],
+        read: [locks.JS_APP, locks.JS_DOC, locks.JS_TOOL],
```
iwehrman: It seems like this action _writes_ to the tool store. Should it hold the write lock?

src/js/actions/transform.js
```diff
@@ -1317,16 +1317,36 @@ define(function (require, exports) {
...
+                return dragUpdatePromise
                     .bind(this)
                     .then(function () {
+                        // This doesn't cause a re-render
+                        this.dispatch(events.tool.SELECT_TOOL_DRAG, { selectedIDs: null });
+
+                        // The history event may arrive before or after the transform event. The
+                        // following ensures correct behavior w.r.t. history amendment by reset
+                        // actions for either case.
+                        return this.transfer("history.newHistoryStateRogueSafe", currentDoc.id);
```
iwehrman: Paging @mcilroyc to confirm that this history action is correct here.

src/js/actions/superselect.js
```diff
@@ -660,7 +661,7 @@ define(function (require, exports) {
                 return descriptor.playObject(toolLib.setToolOptions("moveTool", { "$Abbx": false }))
                     .bind(this)
                     .then(function () {
-                        return this.transfer(click, doc, x, y, diveIn, false);
+                        return this.transfer(click, doc, x, y, diveIn, false, true);
```
iwehrman: It seems suboptimal to transfer to an action (`click`) that starts out by performing a hit test since we already did the hit test. Maybe a subprocedure could be factored out of `click` which takes the hit layers (or more!?) as parameters, and then both `click` and `drag` could transfer to that subprocedure? Seems like we could save at least one redundant hit test that way.

iwehrman:
I definitely can see a performance improvement here, which is great. The structure is kind of disappointing, but I don't currently have a better suggestion. Back to @volfied for some fixes, and waiting on @mcilroyc for a final judgment about the handling of rogue history states.

src/js/actions/transform.js
mcilroyc: This seems fine.  By delaying this in time until after the layer selection is complete, that might increase the chance that the "rogue safe" logic is triggered ... which is to say that this increases the likelihood that the historyState event will be handled before this history action is ... and that's OK.  I can't think of a strong argument for moving this action earlier.  

The layer selection should not affect the history state.

chadrolfs:
Definitely faster than master comparably.  There does still seem to be a delay on initial drag when switching layers when cmd/ctrl key is down.

volfied:
@iwehrman I've made review changes, also cleaned up `drag` so it lets `click` handle selection by itself.

This clean up also fixes #3493.

src/js/actions/superselect.js
```diff
@@ -670,19 +665,20 @@ define(function (require, exports) {
...
+                            );
+                        });
+                } else {
+                    return this.dispatchAsync(events.panel.SUPERSELECT_MARQUEE, { x: x, y: y, enabled: true });
+                }
             });
     };
     drag.action = {
-        reads: [locks.JS_DOC, locks.JS_UI],
+        reads: [locks.JS_DOC],
```
iwehrman: Ugh, sorry, I may have lead you astray last time, but it seems like this _does_ still read the UI store. See above `uiStore.transformWindowToCanvas(x, y)`.

src/js/events.js
```diff
@@ -108,6 +108,7 @@ define(function (require, exports, module) {
         tool: {
             SELECT_TOOL_START: "selectToolStart",
             SELECT_TOOL_END: "selectToolEnd",
+            SELECT_TOOL_DRAG: "selectToolDrag",
```
iwehrman: This is kind of a confusing name given its proximity to the two aforementioned events `SELECT_TOOL_START` and `SELECT_TOOL_END`, which have a completely unrelated meaning to this new `SELECT_TOOL_DRAG` event...

iwehrman:
Two more nits. Back to @volfied.

src/js/actions/superselect.js
volfied: It no longer does, I rely on click to do the transform.

volfied:
Changed the event name to make it clear that it's for superselect tool, thanks @iwehrman ! Back to you.

src/js/actions/superselect.js
```diff
@@ -363,9 +363,10 @@ define(function (require, exports) {
      * @param {number} y Offset from the top window edge
      * @param {boolean} deep Whether to choose all layers or not
      * @param {boolean} add Whether to add/remove layer to selection
+     * @param {boolean} quiet If true, select action will not cause any dispatches
      * @return {Promise.<boolean>} True if any layers are selected after this command, used for dragging
      */
-    var click = function (doc, x, y, deep, add) {
+    var click = function (doc, x, y, deep, add, quiet) {
```
iwehrman: Future work: options-ify the last three parameters on this function, too.

iwehrman:
LGTM.

PR #3494, opened 2016-01-11
Timestamp actions and transfers

iwehrman:
This adds timestamps to the Chromium timeline corresponding to action and transfer execution boundaries. This only happens in debug mode. It results in little yellow lines in the timeline with tooltips to indicate what action is starting or finishing execution:

image

(Above, the line is next to the 3540ms time marker.)

More info: https://developer.chrome.com/devtools/docs/console-api#consoletimestamplabel


volfied:
Useful!

Issue #3436, opened 2015-12-15
Error after adding a mask to an empty layer


ktaki:
New artboard doc. An empty Layer 1 is selected. Add a vector mask to it. Exit vector mask mode. Add a rectangle.

![screen shot 2015-12-15 at 3 28 04 pm](https://cloud.githubusercontent.com/assets/2029597/11827362/7ac0237c-a340-11e5-8bc2-136bcedcb9b8.png)

It looks like Layer 1 was replaced by Rectangle 1, and Rectangle 1 created outside of Artboard 1.  This is wrong.

Subsequent manipulation of the layers will produce errors. For example, moving the Rectangle 1 layer back to inside of Artboard 1 in the Layers panel will result in the following error.

![screen shot 2015-12-15 at 3 31 55 pm](https://cloud.githubusercontent.com/assets/2029597/11827423/073ad856-a341-11e5-8302-6ee66a4bea35.png)

Note that a mask can be added to an empty layer in standard Ps and the above scenario won't produce errors. In DS, we might want to disable this.

ktaki:
Verified the fix.

PR #3496, opened 2016-01-11
make replace layer code vector mask aware

pineapplespatula:
fix for issue #3436

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -1019,7 +1019,7 @@ define(function (require, exports, module) {
         var nextStructure = selected ? this.updateSelection(Immutable.Set()) : this,
             replaceLayer;
 
-        // Default replacement logic is to replace a single, empty, non-background, selected layer
+        // Default replacement logic is to replace a single, empty, non-background, selected layer without a vector mask
         // Allow explicit opt-in or opt-out via the replace param
         if (replace !== false && layerIDs.length === 1) {
             if (Number.isInteger(replace)) {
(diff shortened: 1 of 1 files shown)


src/js/models/layerstructure.js
```diff
@@ -1035,7 +1035,7 @@ define(function (require, exports, module) {
             replace = replaceLayer &&
```
volfied: This expression hurt my brain, can you clean it up a little?

I advise: `replace = replace || (replaceLayer && !replaceLayer.... && ...)` so we can short circuit if replace is already provided.

Also, please update the comment above so it's not stale.

volfied:
One nit, @pineapplespatula

src/js/models/layerstructure.js
pineapplespatula: that changes the logic, replace should be false if there is no replaceLayer.

pineapplespatula:
back to @volfied i broke out the default replacement logic into a separate variable to make it more readable

src/js/models/layerstructure.js
```diff
@@ -1032,10 +1032,9 @@ define(function (require, exports, module) {
             }
 
             // The selected layer should be empty and a non-background layer unless replace is explicitly provided true
-            replace = replaceLayer &&
-                (replace ||
-                (!replaceLayer.isBackground && replaceLayer.isPixel &&
-                replaceLayer.bounds && !replaceLayer.bounds.area));
+            var layerShouldBeReplaced = (!replaceLayer.isBackground && replaceLayer.isPixel &&
+                    replaceLayer.bounds && !replaceLayer.bounds.area && !replaceLayer.vectorMaskEnabled);
+            replace = replaceLayer && (replace || layerShouldBeReplaced);
```
volfied: I think the `replaceLayer` needs to be part of layerShouldBeReplaced, if `replaceLayer` is null, above code will break.

volfied:
:cyclone: @pineapplespatula

pineapplespatula:
:repeat: @volfied

PR #3497, opened 2016-01-12
Expose ReactPerf and make it disabled by default.

shaoshing:
Expose `ReactPerf` to the global for easier access of the React's profiling tool (https://facebook.github.io/react/docs/perf.html). Also removed `ReactPerf.start()`, otherwise it will measure our components endlessly (which will increase rendering time and memory usage in dev mode)

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -121,7 +121,8 @@ define(function (require, exports) {
             window.__PS_ADAPTER__ = adapter;
             window.__FLUX_CONTROLLER__ = _controller;
             window.__LOG_UTIL__ = log;
-            window.__PERF_UTIL = performanceUtil;
+            window.__PERF_UTIL__ = performanceUtil;
+            window.__REACT_PERF__ = ReactPerf;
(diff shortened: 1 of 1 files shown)


iwehrman:
I am highly in favor of both changes. However, in other cases when we've added globals we've prefixed them with underscores. I'm hesitant to say "we must prefix with underscores!!" but I thought I'd check if you thought that was a good idea before hitting merge. I'm OK either way. Assign back to me whenever you're ready!

iwehrman:
CC @volfied in case he has a strong opinion.

volfied:
I'd like to see it be `window.__REACT_PERF__` instead of innocent looking `window.ReactPerf`.

Just so it's obvious we injected it.

shaoshing:
@iwehrman @volfied thanks for the inputs. I've renamed the variable to __REACT_PERF__, and moved it to a more proper location.

iwehrman:
Brilliant.

PR #3499, opened 2016-01-12
Customizable "No options" messages for registered search modules

DivyaPrabhakar:
This creates an option property on registered search modules for a custom string to be shown when no datalist options are available for that category in the search bar. If no value is passed in for the property, a default string will be applied and shown. 

Provides a better solution for some of the concerns raised in #3479. Once this is merged, then we can integrate Adobe Stock search in a much more stable way. 

Note: For some of the payloads, I have added the property `noOptionsString`. For others I have no provided the property. This is to prove to whoever tests this branch that both cases work. Also, a string has been added specific to no options when the Adobe Stock filter is applied. This is in preparation for #3479, as stated above.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -500,20 +500,27 @@ define(function (require, exports, module) {
 
         render: function () {
             var searchStrings = nls.localize("strings.SEARCH"),
-                noMatchesOption = {
-                    id: PLACEHOLDER_ID,
-                    title: nls.localize("strings.SEARCH.NO_OPTIONS"),
-                    type: "placeholder"
-                };
-
-            var filter = this.state.filter,
+                filters = this.state.filter,
                 placeholderText,
                 clearInputBtn;
 
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                lastFilter = filters[filters.length - 1];
+
+            var searchPayload = searchStore.getPayloadFromFilter(lastFilter),
+                noOptionsString = searchPayload ? searchPayload.noOptionsString : null,
+                defaultNoOptions = nls.localize("strings.SEARCH.NO_OPTIONS");
+
+            var noMatchesOption = {
+                    id: PLACEHOLDER_ID,
+                    title: noOptionsString ? noOptionsString : defaultNoOptions,
+                    type: "placeholder"
+                };
+
             // If we have applied a filter, change the placeholder text
-            if (filter.length > 0) {
-                var lastFilter = filter[filter.length - 1],
-                    categoryString = searchStrings.CATEGORIES[lastFilter],
+            if (filters.length > 0) {
+                var categoryString = searchStrings.CATEGORIES[lastFilter],
                     category = categoryString ?
                         categoryString.toLowerCase() : this.state.safeFilterNameMap[lastFilter];
                 placeholderText = searchStrings.PLACEHOLDER_FILTER + category;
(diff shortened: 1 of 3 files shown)


src/js/jsx/search/SearchBar.jsx
```diff
@@ -500,20 +500,29 @@ define(function (require, exports, module) {
...
         render: function () {
             var searchStrings = nls.localize("strings.SEARCH"),
-                noMatchesOption = {
+                filters = this.state.filter,
+                placeholderText,
+                clearInputBtn;
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                registeredSearchModules = searchStore._registeredSearchTypes;
```
volfied: Even though JS allows us to get any variables of stores, it's a bad convention to directly access them. Please implement a getter method in search store that will return you the correct module payload for a given filter and use that to get your `searchPayload`

src/js/jsx/search/SearchBar.jsx
```diff
@@ -500,20 +500,29 @@ define(function (require, exports, module) {
...
+
+            var lastFilter = filters[filters.length - 1];
+
+            var searchPayload = registeredSearchModules[lastFilter];
+            var noOptionsString = searchPayload ? searchPayload.noOptionsString : null;
+            var defaultNoOptions = nls.localize("strings.SEARCH.NO_OPTIONS");
```
volfied: In addition to comment above, instead of doing this in `render`, can you move all these calculations to `getStateFromFlux` method so the no options string to be shown is part of the state?

src/js/jsx/search/SearchBar.jsx
```diff
@@ -500,20 +500,29 @@ define(function (require, exports, module) {
...
+                filters = this.state.filter,
+                placeholderText,
+                clearInputBtn;
+
+            var flux = this.getFlux(),
+                searchStore = flux.store("search"),
+                registeredSearchModules = searchStore._registeredSearchTypes;
+
+            var lastFilter = filters[filters.length - 1];
+
```
volfied: All these vars can be in one big block:
```
var x = ...,
    y = ...,
    z = ...;

File: src/js/jsx/search/SearchBar.jsx [hunk 4]
```diff
@@ -500,20 +500,29 @@ define(function (require, exports, module) {
...
+            var searchPayload = registeredSearchModules[lastFilter];
+            var noOptionsString = searchPayload ? searchPayload.noOptionsString : null;
```
volfied: This line and line below can be consolidated into:
`var noOptionsString = searchPayload ? searchPayload.noOptionsString : nls.localize("strings.SEARCH.NO_OPTIONS");`

volfied:
While correct, I have few qualms about the code style and accessing a private store variable.

Also, for atomicity's sake, can you remove the NO_OPTIONS_STOCK string from this PR and add it to your other PR?

src/js/jsx/search/SearchBar.jsx
DivyaPrabhakar: Sometimes the payload returns null ie. in the case of Layers. There is one registered search module for layers, but then we break down the categories into Pixel Layers, Graphic Layers, etc. The filter "Pixel" is a valid one but there is no specific registered module for just Pixel Layers. Therefore, the `searchPayload` would return null and that is why that line is there to check for the existence of the payload and then line 517 is there to check for the existence of the noOptionsString.

DivyaPrabhakar:
@volfied I have made the changes you suggested and commented back on one. Also NO_OPTIONS_STOCK is taken out of this PR and I will add it to the other one. Back to you.

volfied:
Alrighty then

Issue #3429, opened 2015-12-11
grunt debug - webpack  does not incorporate new JSON changes while running


pineapplespatula:
When you switch between branches with different localization files you get errors on launch

![image](https://cloud.githubusercontent.com/assets/1008531/11752757/174e1968-9ff6-11e5-8412-a23b4d17a395.png)

the current workaround is killing a rerunning grunt debug.

iwehrman:
Fixed by #3501.

PR #3501, opened 2016-01-12
Watch for dictionary file changes

volfied:
This change watches for localization file changes and rebuilds as necessary. Addresses #3429


mcilroyc:
LGTM.  tested, works.

PR #3503, opened 2016-01-12
Layer Panel: faster layer face selection/expansion/visibility

shaoshing:
**Update**
* rebased with master
* description is updated to inlcude result of latest experiment

This PR aims to make the layer selection/expand/collapse/visibility more responsive (UI wise) by promoting the browser's rendering event. Take layer selection as example. Below is the Timeline when select a layer. As you can see, before the browser starts rendering, it has to wait 70ms - 80ms for the Flux cycle and React rendering to be completed. So, what if we use `setState` to trigger React rendering directly, so that the browser can start rendering as soon as possible?

Below is the Timeline after applying the `setState` method. We store the layer selection/expand/visibility in component state, and update the state directly to skip the Flux cycle. The result is as expected, the browser render event got promoted. However, the `paint` event did not get promoted along with the render event, which make the optimization useless because the paint event is when the browser actually fill the pixels. 

The next experiment is to promote the paint event as well. Based on my observation, the paint event starts when the scripting engine is idle. Usually A delay of 5ms - 10ms will be enough to nudge the browser paint the update style/layer. Here is the new Timeline:

Here is the complete evaluation of the optimization.

  | Before   | After | Change
------------- | -------------| ----- | ----
Layer selection | 170ms      | **60ms** | -110ms
AB selection    | 195ms     | **98ms** | -97ms
Collapse AB     | 179ms     | **56ms** | -123ms
Expand AB       | 196ms     | **95ms** | -101ms
Collapse AB (recursively)     | 159ms     | **91ms** | -68ms
Expand AB  (recursively)      | 183ms     | **130ms** | -53ms
AB Visibility   | 186ms     | **110ms** | -76ms
(Numbers are the time needed to complete a frame.)

### DIFF
--- a/src/js/actions/groups.js
+++ b/src/js/actions/groups.js
@@ -70,7 +70,9 @@ define(function (require, exports) {
 
         var targetLayers = layers;
         if (recursive) {
-            layers = layers.flatMap(document.layers.descendants, document.layers);
+            layers = layers.flatMap(document.layers.descendants, document.layers).filter(function (layer) {
+                return layer.isGroup;
+            });
         }
 
         // When collapsing layers, if a visible descendent is selected then the selection
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -120,8 +120,8 @@ define(function (require, exports, module) {
                 this._scrollIntoView(false);
             }
             
-            this.getFlux().store("document").addLayerSelectionListener(this.props.layer.id,
-                this._handleSelectionChange);
+            this.getFlux().store("document").addLayerStateListener(this.props.layer.id,
+                this._handleLayerStateChange);
         },
         
         componentDidUpdate: function (prevProps, prevState) {
@@ -134,7 +134,7 @@ define(function (require, exports, module) {
         },
 
         componentWillUnmount: function () {
-            this.getFlux().store("document").removeLayerSelectionListener(this.props.layer.id);
+            this.getFlux().store("document").removeLayerStateListener(this.props.layer.id);
         },
         
         _scrollIntoView: function (prevSelected) {
@@ -161,18 +161,13 @@ define(function (require, exports, module) {
 
             // Suppress click propagation to avoid selection change
             event.stopPropagation();
-            
-            var nextExpandedState = !this.state.expanded;
-            
-            this.setState({
-                expanded: nextExpandedState
-            });
 
             // Presence of option/alt modifier determines whether all descendants are toggled
             var flux = this.getFlux(),
                 modifierStore = flux.store("modifier"),
                 modifierState = modifierStore.getState(),
-                descendants = modifierState.alt;
+                descendants = modifierState.alt,
+                nextExpandedState = !this.state.expanded;
 
             // The clicked layer may an have out-of-date document models due to
             // the aggressive SCU method in LayersPanel.
@@ -229,7 +224,7 @@ define(function (require, exports, module) {
         _handleLayerClick: function (event) {
             event.stopPropagation();
             this._suppressNextScrollTo = true;
-            
+
             var modifier = "select";
 
             if (event.shiftKey) {
@@ -241,7 +236,7 @@ define(function (require, exports, module) {
                     modifier = "add";
                 }
             }
-            
+
             // The clicked layer may an have out-of-date document models due to
             // the aggressive SCU method in LayersPanel.
             var documentID = this.props.document.id,
@@ -260,9 +255,10 @@ define(function (require, exports, module) {
          * 
          * @param {boolean} selected
          */
-        _handleSelectionChange: function (selected) {
+        _handleLayerStateChange: function (selected, expanded) {
             this.setState({
-                selected: selected
+                selected: selected,
+                expanded: expanded
             });
         },
 
@@ -280,7 +276,14 @@ define(function (require, exports, module) {
             this.setState({
                 visible: !toggled
             });
-            this.getFlux().actions.layers.setVisibility(this.props.document, this.props.layer, !toggled);
+            
+            // Performance Hack: delay the visibility action to promote browser paint event.
+            Promise
+                .bind(this)
+                .delay(0)
+                .then(function () {
+                    this.getFlux().actions.layers.setVisibility(this.props.document, this.props.layer, !toggled);
+                });
         },
 
         /**
@@ -586,7 +589,6 @@ define(function (require, exports, module) {
                 isDragging = this.state.isDragging,
                 isDropTarget = this.state.isDropTarget,
                 dropPosition = this.state.dropPosition,
-                hasChildren = this.props.hasChildren,
                 selected = this.state.selected,
                 expanded = this.state.expanded,
                 visible = this.state.visible;
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -119,6 +119,9 @@ define(function (require, exports, module) {
             if (this.props.document.layers.selected.last() === this.props.layer) {
                 this._scrollIntoView(false);
             }
+            
+            this.getFlux().store("document").addLayerSelectionListener(this.props.layer.id,
+                this._handleSelectionChange);
         },
         
         componentDidUpdate: function (prevProps, prevState) {
@@ -128,6 +131,10 @@ define(function (require, exports, module) {
             }
 
             this._scrollIntoView(prevState.selected);
+        },
+
+        componentWillUnmount: function () {
+            this.getFlux().store("document").removeLayerSelectionListener(this.props.layer.id);
         },
         
         _scrollIntoView: function (prevSelected) {
@@ -223,39 +230,40 @@ define(function (require, exports, module) {
             event.stopPropagation();
             this._suppressNextScrollTo = true;
             
-            var nextSelectedState = true,
-                modifier = "select";
+            var modifier = "select";
 
             if (event.shiftKey) {
                 modifier = "addUpTo";
-                nextSelectedState = true;
             } else if (system.isMac ? event.metaKey : event.ctrlKey) {
                 if (this.state.selected) {
                     modifier = "deselect";
-                    nextSelectedState = true;
                 } else {
                     modifier = "add";
-                    nextSelectedState = false;
                 }
             }
+            
+            // The clicked layer may an have out-of-date document models due to
+            // the aggressive SCU method in LayersPanel.
+            var documentID = this.props.document.id,
+                documentStore = this.getFlux().store("document"),
+                currentDocument = documentStore.getDocument(documentID),
+                currentLayer = currentDocument.layers.byID(this.props.layer.id),
+                options = {
+                    modifier: modifier
+                };
 
-            if (this.state.selected !== nextSelectedState) {
-                this.setState({
-                    selected: nextSelectedState
-                });
-
-                // The clicked layer may an have out-of-date document models due to
-                // the aggressive SCU method in LayersPanel.
-                var documentID = this.props.document.id,
-                    documentStore = this.getFlux().store("document"),
-                    currentDocument = documentStore.getDocument(documentID),
-                    currentLayer = currentDocument.layers.byID(this.props.layer.id),
-                    options = {
-                        modifier: modifier
-                    };
-
-                this.getFlux().actions.layers.select(currentDocument, currentLayer, options);
-            }]
+            this.getFlux().actions.layers.select(currentDocument, currentLayer, options);
+        },
+        
+        /**
+         * Handle select/deselect event
+         * 
+         * @param {boolean} selected
+         */
+        _handleSelectionChange: function (selected) {
+            this.setState({
+                selected: selected
+            });
         },
 
         /**
@@ -578,7 +586,7 @@ define(function (require, exports, module) {
                 isDragging = this.state.isDragging,
                 isDropTarget = this.state.isDropTarget,
                 dropPosition = this.state.dropPosition,
-                hasChildren = this.props.childNodes,
+                hasChildren = this.props.hasChildren,
                 selected = this.state.selected,
                 expanded = this.state.expanded,
                 visible = this.state.visible;
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -72,18 +72,26 @@ define(function (require, exports, module) {
                 dropPosition: null,
                 isDragging: false,
                 dragStyle: null,
-                isEditing: false
+                isEditing: false,
+                selected: this.props.layer.selected,
+                expanded: this.props.layer.expanded,
+                visible: this.props.layer.visible
             };
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
+            var layerFace = function (layer) {
+                // Exclude face attributes that are already represented by the component's state.
+                return layer.face.merge({ selected: null, expanded: null, visible: null });
+            };
+            
             // Drag states
             if (!_.eq(this.state, nextState)) {
                 return true;
             }
             
             // Face change
-            if (!Immutable.is(this.props.layer.face, nextProps.layer.face)) {
+            if (!Immutable.is(layerFace(this.props.layer), layerFace(nextProps.layer))) {
                 return true;
             }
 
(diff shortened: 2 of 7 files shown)


iwehrman:
I'll take a look at this as soon as it's rebased.

shaoshing:
@iwehrman I've updated the PR to show my latest experiment. Please check the updated description.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -147,9 +180,14 @@ define(function (require, exports, module) {
                 documentStore = this.getFlux().store("document"),
                 currentDocument = documentStore.getDocument(documentID),
                 currentLayer = currentDocument.layers.byID(this.props.layer.id);
-
-            this.getFlux().actions.groups.setGroupExpansion(currentDocument, currentLayer,
-                !currentLayer.expanded, descendants);
+            
+            Promise
+                .bind(this)
+                .delay(5)
```
shaoshing: This is the experiment that delays the actions to promote the paint event.

src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -103,10 +103,7 @@ define(function (require, exports, module) {
                     );
 
                     var className = classnames({
-                        "layer-group": layerGroup,
-                        "layer-group__selected": layerGroup && layer.selected,
-                        "layer-group__collapsed": layerGroup && !layer.expanded,
-                        "layer-group__not-visible": layerGroup && !layer.visible
```
shaoshing: I have to move these classes into `LayerFace` due to the `selected/expanded/visible` attributes is not always updated.

shaoshing: As a result, I have to switch to sibling selector `.layer__not-invisible + .layer-group`, and the performance is not as good as child selector `layer-group__not-invisible > .layer-group`. I am observing 10ms - 15ms increased in rendering time.

iwehrman:
Just hit this when activating Vermilion for the first time in a session:
```
Uncaught TypeError: Cannot read property 'layers' of undefined
  (anonymous function) @ document.js:182
  tryCatcher @ bluebird.js:4978
  Promise._settlePromiseFromHandler @ bluebird.js:3042
  Promise._settlePromise @ bluebird.js:3099
  Promise._settlePromiseCtx @ bluebird.js:3136
  (anonymous function) @ bluebird.js:152
  (anonymous function) @ bluebird.js:4276
```

iwehrman:
Quick impression: very nice! Feels good, man! 

Nit: would it be hard to special-case recursive directory expansion? When I do it currently, I see this:
img

shaoshing:
@iwehrman `Cannot read property 'layers' of undefined` this should be fixed. I think you already pull the latest commit and it now works? 

`would it be hard to special-case recursive directory expansion` the easiest way is to drop the optimziation on rendering collapsed layer group at start up. Othweise, it will not easy to synchronize the display.

placegraphichere:
I also see what seems like a 2 step process when OPT+clicking a folder to expand all sub-groups. It is noticeable. The first level will open, then a small pause, and then the second level will open.

shaoshing:
@iwehrman @placegraphichere Thanks for trying! Regarding the recursive expand, I may have a easier solution for this, which is similiar to the selection listener. Will update the PR later for review.

iwehrman:
@shaoshing: maybe a compromise would be something like we do for datalist, which is to initially render just the visible subset of options so that we can open the menu as quickly as possible, but then _immmediately afterward_ render the rest of the options.

iwehrman:
In any case, I'm willing to forego that optimization if 1) it helps solve the problem, and 2) it doesn't significantly impact document-load time.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -147,9 +183,15 @@ define(function (require, exports, module) {
...
                 currentDocument = documentStore.getDocument(documentID),
                 currentLayer = currentDocument.layers.byID(this.props.layer.id);
-
-            this.getFlux().actions.groups.setGroupExpansion(currentDocument, currentLayer,
-                !currentLayer.expanded, descendants);
+            
+            // Performance Hack: delay the expand action to promote browser paint event.
+            Promise
+                .bind(this)
+                .delay(10)
```
iwehrman: I'm curious about how this constant affects things. Does it work the same way with `.delay(0)`? How about `setTimeout(fn, 0)`? In theory, bluebird should now be treating `delay(0)` like `process.nextTick` (i.e., executing asynchronously but _in the same frame_) whereas `delay(n)` for `n > 0` it should be treating like `setTimeout(fn, n)` (i.e., asynchronously and _in a new frame_).

shaoshing: according to my experiment, `delay(0)` and `delay(10)` both serve the same purpose: callback is executed in a new frame. There is also an API `setImmediate` which has a direct goal on solving this problem (https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate); sadly it is only available in IE/Edge

shaoshing:
@iwehrman @placegraphichere I've made the following changes to deal with the delay related to expand.

* expand layer group has delayed on showing the child layers (the icon is expanded while the child layers are still hidden). This is due to the optimization on skipping collapsed layers to reduce initial load time. The solution is as @iwehrman suggested, the root `LayerGroup` will force update the child to render the collapsed layers when it is mounted.

* recursive expand has delay on showing child layers. Now the groups will expand at the same time with the expand/collapse state listener. 

BTW, when testing, you will see the icon flashing quite obviously. This will be addressed in another PR.

src/js/jsx/sections/layers/LayerFace.jsx
iwehrman: There was a bug in bluebird 3.0.x such that zero-delay asynchronous callbacks were executed with `setTimeout` instead of something functionally equivalent to `setImmediate`. Even though `setImmediate` is only available in Edge, it's possible to write a function using different primitives (e.g., `MutationObserver`) with the same semantics. This is what bluebird should have been doing all along, but only started doing again in 3.1.x as a result of a bugfix. This very subtle change in behavior has triggered a few bugs in our code already (in cases where we were inappropriately sensitive to the difference).

Anyway!

src/js/actions/groups.js
```diff
@@ -70,7 +70,9 @@ define(function (require, exports) {
 
         var targetLayers = layers;
         if (recursive) {
-            layers = layers.flatMap(document.layers.descendants, document.layers);
+            layers = layers.flatMap(document.layers.descendants, document.layers).filter(function (layer) {
```
iwehrman: Whoops! Nice catch.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -196,20 +224,19 @@ define(function (require, exports, module) {
         _handleLayerClick: function (event) {
             event.stopPropagation();
             this._suppressNextScrollTo = true;
-
+
```
iwehrman: Bad whitespace change.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -196,20 +224,19 @@ define(function (require, exports, module) {
...
-
-                if (selected) {
+                if (this.state.selected) {
                     modifier = "deselect";
                 } else {
                     modifier = "add";
                 }
             }
-
+
```
iwehrman: Bad whitespace change.

src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -116,7 +136,7 @@ define(function (require, exports, module) {
                                 disabled={this.props.disabled}
                                 document={this.props.document}
                                 layer={layer}
-                                childNodes={layerNode.children}/>
+                                hasChildren={!!layerNode.children}/>
```
iwehrman: Not a problem, but I wonder why you decided to include this explicit prop when you've got the layer as a prop anyway (and this calculation could just be performed inside the layer).

src/js/stores/document.js
```diff
@@ -38,6 +39,11 @@ define(function (require, exports, module) {
          * @type {Object.<number, Document>}
          */
         _openDocuments: null,
+        
+        /**
+         * @type {Map.<number, function>}
+         */
+        _layerIDToStateListenner: null,
```
iwehrman: Typo: should be `_layerIDToStateListener`

src/js/stores/document.js
```diff
@@ -106,6 +112,7 @@ define(function (require, exports, module) {
          */
         _handleReset: function () {
             this._openDocuments = {};
+            this._layerIDToStateListenner = {};
```
iwehrman: The `@type` annotation above indicates that this should be a `Map.<number, function>`, but here it's initialized to an `Object.<number, function>`. I'm generally in favor of using `Map` for lookup tables like this instead of `Object`.

src/js/stores/document.js
```diff
@@ -166,18 +173,37 @@ define(function (require, exports, module) {
...
-                this.emit(changeEventName || "change", nextDocument);
+            
+            var promise = Promise.bind(this);
+
+            // Notify change of layer selection state.
+            if (oldDocument) {
+                promise = promise.then(function () {
+                    oldDocument.layers.layers.forEach(function (layer) {
+                        var nextLayer = nextDocument.layers.byID(layer.id),
+                            callback = this._layerIDToStateListenner[layer.id];
```
iwehrman: I wonder if we could just make our `Layer` models instances of `EventEmitter`?

src/js/stores/document.js
```diff
@@ -166,18 +173,37 @@ define(function (require, exports, module) {
...
-            if (initialized) {
-                // Include the document in the event payload
-                this.emit(changeEventName || "change", nextDocument);
+            
+            var promise = Promise.bind(this);
+
+            // Notify change of layer selection state.
+            if (oldDocument) {
+                promise = promise.then(function () {
+                    oldDocument.layers.layers.forEach(function (layer) {
```
iwehrman: Is this iteration slow on large documents? It always worries me when we iterate over all layers. Would it be worth it to try to optimize this by uninstalling listeners on layers that are, e.g., hidden in the layers panel? (That could well be over-optimization, but hey who knows.)

src/js/stores/document.js
```diff
@@ -166,18 +173,37 @@ define(function (require, exports, module) {
...
+                        var nextLayer = nextDocument.layers.byID(layer.id),
+                            callback = this._layerIDToStateListenner[layer.id];
+                        
+                        if (nextLayer && callback &&
```
iwehrman: If `!nextLayer || !callback` is there a bug? Should we warn or possibly throw in that case? If not, can you add a comment about why it might be expected for one of these to not hold?

src/js/stores/document.js
```diff
@@ -166,18 +173,37 @@ define(function (require, exports, module) {
...
+                // If some selected layer remains uninitialized, a new document will
+                // come along shortly. Wait until then to trigger the change event.
+                var initialized = nextDocument.layers && nextDocument.layers.selected
+                    .every(function (layer) {
+                        return layer.initialized;
+                    });
+
+                if (initialized) {
+                    // Include the document in the event payload
+                    this.emit(changeEventName || "change", nextDocument);
```
iwehrman: If this is just a regular change event, is it bad to delay this emit call?

src/js/stores/document.js
```diff
@@ -166,18 +173,37 @@ define(function (require, exports, module) {
...
+                        
+                        if (nextLayer && callback &&
+                            (layer.selected !== nextLayer.selected || layer.expanded !== nextLayer.expanded)) {
+                            callback(nextLayer.selected, nextLayer.expanded);
+                        }
+                    }, this);
+                }).delay(0); // Performance Hack: delay emitting the `change` event to promote browser paint event.
             }
+            
+            promise.then(function () {
```
iwehrman: Just to be clear, the above `delay(0)` call will not apply to the then-handler below, right?

iwehrman:
Looking great. Just a few more questions and nits. Back to @shaoshing.

iwehrman:
Oh, one more thing: as follow-up work, it would be cool to hook something like this up to the on-canvas UI, so that clicking on a canvas object provides a fast update to the overlays. In theory, the situation for the selection overlays should be analogous to the layers panel.

src/js/jsx/sections/layers/LayerGroup.jsx
shaoshing: Hmm, I don't see a way to check whether or not a `Layer` model has children (`layerNode.children` is an array of `LayerNode` instances), maybe I am missing something?

src/js/stores/document.js
shaoshing: I agree, and we should definitely optimize for the select-all operation: it is extremely slow to select all layers in a large document.

src/js/stores/document.js
shaoshing: Wouldn't this break the idea that `Layer` model is immutable?

src/js/stores/document.js
shaoshing: `!nextLayer` is when the layer is deleted, and `!callback` is when the LayerFace is not mounted yet (the optimization for fast load time). I've added the comment as well.

src/js/stores/document.js
shaoshing: Yes it will. That's how we pause the `change` event for the browser to start rendering faster. This code block is actually not very clear on the purpose so I've refactored it. Please check the change and let me know if you think it is better.

src/js/stores/document.js
shaoshing: no, so I just made the change to always `delay(0)` for this event.

src/js/jsx/sections/layers/LayerGroup.jsx
iwehrman: Eh, it may not actually be better, but you can do it with the `LayerStructure` instance: `document.layers.children(layer)`.

src/js/stores/document.js
iwehrman: Ah, sorry, I missed the assignment above. You're right!

shaoshing:
Thanks for reviewing @iwehrman  :+1:  I've made some change to address your comments, and replied your questions as well. Back to you

src/js/stores/document.js
iwehrman: Yes, but "immutability" is more of a guideline than a rule :) For example, our caching scheme makes crucial use of mutating "immutable" objects. It just does so in a way that we don't expect will violate any assumption about the public interface of the immutable object.

But, again, this really might not be worth doing. Given that you only ever have one listener at a time per layer, this is so simple that whatever duplication exists isn't worth worrying about.

src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -68,7 +72,25 @@ define(function (require, exports, module) {
...
+             */
+            renderCollapsedLayers: React.PropTypes.bool
+        },
+        
+        componentDidMount: function () {
+            if (!this.props.parentLayer) {
+                this.renderCollapsedLayers = true;
+
+                window.setTimeout(function () {
+                    this.forceUpdate.bind(this, null);
```
iwehrman: Wait, what? I have no idea what's going on here. You're binding a function but then throwing it away after a 2 second timeout? I'm confused!

iwehrman:
Only the 2-second timeout is preventing me from hitting the merge button. Back to @shaoshing.

src/js/jsx/sections/layers/LayerGroup.jsx
shaoshing: Yes that's on purpose. At the beginning, we skip rendering collapsed layers for faster loading, then we wait for 2 seconds to perform a `foreUpdate` to render those collapsed layers. I added the explanation at https://github.com/adobe-photoshop/spaces-design/pull/3503/files#diff-690c204937d25a539891aa9f5a9313fcR43, but you are still confused by this, so maybe I should explain more in here as well?

iwehrman: OK, I see now that this is the "wait a bit and then render everything" optimization, which I didn't get before. The comment above makes sense, I just didn't put it together. It might still be nice to have a one-liner at this if statement just to help dummies like me :) But, what still doesn't make sense is the timeout function, which just evaluates `this.forceUpdate.bind(this, null)`. This expression evaluates to a bound function, but it does not call the function or save a reference to it. Should this be `this.forceUpdate.call(this, null)` instead?

shaoshing: Ah, I see what you mean now. I forgot to revet this change when I was experimenting `setTimeout(this.forceUpdate.bind(this, null), 2000)`. Nice catch! I've fixed it and added the comment to clear the confusion.

iwehrman:
Let's give it a whirl!

Issue #3495, opened 2016-01-11
Guide zones flash unnecessarily


iwehrman:
The guide zones are currently shown any time a mousemove event is triggered inside the zone. This causes the guide zones to flash unnecessarily and quite frequently, like when mousing to the document header or to the toolbar. 

![img](https://s3.amazonaws.com/f.cl.ly/items/0T283c2c0A3r1n362V0a/Screen%20Recording%202016-01-11%20at%2003.02%20PM.gif?v=328c1cd1)

Instead, they should only show if the cursor hovers in the zone for some brief but non-trivial amount of time.

iwehrman:
Fixed by #3508.

PR #3504, opened 2016-01-12
Delay guide zone highlighting by debouncing

volfied:
Debounces guide zone hover highlighting, delaying it enough that if the mouse just passes over the area, it doesn't flash.

For guide areas, 3 out of 4 cases we don't need the API to tell us where the mouse is, because it will go over one of our panels so we can cancel the debounced function, for the bottom zone, we will need to use the core `getMousePosition` API (not implemented yet) to check if cursor has left the app frame.

Until then, this prevents flashing in the more visible cases. - #3495


volfied:
Since Core APIs are available, closing this to update.

Issue #2640, opened 2015-09-25
Clicking in guide zone should not create a guide


iwehrman:
If you click in a guide zone, it creates a guide and drops it inside the guide zone. This should be disallowed because it's very easy to accidentally create guides here when trying to grab a new guide from the zone. The desired behavior here:

1. If you drop an existing guide in the guide zone, it should delete the guide.
2. If you drop a fresh guide in the guide zone, it should never be created.

tiburbage:
I have also run into this and was going to log a bug. In some cases, this results in an "orphaned" guide you can't get at again to move or delete.

volfied:
I am working on fixing this, but I don't think dropping a guide on the guide zone should delete it. We don't highlight guide zones during guide drag, so it would confuse the user to delete a guide if they drag it to the edge of visible canvas.

iwehrman:
I think dropping on the guide zone should delete it, but that there should be some feedback about what you're about to do, like changing the cursor in some way.

volfied:
@placegraphichere Ian and I disagree, can you be the tie breaker?

ktaki:
How about deleting it if no dragging happens in the guide zone? Currently, just clicking the area will create a guide and it is hard to re-grab without scrolling the canvas. If the minimum dragging is required, it will avoid mis-creation of a guide.

volfied:
@ktaki, my PR #3506 implements that. We are currently disagreeing on what happens if you drag an existing guide to one of the zone areas.

ktaki:
Oh, OK. Then, I agree with you, @volfied, that moving a guild to the guide zone should not delete the guide. You do not see the guide zone when you are dragging a guide anyway. It is a part of regular canvas.

mcilroyc:
The PR (just merged) prevents guide from being created with a simple click on the guide-zone.  Is this ready to FBNC?

volfied:
@mcilroyc I'd like @placegraphichere's opinion on the second part before test/close.

placegraphichere:
So the fact that you do not see the guide zone when dragging an existing guide over a guide zone to me tells me it should _not_ delete. I would have no feedback that a delete would occur, and it would be confusing.

volfied:
Given Taki and Seth's reply, I'm setting this to FBNC and passing it back to Ian to close (or assign to Taki/Chad to verify)

iwehrman:
[...]

PR #3506, opened 2016-01-12
Create guides on zone drag-out instead of mouse down

volfied:
Instead of immediately starting a guide tracker on mouse down, now we wait for user to drag the mouse out of the guide zone before creating/dragging the guide. This prevents creating guides underneath the zone areas.

Adresses #2640, at least the part I agree with.


mcilroyc:
seems reasonable to me.  works as expected.  merging.

PR #3508, opened 2016-01-13
Delay guide zone highlighting by debouncing

volfied:
On mouse over for the guide zones, we call a debounced function that delays highlight of the zone area by 30 ms. At the end of the delay, we use the `OS.getMousePosition` API to make sure the mouse is still over the zone, and otherwise cancel highlighting.

Addresses #3495

Requires https://github.com/adobe-photoshop/spaces-adapter/pull/181 and playground-dev-mac#971 or later. (CL 1059051))

### DIFF
--- a/src/js/jsx/tools/GuidesOverlay.jsx
+++ b/src/js/jsx/tools/GuidesOverlay.jsx
@@ -66,6 +66,13 @@ define(function (require, exports, module) {
          */
         _scrimGroup: null,
 
+        /**
+         * Throttled function that will eventually highlight an area
+         *
+         * @type {function}
+         */
+        _debouncedHighlight: null,
+
         getStateFromFlux: function () {
             var flux = this.getFlux(),
                 applicationStore = flux.store("application"),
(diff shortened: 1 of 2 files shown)


volfied:
@iwehrman This is now good to merge with updated dependencies.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -233,6 +242,40 @@ define(function (require, exports, module) {
...
+         * @param {number} y mouse Y coordinate
+         * @return {boolean} True if [x,y] is in the zone
+         */
+        zoneUnderLocation: function (zone, x, y) {
+            var zoneLeft = mathUtil.parseNumber(zone.attr("x")),
+                zoneTop = mathUtil.parseNumber(zone.attr("y")),
+                zoneRight = zoneLeft + mathUtil.parseNumber(zone.attr("width")),
+                zoneBottom = zoneTop + mathUtil.parseNumber(zone.attr("height"));
+            
+            return zoneLeft < x && zoneRight > x &&
```
iwehrman: Seems like the edges of the zone should be included? Also, remember `mathUtil.parseNumber` can return null. Presumably if the `zone` were somehow not ready for this, the function will end up returning false? Might want to double-check that.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -233,6 +242,40 @@ define(function (require, exports, module) {
...
         /**
+         * Returns true if the given D3 component is under the given mouse location
+         *
+         * @private
+         * @param {SVGElement} zone Guide zone being checked
+         * @param {number} x mouse X coordinate
+         * @param {number} y mouse Y coordinate
+         * @return {boolean} True if [x,y] is in the zone
+         */
+        zoneUnderLocation: function (zone, x, y) {
```
iwehrman: We usually underscore-prefix private methods.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -233,6 +242,40 @@ define(function (require, exports, module) {
...
+                zoneTop < y && zoneBottom > y;
+        },
+
+        /**
+         * Highlights the given zone, is debounced at component mount
+         *
+         * @private
+         * @param {SVGElement} zone
+         */
+        highlightZone: function (zone) {
```
iwehrman: Same here re: private methods.

iwehrman:
This seems to work well, though honestly I'm not _quite_ sure why it does since I would not have guessed that debouncing is what we want here. It seems like if we provide a steady trickle of mouse events within a guide zone, the zone would never appear, though would we want it to. Still, I can't make this bad hypothetical behavior happen, so I guess it's fine!

Back to @volfied for nits.

src/js/jsx/tools/GuidesOverlay.jsx
volfied: If the zone is not ready for this, it should return false, we don't want to highlight it. I don't see mathUtil.parseNumber returning null in this function given we set those attributes. I'll make the <= changes.

volfied:
@iwehrman Fun fact, `_.throttle` does not have a `cancel` method! I initially used throttle, and cancel was not available - big wtf moment for me. I can actually make that behavior happen, but it requires really good mouse control and because of the 30ms debounce delay, draws the overlay before the user is able to notice the problem.

I went with performance/illusion of correctness instead of complete correctness. I think it's ok.

volfied:
Back to you @iwehrman

iwehrman:
OK!

Issue #3485, opened 2016-01-09
Appearance panel misalignment


iwehrman:
The layer blend mode datalist and layer opacity components are aligned differently in the appearance panel (apparently accidentally) depending on whether a type or vector layer is selected:

![img](https://s3.amazonaws.com/f.cl.ly/items/020Y1J3E3Z251S3e2n3m/Screen%20Recording%202016-01-09%20at%2001.19%20PM.gif?v=9b4d9ddd)

iwehrman:
CC @DivyaPrabhakar

iwehrman:
Fixed by #3509.

PR #3509, opened 2016-01-13
Appearance panel datalist position remains still

DivyaPrabhakar:
References issue #3485


iwehrman:
I love it. I also love that I can see in your gif that the `Radius` component's input is ALSO slightly out of alignment! I'll file a separate issue :grin:

PR #3511, opened 2016-01-13
Styles: more efficient browser rendering when mouse over (layers/libraries list)

shaoshing:
This PR improves the browser rendering performance when mouse over layers/libraries list. 

### Layers List
Before the fix, mouse over a layer face will invalidate large number of unrelated layer faces and cause the browser to spend more time on rendering. The fix is removing the style `.face:hover {z-index: 99}`, and my test shows it doesn't cause other style issue.
layer-panel-hovering-before

After the fix
layer-panel-hovering-after

### Liraries List
The cause is related to the flexbox view: with the combination of `flex-shrink:1` and `flex-grow:1`, it seems force the libraries list to reflow when mouse over its children. 
libraries-before

After the fix
libraries-after

Performance has a significant boost:
Before
libraries-before
After
libraries-after

### DIFF
--- a/src/style/sections/layers/face.less
+++ b/src/style/sections/layers/face.less
@@ -125,7 +125,6 @@ input[type="text"].face__name, input[type="text"]:disabled.face__name {
 // Selection state
 .face:hover {
     background-color: @list-hover-color;
-    z-index: 99;
     
     input[type="text"], input[type="text"]:disabled {
         color: @item-hover;
--- a/src/style/sections/libraries/libraries-section.less
+++ b/src/style/sections/libraries/libraries-section.less
@@ -199,7 +199,7 @@
     }
 
     &__content {
-        flex-shrink: 1;
+        height: 5.4rem;
         flex-grow: 1;
         overflow-y: auto;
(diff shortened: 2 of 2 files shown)


iwehrman:
Strange: I only see very slightly faster times on this branch compared to master. I measured both this and master a few times and each time I got something like:

master: 
image

your branch:
image

I'm perfectly happy to make a small change that results in a small speed-up, of course. I can confirm that the paint flashing is significantly reduced. But it's odd that your measurements seem so different from mine. I suppose it could be due to our different machines? Oh well, we can talk about it in the stand-up tomorrow if either of us have any deep thoughts between now and then.

shaoshing:
Hey @iwehrman I am wondering which panel you were testing? For the layer panel, the slight improvement is expected, I am also observing the same result. For the libraries panel, if you test with a large collection (e.g. Samples Collection April 23 2015), you should be able to see much better improvement.

iwehrman:
Ah, I did not measure the Libraries panel specifically. I believe your numbers, and feel better now that there is an explanation. Thanks!

Issue #3512, opened 2016-01-13
Radius text input slightly out of alignment


iwehrman:
In #3509 @DivyaPrabhakar showed off her alignment fix for the `BlendMode` component's  `TextInput`. She showed too much though, and I can see a similar, minor grid alignment issue with the `Radius` component's `NumberInput`:

![img](https://camo.githubusercontent.com/7468a166410a574460de88741d8d75ce891102af/68747470733a2f2f7777772e64726f70626f782e636f6d2f732f306f673978787a7367307078316d752f6775696465732d73616d652d706f736974696f6e2e6769663f7261773d31)

![image](https://cloud.githubusercontent.com/assets/1075154/12283638/132b4e70-b95d-11e5-828b-bc90f8a1f57d.png)

DivyaPrabhakar:
@iwehrman ![](https://31.media.tumblr.com/a4bbfa577497665658098c49d6197da8/tumblr_inline_n6aan62zxa1r0ql8k.gif)

iwehrman:
Fixed by #3513.

PR #3513, opened 2016-01-13
Adjusts radius text input alignment in appearance panel

DivyaPrabhakar:
References #3512

(Apologies for the UNREGISTERED sprawled over this image)

### DIFF
--- a/src/js/jsx/sections/style/Radius.jsx
+++ b/src/js/jsx/sections/style/Radius.jsx
@@ -165,7 +165,7 @@ define(function (require, exports, module) {
 
         render: function () {
             return (
-                <div className="formline">
+                <div className="formline radius_input">
                     <Label
                         className="label__medium__left-aligned"
                         size="column-4"
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -263,6 +263,10 @@
     flex-shrink: 0;
 }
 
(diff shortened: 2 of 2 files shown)


src/js/actions/search/stock.js
```diff
@@ -0,0 +1,123 @@
+/*
```
iwehrman: I don't think this is supposed to be here :grimacing:

iwehrman:
Also, I recommend checking with @tomwmcrae about buying a copy of [xScope](http://xscopeapp.com/). It's an essential tool for careful UI work, like you're doing. I think he can be convinced.

iwehrman:
Looks amazing, btw!

DivyaPrabhakar:
@iwehrman  oops! Back to you.

iwehrman:
Beautiful!

tomwmcrae:
@DivyaPrabhakar: Sure. Will you also need an Apple Watch to get the full value out of this?

From [xScope](http://xscopeapp.com/) page:
Mirror   Effortlessly see your designs and applications on iOS and Apple Watch while you work on them from your Mac.

DivyaPrabhakar:
@tomwmcrae You said it, not me. I'll expect it on my desk tomorrow. #bestmanagerever

PR #3517, opened 2016-01-13
Only check for document.guidesVisible in guidesOverlay

volfied:
While looking at Timeline, I realized that instead of comparing documents (which, admittedly, will always be different), we should just compare the `guidesVisible` property as it's the only thing we care about in GuidesOverlay. 

This change prevents guidesOverlay from being redrawn everytime something happens in the document, helps with performance a little.


iwehrman:
LGTM!

Issue #3520, opened 2016-01-13
Guide zone flash on creation of new guide


chadrolfs:
move mouse to guide zone (blue hover)
mouse down (no blue hover) + drag (blue hover) then once you leave the zone (no blue hover)

Expect: either it should remain blue hover until exiting the zone or remain no blue hover while drag out of zone.

chadrolfs:
![guidezone-blink](https://cloud.githubusercontent.com/assets/8442631/12307253/ff424364-b9f2-11e5-919c-c601783d2833.gif)

iwehrman:
FBNC to @chadrolfs.

Issue #3270, opened 2015-11-18
Guide policies misaligned


iwehrman:
On a 5k retina iMac with `font-size: 62.5%`, the guide policy areas are consistently misaligned:
![image](https://cloud.githubusercontent.com/assets/1075154/11253032/20afbe4a-8dee-11e5-9d66-abca4899c571.png)

volfied:
hey @iwehrman can you test this to see whether the scroll changes outside those lines?

iwehrman:
I do know that the cursor changes to an unadorned black arrow when it's in the policy area but NOT over the guide itself:

![img](https://s3.amazonaws.com/f.cl.ly/items/0O2g0i1w1f0w2t1l2X1r/Screen%20Recording%202016-01-11%20at%2003.04%20PM.gif?v=71844c61)

If I click and drag in that area, the drag fails silently.

iwehrman:
It's better, but it's still not centered: 
![image](https://cloud.githubusercontent.com/assets/1075154/12339168/6357b8c2-bac8-11e5-8c88-eb3187143e3d.png)

It would be really nice to know why this is so far off...

iwehrman:
Image above is retina; here's non-retina:
![image](https://cloud.githubusercontent.com/assets/1075154/12339206/9e0f0f88-bac8-11e5-8fa5-32ee508ba07a.png)

volfied:
@iwehrman I agree with you that policies are not symmetric around the guide, but these areas are defined in PS and I'm matching them 1-to-1. If I enlarge the policy on the horizontal bottom / vertical right, we start sending PS move drag operations instead of guide drag. If I shrink it on the other side, we lose area of guide that's draggable, where cursor changes to guide move cursor but clicks deselect (since we get the mouse event).

So I'm trying to say this is a non-problem at this point, and only way to make it symmetric is to go into PS and change the guide drag zones there.

iwehrman:
OK, then I'm just going to close this issue and try not to look at those guide zones any longer :stuck_out_tongue_closed_eyes:

PR #3521, opened 2016-01-13
Guide fixes

volfied:
This PR addresses two issues:
#3520 - we make sure to not highlight a guide zone while mouse is down.
#3270 - Increase the right/bottom policy area of guides by one pixel, this allows for symmetrical guide policy zones around the guide.

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -154,13 +154,13 @@ define(function (require, exports) {
                     x: canvasBounds.left,
                     y: Math.floor(guideTL.y - policyThickness - 1),
                     width: canvasBounds.right - canvasBounds.left,
-                    height: Math.ceil(policyThickness * 2 + 1)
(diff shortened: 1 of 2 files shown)


iwehrman:
Something is not right in this branch. After I pull out the first guide, the guide zones don't show any longer. 
img

iwehrman:
Well, after rebasing against master it works again. But, I notice that when dragging a guide, the guide is always created in the guide zone and shown while dragging, which is a little confusing:
img
Is there anything that can be done about that?

volfied:
Fix can come only from core side. We create the guide where user first started dragging, and then start a track on that guide. In the older way, guide zone was acting as a buffer hiding this artifact.

Only workaround I see is us drawing our own temporary guide, but it's bug prone, will not look as good, and will not have the smart popup telling guide location.

iwehrman:
OK. Thanks for humoring me!

PR #3523, opened 2016-01-14
 fix action call to refer to the correct action file

pineapplespatula:
fix spurious delete action that didnt get caught when the function moved


iwehrman:
Sure enough.

PR #3528, opened 2016-01-14
Updated npm-shrinkwrap for sprint 3.1

mcilroyc:
This is the version of npm-shrinkwrap that was pushed to perforce with the sprint 3.1 release.


iwehrman:
I did some light testing; things seem OK. I wonder why the format has apparently changed though? I wonder if this is due to differing node versions? In any case, we might want to specify `engines` in package.json in the future?

Issue #3491, opened 2016-01-11
ShouldComponentUpdate methods for D3 drawing components


volfied:
Our overlays use D3, and they expect React to update the parent component to update themselves. So we may be able to speed things up / reduce noise and unnecessary rendering by implementing a SCU method for these components.

iwehrman:
Fixed by #3531.

PR #3531, opened 2016-01-14
D3 Overlay shouldComponentUpdate methods

volfied:
- For superselect, policy and sampler overlays
- Correct the guides overlay existing SCU
- Change Sampler behavior so it highlgihts/samples leaf layers
- Re-add file watcher that just prints out file changes

Addresses #3491, and some sampler behavior issues.

### DIFF
--- a/src/js/jsx/tools/SuperselectOverlay.jsx
+++ b/src/js/jsx/tools/SuperselectOverlay.jsx
@@ -29,6 +29,7 @@ define(function (require, exports, module) {
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         StoreWatchMixin = Fluxxor.StoreWatchMixin,
+        Immutable = require("immutable"),
         d3 = require("d3"),
         _ = require("lodash");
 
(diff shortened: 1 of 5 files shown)


src/js/jsx/tools/SuperselectOverlay.jsx
```diff
@@ -126,6 +127,21 @@ define(function (require, exports, module) {
...
 
+        shouldComponentUpdate: function (nextProps, nextState) {
+            var lastLayers = this.state.document.layers.selected,
+                nextLayers = nextState.document.layers.selected,
+                lastLayerProps = collection.pluckAll(lastLayers, ["id", "bounds"]),
+                nextLayerProps = collection.pluckAll(nextLayers, ["id", "bounds"]),
+                layersChanged = !Immutable.is(lastLayerProps, nextLayerProps);
+
+            return layersChanged ||
+                this.state.marqueeEnabled !== nextState.marqueeEnabled ||
```
iwehrman: Would it make more sense to do these faster checks first before falling back to the heavier check based on pluck?

iwehrman:
Looks fine otherwise. One Q for @volfied.

volfied:
@iwehrman right back at ya.

iwehrman:
ok!

Issue #3533, opened 2016-01-14
Guide policy areas not updated on displayChanged


iwehrman:
After dragging the window from a non-retina to a retina display:
![image](https://cloud.githubusercontent.com/assets/1075154/12335437/19f4a4ec-bab4-11e5-83bb-c86388498767.png)

iwehrman:
Note that they ALSO don't update when the window is resized, which as an even bigger problem.

chadrolfs:
I swear I wrote this up a while back.  Maybe it was a comment in a PR?  Think it might have been @volfied PR?

chadrolfs:
found it https://github.com/adobe-photoshop/spaces-design/issues/3351#issuecomment-163054277

iwehrman:
Fixed by #3535.

Issue #3351, opened 2015-12-03
D3 + guides error when resizing window


iwehrman:
With guides enabled, if I resize the window to be fairly small I get these warnings:
![image](https://cloud.githubusercontent.com/assets/1075154/11572680/5caf68ee-99b7-11e5-8ead-ab33b3797066.png)

iwehrman:
Over to @chadrolfs to verify.

chadrolfs:
I'm still running into these but maybe only when Show Policy Frames is enabled (?).  I'm also seeing incomplete updates on resize. 

![resize-policyunupdate](https://cloud.githubusercontent.com/assets/8442631/11671784/81e50504-9dbf-11e5-93f7-a782393583df.gif)

chadrolfs:
![screen shot 2015-12-08 at 3 24 07 pm](https://cloud.githubusercontent.com/assets/8442631/11671809/a9d7a53a-9dbf-11e5-8d2e-a9533844cac1.png)

iwehrman:
Back to @chadrolfs to verify again!

chadrolfs:
working fantastically!

PR #3535, opened 2016-01-14
Policy area validation and border update on resize

volfied:
For policy overlay, makes sure area trying to be drawn has positive width/height, skips otherwise, this problem manifested itself if user resized the PS app window to the smallest size possible.

For border policy change listener, we also listen to panel store for canvas rectangle size differences, which helps update all policies on resize, if top left stayed put.

Addresses #3351 and #3533


iwehrman:
:+1::+1::+1:

PR #3537, opened 2016-01-14
Debounce moveToArtboard events with a nonzero delay

mcilroyc:
We already debounce the handling of incoming moveToArtboard events from PS, but it currently uses the default delay of zero.  This seems to *occasionally* handle the first event in a series almost immediately, which results in an error in the case of option+drag artboard duplication (see #1613).

background...
After the end of an option+drag gesture, we receive a series of event from PS in this order: moveToArtboard(several), historyState, move.

moveToArtboard is handled simply by calling (debounced) `layers.resetIndex`.  An index reset will fail, however, if it is performed *after* the duplication has occurred in PS but *before* DS has handled the `move: newDuplicateSheets` event.

By adding a 100ms delay, we increase the chance that the the `resetIndex()` will be enqueued after the move event handler.  This is still a bit janky, but it seems like a safe/quick fix for this infrequent pesky error.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -60,6 +60,12 @@ define(function (require, exports) {
     };
 
     /**
+     * @const
+     * @type {number}
+     */
+    var EVENT_DEBOUNCE_DELAY = 100;
+
+    /**
      * Filter out layers that can't generally be transformed like empty,
      * adjustment and background layers.
      *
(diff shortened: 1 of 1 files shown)


ktaki:
I see. If this is a timing issue, 100 ms delay would help. Tested with the latest pd-dev build. Certainly, it became much harder to reproduce the issue. In fact, I could not break it with option-copying artboards. With single AB, no error occurred. With multiple ABs, it was difficult to select ABs rapid enough. When I tried with groups of layers, and tried to dup many groups at once in succession, I could reproduce the error. I, however, think this is a marginal case, and the solution is acceptable.

Issue #3538, opened 2016-01-14
Using Shift modifier to marquee select layers results in errors


chadrolfs:
cmd/ctrl + n
select rect tool and drag out a shape
hit v for select tool, hold down Shift and marquee select the shape

Result: `Transfer from superselect.marqueeSelect to layers.select failed: Error: Layers.select is passed an invalid modifier`
and then a slew of subsequent errors

iwehrman:
FBNC to @chadrolfs.

PR #3539, opened 2016-01-14
Set select options correctly on marquee select

volfied:
Missed this one during my drag start latency work, oops.

#3538


iwehrman:
Derp, my fault for not catching it.

PR #3540, opened 2016-01-15
Search: placing libraries graphic results in error

shaoshing:
Fix: search for libraries graphic and hit return to place any graphic will result in error. 

I fixed this earlier but got distracted by a bluebird@3.0.6 bug (https://github.com/petkaantonov/bluebird/issues/841) which will silent unhandled errors and exceptions (error is not shown in the console). Now it has been fixed in 3.1.0 and I saw we are already using 3.1.1 in `npm-shrinkwrap.json`, so the problem no longer exists.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "webpack-notifier": "^1.2.1"
   },
   "dependencies": {
-    "bluebird": "^3.0.5",
+    "bluebird": "^3.1.1",
     "classnames": "^2.2.0",
     "d3": "^3.5.9",
     "fluxxor": "^1.7.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "webpack-notifier": "^1.2.1"
   },
   "dependencies": {
-    "bluebird": "^3.0.5",
+    "bluebird": "^3.1.1",
```
volfied: This shouldn't be here (or if you really wanna update bluebird, you also need to update shrinkwrap, but I prefer you do it separately)

volfied: Nvm, I should read the full description.

shaoshing:
wow, that was quick! thanks for merging.

PR #3544, opened 2016-01-16
Babel, generator package updates

iwehrman:
1. The shrinkwrapped version of generator-connection was behind the version in package.json by a minor step. This updates the shrinkwrap to fix that. I don't know why npm chooses a different format for the "from" line in the shrinkwrap file sometimes.... I think it's harmless, but maybe @mcilroyc knows.
2. The the `babel-loader` dependency has been switched from @volfied's fork back to the main repository. I noticed a few weeks ago that the main project had incorporated the small fix from @volfied's branch. I also updated and tested the other babel dependencies while I was in there.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,7 +9,7 @@
     },
     "classnames": {
       "version": "2.2.3",
-      "from": "classnames@2.2.3",
+      "from": "https://registry.npmjs.org/classnames/-/classnames-2.2.3.tgz",
       "resolved": "https://registry.npmjs.org/classnames/-/classnames-2.2.3.tgz"
     },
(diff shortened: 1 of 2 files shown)


volfied:
The from field updating to resolved field sounds like a cache problem where shrinkwrap updates from field with whatever it first resolved to if it's been already installed. One way to test is to shrinkwrap without running install first and see what happens. Either way, I think it's harmless.

Thank you for switching babel-loader. For record keeping, I had a fork of the project since original babel-loader was getting rid of path names during sourcemap creation and causing us to have a flat list of files. But since we changed our sourcemaps to a different format, this is no longer a problem and we can remove the forked dependency.

Merging.

PR #3545, opened 2016-01-18
Automate layer panel performance test 

shaoshing:
This PR adds the ability to run layer panel performance test automatically. The test is accessiable via the menu item `Help > Test Layer Panel Performance`. It works for the flat and hierarchical structure in the LayerPanel. Here is a video of how it works https://www.dropbox.com/s/slalhpmhjjf9a9o/layer-panel-performance-test.mov?dl=0

Benefits of adding the task:

* It includes most testing scenarios (layer selection, art board selection, collapse/expand/visibility), and we can add more in the future.
* It is easeir for us to keep monitoring the performance in each spring.
* It saves time!

PS: would it be a good idae to move the developer snippets into the menu item as well? Maybe under the menu item call `Snippets`?

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -267,6 +267,156 @@ define(function (require, exports) {
     };
 
     /**
+     * Run layer panel performance tests.   
+     *
+     * @private
+     * @return {Promise}
+     */
+    var layerPanelPerformanceTest = function () {
+        var flux = this.flux,
+            applicationStore = flux.store("application"),
+            document = applicationStore.getCurrentDocument(),
+            openDocuments = applicationStore.getOpenDocuments();
+
+        if (openDocuments.size !== 1 || !document.name.match(/vermilion/i)) {
+            window.alert(
+                "To run the performance test, the current document must be " +
+                "the Vermilion file, and there should be only one open document");
+            return Promise.resolve();
+        }
+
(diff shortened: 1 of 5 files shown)


volfied:
Fantastic. Can you add Windows menus as well?

volfied:
Re: Snippets, that's a great idea as well, we should try to find a way to make them toggle-able. Our snippets right now add listeners but don't remove them.

src/js/actions/menu.js
```diff
@@ -267,6 +267,160 @@ define(function (require, exports) {
...
+            artboardElement,
+            artboardIconElement,
+            artboardVisibilityElement,
+            delayBetweenTest = 1500;
+
+        return Promise
+            .bind(this)
+            .then(function () {
+                // Expand all artboards
+                document.layers.roots.forEach(function (root) {
```
iwehrman: It would be better to use promises for this than a `delay` at the end. That is, this should be mapped into an array of promises, and then `Promise.all` should be used to build a single promise that resolves when all the child promises resolve, and then that should be `.then`d with the promise from `deselectAll` below, etc.

iwehrman:
Re: menu item: I suggest creating a new top-level menu called "Debug" or something like that to hold all of our rando developer stuff.

shaoshing:
@volfied I added the menu item for windows too, and refactor the code to address @iwehrman 's comment. About the snippets, I will create a code-cleanup issue with your suggestions.

PR #3546, opened 2016-01-18
Break apart artboardAdders from Select Overlay

volfied:
This is part of the performance improvement work for drag / on canvas interaction. The PR, by itself, does not really affect much, but as we move forward, we can make some gains by avoiding redrawing artboard adders during operations that would not change the artboard bounds. Or we can show them quicker than super select bounds making the app feel faster during artboard drags.

It's also a code clean up, which is nice.

### DIFF
--- a/src/js/jsx/tools/ArtboardAdderOverlay.jsx
+++ b/src/js/jsx/tools/ArtboardAdderOverlay.jsx
@@ -1,0 +1,313 @@
+/*
+ * Copyright (c) 2016 Adobe Systems Incorporated. All rights reserved.
+ *  
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"), 
+ * to deal in the Software without restriction, including without limitation 
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
+ * and/or sell copies of the Software, and to permit persons to whom the 
+ * Software is furnished to do so, subject to the following conditions:
+ *  
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *  
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


src/js/jsx/tools/ArtboardAdderOverlay.jsx
```diff
@@ -0,0 +1,305 @@
+/*
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
iwehrman: 2016.

src/js/jsx/tools/ArtboardAdderOverlay.jsx
```diff
@@ -0,0 +1,305 @@
...
+        /**
+         * Calculates the new artboard bounds in the given direction
+         *
+         * @private
+         * @param {Bounds} bounds Artboard bounds to be copied
+         * @param {string} direction 
+         * @return {Bounds}
+         */
+        _getNewArtboardBounds: function (bounds, direction) {
+            var padding = 100,
```
iwehrman: I don't actually see a problem, but: should this padding be specified in `rems` instead of an absolute pixel dimension?

src/js/jsx/tools/ArtboardAdderOverlay.jsx
```diff
@@ -0,0 +1,305 @@
...
+
+        shouldComponentUpdate: function (nextProps, nextState) {
+            // We check for the cheaper flags before plucking layers from document models
+            if (this.state.vectorMaskMode !== nextState.vectorMaskMode ||
+                this.state.modalState !== nextState.modalState ||
+                this.props.transformString !== nextProps.transformString) {
+                return true;
+            }
+
+            var lastLayers = this.state.document.layers.selected,
```
iwehrman: I wonder if this can go further by saying something like: return false if there were no artboards selected in either the next or previous state.

iwehrman:
Looks good! Back to @volfied for a few questions.

iwehrman:
Oh, also: you marked this as performance. Can you measure any performance changes with this branch?

src/js/jsx/tools/ArtboardAdderOverlay.jsx
volfied: Padding is not related to screen drawing in this function, it's for checking to see if artboards we create on any side, with padding, will not intersect existing artboards.

volfied:
@iwehrman Made changes. I marked it as performance, but there are no performance gains in this PR itself. I will remove that invalid tag.

iwehrman:
Nice cleanup, thanks! And I think the new `sCU` can only help performance, though it may only be a tiny improvement.

Issue #3547, opened 2016-01-18
Marquee select causes unrecoverable error


ktaki:
Open new doc. Add a shape. Select superselect tool. Marquee select the shape layer on canvas.

![screen shot 2016-01-18 at 11 08 21 am](https://cloud.githubusercontent.com/assets/2029597/12400344/47741d9e-bdd4-11e5-845c-3b3835588beb.png)

![screen shot 2016-01-18 at 11 10 59 am](https://cloud.githubusercontent.com/assets/2029597/12400347/4b49bfaa-bdd4-11e5-8132-e664ee95751b.png)

![screen shot 2016-01-18 at 11 11 12 am](https://cloud.githubusercontent.com/assets/2029597/12400351/540c12f0-bdd4-11e5-92c7-4d078bd88372.png)

chadrolfs:
this was just fixed and looks broken again. but this is without Shift key, see related #3538

mcilroyc:
New PR merged

ktaki:
Verified the fix.

PR #3548, opened 2016-01-18
Pass valid modifier to select from marqueeselect

volfied:
Addresses #3547 

I made the modifier explicit, but forgot to set it during marquee select without shift.


mcilroyc:
ok!

Issue #3542, opened 2016-01-15
Unrecoverable error: collapsing group changes selection to uninitialized layer


iwehrman:
1. Open Vermilion.
2. Expand all layers.
3. Select only a leaf layer in the top-most artboard, e.g. "pad copy 4"
4. Reload Design Space.
5. Quickly option-click the icon of the top-most artboard to recursively collapse the artboard.

Result: unrecoverable error.

Hypothesis: upon collapsing the group, the selection is correctly changed to the top-most artboard, but that artboard has not yet been initialized. The action does initialize the newly selected layer, but not until after dispatching `SET_GROUP_EXPANSION`. That handler unsafely accesses uninitialized layer properties, which results in an error. But a) that's just my guess (the console error is inscrutable); and b) that doesn't explain why the error is unrecoverable.

iwehrman:
Here's the initial action log:
```

fluxcontroller.js:835 Enqueuing action panel.updatePanelSizes; 0/0
fluxcontroller.js:841 Executing action panel.updatePanelSizes after waiting 3ms; 1/0
fluxcontroller.js:835 Enqueuing action panel.updatePanelSizes; 1/0
fluxcontroller.js:835 Enqueuing action panel.updatePanelSizes; 1/1
main.js:135 Main component mounted: 324ms
fluxcontroller.js:835 Enqueuing action ui.beforeStartup; 1/2
fluxcontroller.js:835 Enqueuing action documents.beforeStartup; 1/3
fluxcontroller.js:835 Enqueuing action policy.beforeStartup; 1/4
fluxcontroller.js:835 Enqueuing action search.beforeStartup; 1/5
fluxcontroller.js:841 Executing action search.beforeStartup after waiting 0ms; 2/5
fluxcontroller.js:835 Enqueuing action keyevents.beforeStartup; 2/5
fluxcontroller.js:841 Executing action keyevents.beforeStartup after waiting 0ms; 3/5
fluxcontroller.js:835 Enqueuing action libraries.beforeStartup; 3/5
fluxcontroller.js:841 Executing action libraries.beforeStartup after waiting 1ms; 4/5
fluxcontroller.js:835 Enqueuing action menu.beforeStartup; 4/5
fluxcontroller.js:841 Executing action menu.beforeStartup after waiting 1ms; 5/5
fluxcontroller.js:835 Enqueuing action guides.beforeStartup; 5/5
[...]

PR #3552, opened 2016-01-18
Format non-finite values in NumberInput

iwehrman:
Currently `NumberInput` throws if its value is a non-finite number. It's hard to prevent this from ever happening, so instead I'm just changing `NumberInput` to format a non-finite number as the empty string instead of throwing.

This is vaguely related to #3542.


volfied:
I think we should put a log.warn in there to track the cause issues.

volfied:
@iwehrman looks great, but I'd like to leave a note in there to remind us something gone wrong.

iwehrman:
Fixed. Back to @volfied.

Issue #3536, opened 2016-01-14
Scrubbing X/Y values results in layer jumping if layers exist outside AB (auto-canvas calculation)


chadrolfs:
new doc, create shapes inside and outside the artboard
select a shape inside the artboard and scrub X or Y

Result: layer jumps to outside the artboard

chadrolfs:
![coord-jump](https://cloud.githubusercontent.com/assets/8442631/12338518/d315b622-bac4-11e5-8676-a37ebacefc07.gif)

chadrolfs:
this can subsequently end up in disconnecting the layer from it's position on canvas and where it displays on canvas
![layer-position](https://cloud.githubusercontent.com/assets/8442631/12339639/e8154096-baca-11e5-81b2-f34e49d3da39.gif)

iwehrman:
FBNC to @chadrolfs.

PR #3555, opened 2016-01-18
For scrubbing position, use absolute values

volfied:
While scrubbing, we would wrongfully pass the `coalesce` flag as the `translate` flag into `_getMoveLayerActions`, this was causing all scrubs besides the first one to be absolute, and cause the issue seen in #3536 

To fix this issue, I first fixed the flag being passed while scrubbing. However, we still had the problems that if we set absolute position while scrubbing, we would get huge jumps as the layer is being scrubbed across artboard boundaries. So we now use the absolute bounds of the layer to calculate where to scrub it next, which correctly scrubs the layer across boundaries.

Addresses #3536


iwehrman:
LGTM!

Issue #3541, opened 2016-01-15
Cancel quit results in errors and bad state when resuming work


chadrolfs:
cmd/ctrl + n
drag out a guide to dirty the doc
(repeat steps so you have 2 dirty doc)
cmd/ctrl + q
at first prompt to save, click Cancel
then click on Artboard title and try to move it

Result: `Transfer from superselect.click to layers.select failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25922`
lots of other subsequent errors and bad state - can't continue

volfied:
Correction to this bug: It doesn't happen if you click cancel at first prompt, but happens if you don't save first, and cancel second. Seems like PS gathers all the don't saves at first instead of closing documents one by one.

volfied:
Ah, should be a simple fix, PS is in modal state, and sends us a close event later on, but we never handle the close event.

iwehrman:
FBNC to @chadrolfs.

PR #3557, opened 2016-01-19
Handle cmd+q quit gracefully, and a new action flag to prevent reset on failure

volfied:
This PR is written to address #3541, but also implements a new action property. 

When user tries to quit Photoshop with dirty documents open, Photoshop pops up a modal dialog asking user if they want to save, not save, or cancel. There are multiple pathways this can take based on current open documents:

 - Asterisks imply dirty document which will pop the prompt):
 - Cases I don't account for are already handled well / are sub cases

## *Untitled-1 | *Untitled-2

 1. Don't Save Untitled-2 | Cancel Untitled-1

In this case we get a close event from Photoshop with saving:no field for Untitled-2, and nothing else. Which at this point we dispose of the Untitled-2 document in our models and continue

 2. Save Untitled-2 | Cancel Untitled-1

In this case, PS sends us a close event with saving:yes, a save event, and then another close event with saving:no. This is very unexpected, and the first event is equivalent of the next two events. **So as a solution, we ignore `close` events with saving:yes**

## Reason for `allowFailure` flag

This flag was added to action objects, because of this scenario, where it's done quick enough:

1. Create a new document, dirty it (Untitled-1)
2. Create a new document (Untitled-2)
3. Create a new document, dirty it (Untitled-3)
4. Quit Photoshop and click Don't Save really fast

In this scenario, since `initializeLayersBackground` for Untitled-3 is run `whenIdle`. And because Quit dialog is modal, by the time we cancel, Untitled-3 is already gone from Photoshop, the layer get calls to Photoshop fail with an `Unknown exception type`. 

So this error is expected, but should not reset the controller, since it doesn't affect anything. Close events later get emitted to us, where we dispose of our models correctly.

**Side note:** I wanted to keep `allowFailure` as simple and un-intrusive as possible, so during transfers, the parent's flag gets passed down to the transferred functions. This allows us to specify only one action to not reset on failure instead of the entire transfer chain.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -956,6 +956,7 @@ define(function (require, exports) {
      */
     var _makeHandler,
         _openHandler,
+        _closeHandler,
         _selectHandler,
         _saveHandler,
         _pasteHandler,
@@ -1002,6 +1003,21 @@ define(function (require, exports) {
             }
         }.bind(this);
         descriptor.addListener("open", _openHandler);
+
+        _closeHandler = function (event) {
+            if (typeof event.documentID === "number") {
+                // If this document is saved during close, we get both a close event and a save event, 
+                // and then another close event for the closing of the non dirty document. So we ignore the first event
+                var documentSaved = (event.saving && event.saving._value === "yes");
+
+                if (!documentSaved) {
+                    this.flux.actions.documents.disposeDocument(event.documentID, false);
+                }
+            } else {
+                throw new Error("Document closed with no ID");
+            }
+        }.bind(this);
+        descriptor.addListener("close", _closeHandler);
 
         _selectHandler = function (event) {
             var nextDocument,
@@ -1142,6 +1158,7 @@ define(function (require, exports) {
     var onReset = function () {
         descriptor.removeListener("make", _makeHandler);
         descriptor.removeListener("open", _openHandler);
+        descriptor.removeListener("close", _closeHandler);
         descriptor.removeListener("select", _selectHandler);
         descriptor.removeListener("save", _saveHandler);
         descriptor.removeListener("paste", _pasteHandler);
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -746,7 +746,8 @@ define(function (require, exports) {
         reads: [locks.JS_DOC],
         writes: [],
         transfers: [initializeLayers],
-        modal: true
+        modal: true,
+        allowFailure: true
     };
 
     /**
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -196,6 +196,15 @@ define(function (require, exports, module) {
     FluxController.prototype._uiLocked = false;
 
     /**
+     * Indicates whether or not failures in actions should 
+     * be silenced
+     *
+     * @private
+     * @type {boolean}
+     */
+    FluxController.prototype._allowFailure = false;
+
+    /**
      * The set of pending idle task promises.
      *
      * @private
@@ -508,7 +517,7 @@ define(function (require, exports, module) {
 
                                 log.error(message, errMessage);
 
-                                // Any failed transfer triggers a complete controller reset
+                                // Failed transfers trigger a controller reset (unless parent had allowFailure set)
                                 this._resetController(err);
                                 throw err;
                             });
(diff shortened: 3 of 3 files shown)


src/js/actions/documents.js
```diff
@@ -1003,6 +1004,21 @@ define(function (require, exports) {
         }.bind(this);
         descriptor.addListener("open", _openHandler);
 
+        _closeHandler = function (event) {
+            if (typeof event.documentID === "number") {
```
iwehrman: It's almost always better for these handlers to be actions. Is there a good reason for this not to be an action?

src/js/actions/layers.js
```diff
@@ -740,13 +740,19 @@ define(function (require, exports) {
                 if (firstUninitializedLayers.size === 50) {
                     this.whenIdle(initializeLayersBackground, document);
                 }
+            })
+            .error(function (err) {
+                /* jshint multistr: true */
+                log.error("InitializeLayersBackground action failed for document %s, \
```
iwehrman: Nit: I think it would be better to just concatenate two shorter strings here instead of including a JSHint directive, especially since we'd like to get rid of JSHint anyway... 

Also, we use `.catch` elsewhere. Is `.error` a synonym? If so, let's stick with `.catch` for consistency.

src/js/actions/layers.js
```diff
@@ -740,13 +740,19 @@ define(function (require, exports) {
...
+                    it was likely closed by Photoshop - %s", document.name, err.message);
             });
     };
     initializeLayersBackground.action = {
         reads: [locks.JS_DOC],
         writes: [],
         transfers: [initializeLayers],
-        modal: true
+        modal: true,
+        noResetOnFailure: true
```
iwehrman: How about calling this just `allowFailure`? I think it's better not to try to describe exactly what will happen in this case in case we want to adjust it somewhat later on.

src/js/fluxcontroller.js
```diff
@@ -860,8 +874,10 @@ define(function (require, exports, module) {
 
                         log.error("Action " + actionName + " failed:", message);
 
-                        // Reset all action modules on failure
-                        this._resetController(err);
+                        if (!action.action.noResetOnFailure) {
+                            // Reset all action modules on failure
+                            this._resetController(err);
+                        }
                         throw err;
```
iwehrman: It's not obvious to me that you want to throw the err here even if `allowFailure` is set...

src/js/fluxcontroller.js
```diff
@@ -489,6 +493,10 @@ define(function (require, exports, module) {
                                 transferQueue.active(), transferQueue.pending());
                         }
 
+                        if (action.action.noResetOnFailure) {
+                            nextAction.action.noResetOnFailure = action.action.noResetOnFailure;
```
iwehrman: Ew. I do not at all like modifying the action object at runtime. I think it would be better to set a flag in the `FluxController` instance the same way we do for `lockUI`, etc.

iwehrman: I also think that it probably _should_ be the case that `allowFailure` is transitive across the transfer relation. Transfers should be transparent to the caller. I don't know why we would want to allow a `batchPlay` to fail but not a transfer. Often we clean actions up by modularizing them into transferrable actions, and I don't think that should change the semantics of the action.

iwehrman:
Here's an alternative suggestion, for discussion: maybe `whenIdle` should imply `allowFailure`?

src/js/actions/documents.js
volfied: We are calling an action, after filtering out some of the events. And the only reason we're filtering out events is due to buggy behavior in PS. I didn't want to implement an action around that behavior.

src/js/actions/layers.js
volfied: http://bluebirdjs.com/docs/api/error.html

I specifically wanted to log on operationalErrors fail. If you still think that's a bad idea, I will move to `catch`.

volfied:
I don't think that's a good idea, especially with `allowFailure` change we're quietly failing. Someone not familiar with the code may use `whenIdle` and not see bugs this way.

volfied:
@iwehrman I simplified/rebased. We now silently continue on errors by initializeLayersBackground.

iwehrman:
Much simpler, I like it. Thanks!

Issue #3550, opened 2016-01-18
Color Picker offset (if not enough bottom space) is now at top and scrolling explodes the Color Picker controls


chadrolfs:
new doc, drag out a shape, apply some effects, resize app window so there is lack of space at bottom
click on color picker of one of the top effects (so that there is an effect below it)

Bug 1: color picker opens at the top of the Properties panel, it should be just above the color picker swatch you click on.

next step: scroll down the effects panel and mouse over an attribute field

Bug 2: the color picker controls spread out (explode)

chadrolfs:
![color-picker-offsetscroll](https://cloud.githubusercontent.com/assets/8442631/12401898/55be9640-bdde-11e5-8cfb-d58a46178675.gif)

iwehrman:
FBNC to @chadrolfs.

Issue #3440, opened 2015-12-18
Clipped layer effect swatch + incorrect color picker position


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/11904125/dfc59354-a572-11e5-97cc-b41c69d44525.png)

To repro, reduce window height so that the second effect needs slightly more space than is available in the effects panel.

iwehrman:
Related #3550.

volfied:
I'm closing this as clipping no longer occurs (unless the swatch is at the bottom and it's the Export panel divider that clips it). Other issue is being tracked in #3550

iwehrman:
Over to @chadrolfs for testing + verification.

PR #3568, opened 2016-01-19
Add bottom dialog pointer to color picker and fix positioning

volfied:
Using [CSS Mixin Guards](http://lesscss.org/features/#mixin-guards-feature), we now position the dialog pointer either at the top or the bottom of color picker dialog.

Also, addressing #3550, we now position the color picker correctly when there is no room below the clicked color input and color picker appears above it.

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -254,17 +254,16 @@ define(function (require, exports, module) {
                         offsetParentBounds = this._target.offsetParent.getBoundingClientRect(),
                         placedDialogTop = Math.round(targetBounds.bottom) - Math.round(offsetParentBounds.top),
                         placedDialogBottom = placedDialogTop + Math.round(dialogBounds.height),
+                        pointerAbove = false,
+                        pointerBelow = false,
                         topPosition,
-                        bottomPosition,
-                        placedAbove;
+                        bottomPosition;
                     
                     // If there is space below the target, place the dialog below.
                     if (placedDialogBottom <= clientHeight) {
-                        placedAbove = false;
+                        pointerAbove = true;
                         topPosition = placedDialogTop + "px";
                     } else {
-                        placedAbove = true;
-                        
                         // Need to account for element margin
                         var dialogComputedStyle = window.getComputedStyle(dialogEl),
                             dialogMarginTop = math.pixelDimensionToNumber(dialogComputedStyle.marginTop),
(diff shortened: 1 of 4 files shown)


src/style/panel.less
```diff
@@ -383,7 +383,7 @@
     width: 100%;
     
     &:before {
-        .dialog-pointer(5.4rem);
+        .dialog-pointer(5.4rem, true);
```
volfied: I have to add this because there are no default parameters with less mixins.

src/style/shared/dialog.less
```diff
@@ -190,10 +190,20 @@ dialog::backdrop {
     width: 100%;
 }
 
-.dialog-pointer(@left) {
+.pointerMixin (@atTop) when (@atTop) {
```
volfied: This is equivalent to `@atTop === true` in JS

src/style/shared/dialog.less
```diff
@@ -190,10 +190,20 @@ dialog::backdrop {
...
+    top: -0.9rem;
+    transform: rotate(45deg);
+}
+
+.pointerMixin (@atTop) when (default()) {
```
volfied: This is all the cases not accounted for `.pointerMixin`

src/style/shared/dialog.less
```diff
@@ -190,10 +190,20 @@ dialog::backdrop {
...
+    bottom: -0.9rem;
+    transform: rotate(-135deg);
+}
+
+.dialog-pointer(@left, @atTop) {
+    .pointerMixin(@atTop);
```
volfied: And this is how you call a prepared mixin, adding the properties it calculates into the class.

iwehrman:
Here's a failure case involving a short window:
image

iwehrman:
Note that in the case above I think it should just show neither pointer.

volfied:
@iwehrman thanks for the catch, I changed the class names to be more explicit and fixed the issue so it doesn't draw any pointers if it's going to float.

src/js/jsx/shared/Dialog.jsx
```diff
@@ -358,8 +360,10 @@ define(function (require, exports, module) {
         render: function () {
             var children,
                 globalClass = (this.props.position === POSITION_METHODS.CENTER) ? "dialog__center" : "dialog__target",
+                // We reverse the pointerAbove/Below classes as they remove the dialog pointer rather than add it
                 classes = classnames(globalClass, this.props.className || "",
-                    { "placed-above": this.state.placedAbove }),
+                    { "pointer-above": !this.state.pointerBelow },
+                    { "pointer-below": !this.state.pointerAbove }),
```
iwehrman: Hmm, this is kind of confusing. Is it possible for these two sets of names to be more in sync?

volfied: Fixed it, thanks.

volfied:
Back to you, @iwehrman

iwehrman:
image

volfied:
After a remote debugging session, we realized this issue happens when the swatch is half rendered outside the panel bounds (bottom one in the screenshot). Making a note here for myself.

iwehrman:
A fix for this may also address #3440.

volfied:
@iwehrman I can't reproduce this again, I've tried at all window sizes / color picker locations.

iwehrman:
Hmm, I also can't repro at the moment. I guess that means it's fixed!

Issue #3567, opened 2016-01-19
Less stops re-building after first error on `grunt debug`


volfied:
Which leaves you wondering why your less changes are no longer appearing.

iwehrman:
Fixed by #3569.

PR #3569, opened 2016-01-19
Reload style watcher on changes so fails don't stop it

volfied:
Addresses #3567, re-starting less watcher every time files change so it doesn't stop working after an error.

Since less does not do incremental re-build, there is no loss from restarting it.


iwehrman:
Yeah!

Issue #3558, opened 2016-01-19
Descenders are clipped a little in Layer names on Retina


designedbyscience:
So close.

<img width="72" alt="vermilionartboards_psd___16_7___image__rgb_8____" src="https://cloud.githubusercontent.com/assets/22372/12427085/26024f28-beab-11e5-9611-784a217c8f3b.png">

iwehrman:
CC @shaoshing: maybe this was a real issue? See discussion in #3553.

iwehrman:
@designedbyscience: can you confirm that if you restart `grunt debug` that this still happens?

designedbyscience:
Yes, double-checked. Looks good on non-Retina.

shaoshing:
Looking at it now.

iwehrman:
I still can't repro, but I do see the problem in the Libraries panel:
![image](https://cloud.githubusercontent.com/assets/1075154/12430460/98cce506-bea4-11e5-8f74-77d9b14b62fd.png)

shaoshing:
Confirmed the issue in small screen resolution (seeing this in 15-inch mbp)

iwehrman:
(Note that we don't mark a bug as FBNC until the fix is merged into main.)

designedbyscience:
@shaoshing fixed it real good.

PR #3571, opened 2016-01-20
Layer Panel: fix style issues

shaoshing:
This PR fix the following style issues:

#3558 Descenders are clipped a little in Layer names on Retina. Happens on some layer elements when `font-size` is `@base-8` (when `device-width <= 1440px`). 

screen shot 2016-01-19 at 5 29 40 pm
(the first line is not clipped, the sconed one is clipped slightly)

Another issue I found was that when element's height has decimal  (e.g. 25.1875px), their layout may be inconsistent in the screen. The picture below demonstrate the problem.

after
(when face's height is 25.1875px, the two folder icons has different relative position)

after-fixed
(If change height to 25px, everything looks fine.)

As a result, when double-click to edit title, some input will shift up by 1px.
layer-panel-offset

I cannot find an easy way to fix this, except for using the `@media` rule to tweak the face's rem height for a rounded pixel number in different screen resolution. @designedbyscience can you review and let me know if you will approve the solution? :hammer:

### DIFF
--- a/src/style/sections/layers/face.less
+++ b/src/style/sections/layers/face.less
@@ -82,14 +82,10 @@
 
 input[type="text"].face__name {
     font-size: 1.2rem;
-    color: @item-active;
+    color: @item;
     margin-top: -0.1rem;
     width: 100%;
-}
-
-input[type="text"].face__name:focus {
     border-bottom: solid @hairline @focus-highlight;
-    color: @item;
     -webkit-user-select: text;
 }
 
@@ -118,16 +114,6 @@ input[type="text"].face__name:focus {
     .face__name, 
     &:hover .face__name {
         color: @item;
-    }
-    
-    input[type="text"].face__name {
-        border-bottom: solid @hairline transparent;
-    }
-
-    input[type="text"].face__name:focus  {
-        border-bottom: solid @hairline @focus-highlight;
-        color: @item;
-        -webkit-user-select: text;
     }
 }
 
--- a/src/style/sections/layers/face.less
+++ b/src/style/sections/layers/face.less
@@ -73,7 +73,7 @@
     flex-grow: 1;
     flex-shrink: 1;
     -webkit-user-select: none;
-    white-space: nowrap;
+    white-space: pre;
     text-overflow: ellipsis;
     overflow-x: hidden;
     height: 2rem;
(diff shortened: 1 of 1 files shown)


shaoshing:
Another thing I noticed is that, Chrom is not using the correct font-size.

font-size
(`font-size: 50%` should equals to 8px, but it is 9px in the HTML element)

font-size-2
(Try change to 44%, but it is still using 9px)

designedbyscience:
@shaoshing Still looking, but there is still some clipping on the highlighted layer name (on Retina, you can see it in your gif above actually)

shaoshing:
@designedbyscience regarding text got clipped in the input, I noticied it yesterday and was doing some research.

First, I traced back to where we were using the flat structure, and found the exact same issue:

before - 9 6px
(input font-size is 9.6 px)
before - 9 6px editing
(in editing mode)

Then I tried different `font-size` in latest master, and found it happens when the input has the `font-size` between 9 and 10 plus when it is using the `Adobe Clean` type face.

##### `font-size: 8px` input text is correctly displayed
8px

##### `font-size: 9.6px` input text is clipped
9 6px

##### `font-size: 10px` input text is clipped
10px

##### `font-size: 11px` input text is correctly displayed
11px

##### `font-size: 9.6px` with default system font, input text is correctly displayed
9 6 px - regular font

Unfortunately I couldn't find any CSS attribute that controls the bottom space in text input (I tried `padding/box-sizing/line-height/height ...`). I am wondering if you have fought with this issue before, and do you have any suggestion on how we can resolve or bypass it?

src/style/sections/layers/face.less
```diff
@@ -73,22 +73,19 @@
     flex-grow: 1;
     flex-shrink: 1;
     -webkit-user-select: none;
-    white-space: nowrap;
+    white-space: pre;
```
shaoshing: This is changed to `pre` to preserve multiple spaces in layer name.

shaoshing:
Update: @designedbyscience I've added fix for the jump issue on non-retina screen.

designedbyscience:
Woohoo! Looks good. 

a-b-yay

placegraphichere:
this is great, thanks for diving in and fixing this @shaoshing

Issue #3502, opened 2016-01-12
Layer dragging enabled during rename


iwehrman:
Layer dragging is enabled while renaming the layer, but it shouldn't be because that conflicts with the text-selection dragging:

![img](https://s3.amazonaws.com/f.cl.ly/items/33263L1D392E2M0g0204/Screen%20Recording%202016-01-12%20at%2012.13%20PM.gif?v=c0ce36b3)

iwehrman:
Fixed by #3572

PR #3572, opened 2016-01-20
Disable drag-and-drop when editing layer name.

shaoshing:
Fix for issue #3502

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -71,16 +71,14 @@ define(function (require, exports, module) {
                 isDropTarget: false,
                 dropPosition: null,
                 isDragging: false,
-                dragStyle: null
+                dragStyle: null,
+                isEditing: false
             };
         },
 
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -153,6 +151,12 @@ define(function (require, exports, module) {
             this.getFlux().actions.groups.setGroupExpansion(currentDocument, currentLayer,
                 !currentLayer.expanded, descendants);
         },
+        
+        _handleLayerNameInput: function (focused) {
```
volfied: This requires JSDoc. I would also name it something more explicit about what it is doing. It's confusing, to me at least, to see a function name handleLayerNameInput which does nothing with the layer name being inputted.

volfied:
One nit, looks great otherwise!

shaoshing:
Thanks for reviewing. I added JSDoc and switched to a better name for the focus event handler (`_handleInputFocusChange`). Back to you @volfied

volfied:
Let'er rip!

Issue #3554, opened 2016-01-18
Size, Position and Radius shouldComponentUpdate are broken


iwehrman:
All return `this.state !== nextState`. This does not do what it seems like it should do. In particular, if the same object is used for state then this will always return true, even if the properties have changed. And, if a new object is used for state then this will always return false even if all the properties are the same. I don't know of any particular bug that this causes, but at some point it almost certainly will cause one!

iwehrman:
Tentatively over to @volfied since I _think_ he wrote these lines in the first place. (My `git blame` reading could be wrong, though.)

volfied:
@iwehrman I've written it this way, because we calculate the state out of props at `componentWillReceiveProps`. Both components do not use any of the props to render, and rely on state to render everything. I'll think about it a bit more, but it was a conscious decision on my part which I ran by React people.

iwehrman:
I don't get it. Depending on the implementation of `setState` this will always return true or it will always return false. Remember that state is a mutable object. This would be a great implementation if state were an `Immutable` object.

chadrolfs:
If this is a case where it would show up to users as a case where the object appearance on canvas is separate from it's location (coordinates/highlight) then I've hit this recently a few times but haven't been able to reliably repro.  Taki has hit it as well but can't reliably repro.  Not sure if this is would cause this or if that's a separate issue.

iwehrman:
This would only manifest as incorrect values in the Size, Position or Radius controls in the transform panel.

iwehrman:
Fixed by #3573.

PR #3573, opened 2016-01-20
Refined SCU for Size, Position and Radius checking each state member

volfied:
Addresses #3554, correctly implementing `shouldComponentUpdate` for Size, Position and Radius components.

### DIFF
--- a/src/js/jsx/sections/transform/Position.jsx
+++ b/src/js/jsx/sections/transform/Position.jsx
@@ -53,7 +53,12 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            return this.state !== nextState;
+            return !this.state ||
+                this.state.disabled !== nextState.disabled ||
+                !Immutable.is(this.state.xValues, nextState.xValues) ||
+                !Immutable.is(this.state.yValues, nextState.yValues) ||
+                !Immutable.is(this.state.absoluteXValues, nextState.absoluteXValues) ||
+                !Immutable.is(this.state.absoluteYValues, nextState.absoluteYValues);
         },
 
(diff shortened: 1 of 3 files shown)


src/js/jsx/sections/transform/Position.jsx
```diff
@@ -53,7 +53,12 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            return this.state !== nextState;
+            return !this.state ||
```
shaoshing: Do you think `getInitialState` will handle this better? Then you don't need the state checker here. Also, if the state is never set (just guessing), the component will always re-rendered upon receiving new props.

BTW, I see other state like `scrubX` `scrubY` and they are not check in the SCU, is it intentional?

src/js/jsx/sections/transform/Size.jsx
```diff
@@ -47,7 +47,12 @@ define(function (require, exports, module) {
         shouldComponentUpdate: function (nextProps, nextState) {
             // Calculations that are usually done here are done in
             // componentWillReceiveProps
-            return this.state !== nextState;
+            return !this.state ||
```
shaoshing: Same question about the state existence checker.

src/js/jsx/sections/transform/Position.jsx
volfied: I'll give getInitialState a try, when I used it before, it was leaving the component unrendered, but that might have been the side effect of using `===`

`scrubX/scrubY` are states that don't affect rendering, and everytime we scrub, component receives new props to re-render. So it is intentional.

iwehrman:
Looks great, thanks!

PR #3574, opened 2016-01-20
Embed less sourcemaps with correct filepath

volfied:
`outputSourceFiles` embeds the files into the sourcemap, but `sourceMapRootpath` actually tells it where to find the real files by going from `www/build` to `www`


volfied:
This doesn't correctly work due to us using `import (multiple)` less directive. All the files but main and reset.less get imported multiple times with different color_stop class names, which causes source maps to fail and map everything to `main.less`.

I think as a step forward, this is ok to merge, and makes the problem more visible where I hope someone with more `less` skills can solve.

designedbyscience:
Makes sense to me to merge

Issue #3561, opened 2016-01-19
debug Help menu command > Test Action Failure... leaves you in unrecoverable states


chadrolfs:
DS should recover after running this.

volfied:
@chadrolfs please confirm the fix.

PR #3577, opened 2016-01-20
Anticipate null documents in SuperselectOverlay and ArtboardAdderOvelay

iwehrman:
Recent overlays changes caused them to throw in `shouldComponentUpdate` when the document state is null. This would happen when the controller resets, which causes all unhandled errors to be unrecoverable errors. The intent of this change is just to handle that additional case more gracefully.

Addresses #3561.


volfied:
Oops, my bad!

Issue #3510, opened 2016-01-13
Opening multiple .PSDs results in errors


chadrolfs:
double-click empty area of workspace to invoke open dialog
multi-select several pre-saved .PSD docs

Result: `Transfer from type.initFontList to layers.resetLayers failed: Error: Could not amend history, document not properly initialized: 2073`

several `Failed to deregister dialog`

`Uncaught TypeError: Cannot read property 'call' of undefined`

chadrolfs:
![open-mult](https://cloud.githubusercontent.com/assets/8442631/12281273/8fbc6b2c-b949-11e5-8940-e431109ed15e.gif)

chadrolfs:
Note: if I open up each file one by one, I get no errors.

chadrolfs:
On Windows, I get 
![ice_screenshot_20160112-164444](https://cloud.githubusercontent.com/assets/8442631/12281463/37a6b1c0-b94b-11e5-8367-ded108e8bf0d.png)

DivyaPrabhakar:
@chadrolfs I just merged @volfied's PR addressing this issue.

chadrolfs:
Yusssss!  Fixed.  Seeing these warnings but that's ok  `[History] Could not amend history, document history not found`

PR #3578, opened 2016-01-20
Don't handle a history event if document is not current

volfied:
When we open multiple documents at once, we get multiple `open` and `historyState` events. When we try to handle the `historyState` events for the non-active documents, because they're not allocated/initialized yet, `_handleHistoryAmendment` fails. With this PR, we check to make sure the history we're amending is of the active document.

This works for the other documents, because during allocation we initialize history. Addresses #3510

### DIFF
--- a/src/js/actions/type.js
+++ b/src/js/actions/type.js
@@ -726,6 +726,7 @@ define(function (require, exports) {
      */
     var initFontList = function (force) {
         var fontStore = this.flux.store("font"),
+            historyStore = this.flux.store("history"),
             fontState = fontStore.getState(),
             initialized = fontState.initialized;
 
@@ -761,7 +762,7 @@ define(function (require, exports) {
                     .getOpenDocuments()
                     .filter(function (document) {
                         // Skip uninitialized documents
-                        return document.layers;
+                        return document.layers && historyStore.hasInitializedHistory(document.id);
                     })
                     .map(function (document) {
                         var layers = document.layers.all,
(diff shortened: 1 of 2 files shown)


iwehrman:
It's unclear to me that it is safe to ignore these events for non-current documents. Is it possible to get a history event for an initialized but recently deactivated document? Any thoughts, @mcilroyc?

mcilroyc:
I'm a little confused. We already _try_ to ignore history states events for inactive documents.  see [here](https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/actions/history.js#L349-L352). That action dispatches a `events.history.PS_HISTORY_EVENT` which is _not_ handled by `_handleHistoryAmendment`, but which is what you modified in this PR.  @volfied Can you set me straight?  For this issue, what action is causing `_handleHistoryAmendment`?

volfied:
I will get back to you guys with a detailed cause of the issue and how my fix helps. Or figure out why my fix is not right and correct it.

volfied:
@iwehrman and @mcilroyc based on our discussion, I've made the changes so we skip resetLayers for documents with no history, also `_handleHistoryAmendment` only warns when a document with no history is trying to amend history.

src/js/actions/type.js
```diff
@@ -726,6 +726,7 @@ define(function (require, exports) {
      */
     var initFontList = function (force) {
         var fontStore = this.flux.store("font"),
+        historyStore = this.flux.store("history"),
```
mcilroyc: +1 tab stop

mcilroyc:
This looks right and safe to me.  one spacing nit.

Issue #3579, opened 2016-01-20
(debug) Test Reset Failure... appears twice under the help menu on Windows


chadrolfs:
Test Reset Failure... appears twice under the help menu on Windows and Test Transfer Failure... is missing so I'm guessing one of those two is labeled wrong

PR #3580, opened 2016-01-20
Fix a duplicate menu entry on Windows.

iwehrman:
There are two `Help > Test Reset Failure` entries on Windows. One should instead be `Help > Test Transfer Failure`.

Addresses #3579.

Over to @chadrolfs to test and merge.


chadrolfs:
Everything looks in order and functions as expected.

Issue #3563, opened 2016-01-19
Size scrubbing with no selected layers fails


iwehrman:
```
Action transform.setSize failed: Error: value conversion failed: Argument 1 could not be converted. Error: v8 numeric value that is 'not a number'. There is no Cef representation for this value type
    at Descriptor._batchPlayImmediate (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:1067:22)
    at Descriptor.batchPlay (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:1200:26)
    at Descriptor.batchPlayObjects (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:1225:22)
    at Descriptor.playObject (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:1048:22)
    at Object.setSize (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:75238:39)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:60076:45)
From previous event:
    at FluxController._applyAction (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:60075:15)
    at null.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:60165:30)
    at AsyncDependencyQueue._processNext (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:79603:28)
    at AsyncDependencyQueue.push (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:79449:15)
    at Object.<anonymous> (file:///Users/ian/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:60159:43)
[...]

PR #3582, opened 2016-01-20
Fix scrubbing on no layers / failures

volfied:
#3563 is addressed with dcb4575, where we set the scrub value to null so we skip scrubbing it.

Either way, when scrub failed, the scrubby state would never be reset, causing extra mouse moves to keep things broken. 2a2cd70 fixes that by listening to `"reset"` events from application store, which gets emitted when controller resets.


iwehrman:
LGTM!

Issue #3584, opened 2016-01-20
Closing a doc in the vector mask mode will not exit the mode


ktaki:
Open a AB doc. Select superselect tool. Select the AB. Hit M. Close the doc. (Note that you are still in the mode). Change the tool to a shape tool.

![screen shot 2016-01-20 at 3 45 26 pm](https://cloud.githubusercontent.com/assets/2029597/12467056/59e1ad42-bf8d-11e5-873e-77c3e8161596.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3587, opened 2016-01-21
exit Vector Mask Mode on Document Close

pineapplespatula:
fix for issue #3584

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -669,23 +669,26 @@ define(function (require, exports) {
                 interactionMode: descriptor.interactionMode.DISPLAY
             };
 
-        return this.transfer("panel.cloak")
-            .bind(this)
-            .then(function () {
-                return descriptor.playObject(closeObj, playOptions);
-            })
-            .then(function (result) {
-                var isDocumentSaved = result.saving && result.saving._value === "yes";
-                
-                return this.transfer(disposeDocument, document.id, isDocumentSaved);
-            }, function () {
-                // the play command fails if the user cancels the close dialog
-            });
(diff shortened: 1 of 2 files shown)


src/js/actions/documents.js
```diff
@@ -680,12 +680,15 @@ define(function (require, exports) {
                 return this.transfer(disposeDocument, document.id, isDocumentSaved);
             }, function () {
                 // the play command fails if the user cancels the close dialog
+            })
+            .then(function () {
+                return this.transfer(toolActions.changeVectorMaskMode, false);
```
iwehrman: Does this really need to wait for all the other operations to finish? It seems like it can happen concurrently with everything else and be start immediately with the action.

pineapplespatula:
@iwehrman :recycle:

src/js/actions/documents.js
```diff
@@ -669,6 +669,8 @@ define(function (require, exports) {
                 interactionMode: descriptor.interactionMode.DISPLAY
             };
 
+        this.transfer(toolActions.changeVectorMaskMode, false);
```
iwehrman: We still need to block the close action from completing until this transfer finishes. My previous nit was that we were executing this sequentially and only after everything else had finished. Instead we should execute transfers as early as possible and with as much concurrency as possible, but still block the action until all have completed.

iwehrman: So, you might want to do something like:

```javascript
var p1 = this.transfer("a"),
    p2 = this.transfer("b");

return Promise.join(p1, p2);
```

pineapplespatula:
@iwehrman :repeat:

iwehrman:
Looks good, thanks!

Issue #3532, opened 2016-01-14
Search and Export document header buttons are broken


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/1g421J3t2z1x0P3w3j0C/Screen%20Recording%202016-01-14%20at%2011.24%20AM.gif?v=ff77c5d1)

pineapplespatula:
@DivyaPrabhakar  and I couldn't repro this. do you have any other steps?

iwehrman:
CC @chadrolfs: I think you saw this too, can you help?

DivyaPrabhakar:
@chadrolfs @pineapplespatula @iwehrman This was my fault -- I needed to update my bluebird version. Now I see the issue. Off to fix it!

iwehrman:
Hmm... do you think this is somehow related to the updated version of Bluebird?

DivyaPrabhakar:
@iwehrman What is happening is that upon initialization, in dialog.jsx we add an event listener to handle a window click if the parent component passes in a prop called dismissOnWindowClick. Therefore when we click on any of the document header icons, part of the `_handleWindowClick` function is to call the `toggle` dialog function.

iwehrman:
You said earlier that you couldn't see the problem before you upgrade your version of bluebird, though. Is it true that this works correctly with an older version of bluebird and not with the latest version?

DivyaPrabhakar:
@iwehrman -- Yes that is true.

DivyaPrabhakar:
Update on research for this issue: 

When any dialog is launched, the `render` function within `dialog.jsx` is executed with some amount of the set of dialog ids. During a cycle, the `render` function checks each of the dialog ids for whether they are open or closed and then renders it as such. 

When a button in the document header is clicked, the `render` function within `dialog.jsx` is being called in three separate render cycles as explained above: 

First cycle -- some dialog ids are processed and the search bar opens
Second cycle -- some dialog ids are processed and the SVG for the search bar appears
Third cycle -- some dialog ids are processed and the dialog prompted via the document header closes

When the same dialog is prompted via the menu or keyboard shortcut, the third cycle does not occur. Something in particular is happening with the `documentheader` component. The subset of dialog ids that are being processed in the third cycle are: 

- `first-launch__dialog`
- `keyboard-shortcut__dialog`
- `search-bar__dialog`
- `exports-panel__dialog`
- `dialog-export-popover`
- `dialog-libraries`

[...]

PR #3588, opened 2016-01-21
Put a timer on adding event listeners to the window

DivyaPrabhakar:
References #3532 

With the previous version of bluebird, the call to `_addListeners()` was working asynchronously behind the scenes. Therefore the click events on the document header we handled before an event listener was added on the window. With this recent change to bluebird, https://github.com/petkaantonov/bluebird/issues/915, this became synchronous and the event listener on the window was added before the click to open the document header buttons were interpreted, resulting in the automatic toggle. This PR makes the addition of event listeners in the `componentDidUpdate` asynchronous again.

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -128,7 +128,12 @@ define(function (require, exports, module) {
                     dialogEl.showModal();
                 }
 
-                this._addListeners();
+                // Ensure that handlers are added asynchronously to avoid incorrectly
+                // handling the event that opened the dialog initially. See #3532.
+                window.setTimeout(function () {
+                    this._addListeners();
+                }.bind(this), 0);
+
                 this._positionDialog(dialogEl);
                 this.props.onOpen();
             } else if (!this.state.open && prevState.open) {


iwehrman:
In a little more detail, here's what we found was happening:

1. A click event on the document header buttons was handled (in the bubble phase) by the button click handler. This click handler executed an action to toggle a dialog open.
2. The handler is of course run synchronously as part of the browser's event propagation. The action execution was _also_ being processed in the same frame (assuming there were no blocking tasks executing in the queue).
3. The dialog action prompts the dialog store to emit a change event (in the same frame), which prompts the search or export dialog to re-render as "open" (in the same frame), which causes the dialog to attached click handlers to the `window` (in the bubble phase, and in the same frame), which will close the dialog when the `window` receives a click event.
4. Event propagation continues, eventually bubbling back up to the `window`, which now has a `click` event handler which dismisses the dialog. :black_large_square: 

Our hypothesis is that action execution was previously running asynchronously and in a different frame from the event propagation, but that a recent bluebird upgrade changed this so that it started running in the same frame. And, indeed, petkaantonov/bluebird#915 supports this hypothesis. The "fix" is to ensure that the important part of the dialog open work happens in a different frame to ensure that any event handler that prompted a dialog open is not processed by the event handlers attached by the dialog, which are all intended to only act upon later events.

Issue #3267, opened 2015-11-18
Entering Vector Mask mode on a masked group countaining a shape object selects the path associated with the shape instead of the mask on the group


iwehrman:
(to reduce later confusion, note that @pineapplespatula originally filed this from my account, and he's the best person to talk about about STR, etc.)

pineapplespatula:
I cannot reproduce this anymore, @chadrolfs would you mind seeing if you can repro it?

chadrolfs:
@iwehrman can you provide steps or a test file for this?  I'm confused about STR.  Currently you can't have a mask on a shape layer.

pineapplespatula:
You can have a mask on the group though. 

![screen shot 2016-01-22 at 5 06 56 pm](https://cloud.githubusercontent.com/assets/1008531/12527042/d34ef278-c129-11e5-8368-3ec54eef5bd9.png)

pineapplespatula:
![screen shot 2016-01-22 at 5 07 51 pm](https://cloud.githubusercontent.com/assets/1008531/12527049/e3e1408c-c129-11e5-9ba2-85b6e91f6dfe.png)

chadrolfs:
I can repro.
![mask-shapepath](https://cloud.githubusercontent.com/assets/8442631/12561813/d2492d7e-c356-11e5-9dab-6c5c27b14a70.gif)

chadrolfs:
This is very confusing and problematic if a user is trying to delete the mask.  Entering mask mode in this case and hitting delete will do nothing.  Bumping to High.

pineapplespatula:
i created Jira bug PS-1080

chadrolfs:
Thanks for filing the JIRA bug, however I'm concerned that JP won't see this bug because he is on a different molecule team with their unique JIRA dashboard that only shows tasks/bugs assigned to that team.  Could you reach out to him to see if 1) he's aware of it, 2) if he's planning on fixing it, and 3) what his rough timeline is (when will he get to it)?  Thanks.

iwehrman:
Alternatively, re-assign to @jsbache so that he can triage it when he returns.

pineapplespatula:
Talked to JP, he wont be able to get to it for a while I'm going to reassign this to @jsbache for now

chadrolfs:
:+1: thanks

pineapplespatula:
[...]

PR #3590, opened 2016-01-21
use top most layer to avoid group confusion

pineapplespatula:
It seems like in some cases our number of selected layers is different than expected when groups are selected. Using the top layer instead should resolve that issue. 

fix for issue #3267

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -612,7 +612,7 @@ define(function (require, exports) {
         }
 
         var currentLayers = currentDocument.layers.selected,
-            currentLayer = currentLayers.first();
+            currentLayer = currentLayers.last();
 
         // vector mask mode requires an active layer
         if (!currentLayer) {


iwehrman:
If we have a bug in which having one group selected results in `document.layers.selected` returning more than that one layer, that is the bug which should be fixed. I am not in support of this change as a workaround for that bug, at least not without additional information.

pineapplespatula:
I cant reproduce the bug anymore, i think that using last instead of first is the expect user behavior if we allowed multiple selections. since it would be the top of the layer selection, ( and the first from a user perspective), but you are right that it doesn't address the core issue here of our selection state being incorrect. 

TL:DR I think this should be merged but don't close the bug.

iwehrman:
@pineapplespatula: I suggest we discuss this in the standup tomorrow.

pineapplespatula:
Closing until we can actually repro the bug

Issue #3137, opened 2015-10-26
Exporting assets from a modal tool state fails


iwehrman:
0. Make a text layer, configure at least one asset.
1. Edit the text layer.
2. Click the "export now" button in the export panel:

```
Executing action export.exportLayerAssets after waiting 1ms; 1/4
fluxcontroller.js:472 Transfer from policy.suspendAllPolicies to policy.suspendPolicies failed: Error: Policies are already suspended
    at Object.suspendPolicies (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/policy.js:245:35)
    at FluxController._applyAction (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:568:36)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:455:37)
    at AsyncDependencyQueue._processNext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:291:27)
    at AsyncDependencyQueue.push (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:143:14)
    at Object.FluxController._makeActionReceiver.Object.create.transfer.value (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:445:42)
    at Object.suspendAllPolicies (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/actions/policy.js:291:36)
    at FluxController._applyAction (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:568:36)
    at null.<anonymous> (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/fluxcontroller.js:455:37)
    at AsyncDependencyQueue._processNext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/src/js/util/async-dependency-queue.js:291:27)
[...]

PR #3592, opened 2016-01-21
Disable quick direct export in modal state

iwehrman:
1. The quick export button in the export panel is now disabled in modal states.
2. Component state that forcibly disables export is now reset when the controller resets to prevent it from being stuck in a disabled state.

Addresses (and there is more discussion in) #3137.


volfied:
:+1:

Issue #3281, opened 2015-11-19
focus lost after setting color via color picker


chadrolfs:
new doc
drag out 2 rect shapes
click on Fill color and click to set color
press Shift + Tab

Result: focus switches to other rect shape

chadrolfs:
![colorpicker-lostfocus](https://cloud.githubusercontent.com/assets/8442631/11258518/9772834e-8e0e-11e5-9435-657f5288dbde.gif)

chadrolfs:
happens on 16.1 GM (r.301)

iwehrman:
Sliders eventually need to be made focusable.

iwehrman:
Or, on mouse up, refocus whatever was focused before.

iwehrman:
FBNC to @chadrolfs.

PR #3593, opened 2016-01-21
Support same key diff. modifier dialog closers, add shift+tab to ColorInput

volfied:
Addresses #3281, problem was that we had Tab with no modifiers as a closer, but not shift+tab. And on `Map` focus, it was body that had the focus, so shift tab would be handled by active tool, while tab handled by dialog.

Then I realized, we don't allow same key shortcuts, but this was same key / different modifier. So I changed that code to support same key different modifier shortcuts, using modifiers as a way to come up with unique ID.

### DIFF
--- a/src/js/jsx/shared/Dialog.jsx
+++ b/src/js/jsx/shared/Dialog.jsx
@@ -34,7 +34,8 @@ define(function (require, exports, module) {
 
     var os = require("adapter").os;
 
-    var math = require("js/util/math");
+    var math = require("js/util/math"),
+        keyUtil = require("js/util/key");
 
     /**
      * Valid methods of positioning the dialog
@@ -329,8 +330,10 @@ define(function (require, exports, module) {
             if (this.props.dismissOnKeys && _.isArray(this.props.dismissOnKeys)) {
                 var flux = this.getFlux();
                 this.props.dismissOnKeys.forEach(function (keyObj) {
+                    var modifiers = keyObj.modifiers || {};
+
                     flux.actions.shortcuts.addShortcut(keyObj.key,
-                        keyObj.modifiers || {}, this.toggle, this.props.id + keyObj.key, true);
+                        modifiers, this.toggle, this.props.id + keyUtil.modifiersToBits(modifiers) + keyObj.key, true);
                 }, this);
             }
         },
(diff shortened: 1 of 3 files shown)


src/js/jsx/shared/Dialog.jsx
```diff
@@ -324,8 +324,10 @@ define(function (require, exports, module) {
             if (this.props.dismissOnKeys && _.isArray(this.props.dismissOnKeys)) {
                 var flux = this.getFlux();
                 this.props.dismissOnKeys.forEach(function (keyObj) {
+                    var modifiers = keyObj.modifiers || {};
+
                     flux.actions.shortcuts.addShortcut(keyObj.key,
-                        keyObj.modifiers || {}, this.toggle, this.props.id + keyObj.key, true);
+                        modifiers, this.toggle, this.props.id + Object.keys(modifiers).join("+") + keyObj.key, true);
```
iwehrman: The order of `Object.keys` can be a little hard to predict. Would it be similarly convenient and more reliable to use the modifier bits here instead?

iwehrman:
I agree that this improves the situation with shift-tab, but it still seems like the tab behavior is not correct here. If the input is focused and I drag the slider, it seems like the input should still be focused. (Or it seems like the slider should have focus.) What do you think?

volfied:
I think I like the concept of leaving the focus on the already focused input, but not focusing the sliders or the map. I'll see what I can do in this regard, it'll complicate the change. Meanwhile, @iwehrman, please let me know if that sounds reasonable to you.

volfied:
@iwehrman This is now ready for review, I added some hacky bookkeeping to keep track of last focused component in color picker, and upon mouse up on the sliders or the color map, return the focus back to it.

@placegraphichere it would be nice to get your input on this interaction, you can test it out by checking out this branch.

iwehrman:
Hmm, I'm at 4d4de9de30db4ba96daec979348bcc7873f63380 but I'm not seeing the new behavior. I still lose focus entirely after dragging the map slider:
img

iwehrman:
Ugh, disregard that. I have a `grunt` problem. Code looks great and, now that I've recompiled, it actually works :grimacing:

Issue #3525, opened 2016-01-14
Search filter icon alignment is weird


iwehrman:
This looks weird/bad to me:

![image](https://cloud.githubusercontent.com/assets/1075154/12312001/82fff7a6-ba10-11e5-9e18-0f4db044dfdd.png)

The spacing between the search icon, the filter icon, and the text input feels clumsy.

iwehrman:
I think this needs some love from @placegraphichere. One possibility might be to simply replace the magnifying glass icon with the filter icon when the filter is active. There are many other possibilities!

iwehrman:
Fixed with extreme prejudice by #3594.

Issue #2848, opened 2015-10-07
Confusing icon 'x' in the search dialog


ktaki:
Most of people think that 'x' in the search dialog is for dismissing the dialog. It is, however, meant to be clearing the input. The icon need to be modified to avoid confusion.

iwehrman:
This is the pattern in use:
![image](https://cloud.githubusercontent.com/assets/1075154/10326506/d9d446e4-6c50-11e5-89da-c498d4a1eeba.png)

iwehrman:
@placegraphichere: do you think this is a bug? Can you either a) articulate exactly how you think this should work and then assign to @DivyaPrabhakar; or, if it already works exactly as intended, b) close the bug?

iwehrman:
@ktaki: @placegraphichere has updated the icon to be less confusing. Let us know what you think!

chadrolfs:
As in, there is no more `x`?

iwehrman:
It's a different X that may look more like the Mac OS "clear input" icon compared to a "close window" icon.

ktaki:
Yes, that is better.

PR #3594, opened 2016-01-21
Search Bar visual clean up

placegraphichere:
Fixes for
- #2848 : Using new icon that is more similar to the “clear search”
pattern
- #3525 : Adjusted icon position and other input position to lock it in
place
- Also stopped the hover state from shifting on search results due to
mismatched border size

screen shot 2016-01-21 at 4 02 21 pm

search-hover


iwehrman:
:hamburger: :fries:

Issue #3519, opened 2016-01-13
Search UI loses focus when re-focusing app


iwehrman:
1. Open the search UI. The search UI has focus.
2. Cmd-tab to another application.
3. Cmd-tb back to Design Space.

Result: sometimes, but not always, the search UI has lost focus.

The search UI should always have focus while the search UI is open.

iwehrman:
Strangely, it seems to lose focus _every other time_.

volfied:
@chadrolfs can you verify the fix?

chadrolfs:
works great on both Mac and Win!

PR #3596, opened 2016-01-21
SuperHack: force search UI to maintain focus when activating application

iwehrman:
This works around what I assume is a Chromium bug that is causing the active text input to be blurred when the application window is activated. I don't know for sure that this is a Chromium bug, but I can't figure out any other reason why this would be happening. This is surely a SuperHack.

Addresses #3519 miserably.


volfied:
Well, it's marked superhack, and seems to work!

Issue #3543, opened 2016-01-15
Automatically turn off "Add copy to to copied layers and groups"


placegraphichere:
We should not add "copy" to all duplicated layers. Its annoying. Either maintain the setting a user has or just turn it off by default in Ds. 

Received through user feedback

![screen shot 2016-01-14 at 7 55 40 pm](https://cloud.githubusercontent.com/assets/3229167/12343428/21ee97d0-baf9-11e5-87b6-46675abb7ce9.png)

Lemme know if this should go in JIRA

iwehrman:
This doesn't have to happen in the current sprint @volfied but if you can find a simple flag to pass to our action descriptor, this would be nice.

volfied:
I can fix this in a simple way by exposing that option to the UAppElement property, since it's globally available in Photoshop. Then we can set it ourselves on startup to the value we like.

volfied:
I will work with Jesper on making sure this property is bounced back to existing value in Classic.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Personally, I didn't see a problem with the way it was before this change - DS would honor whatever the standard PS pref was so if you didn't want 'copy' added, just turn the pref off, enter DS and work away. The old way, you had the option either way - this way, we're forcing the user to user to work with it off.   

Other than forcing it off by default with no other option to turn it back on, the only other workflow problem with this is with Export.  You can have 5 artboards or layers with the same titled/name (e.g. Artboard 1), but only 1 will get exported.

volfied:
@chadrolfs that is an important point. @mcilroyc is this going to break export, as in is export reliant on layers having different names?

iwehrman:
Also, @placegraphichere, can you jump in here and argue for not exposing this pref (via standard PS) in Design Space?

volfied:
By the way, I think that export code should gracefully handle multiple layers with the same name, maybe dropping down to parent names to show shortest unique path in the exported file name. After all, this option of appending "copy" is something the user can achieve themselves by renaming layers, as well.

mcilroyc:
I agree that export should be better about multiple layers with the same name, so it is irrelevant to this conversation.

chadrolfs:
I think that issue can be tracked via #2274

Issue #3246, opened 2015-11-13
Placing smart object should skip the transform modal state


iwehrman:
There's already a standard PS preference for this:
![image](https://cloud.githubusercontent.com/assets/1075154/11156416/a6b446ce-8a00-11e5-9b36-aa4751a3484b.png)

iwehrman:
As of #3598, we now skip the transform mode by default when placing. I am wary that this will cause some bugs, mainly because we've had to put in so many hacks and workarounds to deal with placing objects already. FBNC to @ktaki, but testing from @chadrolfs would be great too!

ktaki:
Looks good to me. Turn it over to @chadrolfs.

chadrolfs:
Still sometimes gets stuck on LLSO, but that's tracked here #3575. Secondarily also relates to #3102.  Otherwise, this looks good to me as well.

PR #3598, opened 2016-01-22
Set appendCopy and skipTransform on startup

volfied:
This will require pg-dev-mac#996 or later builds. Also will require https://github.com/adobe-photoshop/spaces-adapter/pull/182, and will address #3246 and #3543 

I've exposed the Layers Panel option "Add "copy" to layer names" in Photoshop to us, and added an lib/application method to set it. I've also set it up as a auto-restored property. So now, when we option drag / duplicate layers, they don't get "copy" added to their names. #3543

Skip transform when placing SOs from library was already an existing preference in Photoshop, so I've added a new `setGlobalPreferences` API to lib/application to set it. This was a conscious decision, instead of having one function to set an option, for future uses of this API. #3246

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -32,6 +32,7 @@ define(function (require, exports) {
         descriptor = require("adapter").ps.descriptor,
         documentLib = require("adapter").lib.document,
         layerLib = require("adapter").lib.layer,
+        appLib = require("adapter").lib.application,
         OS = require("adapter").os;
 
     var Layer = require("js/models/layer"),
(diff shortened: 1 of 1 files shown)


src/js/actions/layers.js
```diff
@@ -2010,11 +2011,16 @@ define(function (require, exports) {
         }.bind(this);
         descriptor.addListener("delete", _deleteHandler);
 
-        return Promise.resolve();
+        var setAppendCopyPromise = descriptor.playObject(appLib.setAddCopyToLayerNames(false)),
```
iwehrman: Any reason not to `batchPlay` these commands?

src/js/actions/layers.js
```diff
@@ -2010,11 +2011,16 @@ define(function (require, exports) {
...
+            setSkipTransformPromise = descriptor.playObject(appLib.setGlobalPreferences({
+                skipTransformSOFromLibrary: true
+            }));
+
+        return Promise.join(setAppendCopyPromise, setSkipTransformPromise);
     };
     beforeStartup.action = {
         reads: [],
-        writes: []
+        writes: [locks.JS_APP]
```
iwehrman: Seems like this should be `PS_APP`?

iwehrman:
Nits 2 @volfied.

volfied:
@iwehrman thanks for the review. I realized we can call these in afterStartup since they don't affect our rendering and are not needed to change immediately.

iwehrman:
:+1:

PR #3601, opened 2016-01-22
Add missing px unit to the media queries.

shaoshing:
I forgot the px unit in my previous PR so the recent style fix will not work correctly in smaller screen (my bad :( )


iwehrman:
OK!

Issue #3549, opened 2016-01-18
Color picker swatch on 2nd applied Effect is clipped at bottom


chadrolfs:
<img width="288" alt="screen shot 2016-01-18 at 12 12 21 pm" src="https://cloud.githubusercontent.com/assets/8442631/12401672/e48e0556-bddc-11e5-8093-4d14512a3a7f.png">

chadrolfs:
Still an issue but not just with 2nd applied effects.  
![colorpicker-clip](https://cloud.githubusercontent.com/assets/8442631/12594492/3aa8ed9e-c42b-11e5-9138-7a5f16c4c991.gif)

chadrolfs:
also check on Windows machine (particularly low res).

chadrolfs:
at reasonable resolutions, this is resolved.  However, at smaller resolutions (on mac thunderbolt at 1280x720 or on PC at 1366x768) there is still clipping. 
PC:  
![icon-clipping](https://cloud.githubusercontent.com/assets/8442631/12728666/c462d3d2-c8d7-11e5-948b-d2cae3cc2e1f.png)

chadrolfs:
Mac: 
![screen shot 2016-02-01 at 11 37 59 am](https://cloud.githubusercontent.com/assets/8442631/12728782/476d7016-c8d8-11e5-8c13-ab025d0695c9.png)

chadrolfs:
downgrading to medium priority since most of our users are probably not using that low of a monitor resolution.

PR #3602, opened 2016-01-22
Padding on formlines first-child removed to fix clipping of color swatch

DivyaPrabhakar:
References issue #3549 

Before: 

After: 

One thing to note, is that the padding in other panels changes slightly with this fix. Can @placegraphichere or @designedbyscience give their opinion about the change? I made this image to try and show the slight difference. If this doesn't help, you can just check out toggling the CSS change on this branch.

### DIFF
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -134,7 +134,7 @@
 
 .formline__no-padding {
     padding-top: 0.2rem;
-    padding-bottom: 0.2rem;    
+    padding-bottom: 0.1rem;    
 }
 
 .formline:first-child {


mcilroyc:
@placegraphichere for quick design review

placegraphichere:
I'd like not see the larger panel section headers shown in the new screen shot. It seems like we're adding quite a bit to the bottom of each.

DivyaPrabhakar:
This is what it looks like now -- thoughts @placegraphichere?

Before: 


After:

placegraphichere:
@DivyaPrabhakar nice

DivyaPrabhakar:
@mcilroyc If you like it, merge it.

mcilroyc:
sure!

Issue #3599, opened 2016-01-22
Layers panel icons flash when they swap 


placegraphichere:
Layer icons like the artboard and group icons that have different states will often flash blank before the other state is loaded. It looks odd. We should see if we can avoid this.

![icon-flash-when-selecting](https://cloud.githubusercontent.com/assets/3229167/12499948/61bb4928-c074-11e5-9f07-76890f9e8468.gif)

iwehrman:
Related to or a duplicate of #2020?

iwehrman:
Please verify and close, @placegraphichere.

placegraphichere:
This feels so much better, thanks for getting this done @shaoshing!

Issue #2020, opened 2015-08-13
Most icons blink when a document is opened/closed


iwehrman:
Notably, not the toolbar icons:
![blink](https://s3.amazonaws.com/f.cl.ly/items/1v1C3Z3t1F0e0z2A300h/Screen%20Recording%202015-08-13%20at%2011.54%20AM.gif)

designedbyscience:
So, that's the same toolbar but different everything else (they are blinking because we are creating new versions of all of them)

chadrolfs:
Agree with @designedbyscience.  Only icons associated with new attributes blink.  Tab icons and tab titles don't blink.  And closing doesn't blink.  It's definitely better than it was.  @iwehrman do you think this issue is still medium priority based on current behavior?

chadrolfs:
reviewed - not M3

iwehrman:
Yeah, this is a real bummer, but I think we're out of ideas anyway. Off to M4land!

chadrolfs:
I'm still seeing this occur every so often now (not every time) if I repeatedly do cmd/ctrl + n.

chadrolfs:
almost totally gone though

iwehrman:
Still happens every time for me!

iwehrman:
Fixed by #3611!

PR #3611, opened 2016-01-26
Use inline SVG icon to avoid flashing

shaoshing:
**`npm install` is required**

This PR use inline SVG icon (via WebPack) to avoid icon flashing in the following scenarios: 

* expand/collapse layer group #3599 
* switch document #2020 

#### Load time
I observe no impact on the load time or initial display time (time until the layer panel is fully displayed).

#### Layer Panle Performance Comparision

Before and After has the same performance by comapring the results of the Layer Panel Performance Test

  | reference   | inline 
------------- | -------------| ----- 
Layer selection | 22ms      | 19ms 
Collapse AB     | 45ms     | 43ms
Expand AB       | 93ms     | 94ms
(numers are rendering time only)


designedbyscience:
I like this now that we're using WebPack and can still use files for ease, but Web Pack will inline everything for us... instead of just doing this here, could we update SVGIcon for this to work everywhere?

shaoshing:
@designedbyscience this PR is working as you suggested. `SVGIcon` will inline the SVG by default. We can still manage the files in the same way without change, and WebPack will compile the files once they are changed.

shaoshing:
I saw some icons are a little off their original position. need more work to fix them.

iwehrman:
I ran `npm install` and restarted `grunt debug` but I still get this runtime error in the `SVGIcon` render method:
```
Transfer from documents.beforeStartup to documents.initActiveDocument failed: Error: Cannot find module './ico-libraries-cc.svg'.
    at file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:33728:42
    at webpackContextResolve (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:33728:90)
    at webpackContext (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:33725:30)
    at React.createClass.render (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:33530:48)
```

src/js/jsx/shared/SplitButton.jsx
```diff
@@ -52,8 +52,7 @@ define(function (require, exports, module) {
             
             var buttonIcon;
             if (this.props.iconId) {
-                buttonIcon = (<SVGIcon viewBox="0 0 16 16" CSSID={this.props.iconId}
-                    iconPath={this.props.iconPath}/>);
```
iwehrman: Why don't we need this prop any longer? (Mainly I just don't understand what the use of `CSSID` is versus `iconPath`.)

shaoshing: After switching to inline icons, it is no longer needed to refer the icon via its path. Another reason is that `iconPath` seems not in use any more.

shaoshing:
@designedbyscience @iwehrman @placegraphichere I've fixed all known style issues due to the change of SVG. I also checked the icons under different font sizes (base-8 and base-11) and they are all displaying correctly. So, this PR is now ready for merging. Since the `SVGIcon` is now much simpler with a most of the code removed, can you help review my change to make sure that I did not break anything? :) 

BTW, the icons are correct without applying the `viewBox` property, so I am wondering if I should remove the property completely in other components, or leave them unchanged just in case?

iwehrman:
Everything looks great! Please also take a look @designedbyscience and @placegraphichere, but I'm mergin' away.

placegraphichere:
I am excited about this

Issue #3467, opened 2016-01-06
Search bar causes localization error on filter drill down for libraries


volfied:
STR: 
 1 - cmd+f to bring up search
 2 - type name of one of the libraries
 3 - Press enter to search in that library
 4 - nls throws saying translation can not be found for library ID.

PR #3612, opened 2016-01-26
Bug fix: search bar localization error

shaoshing:
Fix for issue #3467

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -212,26 +212,27 @@ define(function (require, exports, module) {
         /**
          * Removes words from Datalist input that are contained in the ID
          *
-         * @param {string} id ID of selected item
+         * @param {Array.<string>} id ID of selected item
          */
         _updateDatalistInput: function (id) {
             if (id) {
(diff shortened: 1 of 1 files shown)


volfied:
This looks like it's curing the symptom rather than the cause. We have safeFilterNameMap to figure out what we can remove from the id. Can you explain why this change had to be done in the function, essentially quieting down the error, instead of fixing it in `search/libraries.js` where we're supplying the non-localizable strings?

shaoshing:
Hi @volfied, that's a great question. The essential cause is that the ID is a mix of localizable/non-localizable strings. For example, here are some possible values of the `id` param:

```js
[""] // not localizable
["RECENT_DOC"]
["ALL_LAYER", "ARTBOARD"]
["LIBRARY", "123456.ABCD.ABCD.ABCD.123456"] // ID is not localizable
```

The original way was to translate the text, then look for the name from `safeFilterNameMap`. If neither of them works, do nothing. This used to be working until we changed the behavior of `nls.localize` to raise an error if translation not found. So the easiest fix for me is to catch the translation error.

volfied:
Thanks for the explanation.

Issue #3613, opened 2016-01-26
Transform inputs blink when creating a new document


iwehrman:
After #3611, none of icons blink when creating a new document (:tada:). The inputs in the transform panel are still blinking though, presumably for a very different reason:
![img](https://s3.amazonaws.com/f.cl.ly/items/3V391B2M1x1h0G3g0e2M/Screen%20Recording%202016-01-26%20at%2010.37%20AM.gif?v=6225b72c)

iwehrman:
Fixed by #3616.

PR #3616, opened 2016-01-26
Prevent Size and Position components from blinking on mount

iwehrman:
The `Size` and `Position` components currently blink when first showing a document. This is entirely due to quirks in their `shouldComponentUpdate` methods. In particular, it seems like the components weren't rendering until their first state _change_ after mounting.

@volfied: I ended up removing a bunch of code that you wrote recently, so I'd like your feedback on this to make sure I didn't screw something up. I've removed some of your optimizations because I think they were partially responsible for the blinking.

Addresses #3613.

### DIFF
--- a/src/js/jsx/sections/transform/Position.jsx
+++ b/src/js/jsx/sections/transform/Position.jsx
@@ -53,8 +53,9 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            return !this.state ||
+            return this.props.active !== nextProps.active ||
                 this.state.disabled !== nextState.disabled ||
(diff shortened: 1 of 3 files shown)


src/js/jsx/sections/transform/Position.jsx
```diff
@@ -99,6 +83,10 @@ define(function (require, exports, module) {
             });
         },
 
+        componentWillMount: function () {
+            this.componentWillReceiveProps(this.props);
```
volfied: Can you add a comment explaining why you need to call this life-cycle function here? I'm sure it'll confuse someone later.

iwehrman: Fixed.

Issue #3619, opened 2016-01-27
Unrecoverable error initializing background documents for search


iwehrman:
```
Enqueuing action search.toggleSearchBar; 0/0
Executing action search.toggleSearchBar after waiting 0ms; 1/0
Enqueuing action documents.initializeDocuments; 1/0
Finished action search.toggleSearchBar in 30ms with RTT 30ms; 1/1
Executing action documents.initializeDocuments after waiting 19ms; 1/0
Enqueuing action shortcuts.addShortcut; 1/0
Transfer from documents.initializeDocuments to documents.updateDocument failed: TypeError: Cannot read property 'layers' of undefined
    at Fluxxor.createStore.setDocument (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:80628:36)
    at Fluxxor.createStore._documentUpdated (file:///Users/wehrman/Desktop/photoshop-Release_x86_64/Plug-Ins/Spaces/www/build/spaces-design-en.js:80667:19)
```

This also seems to send `Adobe Spaces Helper` into a divergence, which is extra bad.

iwehrman:
Fixed by #3620.

PR #3620, opened 2016-01-27
Do not attempt to dispatch layer-selection-changed events for uninitialized documents

iwehrman:
What it says! This was causing an unrecoverable error when opening the search dialog with an uninitialized document (i.e., one for which `document.layers === null`).

@shaoshing: can you let me know if this looks like the right fix?

Addresses #3619.


shaoshing:
This looks perfect! Thanks for fixing the bug.

PR #3621, opened 2016-01-27
LayerPanel: fix broken layer face drag-and-drop

shaoshing:
This PR fixes the drag-and-drop in the LayerPanel by changing the type of drag target in the `Draggable` component from `Layer` model to layer id, so that the `Draggable` won't hold a stale layer model. 

More details about the fix:

In the recent layer panel PR, `LayerFace` stores some `Layer` model attributes (`selected/expanded/visible`) in its state. When it receive a new layer prop, it will skip rendering if the change is the same as its state. As a result, the `Draggable` instance will hold the old version because the component is not re-rendered. This makes it fail when `Draggable` component needs to check whether it is the drag target when a drag event starts, which will always return false. The solution is to switch to layer id, which is a number and will never stale.

### DIFF
--- a/src/js/jsx/sections/layers/LayerFace.jsx
+++ b/src/js/jsx/sections/layers/LayerFace.jsx
@@ -35,7 +35,6 @@ define(function (require, exports, module) {
     
     var system = require("js/util/system"),
         svgUtil = require("js/util/svg"),
-        collection = require("js/util/collection"),
         nls = require("js/util/nls");
 
     var Draggable = require("js/jsx/shared/Draggable"),
@@ -453,19 +452,21 @@ define(function (require, exports, module) {
         _handleBeforeDragStart: function () {
             // Photoshop logic is, if we drag a selected layers, all selected layers are being reordered
             // If we drag an unselected layer, only that layer will be reordered
-            var draggedLayers = Immutable.List([this.props.layer]);
+            var draggedLayerIds = Immutable.List([this.props.layer.id]);
 
             if (this.state.selected) {
-                draggedLayers = this.props.document.layers.selected.filter(function (layer) {
+                draggedLayerIds = this.props.document.layers.selected.filter(function (layer) {
                     // For now, we only check for background layer, but we might prevent locked layers dragging later
                     return !layer.isBackground;
-                }, this);
+                }).map(function (layer) {
+                    return layer.id;
+                });
             }
-            
+
             this._isDragEventTarget = true;
             this.getFlux().actions.panel.disableTooltips();
             
-            return { draggedTargets: draggedLayers };
+            return { draggedTargets: draggedLayerIds };
         },
 
         /**
@@ -510,7 +511,7 @@ define(function (require, exports, module) {
          * @private
          * @type {Droppable~onDrop}
          */
-        _handleDrop: function (draggedLayers) {
+        _handleDrop: function (draggedLayerIds) {
             if (!this.state.isDropTarget) {
                 return Promise.resolve();
             }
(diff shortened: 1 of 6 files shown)


src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -554,8 +554,9 @@ define(function (require, exports, module) {
          * @private
          * @type {Droppable~onDragTargetMove}
          */
-        _handleDragTargetMove: function (draggedLayers, dragPosition) {
-            var isDropTarget = !draggedLayers.includes(this.props.layer),
+        _handleDragTargetMove: function (draggedLayerIds, dragPosition) {
+            var draggedLayers = this.props.document.layers.byIDs(draggedLayerIds),
```
volfied: Instead of adding a new layerStructure method, can you use
```
var layerTree = this.props.document.layers,
     draggedLayers = draggedLayerIds.map(layerTree.byID, layerTree),
    ....
```

volfied: You can also filter afterwards to make sure layer model exists, but I'm not a fan of adding yet another method to layerstructure that can be done with a map and a filter, and it will keep things Immutable.

src/js/jsx/sections/layers/LayerFace.jsx
```diff
@@ -554,8 +554,9 @@ define(function (require, exports, module) {
          * @private
          * @type {Droppable~onDragTargetMove}
          */
-        _handleDragTargetMove: function (draggedLayers, dragPosition) {
-            var isDropTarget = !draggedLayers.includes(this.props.layer),
+        _handleDragTargetMove: function (draggedLayerIds, dragPosition) {
```
volfied: Also, can you fix the JSDocs while you are here?

src/js/jsx/sections/libraries/LibrariesPanel.jsx
```diff
@@ -128,6 +128,7 @@ define(function (require, exports, module) {
 
             var flux = this.getFlux(),
                 document = flux.store("application").getCurrentDocument(),
+                draggedLayers = document.layers.byIDs(draggedLayerIds),
```
volfied: I see, you're using it in two places. Cancel my comment above.

src/js/models/layerstructure.js
```diff
@@ -609,6 +609,20 @@ define(function (require, exports, module) {
...
     };
+    
+    /**
+     * Get a list of Layer models by layer IDs.
+     *
+     * @param {Immutable.Iterable.<number>} ids
+     * @return {Immutable.Iterable.<Layer>}
+     */
+    LayerStructure.prototype.byIDs = function (ids) {
+        return ids.reduce(function (layers, id) {
```
volfied: It would be better to use map/filter here instead of reduce, since you're creating n Immutable.Lists every time you call this function.

```
return ids
    .map(this.byID, this)
    .filter(function (layer) {
        return layer;
    });
```

or something to that effect.

volfied:
One nit, and one JSDoc nit, but looks great!

src/js/models/layerstructure.js
iwehrman: I also really don't like putting something like this so deep in the model that filters out null layers. Non-existent layer IDs are usually a sign of a bug.

shaoshing: I agree with that subtle behavior on removing null layers. I have changed the function to include null if any of the IDs is not found. Thanks @volfied for the code snippet, it is very useful!

src/js/jsx/sections/layers/LayerFace.jsx
shaoshing: JSDoc fixed

shaoshing:
@volfied Thanks for reviewing. Fixed the JSDoc and addressed the `Layer#byIDs` function. Back to you.

PR #3622, opened 2016-01-27
LayerPanel: cancel forceUpdate when LayerGroup will unmount.

shaoshing:
If you create a document and close it immediately, it will raise an error complaining `this.forceUpdate` not allowed on a unmounted instance. This PR fixes the error by cancelling the timer callback when the instance is unmounted.

### DIFF
--- a/src/js/jsx/sections/layers/LayerGroup.jsx
+++ b/src/js/jsx/sections/layers/LayerGroup.jsx
@@ -41,15 +41,20 @@ define(function (require, exports, module) {
 
         /**
          * This is used by the root LayerGroup to control whether or not its child LayerGroups 
-         * should render their collapsed layers. At initial rendering, `renderCollapsedLayers` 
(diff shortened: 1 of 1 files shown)


iwehrman:
I have an additional suggestion: instead of using `setTimeout` with an arbitrary 2-second timeout, what about using `requestIdleCallback` instead? That can be canceled similarly, and won't run unless the browser is idle. https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback

What do you think?

shaoshing:
@iwehrman I actually tried the `requestIdleCallback` but have to give up the idea because it fires right away and slow down the load time.

iwehrman:
OK. If you can think of any other way to get rid of this arbitrary timeout, I would really like to get rid of it. The reason is that 2s might work well on your computer, but it might be bad on a much slower or much faster computer. 

In any case, this small change looks good!

PR #3623, opened 2016-01-27
Grunt: better notification for WebPack and LESS

shaoshing:
**needs `npm install`**

Show notification for LESS task. It will show both successful and error message.
screen shot 2016-01-27 at 1 56 35 pm
screen shot 2016-01-27 at 1 57 20 pm

* Show timestamp in the WebPack's successful build message. This would be easier to know when the latest build is done if you make changes constantly.
screen shot 2016-01-27 at 1 56 21 pm


volfied:
Why not.

Issue #3136, opened 2015-10-26
Numbers are not correctly localized


fzhu1978:
Waston Bug:
https://watsonexp.corp.adobe.com/#bug=4078212

Problem: The numbers are not correctly localized on Design Space (Preview).

Method:
1. Install PS 20151022.m.268 build on DEU OS.
2 Switch to Design Space (Preview) work space.
3. New a doc and add the shape value.
4. Change the value of the shape properties, check the display of the numbers.

Result: The comma in the numbers display as dot. It is not localized. And input "," instead of "dot", it is un-accept.

Expected: The comma should be display well.

Note:
1. This issue occur on both Win and Mac on DEU and FRA OS.

fzhu1978:
@iwehrman 
Can you evaluate if we can fix for 16.1.2?

volfied:
@iwehrman you said this is partially addressed by #3627, can you comment on what's left / update the bug.

iwehrman:
Still missing: respecting the system preferences for the decimal separator when it differs from the locale default, and possibly also correctly formatting percents.

ktaki:
Also missing is a manual number input support. Currently fields do not accept numbers with comma as a decimal separator.

iwehrman:
After #3692, the only thing missing that I know of is % formatting. (In some locales, a space precedes the %.) We should probably just file a different bug for that, though. Over to @ktaki to verify.

ktaki:
New implementation is if the decimal separator is set to other than a period in the system prefs, it is set to use a comma. Windows offers an apostrophe as a choice, and standard Ps honors the selection.
According to wikipedia, however, it is only used in Spanish handwriting. The current implementation is straightforward, works well, and serves the purpose. 

Note for inputting numbers: 
If the decimal separator is a period, numerics with a comma as a decimal separator will be rejected.
If the decimal separator is a comma, both a period and a comma are accepted in numerical input, but the numerics are displayed with a comma as the decimal separator once accepted.

Closing this as fixed. Will address a space preceding % separately.

ktaki:
Apparently, a space preceding % is shown in standard Ps in French and German. DS does not have the space. So, this is non-issue for DS.

PR #3627, opened 2016-01-27
Basic number localization

iwehrman:
This provides extremely basic number localization for `NumberInput` mainly, and also for the scale values used by export. The display of numbers is handled by the browser by the [`toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) method of `Number` objects, and relies on the browser's locale being correctly set, as it has been. The somewhat more complicated issue is parsing arithmetic expressions in `NumberInput`. This is handled with a hack: we determine if the browser wants to format decimals with a comma, and if so we swap out commas for periods and semicolons (which often separate arguments to mathematical functions in the same locale) with commas. Hence, you `sum(1,5; 2,5)` is translated to `sum(1.5, 2.5)` before a handoff to `mathjs.eval`. After that, it's translated back using `toLocaleString` again. This doesn't seem great, but it's the approach suggested by the mathjs developer. See josdejong/mathjs#421 and http://mathjs.org/examples/browser/custom_separators.html.html.

Note that this does _not_ attempt to localize color expressions returned by `tinycolor`. So, for example, you'll see `rgba(0, 0, 0, 0.5)` instead of, say, `rgba(0; 0; 0; 0,5)`. This is because these are intended to be CSS color expressions, which by definition are not localized. This might be debatable, of course, but I thought I'd start here without digging a hole that didn't seem to need digging.

I'm also vaguely aware that in some locales, percentages are formatted slightly differently, like `50 %` instead of `50%`. I'll handle that, and any other issues that come up, separately.

Partially addresses #3136.

### DIFF
--- a/src/js/jsx/sections/export/ExportAllPanel.jsx
+++ b/src/js/jsx/sections/export/ExportAllPanel.jsx
@@ -72,7 +72,7 @@ define(function (require, exports, module) {
                 var stable = asset.status === ExportAsset.STATUS.STABLE,
                     requested = asset.status === ExportAsset.STATUS.REQUESTED,
                     errored = asset.status === ExportAsset.STATUS.ERROR,
-                    assetTitle = asset.scale + "x";
+                    assetTitle = asset.scale.toLocaleString() + "x";
 
                 var assetClasses = classnames({
                     "exports-panel__layer-asset": true,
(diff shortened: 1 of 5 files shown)


volfied:
This looks good to me, but I'm going to wait for @mcilroyc to review the export code to make sure it's kosher.

mcilroyc:
Export stuff looks OK to me.  One concern that occurred to me is that this will put commas in the filename (via the scale suffix), but that seems OK upon a quick test.

iwehrman:
I think this is worth merging, and I do not think the ball is currently in my court. What say you @volfied or @mcilroyc?

PR #3629, opened 2016-01-27
Scan and load actions and stores using webpack contexts

volfied:
Instead of using index.js files for stores and actions, this change uses [Webpack contexts](https://github.com/webpack/docs/wiki/context) to scan the two folders at compile time and dynamically bundle files.

This has the added benefit of not having to maintain index.js files for our actions and stores.

One change that had to happen to achieve this was change action name strings like `toolSuperselectType` to `tool.superselect.type` where it's a dot separated path of the file.


src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
+     * Automatically loads all js files in ./actions folder
+     * replacing their paths with dot separated keys
+     *
+     * @param {function} requireContext require.context call of Webpack pointing to the actions folder
+     *
+     * @return {function()} when called, returns all action modules loaded
+     */
+    var requireActions = function (requireContext) {
+        return requireContext.keys().reduce(function (actionMap, actionKey) {
+            var actionId = actionKey.substring(2, actionKey.indexOf(".js")).replace(/\//gi, ".");
```
shaoshing: Can you add example values of the `actionKey` and `actionId` as comment? I think it is easier to understand the purpose of the string transformation with an example.

src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
+     * @return {function()} when called, will initialize all stores
+     */
+    var requireStores = function (requireContext) {
+        return requireContext.keys().reduce(function (storeMap, storeKey) {
+            var storeId = storeKey.substring(2, storeKey.indexOf(".js")),
+                Store = requireContext(storeKey);
+
+            storeMap[storeId] = new Store();
+            return storeMap;
+        }.bind(this), {});
```
shaoshing: Looks like `this` is not necessary here?

src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
+    };
+
+    /**
+     * Raw (unsynchronized) action modules indiced by dot separated path as IDs
+     *
+     * @type {objects}
+     */
+    var actionIndex = requireActions(require.context("./actions", true, /^\.\/.*\.js$/));
```
shaoshing: What do you think about moving the context creation code into the `requireActions` function?

shaoshing:
Hey @volfied that's a very nice use of the WebPack's context feature! I just have a few questions, otherwise, everything looks excellent. Back to you.

src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
+
+            storeMap[storeId] = new Store();
+            return storeMap;
+        }.bind(this), {});
+    };
+
+    /**
+     * Raw (unsynchronized) action modules indiced by dot separated path as IDs
+     *
+     * @type {objects}
```
iwehrman: That isn't a valid type expression.

src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
+            return actionMap;
+        }, {});
+    };
+
+    /**
+     * Automatically loads all js files in ./stores folder
+     *
+     * @param {function} requireContext require.context call of Webpack pointing to stores folder
+     *
+     * @return {function()} when called, will initialize all stores
```
iwehrman: I'm not sure whether `function` or `function()` is the correct type expression, but in any case it doesn't seem like this returns a function.

src/js/fluxcontroller.js
```diff
@@ -34,14 +34,53 @@ define(function (require, exports, module) {
...
         performance = require("./util/performance"),
         log = require("./util/log");
 
     /**
+     * Automatically loads all js files in ./actions folder
+     * replacing their paths with dot separated keys
+     *
+     * @param {function} requireContext require.context call of Webpack pointing to the actions folder
+     *
+     * @return {function()} when called, returns all action modules loaded
```
iwehrman: This also does not seem to return a function.

volfied:
@iwehrman and @shaoshing thank you for your reviews. I've cleaned up the code and am actually using contexts a bit smarter and integrated to flux controller now.

Please re-review.

src/js/fluxcontroller.js
```diff
@@ -958,6 +967,23 @@ define(function (require, exports, module) {
     };
 
     /**
+     * Given a webpack context of stores, returns all Flux initialized stores
+     *
+     * @param {object} context all the store files in a context
+     *
+     * @return {object} An object of initialized stores
```
iwehrman: It would be nice to provide stronger type annotations here. In particular, when objects are used as maps they can and should be annotated like `Object.<T1, T2>` (note the capital O).

volfied:
@iwehrman right back at ya.

iwehrman:
LGTM.

PR #3632, opened 2016-01-28
Padding changed to avoid clipping on color swatches

DivyaPrabhakar:
References issue #3549  

- [x] @placegraphichere to sign off on the design

### DIFF
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -133,7 +133,6 @@
 }
 
 .formline__no-padding {
-    padding-top: 0.2rem;
     padding-bottom: 0.1rem;    
 }
 


src/style/panel.less
```diff
@@ -133,12 +133,11 @@
 }
 
 .formline__no-padding {
-    padding-top: 0.2rem;
     padding-bottom: 0.1rem;    
 }
 
 .formline:first-child {
-    padding-top: 0;
+    padding-top: 0rem;
```
iwehrman: I don't think this is valid CSS? It probably works either way, but I was under the impression that 0 should always be unit-less.

placegraphichere:
@iwehrman units should not matter, but we should remove it. @DivyaPrabhakar I will review

iwehrman:
I think this is referencing the wrong issue? #3459 was about the scripting time needed for document activation.

DivyaPrabhakar:
@placegraphichere I have removed the units so back to you. 
@iwehrman You are right, it references issue #3549. I have changed it.

Issue #3636, opened 2016-01-29
Inner Shadow should be listed above Color Overlay


ktaki:
In Effects panel, the order should be

Stroke
Inner Shadow
Color Overlay
Drop Shadow

![screen shot 2016-01-28 at 3 57 22 pm](https://cloud.githubusercontent.com/assets/2029597/12662877/2fd8c96c-c5d8-11e5-9190-be565bed3b29.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3637, opened 2016-01-29
Reordered the Effects to be in more logical order

placegraphichere:
Reordered the Effects to be in more logical order

Fix for #3636 

removed some white spaces


placegraphichere:
screen shot 2016-01-28 at 6 05 20 pm

iwehrman:
It's not immediately clear to me why this is a better order, but I'll take your word for it.

placegraphichere:
It implies the visual sorted order the effects will be applied in. Stroke is above inner shadow which is above color which is above drop shadow. The list now goes along with that visual sort order

Issue #3639, opened 2016-01-29
synchronization.debounce ignores promise


iwehrman:
On master, [`synchronization.debounce`](https://github.com/adobe-photoshop/spaces-design/blob/5f40f5dbdedcf4a405f0037142278db9ff5933f1/src/js/util/synchronization.js#L97) doesn't actually do anything with the promise returned from the supplied function. At best it is equivalent to `_.debounce`. At worst, it's broken. I'm not sure yet. In the short term, I'm removing the `promise` variable to quiet the linter, but we should either fix this or remove it entirely.

iwehrman:
Fixed by #3641.

PR #3640, opened 2016-01-29
Replace jshint with eslint

iwehrman:
This PR replaces our `jshint` grunt task with an improved `eslint` grunt task. eslint is better for three reasons:

1. it has better support for ES2015 syntax
2. it works better than jshint, finding many errors that jshint should have caught but didn't
3. it works better with SublimeLinter, so it's now actually possible to do real-time linting in ST3.

As a result of 2) above, many indentation errors had to be fixed. eslint also caught a number of instances of unused variables, some of which were genuine bugs. (Related to that, see #3639.)

To make the linters work correctly with ST3, use Package Control to install Babel, SublimeLinter, SublimeLinter-jscs and SublimeLinter-contrib-eslint. Next, globally install jscs and eslint globally with `npm install -g jscs eslint`. You should verify that the globally installed versions are approximately the same as the dependencies of our grunt tasks. (I.e., if you had an old version installed globally, `npm install` won't necessarily update it.)


src/js/actions/libraries.js
```diff
@@ -464,9 +464,9 @@ define(function (require, exports) {
                 newElement = currentLibrary.createElement(currentLayer.name, ELEMENT_LAYERSTYLE_TYPE);
 
                 return Promise.fromCallback(function (done) {
-                        var representation = newElement.createRepresentation(REP_LAYERSTYLE_TYPE, "primary");
-                        representation.updateContentFromPath(stylePath, false, done);
-                    })
+                    var representation = newElement.createRepresentation(REP_LAYERSTYLE_TYPE, "primary");
+                    representation.updateContentFromPath(stylePath, false, done);
+                })
                     .then(function () {
```
shaoshing: This looks wrong to me. Maybe we should remove the indents of the below code block as well.

src/js/actions/libraries.js
```diff
@@ -937,8 +937,8 @@ define(function (require, exports) {
         }
 
         return Promise.fromCallback(function (done) {
-                representation.getContentPath(done);
-            })
+            representation.getContentPath(done);
+        })
             .bind(this)
```
shaoshing: Same question about the indention.

src/js/actions/libraries.js
shaoshing: Or we can do it this way. Depends on our preference of coding style.

```
return Promise
    .fromCallback(function (done) {

    })
    .then(function (done) {

    });
```

src/js/actions/search/libraries.js
```diff
@@ -144,15 +144,15 @@ define(function (require, exports) {
...
+                var uiStore = this.flux.stores.ui,
+                    panelStore = this.flux.stores.panel,
+                    centerOffsets = panelStore.getState().centerOffsets,
+                    midX = (window.document.body.clientWidth + centerOffsets.left - centerOffsets.right) / 2,
+                    midY = (window.document.body.clientHeight + centerOffsets.top - centerOffsets.bottom) / 2,
+                    location = uiStore.transformWindowToCanvas(midX, midY);
                     
-                    selectPromise
+                selectPromise
                         .bind(this)
```
shaoshing: Maybe we should remove the extra indents below as well? Wondering why Eslint is not reporting it.

src/js/jsx/sections/libraries/LibraryList.jsx
```diff
@@ -210,28 +210,28 @@ define(function (require, exports, module) {
...
-                        commandPromise = libraryActions.createLibrary(libraryName)
+            case _CREATE_LIBRARY:
+                if (libraryName.length !== 0) {
+                    commandPromise = libraryActions.createLibrary(libraryName)
                             .bind(this)
                             .then(function () {
                                 if (this.props.onCreateLibrary) {
                                     this.props.onCreateLibrary(false);
                                 }
                             });
```
shaoshing: These also have extra indents.

src/js/jsx/tools/GuidesOverlay.jsx
```diff
@@ -301,17 +301,17 @@ define(function (require, exports, module) {
...
+                            .classed("guide-edges__hover", false);
                             
-                            self._currentMouseDown = true;
+                        self._currentMouseDown = true;
                             
-                            d3.event.stopPropagation();
-                        })
+                        d3.event.stopPropagation();
+                    })
                         .on("mouseout", function () {
```
shaoshing: :point_left: :)

src/js/jsx/sections/libraries/assets/AssetPreviewImage.jsx
```diff
@@ -110,8 +110,8 @@ define(function (require, exports, module) {
                     librariesAction.RENDITION_GRAPHIC_SIZE : librariesAction.RENDITION_DEFAULT_SIZE;
 
             this._renditionUpdate = Promise.fromCallback(function (cb) {
-                    element.getRenditionPath(renditionSize, cb);
-                })
+                element.getRenditionPath(renditionSize, cb);
+            })
                 .bind(this)
```
shaoshing: :point_left: indent

src/js/jsx/sections/nodoc/ArtboardPresets.jsx
```diff
@@ -64,7 +64,7 @@ define(function (require, exports, module) {
                             <span>{template.width} x {template.height}</span>
                         </li>
                     );
```
shaoshing: :point_left: indent

src/js/models/exportasset.js
```diff
@@ -73,8 +73,8 @@ define(function (require, exports, module) {
                 return formatOption.format === format && formatOption.quality === quality;
             });
         }, function (format, quality) {
-            return format + quality;
-        });
```
shaoshing: maybe this should be 

```
    var _findFormatOptionIndex = _.memoize(
        function (format, quality) {
            return FORMATS_OPTIONS.findIndex(function (formatOption) {
                return formatOption.format === format && formatOption.quality === quality;
            });
        }, function (format, quality) {
            return format + quality;
        }
    );
```

src/js/models/layerstructure.js
```diff
@@ -1103,10 +1103,10 @@ define(function (require, exports, module) {
...
             };
         }.bind(this),
-        {
-            layers: nextStructure.layers,
-            index: nextStructure.index
-        });
+            {
+                layers: nextStructure.layers,
+                index: nextStructure.index
+            });
```
shaoshing: The object is the second param of the `reduce` function. How about this?

```
}.bind(this), { layers: nextStructure.layers, index: nextStructure.index })
```

src/js/util/gridsort.js
```diff
@@ -105,8 +105,8 @@ define(function (require, exports, module) {
         // sort left-to-right, then reduce them into a grid.
         // note the use of '_.ary()' to pass only the first two reducer args into _addLayerToRow
         var grid = layers.filter(function (layer) {
-                return layer.bounds && !layer.bounds.empty;
-            })
+            return layer.bounds && !layer.bounds.empty;
+        })
             .sort(function (a, b) {
```
shaoshing: Extra indents.

shaoshing:
@iwehrman thanks for upgrading and fixing all the issues. `eslint` looks pretty sweet! :+1: Before merging, I noticed some new idention problems due to the changes, so back to you.

src/js/actions/libraries.js
volfied: Yeah this happens in multiple places, I think the format @shaoshing suggested above is the cleanest way.

iwehrman:
@shaoshing @volfied: thanks for taking a look. In some cases I agree that the indentation is incorrect but in most cases I think eslint is putting things in the "correct" place. In particular, I think this is "correct":

```javascript
var foo = new Promise(function (resolve) {
    resolve();
})
    .bind(this);
```

At least, I think this formatting is a valid consequence of the four-space indentation rule that eslint is enforcing. At first I didn't like it, but now I think it's actually reasonable. In any case, I'm not sure how I would explain to eslint that it should be any other way. (JSCS is agnostic about this; it accepts either indentation.)

I'll look into fixing the other issues, but in the meantime, what's your reaction to that?

iwehrman:
(Also, I agree with the @shaoshing's suggested formatting workaround when it's possible.)

src/js/actions/search/libraries.js
iwehrman: Good question...

src/js/models/exportasset.js
iwehrman: ESLint didn't like that, so I went with this instead:
```javascript
    var _findFormatOptionIndex = _.memoize(
        function (format, quality) {
            return FORMATS_OPTIONS.findIndex(function (formatOption) {
                return formatOption.format === format && formatOption.quality === quality;
            });
        },
        function (format, quality) {
            return format + quality;
        });
```

iwehrman:
OK, I've used @shaoshing's suggested workaround for all the oddities he pointed out. Back to @shaoshing!

shaoshing:
Superb!

iwehrman:
But don't forget adobe-photoshop/spaces-adapter#183 :)

PR #3641, opened 2016-01-29
Get rid of synchronization.debounce and use _.debounce instead

volfied:
After realizing that in Mar 19, 2015, we got rid of the async waiting nature of our own debounce function, and it's identical to `_.debounce`, this cleans it up so we use lodash.

Addresses #3639, @iwehrman please review. @mcilroyc it would be nice to have your eyes on this, as well.

### DIFF
--- a/src/js/fluxcontroller.js
+++ b/src/js/fluxcontroller.js
@@ -68,14 +68,6 @@ define(function (require, exports, module) {
     var THROTTLED_ACTION_SUFFIX = "Throttled";
 
     /**
-     * Suffix used to name debounced actions.
-     *
-     * @const
-     * @type {String}
-     */
-    var DEBOUNCED_ACTION_SUFFIX = "Debounced";
-
-    /**
      * Maximum delay after which reset retry will continue
      *  before failing definitively.
      *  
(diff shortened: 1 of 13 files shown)


mcilroyc:
I've never believed that using debounce with a `wait` of zero was a reasonable thing to do ... or least that it is never what we want to do in design space.  Now may be as good a time as any to argue about this.  There's only a couple of places that still provide no explicit wait value (thus defaulting to zero).

mcilroyc:
I suppose I forgot to mention: this all looks correct to me as a direct refactoring.  I guess we never used the "immediate" feature which would have been equivalent to the the "leading" lodash option.  Just saying that out loud.

src/js/fluxcontroller.js
```diff
@@ -927,7 +927,7 @@ define(function (require, exports, module) {
                 debouncedName = name + DEBOUNCED_ACTION_SUFFIX,
                 synchronizedAction = this._synchronize(namespace, module, name),
                 throttledAction = synchronization.throttle(synchronizedAction),
-                debouncedAction = synchronization.debounce(synchronizedAction);
+                debouncedAction = _.debounce(synchronizedAction);
```
iwehrman: Re: @mcilroyc's concerns about zero-delay debouncing, which I completely agree with, I wonder if we shouldn't just take this opportunity to remove these auto-debounced actions as well.

iwehrman: Also, a quick grep did not turn up any instances in which we're actually using those actions.

volfied:
@mcilroyc @iwehrman Thank you for reviews. I've cleaned up the actions code so we no longer have debounced suffix, and also the untimed debounce call for artboard Transform handler.

iwehrman:
LGTM, thanks!

Issue #3628, opened 2016-01-27
Datalists colors are inconsistent (selections hard to read on some themes)


chadrolfs:
compare search, cc libraries and blend modes datalists (effect and appearance).

blend modes is lighter and highlight is harder to read.

chadrolfs:
<img width="184" alt="screen shot 2016-01-27 at 3 48 51 pm" src="https://cloud.githubusercontent.com/assets/8442631/12632602/faf80ac0-c50d-11e5-9257-0b44ce5d9476.png">
<img width="190" alt="screen shot 2016-01-27 at 3 49 14 pm" src="https://cloud.githubusercontent.com/assets/8442631/12632601/faf77c54-c50d-11e5-8df7-863e02604139.png">

chadrolfs:
related to #3345

iwehrman:
Might need some input from @placegraphichere here first.

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
yeppers

PR #3642, opened 2016-01-29
Coordinated datalist and highlight colors for search, effects, appear…

DivyaPrabhakar:
Refers to issue #3628 and #3739 

All the dialogs should have the same background colors and outlines: 

Final datalist dialog: 

Final keyboard shortcut dialog: 

Final export dialog: 

Final introduction screen dialog:

### DIFF
--- a/src/style/search/search-bar.less
+++ b/src/style/search/search-bar.less
@@ -31,7 +31,6 @@
     background-position: 1.7rem 1.5rem;
     top: 25%;
     left: 50%;
-    transform: none;
     transform: translate(-50%, 0); /*this is important so it won't bounce when typing in the search field */
     
     width: @search-bar-width;
@@ -131,7 +130,8 @@
 }
 
 .search-bar__dialog .select__option {
-    padding: 0.5rem 0.8rem 0.7rem 0.8rem;
+    padding: 0.5rem 0.8rem 0.7rem 0.7rem;
+    margin: 0;
     border-left: .3rem solid transparent;
     display: flex;
     overflow: hidden;
@@ -156,9 +156,9 @@
 }
 
 .search-bar__dialog .select__option__selected {
-    color: @text-highlight;
+    color: @text-focus;
     border-left: .3rem solid @focus-highlight;
-    background-color: @bg-app;
+    background-color: @bg-dialog;
     
     svg {
         fill: @focus-highlight;
--- a/src/style/shared/colors-medium.less
+++ b/src/style/shared/colors-medium.less
@@ -26,6 +26,7 @@
 @bg-border: #383838;
 @bg-panel: #535353;
 @bg-alt: #424242;
+@bg-dialog: #282828;
 
 //text colors
 @item: #e6e6e6;
@@ -35,7 +36,7 @@
 @item-disabled: #757575;
 
 //icon colors 
-@icon-hover: #f0f0f0;
+@icon-hover: #F0F0F0;
 @icon-active: #D7D7D7;
 @icon-inactive: #DDDDDD;
 @icon-disabled: #919191;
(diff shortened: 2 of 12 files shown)


placegraphichere:
So the `var` we're using is for "text highlighting". But in this instance we are trying to set the color of the text itself. I think we need a new color for the text: color. 

The contrast for the background color on datalists is also problematic trying to find a color that suits the darker color but is useful elsewhere in the app for that color theme.

I'll add some more info on this shortly.

DivyaPrabhakar:
@placegraphichere When you hover over datalist options, isn't that highlighting the text? I am confused about what you mean by we are "setting the color of the text" -- can you explain?

iwehrman:
I think this is on @placegraphichere's plate currently, so assigning to him.

src/style/search/search-bar.less
```diff
@@ -29,10 +29,11 @@
     background: no-repeat left @bg-panel;
     background-size: @text-xxlarge;
     background-position: 1.7rem 1.5rem;
-    top: 25%;
-    left: 50%;
+    top: round(25%);
```
iwehrman: This looks suspicious to me. I would guess that this is equivalent to `25%`, but I'm not sure. Registering this concern for posterity.

iwehrman:
The main blocker is that the search bar is not correctly positioned in this branch. It seems to be positioned in the middle of the screen instead of in the middle of the canvas:
image

iwehrman:
Colors look great. Back to @DivyaPrabhakar to fix the positioning issue.

src/style/search/search-bar.less
designedbyscience: Agreed...this is a LESS function which doesn't seem to make it into the final CSS. It is translated to 25%...

iwehrman:
I've amended the commit to fix the positioning problem, which I think was due to repeated `transform` rules. I think this is ready to merge, but I'm unassigning myself so that someone else can double-check it.

placegraphichere:
I'm still seeing the gap between the bar and edge of the hovered item and the search input. I thought we got this sorted, but it's still not. I believe this is due to the overall dialog positioning being on a half px.

screen shot 2016-02-10 at 5 37 19 pm

screen shot 2016-02-10 at 5 39 00 pm

It'd be great to just round this position to a whole px. Otherwise, this is all great stuff.

placegraphichere:
So on a retina, if you just resize the app frame to be slightly smaller on an odd pixel (1907x 1178 was what I resized it to), usually you'll see this occur. If it's full screen, it will look correct.

iwehrman:
AFAICT this is back in @DivyaPrabhakar's court.

iwehrman:
FWIW, I can't repro this issue. I tried to get the window as close as possible to the resolution that @placegraphichere suggested, but it still looks good to me.

ktaki:
Tried with various resolutions and app windows sizes on MacBook Pro retina. I do not see the gap @placegraphichere mentioned either.

DivyaPrabhakar:
@placegraphichere Can you can sign off on the changes to the keyboard shortcut dialog? After that, I can offer this PR up to the team so someone can review and merge it.

src/style/shared/colors-medium.less
```diff
@@ -59,6 +60,7 @@
 
 @focus-highlight: #3691ff;
 @text-highlight: #1B8BE9;
+@text-focus: #1473e6;
```
volfied: For code cleanliness sake, can you make sure hex letters are uppercase like the examples above?

src/style/shared/colors-original.less
```diff
@@ -26,6 +26,7 @@
 @bg-border: #CCCCCC;
 @bg-panel: #f0f0f0;
 @bg-alt: #DBDBDB;
+@bg-dialog: #FFF;
```
volfied: Is this short hand for FFFFFF or F0F0F0 or 0F0F0F? You get my point :)

DivyaPrabhakar:
@volfied style changes made

iwehrman:
LGTM!

PR #3643, opened 2016-01-29
Add react-addons-perf to shrinkwrap / deps

volfied:
On a fresh clone, `npm install` fails, because React just released 0.14.7, and `react-addons-perf`is not shrinkwrapped (shrinkwrap does not affect dev dependencies)

So I moved it to dependencies, and also updated all the other React dependencies to 0.14.6

This is painful.


shaoshing:
LGTM!

PR #3645, opened 2016-01-29
Correctly reference nested action paths

### LINKED ISSUE #3644: Switch to DS with Pen tool selected will result in unrecoverable error
Select Pen tool in standard Ps and switch to DS, then load current master. It does not matter a doc is opened or not.

screen shot 2016-01-29 at 2 45 39 pm

volfied:
Not the best change, but gets things back to working status.

Maybe instead of dot separated paths, we should use nested paths to camelCasedNestedPaths... but then there are chances of clashes.

For now, this will fix #3644


mcilroyc:
Good enough to get us through...

PR #3646, opened 2016-01-29
Allow nested actions

volfied:
Continuation of #3645, this will allow us to refer to actions by dot separated path names in `this.flux.actions` object.

So instead of `this.flux.actions["tool.type"].select`, we can type `this.flux.actions.tool.type.select`

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -35,6 +35,7 @@ define(function (require, exports) {
     var events = require("js/events"),
         locks = require("js/locks"),
         system = require("js/util/system"),
+        objUtil = require("js/util/object"),
         log = require("js/util/log"),
         headlights = require("js/util/headlights"),
         policyActions = require("./policy"),
@@ -428,7 +429,7 @@ define(function (require, exports) {
             actionModuleName = actionNameParts[0],
             actionName = actionNameParts[1],
             actionNameThrottled = actionName + "Throttled",
-            actionThrottled = this.flux.actions[actionModuleName][actionNameThrottled];
+            actionThrottled = objUtil.getPath(this.flux.actions, actionModuleName)[actionNameThrottled];
 
         return actionThrottled;
     };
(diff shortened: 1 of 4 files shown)


src/js/fluxcontroller.js
```diff
@@ -958,9 +970,23 @@ define(function (require, exports, module) {
...
-            var rawModule = modules[moduleName];
+            var rawModule = modules[moduleName],
+                syncModule = this._synchronizeModule(moduleName, rawModule);
+
+            var modulePathArray = moduleName.split("."),
+                modulePath = _.initial(modulePathArray),
+                moduleBasename = _.last(modulePathArray),
+                root = exports;
+
+            modulePath.forEach(function (path) {
```
iwehrman: Can you add a little comment to explain what is going on here?

src/js/actions/menu.js
```diff
@@ -428,7 +429,7 @@ define(function (require, exports) {
             actionModuleName = actionNameParts[0],
             actionName = actionNameParts[1],
             actionNameThrottled = actionName + "Throttled",
-            actionThrottled = this.flux.actions[actionModuleName][actionNameThrottled];
+            actionThrottled = objUtil.getPath(this.flux.actions, actionModuleName)[actionNameThrottled];
```
iwehrman: Would this ideally be `FluxController.prototype.getModule` instead of `objUtil.getPath`?

iwehrman: Just asking because I really wonder if we should make it easier to get ahold of the main `FluxController` instance somehow...

iwehrman:
One request for documentation, one question. Back 2 @volfied.

src/js/actions/menu.js
volfied: Yes, ideally. But as you said, it's not exposed. I think exposing FluxController will allow us to shoot ourselves in the foot easier, and not having it readily available makes us limit calls like this, which is a good thing.

volfied:
I've actually found a bug (that we wouldn't have hit with current webpack code) in module synchronization.

Fixed it, with an explanation.

Back to you @iwehrman

iwehrman:
Okie doke.

PR #3647, opened 2016-02-01
Panel/Column Toggle Animations

designedbyscience:
I've updated a previous branch to work with more recent changes to spaces-design. For now, I will just tag this as Review Only

Below are some timelines for expanding/collapsing the columns with OSR on and off
**OSR: true**
Non-Retina
osr_true_non-retina

Retina
osr_true_retina

**OSR: false**
Non-Retina
osr_false_non-retina

Retina
osr_false_retina

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -414,11 +414,13 @@ define(function (require, exports, module) {
                 }
             } else if (this.state.recentFilesInitialized) {
                 return (
-                    <div className="panel-set panel-set__active null-state">
-                        <PanelColumn ref="panelSet" visible="true">
-                            <RecentFiles recentFiles={this.state.recentFiles} />
-                            <ArtboardPresets />
-                        </PanelColumn>
+                    <div className="panel-set__container panel-set__active null-state">
+                        <div ref="panelSet" className="panel-set">
+                            <PanelColumn ref="panelSet" visible="true">
+                                <RecentFiles recentFiles={this.state.recentFiles} />
+                                <ArtboardPresets />
+                            </PanelColumn>
+                        </div>
                         <IconBar />
                     </div>
                 );
--- a/src/style/panel.less
+++ b/src/style/panel.less
@@ -28,9 +28,6 @@
 
 /* -- PROPERTIES -- */
 .panel-set {
-    height: 100%;
-    max-height: 100%;
-    display: flex;
     box-shadow: inset @panel-border-width 0 0 @bg-border-for-canvas;
 }
 
@@ -84,17 +81,29 @@
     box-shadow: inset 0 (0.2rem) 0 @bg-border;
     border-right: @panel-border-width solid @bg-border;
     z-index: 1;
-    display: none;
-    position: relative;
+    left: 0;
+    transition: left 0.2s linear;
+    position: absolute;
     flex-direction: column;
+    // display: none;
 }
 
 .panel__visible + .panel:nth-child(2){
     border-right: @panel-border-width solid @bg-border;
+    left: 32rem;
 }
 
 .panel__visible{
+    transform: translateX(0);
     display: flex;
+}
+
+.panel + .panel__visible:nth-child(2) {
+    left: 0;
+}
+
+.panel.panel__visible + .panel__visible:nth-child(2) {
+    left: 32rem;
 }
 
 .panel__element {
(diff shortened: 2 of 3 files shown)


designedbyscience:
Sidenote: also tried this morning to animate `translate` instead of `left`

Doesn't look as good and no major difference in framerate

Timeline for Retina, use_osr: false

screen shot 2016-02-02 at 11 39 19 am

designedbyscience:
I called this review only mostly because it doesn't perform that well, and we'd need to live with that in master if merged

placegraphichere:
I'm super happy we're looking at this. That said, I agree with @designedbyscience that this is probably not performant enough (at this point) to be merged.

src/js/jsx/PanelSet.jsx
```diff
@@ -414,11 +414,13 @@ define(function (require, exports, module) {
                 }
             } else if (this.state.recentFilesInitialized) {
                 return (
-                    <div className="panel-set panel-set__active null-state">
-                        <PanelColumn ref="panelSet" visible="true">
-                            <RecentFiles recentFiles={this.state.recentFiles} />
-                            <ArtboardPresets />
-                        </PanelColumn>
+                    <div className="panel-set__container panel-set__active null-state">
+                        <div ref="panelSet" className="panel-set">
```
iwehrman: Seems like a bug for two different elements to have the same `panelSet`ref?

src/style/panel.less
```diff
@@ -84,19 +81,31 @@
     box-shadow: inset 0 (0.2rem) 0 @bg-border;
     border-right: @panel-border-width solid @bg-border;
     z-index: 1;
-    display: none;
-    position: relative;
+    left: 0;
+    transition: left 0.2s linear;
+    position: absolute;
     flex-direction: column;
+    // display: none;
```
iwehrman: Stale comment.

iwehrman:
The appearance panel doesn't animate when collapsing like the other panels.

iwehrman:
When the libraries panel collapses down, its contents seem to be hidden before the animation starts, which seems weird:
img

iwehrman:
The same basic thing seems to happen with the export panel, though it collapses up (for some reason):
img

iwehrman:
Closing the last open file flashes pretty badly:
img

iwehrman:
Back into the animation mines!

designedbyscience:
Updated, however there is an issue with cloaking currently

designedbyscience:
Updated again, cloaking issues fixed. Pulled some of the logic about panel state and CSS classes into the panel set

designedbyscience:
To check the animations, you need to add a feature flags object to the `manifest.json` with a property `use_animations:true`

iwehrman:
Needs moar rebase :8ball:

src/js/jsx/Main.jsx
```diff
@@ -148,18 +137,20 @@ define(function (require, exports, module) {
...
-            var hasDocument = this.getFlux().store("application").getCurrentDocument(),
-                colorStop = this.state.colorStop || this.props.initialColorStop,
+            /* global _spaces */
+            var colorStop = this.state.colorStop || this.props.initialColorStop,
                 stopClass = "color-stop__" + colorStop.toLowerCase();
 
+            // To enable animations, add a feature flags object to manifest.json and then
+            // set the property use_animations:true in that object. You will need to restart Photoshop
+            // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+            var useAnimations = !!(_spaces.feature_flags && _spaces.feature_flags.use_animations);
```
iwehrman: I think this can just be a constant instead of recalculating it on every render?

iwehrman:
It also seems like the document header does not animate when opening a panel; it just gets pushed over in one step, and then the panel position is animated. I guess I'd be OK with merging it anyway if it were off by default, but the animation still seems to be enabled despite a lack of `feature_flags` entries in my settings.json file. Any ideas?

designedbyscience:
1) There was an error that I let some things animate despite the flag (fixed now). This also accounted for the header not animating when the panels did
2) It is the `manifest.json` file, not settings.json, where you want to add that feature flag

designedbyscience:
Back to @iwehrman.

I also have another branch I started to particularly improve the open/close section animation, where I can explicitly animate height rather than fool around with min-height

PR #3648, opened 2016-02-01
Make grunt tasks check for dependencies

shaoshing:
**`npm install` is required**

This PR uses the `grunt-check-dependencies` package to add the ability to check for package dependencies before running some grunt tasks, and will inform you when `npm install` is needed. Below are the tasks that will check dependencies:

* grunt test (npm test)
* grunt compile
* grunt debug
* when running `grunt debug`, it will keep monitoring the `package.json` file for change of dependencies.

screen shot 2016-02-01 at 11 43 02 am

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -184,6 +184,10 @@ module.exports = function (grunt) {
                 options: {
                     interval: 500
                 }
+            },
+            dependencies: {
+                files: ["package.json"],
+                tasks: ["checkDependencies"]
             }
         },
         // Build tasks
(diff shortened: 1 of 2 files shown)


volfied:
Is this going to work offline? It would suck to block the build progress if there are errors with package checking, can you make it warn only instead?

shaoshing:
@volfied that's a good question. The package itself does not support silenced warning, but you can use the `--force` option to ask Grunt to ignore the warnings. Given that we rarely work without internet, I think this is a good enough workaround? Please let me know your thoughts.

volfied:
Alright, I'm semi-convinced this helps. We'll see how it goes!

PR #3649, opened 2016-02-01
More modal-safe actions

iwehrman:
1. Mark most actions that do not write PS doc or app state as modal safe. Playing `get` commands _should_ work even when PS is in a modal state, and the adapter should not require those to be played with the `canExecuteWhileModal` flag. Small consequence: you can now pan the canvas without ending the modal state.
2. Make `setGroupExpansion` and `setColorStop` modal safe even though they do write PS doc state.
3. Instrument the controller so that it fails fast if there are duplicate read/write locks, or unexpected action properties, both of which are indicative of typos. 
4. Fix a few typos which caused both of the aforementioned bug instances, and also add a few other missing locks.

### DIFF
--- a/src/js/actions/guides.js
+++ b/src/js/actions/guides.js
@@ -225,7 +225,7 @@ define(function (require, exports) {
     };
     createGuideAndTrack.action = {
         reads: [locks.JS_UI],
-        writes: [locks.JS_DOC]
+        writes: [locks.JS_DOC, locks.PS_DOC]
     };
 
     /**
@@ -298,7 +298,7 @@ define(function (require, exports) {
     };
     deleteGuide.action = {
         reads: [],
-        writes: [locks.JS_DOC],
+        writes: [locks.JS_DOC, locks.PS_DOC],
         transfers: [resetGuidePolicies]
     };
 
@@ -405,7 +405,8 @@ define(function (require, exports) {
     };
     queryCurrentGuides.action = {
         reads: [locks.PS_DOC],
-        writes: [locks.JS_DOC]
+        writes: [locks.JS_DOC],
+        modal: true
     };
 
     // Event handlers for guides
@@ -458,9 +459,9 @@ define(function (require, exports) {
         return Promise.resolve();
     };
     beforeStartup.action = {
-        modal: true,
         reads: [],
         writes: [],
+        modal: true,
         transfers: []
     };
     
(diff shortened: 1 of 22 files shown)


src/js/actions/guides.js
```diff
@@ -225,7 +225,7 @@ define(function (require, exports) {
     };
     createGuideAndTrack.action = {
         reads: [locks.JS_UI],
-        writes: [locks.JS_DOC]
+        writes: [locks.JS_DOC, locks.PS_DOC]
```
iwehrman: This and the next action were missing a needed `PS_DOC` lock.

src/js/actions/transform.js
```diff
@@ -1370,7 +1370,7 @@ define(function (require, exports) {
         }
     };
     handleTransformLayer.action = {
-        read: [locks.JS_APP, locks.JS_DOC],
```
iwehrman: Incorrect action property typo.

src/js/actions/transform.js
```diff
@@ -779,7 +779,7 @@ define(function (require, exports) {
     };
     flipY.action = {
         reads: [],
-        writes: [locks.JS_DOC, locks.JS_DOC],
+        writes: [locks.JS_DOC, locks.PS_DOC],
```
iwehrman: Redundant lock typo.

src/js/actions/type.js
```diff
@@ -817,7 +817,7 @@ define(function (require, exports) {
     beforeStartup.action = {
         reads: [],
         writes: [],
-        modal: []
+        modal: true
```
iwehrman: Potential future work: manually check the types of these properties to guard against typos like the ones in this file.

src/js/actions/superselect.js
```diff
@@ -449,6 +449,7 @@ define(function (require, exports) {
     click.action = {
         reads: [locks.PS_DOC, locks.JS_DOC, locks.JS_TOOL, locks.JS_UI],
         writes: [],
+        modal: true,
```
volfied: These all write to the doc state (change selection).

volfied:
Just one question @iwehrman but rest looks good.

Do we have any plans on warning when transferring from modal safe to non modal safe functions?

src/js/actions/superselect.js
iwehrman: No, they transfer to actions that write to the doc state, which is why they don't have write locks. Marking them as modal delays canceling the modal state until the transfer, if that actually occurs.

volfied:
Ok, confusion cleared!

Issue #3650, opened 2016-02-02
Cannot add effects


ktaki:
Open new doc, add a shape, click '+' in the Effects panel, and choose an effect (e.g., Stroke).

Uncaught TypeError: Cannot read property 'addEffect' of undefined

![screen shot 2016-02-01 at 4 58 19 pm](https://cloud.githubusercontent.com/assets/2029597/12736453/5f067876-c905-11e5-8d90-382e21119940.png)

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3651, opened 2016-02-02
Use actions.layereffects with correct casing

volfied:
Addresses #3650, we were using `actions.layerEffects`, but file name is `layereffects` so casing had to be changed.

### DIFF
--- a/src/js/jsx/tools/SamplerOverlay.jsx
+++ b/src/js/jsx/tools/SamplerOverlay.jsx
@@ -458,7 +458,7 @@ define(function (require, exports, module) {
                         .on("click", applyTypeStyleFunc);
                 } else if (sample.type === "layerEffects") {
                     var duplicateLayerEffectsFunc = function () {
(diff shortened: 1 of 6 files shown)


src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -458,7 +458,7 @@ define(function (require, exports, module) {
                         .on("click", applyTypeStyleFunc);
                 } else if (sample.type === "layerEffects") {
                     var duplicateLayerEffectsFunc = function () {
-                        fluxActions.layerEffects.duplicateLayerEffects(this.state.document, targetLayers, sample.value);
+                        fluxactions.layereffects.duplicateLayerEffects(this.state.document, targetLayers, sample.value);
```
iwehrman: I think the first variable should not have been renamed.

volfied:
Sorry @iwehrman, didn't finish my tea yet! :$

I rebased

PR #3652, opened 2016-02-02
Unify mask commands to all have the same result

pineapplespatula:
Adding code to the two paths into vector masks so that they have the same results depending upon the layer selection

### DIFF
--- a/src/js/models/layerstructure.js
+++ b/src/js/models/layerstructure.js
@@ -581,11 +581,30 @@ define(function (require, exports, module) {
         "selectedLayersCanHaveVectorMask": function () {
             var layers = this.selected,
                 curLayer = layers.first();
-            return !layers.isEmpty() &&
+            return layers.size === 1 &&
                 !curLayer.isGroupEnd &&
                 !curLayer.isVector &&
                 !curLayer.isBackground &&
                 !curLayer.locked;
+        },
+
+        /**
+         * Determine whether a vector mask can be created from a shape on the current layer selection
+         *
+         * @return {boolean} if the layer selection can handle a vector mask
+         */
+        "selectedLayersCanMakeVectorMaskFromShape": function () {
+            var layers = this.selected;
+            return layers.size === 2 &&
+                layers.some(function (curLayer) {
+                    return curLayer.isVector;
+                }) &&
+                layers.some(function (curLayer) {
+                    return !curLayer.isGroupEnd &&
+                    !curLayer.isVector &&
+                    !curLayer.isBackground &&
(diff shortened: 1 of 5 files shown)


src/js/models/menubar.js
```diff
@@ -274,14 +274,18 @@ define(function (require, exports, module) {
...
-                (document.layers.selectedNormalized.size === 2) &&
+                (((document.layers.selectedNormalized.size === 2) &&
                 (document.layers.selected.some(function (layer) {
                     return layer.isVector;
                 })) &&
                 (document.layers.selected.some(function (layer) {
                     return document.layers.canSupportVectorMask(layer) &&
                         !layer.vectorMaskEnabled;
-                }))
+                }))) ||
```
iwehrman: There sure are a lot of parentheses here...

iwehrman:
Does this address a particular bug? Can you say a little more about what you mean about mask commands having the "same result"?

pineapplespatula:
this is in reference to the first check mark on my JIRA task for vector masks. basically regardless of how you enter mask mode you will get a the correct result based upon how many layers you have selected. 

CMD M with 1 layer selected enters mask mode. 
m or toolbar button with a shape and pixel layer selected creates a vector mask from the shape

the thing that isnt done here is dynamically changing the menu text to reflect the resulting action.

src/js/models/layerstructure.js
```diff
@@ -582,6 +582,7 @@ define(function (require, exports, module) {
             var layers = this.selected,
                 curLayer = layers.first();
             return !layers.isEmpty() &&
+                layers.size === 1 &&
```
iwehrman: This conjunct implies the previous one, so the previous one can be omitted.

src/js/models/layerstructure.js
```diff
@@ -589,6 +590,26 @@ define(function (require, exports, module) {
...
 
         /**
+         * Determine whether a vector mask can be created from a shape on the current layer selection
+         *
+         * @return {boolean} if the layer selection can handle a vector mask
+         */
+        "selectedLayersCanMakeVectorMaskFromShape": function () {
+            var layers = this.selected;
+            return !layers.isEmpty() &&
+                layers.size === 2 &&
```
iwehrman: This conjunct implies the previous one, so the previous one can be omitted.

iwehrman:
Thanks for clarifying. I found some nits. So, back to @pineapplespatula for that.

pineapplespatula:
addressed nits @iwehrman  :repeat:

Issue #3653, opened 2016-02-02
Marquee select does not work


ktaki:
Open new doc. Create a few shape layers. Try marquee select. No shapes are selected.

chadrolfs:
this just broke.  I was working this morning

ktaki:
Verified the fix.

PR #3654, opened 2016-02-02
Re-add marquee rectangles

volfied:
Looks like we got a little over zealous with the lint cleanup in #3640 and removed the marquee rectangles from the overlay, breaking marquee select.

This re-adds them, addressing #3653


mcilroyc:
LGTM

Issue #3581, opened 2016-01-20
Shortcut to undo fails when reordering layer groups


chadrolfs:
create 3 shape layers 
select one at a time (in the layers panel) and cmd + g so that each shape is in their own group
collapse all 3 groups
drag the middle group above the top group (reorder the stacking order of the groups)
cmd + z

Result: nothing
workaround: use the menu command

iwehrman:
So it works completely correctly if you choose `Edit > Undo`?

chadrolfs:
yup

chadrolfs:
seems to work fine on windows

chadrolfs:
Here's another case where it fails:
create two shapes, select them both, click on one of the combine icons (e.g. Unite), cmd+z - doesn't work.

volfied:
Seems like selecting the layers in the panel is the cause of this, we're probably stopping propagation.

volfied:
This is due to fact that we call acquireFocus on mousedown for LayerFace's TextInput, but never call releaseFocus. Not sure how to fix it, but I'm leaving this note here so the next dev doesn't have to triage it.

iwehrman:
FBNC to @chadrolfs.

PR #3657, opened 2016-02-03
Add acquireFocusOnMouseDown prop to Focusable, set to false in LayerFace

volfied:
When the name of the layer in a LayerFace was clicked on, we would immediately call `OS.acquireKeyboardFocus` due to the `mousedown` listener in Focusable mixin (which is the mixin of TextInput showing layer name). But we would never call `OS.releaseKeyboardFocus` because those were only called on commit/cancel/finish methods.

For TextInputs like layer name, we don't want to acquire focus unless the user double clicks to start editing. We instead acquire focus on `_beginEdit` function which is what happens when user double clicks.

Addresses #3581

### DIFF
--- a/src/js/jsx/mixin/Focusable.js
+++ b/src/js/jsx/mixin/Focusable.js
@@ -24,13 +24,24 @@
 define(function (require, exports, module) {
     "use strict";
 
-    var ReactDOM = require("react-dom");
+    var ReactDOM = require("react-dom"),
+        React = require("react");
 
     var OS = require("adapter").os;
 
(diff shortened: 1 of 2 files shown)


iwehrman:
Do we need to explicitly call `acquireFocus` once the `TextInput` becomes editable?

volfied:
No @iwehrman, _beginEdit calls acquireFocus. It is called by both click, focus and doubleclick based on different scenarios.

iwehrman:
Roger. In that case: looks good!

PR #3659, opened 2016-02-03
User interaction batchPlay option

iwehrman:
1. Set the `isUserInteractionCommand` batchPlay option when coalescing `transform.setRadius`, `shapes.setFill*`, `shapes.setStroke*` and `type.setType*` actions. (I think this only affects the color picker and the radius slider, at the moment, since the other actions that could be affected by this change aren't ever coalesced.) I can't actually measure any performance difference with this change, but @jsbache wants it and I want him to be happy! 
2. Also, just because, double the menu-update debounce timeout from 200ms to 400ms to make it less likely to trigger during user interactions. 200ms seems a little optimistic for a lot of computers...


mcilroyc:
LGTM.  cc; @chadrolfs

PR #3661, opened 2016-02-03
document load time: CC Libraries optimization

shaoshing:
This PR includes the following optimizations on CC Libraries to improve performance.

* Defer CC Libraries loading/rendering process to improve startup performance. Before, CC Libraries is loaded in the `beforeStartup` callback, which slows down the startup process and triggers libraries panel rendering, which slow down the startup further. This process can be deferred as the chance of using the librarise panel right after startup is small. 
* Prevent the Libraries panel DOM from removing when closing the last document. Otherwise, we have to waste time in rendering the panel again when creating a new document.
* Pre-render the Libraries panel in no-doc state. While in no-doc state, the libraries panel can be pre-rendered (by React) when the data is ready. 

Below is the result of the before/after test in various scenarios:

|  | Before | After | Diff |
| --- | --- | --- | --- |
| no-doc startup | 869ms | **758ms** | -111ms |
| empty doc startup | 1741ms | **828ms** | -903ms |
| vermillion startup | 2562ms | **1640ms** | -922ms |
| no-doc -> empty doc (scripting) | 691ms | **252ms** | -439ms |
| no-doc -> empty doc (rendering) | 71ms | **54ms** | -17ms |
| empty doc -> no-doc (scripting) | 208ms | 208ms | 0 |
| empty doc -> no-doc (rendering) | **5ms**  | 9ms | 4ms |
(selected library: `Samples Collection April 23 2015`)

### DIFF
--- a/src/js/actions/libraries.js
+++ b/src/js/actions/libraries.js
@@ -55,13 +55,6 @@ define(function (require, exports) {
         historyActions = require("./history"),
         preferencesActions = require("./preferences"),
         LibrarySyncStatus = require("js/models/library_sync_status");
-
-    /**
-     * The external CC Libraries API is loaded here asynchronously, used in beforeStartup
-     *
-     * @type {Promise}
-     */
-    var apiLoadPromise = Promise.promisify($S)("file://shared/libs/cc-libraries-api.min.js");
         
     /**
      * For image elements, their extensions signify their representation type
@@ -1273,14 +1266,35 @@ define(function (require, exports) {
         _toolModalStateChangedHandler = function (event) {
             var isPlacingGraphic = this.flux.store("library").getState().isPlacingGraphic,
                 modalStateEnded = event.state && event.state._value === "exit";
-
+        
             if (isPlacingGraphic && modalStateEnded) {
                 this.flux.actions.libraries.handleCompletePlacingGraphic();
             }
         }.bind(this);
         descriptor.addListener("toolModalStateChanged", _toolModalStateChangedHandler);
 
-        return apiLoadPromise
+        return Promise.resolve();
+    };
+    beforeStartup.action = {
+        reads: [locks.JS_PREF],
+        writes: [locks.JS_LIBRARIES, locks.CC_LIBRARIES]
+    };
+
+    /**
+     * Libraries loadedCollections event handlers.
+     *
+     * @private
+     * @type {function}
+     */
+    var _handleLibrariesLoadedHelper;
+
+    /**
+     * After startup, load the libraries
+     *
+     * @return {Promise}
+     */
+    var afterStartup = function () {
+        return Promise.promisify($S)("file://shared/libs/cc-libraries-api.min.js")
             .timeout(3000, "CC Libraries API load timeout, please don't restart and notify the chatroom!")
             .bind(this)
             .then(function () {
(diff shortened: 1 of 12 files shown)


chadrolfs:
:moneybag:

src/js/stores/panel.js
```diff
@@ -248,6 +248,7 @@ define(function (require, exports, module) {
             var centerOffsets = this.getCenterOffsets(),
                 windowWidth = window.document.body.clientWidth,
                 windowHeight = window.document.body.clientHeight;
+
```
iwehrman: :space_invader:

(Weird that nothing caught this...)

shaoshing: I think that's because the new line has spaces, which is not considered as extra line break.

shaoshing: Maybe it is time to remove all trailing spaces and add validation for it.

iwehrman:
One concern with this is that the rest of the panel contents fade in when they're ready, but the Libraries panel now does not, and it feels just a little off. Can you do some experiments to try to improve this? Like, try fading in the libraries panel content when it's ready?

iwehrman:
Or, alternatively, rendering a spinner when first showing the UI and replacing that spinner with the contents when ready?

shaoshing:
Good idea. I will experiment with the spinner loader if content is not ready

iwehrman:
Sorry, I started reading through this and then got distracted, but besides my one gripe I think this looks great. I'm ready to merge if we can improve the loading experience just a little bit.

shaoshing:
@iwehrman I experimented and compared the `fade-in` and `spinner` solutions, and I think `face-in` has better user experience in this case. The libraries panel usually can finish loading and rendering within within 500ms (even for large collection, like `Samples Collection April 23 2015`, the time is still within 1200ms range). In such short amount of time, the `spinner` seems to be distracting. Also, it makes me feel that the panel is slower, compare to fading-in the content when ready. The reason might be because moving object can attract attension easier.

I've push my changes with the fade-in solution, please try it out and let me know what you feel.

iwehrman:
I'm not surprised that the spinner was distracting, but I'm glad you tried both options. I think the current behavior is great. I'm ready to merge except for that one additional typo. :+1: :+1:

shaoshing:
@iwehrman I just addressed the typo. Please review again.

src/js/actions/libraries.js
```diff
@@ -1344,49 +1359,33 @@ define(function (require, exports) {
...
+                _handleLibrariesLoadedHelper = function () {
+                    return this.flux.actions.libraries.handleLibrariesLoaded();
+                }.bind(this);
+                ccLibraries.addLoadedCollectionsListener(_handleLibrariesLoadedHelper);
+                
+                // `ccLibraries` will not emit the `LoadedCollections` event when the user sign out Creative Cloud
+                // before it is loaded. We have to check the size of libraries once, and call the event handler
+                // when it is empty.
+                setTimeout(function () {
+                    if (ccLibraries.getLoadedCollections.length === 0) {
```
iwehrman: I think still needs to be `getLoadedCollections().length`?

iwehrman:
LGTM!

shaoshing:
Thanks for merging!

Issue #3615, opened 2016-01-26
Opacity/Blend Mode misalignment


iwehrman:
These components are currently aligned differently depending on the current layer selection:
![img](https://s3.amazonaws.com/f.cl.ly/items/0d3D3Z2O123l3U3g3K2o/Screen%20Recording%202016-01-26%20at%2012.09%20PM.gif?v=da829b9f)

chadrolfs:
almost.  shapes are the odd ball now.
![align-almost](https://cloud.githubusercontent.com/assets/8442631/13716196/75083680-e78e-11e5-8d36-82dd13ff5b87.gif)

Issue #3147, opened 2015-10-27
Appearance panel should not be disabled in mixed states


iwehrman:
This is a regression. If you select a shape and a text layer, all the controls should still be active. Controls that only apply to vector layers should still only apply to the subset of selected vector layers, etc.

chadrolfs:
@DivyaPrabhakar is working on this. #3662

Issue #2457, opened 2015-09-14
Appearance and Effects panels should remain even when no layers are selected


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/9859811/3069171c-5add-11e5-938e-45e7745879d3.png)

This is related to #2152. Not only should the export panel **not** change position based on layer selection, I think the other panels should remain in the UI, even when they're empty.

designedbyscience:
I disagree with this, but I think a holistic pass needs to be made, design-wise

chadrolfs:
reviewed - probably not M3

We would need fix in the next few days for UIF.  Out of time.

DivyaPrabhakar:
Fixed by #3662

PR #3662, opened 2016-02-03
Mixed Appearance panel shows controls for valid layers in selection

DivyaPrabhakar:
Fixes issues #3777, #3756,  #3615, #3147, and #2457

Issue #3777:

Before:

After: I have made all disabled inputs use the specified CSS color for each stop `@underlines-disabled`. In the examples below, blend mode and opacity are enabled and the rest of the type settings are disabled, just to give an idea of the different states relative to one another. @placegraphichere thoughts?

Issue #3756

Opacity and Blend Mode Alignment
Before:

After:

Color Picker Inputs Alignment
Before: 

After: 

Radius Input Alignment: 
Before: 

After: 

Issue #3615 
Before: 

After: 

Issue #3147 
Before: 

After: 

Issue #2457 @placegraphichere Can I get your opinion about this too? 

Before: 

After:


DivyaPrabhakar:
@volfied or @chadrolfs Can you make sure the now activated fields are the correct ones? Is there anything that should not be active? One thing I saw on `master` also is when two text layers are selected `Mixed Font Styles` is active but there are no options to choose from. Is that correct behavior?

ktaki:
Hi @DivyaPrabhakar Let me chime in. 

Found an issue. Create a rectangle and a text layer. Change the color of the rectangle. Select both layers. Now the color of rectangle in the panel is wrong.

screen shot 2016-02-03 at 2 44 11 pm

When two text layers are selected, the current behavior looks right though more testing is needed.

screen shot 2016-02-03 at 2 49 49 pm

chadrolfs:
A couple more things (leveraging Taki's screenshots above):
- When mixed state includes a shape and text, you can change the color of the text but not the shape
- In this mixed state ^^, blend and opacity are out of alignment
- There's no way that I can get the panel in to full view (without scrolling).  Collapsing Effects and Export collapse up

DivyaPrabhakar:
@ktaki and @chadrolfs 

1.) The appearance panel does not scroll anymore in the mixed state


2.) The blend modes and opacity are now aligned 


3.) The color picker for vector shapes is fixed

src/js/jsx/sections/style/VectorFill.jsx
```diff
@@ -181,44 +181,52 @@ define(function (require, exports, module) {
                     document={this.props.document}
                     layers={this.state.layers}
                     fill={this.state.fill} />);
+
+            if (this.props.document.layers.selected) {
```
iwehrman: This is always truthy. Maybe you want `!this.props.document.layers.selected.isEmpty()`?

src/js/jsx/sections/style/VectorFill.jsx
```diff
@@ -181,44 +181,52 @@ define(function (require, exports, module) {
...
+                var disabledVectorFill = this.props.document.layers.selected.filter(function (layer) {
```
iwehrman: This is legal and functionally correct (due to [variable hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting)), but for clarity I think we would prefer something like this:
```javascript
var disabledVectorFill;
if (this.props.document.layers.selected.isEmpty()) {
    disabledVectorFill = false;
} else {
    disabledVectorFill = // filter...
}
```

iwehrman:
It may well be inconvenient to implement this with our current component structure, but I definitely do not think showing a fill color swatch row AND a type color swatch row is the correct design when some type layers and some vector layers are selected. There should just be one swatch/blend-mode row, and the behavior should be intuitively uniform: the swatch is used to change the type AND vector colors simultaneously, as appropriate for the given selected layer. Why should the user care that these two swatches are wired up to different underlying actions? Instead it should be possible to change all the colors in a single operation.

DivyaPrabhakar:
I think @iwehrman is right. Initially I was thinking that the current approach made sense because the user cases for choosing multiple layers at a time (ranked by priority) is 1.) To make a property change to all selected layers 2.) To move the selected layers on the canvas. 

In the case that the user selects the layers to move them around, my initial hesitation was that we can't assume they want to apply a change to a property (like fill) to all the layers selected at once, so we need to provide the separated text color and vector fill options. But in that case, we already know the primary objective for the user is to change the placement of the layers, not change a property so it doesn't matter if we provide one or two different fill options anyway. Another thing I saw is that when the vector fill and text color are enabled, the stroke swatch is below the text. I think it should be paired with the vector fill swatch. 

I will try and implement the one swatch approach and add it as a separate commit to this branch so people can try both ways and we can decide from there.

chadrolfs:
sounds good @DivyaPrabhakar, let @ktaki and I know when the new branch is ready and we'll give it a once over.

ktaki:
@DivyaPrabhakar Toggle Fill seems not functional.

screen shot 2016-03-04 at 9 55 26 am

ktaki:
The Sampler is broken in this branch. Make sure this isn't a side-effect of your change.

chadrolfs:
Comments on new branch (mixed-appearance-refactoring-2):

Hitting Tab key after editing a field in mixed selection state results in an error:
screen shot 2016-03-04 at 11 39 44 am

chadrolfs:
rotating a mixed selection changes the attributes but shouldn't
mixed-rotate

DivyaPrabhakar:
@ktaki and @chadrolfs -- I have updated the PR description with all the bugs that this PR should fix -- mainly UI fixes. Also the problems that you both identified before should @ktaki and @chadrolfs -- The problems that you both identified before should be fixed: 

1.) Color swatch keeps the fill when the selections are rotated


2.) Sampler tool works 


3.) Hitting Tab key after editing Typeface field in mixed selection state results in an error:Hitting Tab key after editing Typeface field in mixed selection state results in no error now.

chadrolfs:
color picker is blinks mixed:
blinky-cp

chadrolfs:
interplay between the opacity in the color picker and the panel opacity is funky with a mixed selection that includes a pixel layer.

chadrolfs:
changing the tracking fails:
`Action type.setTracking failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Set” is not currently available.`

chadrolfs:
same thing with Alignment:
`Action type.setAlignment failed: Error: Photoshop returned an error when playing command number: 1 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Set” is not currently available.`

ktaki:
The issues with manipulating alpha value could be caused by the issue reported in #3786 in which the alpha value for the text layers is treated differently from that of shape layers.

ktaki:
I did not find any other new issues.

src/js/jsx/sections/style/AppearancePanelSections.jsx
```diff
@@ -0,0 +1,491 @@
...
+                    );
+                };
+
+                if (textLayers.size === 0 && vectorLayers.size > 0) {
+                    colors = this.state.vectorFill.colors;
+                } else {
+                    if (characterStyles && !characterStyles.isEmpty()) {
+                        var textColors = collection.pluck(characterStyles, "color");
+                        var vectorColors = this.state.vectorFill.colors;
+                        var temp = textColors.concat(vectorColors);
```
DivyaPrabhakar: @iwehrman Can you help me here? The problem that happens is that `textColors` is in the same format as `vectorColors` but apparently the `r, g, b, a,` values of `textColors` is not rounded to the nearest integer and `vectorColors` are. Should I just call `Math.round()` on each of these values or am I missing something if I do that? Just not sure because we clearly handle text colors and vector fills differently so I wanted to double check. @pineapplespatula You worked with type right? So maybe you know better? 

The end goal of this is getting all the vector fills and all the text colors, and putting them into one final data structure that is of the same type as `textColors`.

DivyaPrabhakar:
@chadrolfs I fixed the errors that were occurring on alignment and tracking and I fixed part of the opacity issue -- before if you used the slider, only text and vector layers would change opacity, but now all layers change opacity. Also the user can change opacity via the opacity input field i.e. type opacity = 30 and all the layers will adjust accordingly. 

2 known problems still exist -- 1.) When the opacity slider is adjusted, the opacity input field does not reflect this change. This is because of the open bug @ktaki mentioned above, and we have decided not to fix that bug. 2.) The color swatch overlay goes to mixed state when the user rotates the layers. I know why this is happening and was working on the fix but I need a little help from @iwehrman or @pineapplespatula, and I have commented on that line and will wait for their feedback. 

The rest of the issues both you and @ktaki found should have been addressed. I will let you know right when a decision on problem no. 2 is made. Sorry for the delay on this, it took me some time to narrow down the problem 2.

DivyaPrabhakar:
@chadrolfs The rotation error should be fixed now!

chadrolfs:
OK, I think this'll do it!  Thanks for all your hard work on this.  Ok to merge!

src/js/actions/sampler.js
```diff
@@ -211,16 +211,12 @@ define(function (require, exports) {
      * @return {Promise}
      */
     var applyColor = function (doc, layers, color) {
```
volfied: so what's the point of passing layers to this now, if we're just gonna ignore it?

src/js/models/color.js
```diff
@@ -96,9 +96,9 @@ define(function (require, exports, module) {
         }
 
         return new Color({
-            "r": rgb.red,
-            "g": green,
-            "b": rgb.blue,
+            "r": mathjs.round(rgb.red),
```
DivyaPrabhakar: FYI: This is because colors for type layers were returning as floating points and vector layers colors were integers. Created a problem when we called .uniformValue() on a group of type and vector colors, even though they were the same.

volfied:
Looks good to me, since it's also heavily tested, I saw no blaring errors.

PR #3666, opened 2016-02-03
Upgrade spaces-adapter

iwehrman:
Upgrade spaces-adapter to v1.2.1, and update the shrinkwrap file accordingly. I believe the from/resolved changes are just noise, but please double-check my arithmetic @mcilroyc.


mcilroyc:
looks fine

PR #3671, opened 2016-02-04
More reliable click modifier tracking

iwehrman:
With this PR, properties on click events are used to infer modifier state instead of using the modifier store, which is less reliable. It's less reliable because it relies on receiving `adapterFlagsChanged` events, which can't be dispatched when the application is not active. So, e.g., if another window is activate, and then some modifiers are held, and then the Design Space window is activated, the modifier store's state will be incorrect. In some cases, like `SuperSelectOverlay`, we actually do want to use the functionality provided by the modifier store because we want to re-render views when modifier state changes in absence of other mouse events. But, in the case of the `LayerFace` and `IconBar` click handlers below, it's both simpler and more reliable to use the modifier properties from the click event.

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -193,14 +193,18 @@ define(function (require, exports, module) {
                 !Immutable.is(this.state.activeDocument, nextState.activeDocument);
         },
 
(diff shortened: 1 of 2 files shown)


volfied:
Simple enough.

Issue #3678, opened 2016-02-05
Closing files rapidly in succession will cause an error


ktaki:
Action documents.close failed: TypeError: Cannot read property 'id' of null

The error can be seen closing large documents in succession by repeated Cmd/Ctrl-w, but easiest method for reproducing this error is the following.

1. Create two new documents by Cmd/Ctrl-n
2. Closing the two files by keep pressing Cmd/Ctrl-w

![screen shot 2016-02-05 at 11 05 40 am](https://cloud.githubusercontent.com/assets/2029597/12859260/dc05be20-cc09-11e5-93d4-abbde35dca34.png)

Note: A similar issue was reported and fixed in the issue #1813.

chadrolfs:
I ran into this once yesterday but cnr on subsequent attempts.  Thanks for tracking this down.

iwehrman:
I can repro. The problem is that the action assumes that a valid document model exists because the menu command is supposed to be disabled when that isn't the case. But the menu is not rebuilt for a few hundred milliseconds after the models are updated so that updates can be grouped, thus reducing unnecessary work. As a result, the `File > Close` command is enabled when it shouldn't be after the last document is closed, and so the user is allowed to execute a `documents.close` action in a state that violates that action's precondition. Simple fix coming right up...

volfied:
FBNC to @ktaki

ktaki:
Verified the fix.

PR #3682, opened 2016-02-07
Guard against null documents in documents.close

iwehrman:
This PR adds a guard against null document model references in `document.close`. This is necessary because the menu system can't guarantee that this action will only be played when a valid document model exists because the menu system's rebuilds are debounced. See discussion in #3678. We should be on the lookout for other instances of this unsound assumption.

Addresses #3678.


volfied:
Ok!

PR #3683, opened 2016-02-08
Two shouldComponentUpdate fixes

iwehrman:
1. Implement `shouldComponentUpdate` in `Dialog`. This is not as easy as it should be because `Dialog` components render in two passes. To make this possible, I've added a flag to each instance of which makes it possible to track the progress of these two-pass renders.
2. Fix the `shouldComponentUpdate` implementations in Fill.jsx. We were passing in a plain objects as the `fill` prop to the components defined in `Fill.jsx`, but testing for equality with `Immutable.is`. (It would be nice if this threw an error?) I worked around the problem by just creating a custom struct to hold the properties in that previously mutable object.

Both of these components were being rendered unnecessarily when calling `initializeBackgroundLayers`. The former was also causing _all_ dialogs to re-render whenever one dialog was opened.

Addresses: concerns.

### DIFF
--- a/src/js/jsx/sections/style/VectorFill.jsx
+++ b/src/js/jsx/sections/style/VectorFill.jsx
@@ -44,6 +44,17 @@ define(function (require, exports, module) {
         classnames = require("classnames");
 
     /**
+     * Immutable object that summarizes downsampled fills.
+     *
+     * @constructor
+     */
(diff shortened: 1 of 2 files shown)


volfied:
Yuck, but it works!

PR #3684, opened 2016-02-08
Reduce cost of updating panel sizes

iwehrman:
This PR makes three changes to avoid unnecessary action calls to update the cloaking rectangle in the adapter:

1. Factor `commitPanelSizes` out of `updatePanelSizes`. The former commits changes to the adapter, while the latter only updates the bounds in the panel store. 
2. Do not call `commitPanelSizes` until all panels have been initialized. This should make the calls that happen before `beforeStartup` less costly.
3. Once all panels are initialized, use a debounced version of `commitPanelSizes` to avoid committing unnecessarily. This should avoid repeatedly committing stale panel size values when, e.g., dragging a window from one display to another.

@shaoshing: I haven't convinced myself that this leads to any concrete improvements. Can you take a look and make a judgement call about whether this extra complexity is warranted? I'm not sure myself yet.

### DIFF
--- a/src/js/actions/panel.js
+++ b/src/js/actions/panel.js
@@ -193,29 +193,62 @@ define(function (require, exports) {
     };
 
     /**
+     * Commit panel-sizes changes to the plugin, including updating the transform
+     * and the cloaking rectangle.
+     *
+     * @return {Promise}
+     */
+    var commitPanelSizes = function () {
+        var transformPromise = this.transfer("ui.updateTransform"),
+            panelState = this.flux.stores.panel.getState(),
+            centerOffsets = panelState.centerOffsets,
+            offsetsPromise = adapterUI.setOverlayOffsets(centerOffsets)
+                .bind(this)
+                .then(function () {
+                    return this.transfer(setOverlayCloaking);
+                });
+
+        return Promise.join(transformPromise, offsetsPromise);
+    };
+    commitPanelSizes.action = {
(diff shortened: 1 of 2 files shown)


shaoshing:
This looks good! It has reduced the number of `adapterUI` calls from 8 to 2 during startup. I don't see significant performance boost either (I saw the `adapterUI` calls are completed within 1ms), but it definitely make the action calls more clear (and less console outputs), so I think we should get this in.

src/js/actions/panel.js
```diff
@@ -193,29 +193,62 @@ define(function (require, exports) {
...
+        writes: [locks.PS_APP],
+        transfers: [setOverlayCloaking, "ui.updateTransform"],
+        modal: true
+    };
+
+    /**
+     * A debounced function which calls the commitPanelSizes action.
+     *
+     * @private
+     * @type {function(FluxController)}
```
shaoshing: Maybe this should be `@param {FluxController}`

shaoshing:
@iwehrman I am ready to merge this, just got one question about the JSDoc. Back to you.

src/js/actions/panel.js
iwehrman: Good call, thanks.

iwehrman:
Thanks for the review. JSDoc issue addressed. Back to @shaoshing!

shaoshing:
Sweet!

Issue #3658, opened 2016-02-03
Color picker slider position doesn't update on undo


iwehrman:
1. Change the color of a shape by dragging the slider on the color picker map.
2. Undo.

Result: the color of the shape is updated on canvas, but neither the color slider position nor the color input are updated.

Bonus: If you change the color to be pure black with the slider in the bottom right corner, undo moves the slider to the bottom left corner. 

![image](https://cloud.githubusercontent.com/assets/1075154/12770299/3063909c-c9d4-11e5-9bcd-2780d1617ab9.png)

mcilroyc:
More/different badness: After making several color changes, `undo` changes the color picker but incorrectly.  This may be `coalesce` related?
![ss](http://cl.ly/1b2C231A1W0A/Screen%20Recording%202016-02-03%20at%2010.23%20AM.gif)

mcilroyc:
bisect indicates first bad commit: 5d7ae63e6c4f9b410f52f9be2fbad1b8ccd9f09d

mcilroyc:
You break it, you bought it.  @shaoshing can you please take a look at this since it *seems* that your "Delay running action to promote browser painting" performance hack has significantly regressed the color picker "timeTravel" handling.  You can assign it back to me if you think we need to re-jigger the way this works wrt to history ... but I figured you're the best candidate to make progress on this right now.

mcilroyc:
FBNC.  Please review @ktaki

ktaki:
@iwehrman: You stated "Bonus: If you change the color to be pure black with the slider in the bottom right corner, undo moves the slider to the bottom left corner."

Currently, if RGB is set to (0,0,0), then HSB is forced to be set to (0,0,0) once color picker is closed. Is this the behavior you want? Or, do you want to HSB to be reset only with undo?

iwehrman:
I don't really care where the slider is when you open the picker with HSB = (0,x,0). The problem is that if I move the slider to, say, (0,50,0) in one step, and then move it to say, red, in a second step, when I undo it goes back to HSB = (0,0,0) instead of (0,50,0). These all mean "black" but the slider goes to a different place in the map. It's relatively a minor issue.

ktaki:
[...]

PR #3685, opened 2016-02-08
Bug fix: document `change` event should be emitted before history `timetravel`

shaoshing:
When `timetravel` event is fired, ColorInput requires its prop to be updated so that it can re-render based on the current props, and the update is triggered by document store's `change` event. Below is the expected working order:

1. Undo command executed
2. document store emits `change` event
3. ColorInput receives updated props, but skip rendering
4. history store emits `timetravel` event
5. ColorInput handles `timetravel` and re-render 

With our recent optimization on browser paint, document store's `change` event is delayed to allow browser starts rendering ASAP, and this hack disrupts the order, and now it becomes: `1 - 4 - 5 - 2 - 3`

This PR use promise to ensure the correct event order.

Fix for #3658

@mcilroyc can you review and let me know if you approve the fix?

### DIFF
--- a/src/js/stores/document.js
+++ b/src/js/stores/document.js
@@ -157,11 +157,14 @@ define(function (require, exports, module) {
          * @param {boolean=} dirty Whether to set the dirty bit, assuming the model has changed
          * @param {boolean=} suppressChange Whether or not to suppress the change event
          * @param {string=} changeEventName The event to emit. Default: "change".
+         * @return {Promise}
          */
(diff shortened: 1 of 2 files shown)


mcilroyc:
This seems to fix the issue, and your description makes sense.  

In reviewing this, it occurred to me that for the *general* use of `setDocument` there could be unintended consequences of adding the asynchronous delay prior to emitting the "change" event.  All the event handlers that call setDocuments will now return before the event is emitted.  That was introduced prior to this PR, and seems to be working anyway, so just noting it here for the heck of it.

Issue #3505, opened 2016-01-12
Swap breaks sometimes with one layer in and one layer outside Artboard


chadrolfs:
cmd+n, zoom out x3, 
drag a LLSO onto canvas inside AB, 
then select rect tool and draw a rect outside AB, 
select both, then swap (repeat swap if needed)
or next, 
move both layers outside AB and then swap

chadrolfs:
![swap-broke2](https://cloud.githubusercontent.com/assets/8442631/12277995/64fb3b7c-b933-11e5-9f8d-0b652504ad35.gif)
![swap-broke3](https://cloud.githubusercontent.com/assets/8442631/12277994/64fae636-b933-11e5-9c79-6ef5b7115d11.gif)

chadrolfs:
1st case is not always reliable to repro.  Edge case

volfied:
This is happening if you click swap fast enough that we enqueue two `swapLayersCurrentDocument` actions, and we don't get enough time to process handleCanvasShift in between.

It's not possible to repro this if you click swap slow enough.

iwehrman:
FBNC to @chadrolfs

chadrolfs:
same steps as above but after you swap a few times, move the selection a bit and then swap again.

![swap-fail2](https://cloud.githubusercontent.com/assets/8442631/12934522/fd41ab8e-cf43-11e5-9cdb-7d443fe5823b.gif)

PR #3686, opened 2016-02-08
Reset bounds of layers that moved during swap, as canvas shift can happen

volfied:
Addresses #3505 

We set autoExpandDisabled during our action play, but as soon as we finish and re-set it, PS handles canvas shift, and sends us the event, which we debounce. Because we debounce it, if we were to call swap again, we would be trying to swap layers at a wrong location, and cause the bug. So now as part of swap, we call resetBounds on all selected layers. This doesn't add that much to action time, so didn't seem like a bad performance hit.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -564,7 +564,10 @@ define(function (require, exports) {
                 return descriptor.endTransaction(transaction);
             })
             .then(function () {
-                return this.transfer(layerActions.resetIndex, document);
+                var boundsPromise = this.transfer(layerActions.resetBounds, document, document.layers.allSelected),
+                    indexPromise = this.transfer(layerActions.resetIndex, document);
+
+                return Promise.join(boundsPromise, indexPromise);
             });
 
         return Promise.join(dispatchPromise, swapPromise);
@@ -572,7 +575,7 @@ define(function (require, exports) {
     swapLayers.action = {
         reads: [],
         writes: [locks.PS_DOC, locks.JS_DOC],
-        transfers: [layerActions.resetIndex],
+        transfers: [layerActions.resetIndex, layerActions.resetBounds],
         post: ["verify.layers.verifySelectedBounds"]
     };
 


src/js/actions/transform.js
```diff
@@ -564,7 +564,10 @@ define(function (require, exports) {
                 return descriptor.endTransaction(transaction);
             })
             .then(function () {
-                return this.transfer(layerActions.resetIndex, document);
+                var boundsPromise = this.transfer(layerActions.resetBounds, document, document.layers.allSelected),
```
iwehrman: We should really rename that property to something like `descendantsOfSelected`. It's kind of hard to guess what `allSelected` means compared to just `selected`. (Not necessary for this PR though.)

src/js/actions/transform.js
```diff
@@ -564,7 +564,10 @@ define(function (require, exports) {
...
+                    indexPromise = this.transfer(layerActions.resetIndex, document);
```
iwehrman: Just to be clear, do you not care which order these two actions are processed in? Although they might execute in an apparently deterministic order here, we shouldn't rely on that if there is some correctness constraint based on order. One reason why it might possibly matter is w.r.t. history.

iwehrman:
One question, but looks good generally. Back to @volfied.

iwehrman:
You know, I wonder if we could filter out non-artboard group layers in `resetBounds` as an optimization since we don't track bounds for plain groups anyway.

iwehrman:
(Work in progress on that here: https://github.com/adobe-photoshop/spaces-design/compare/iwehrman/reset-fewer-bounds)

src/js/actions/transform.js
volfied: In this case, both dispatches dispatch by ID, so it should be fine. History also gets amended correctly for both cases, since they don't cause changes in PS history.

volfied:
@iwehrman Answered your question, I think the change is safe.

iwehrman:
:+1: :+1:  :+1:

Issue #3679, opened 2016-02-05
Opening large documents in one open command will cause an error


ktaki:
Uncaught TypeError: Cannot read property 'id' of null

Method:
0. Reset DS
1. Locate Vermilion doc and create a copy of it. (i.e., two Vermilions)
2. Cmd/Ctrl-o.
3. Select the original Vermilion doc and its copy at the same time.
4. Click Open.

![screen shot 2016-02-05 at 1 34 17 pm](https://cloud.githubusercontent.com/assets/2029597/12859840/a3e963a8-cc0d-11e5-96b2-782762e4586f.png)

iwehrman:
I can't reproduce, but looking at `allocateDocument` I'm not surprised that something could go awry here: we call `_getSelectedDocumentID` to allocate the "current" document, regardless of whether that relates to the document ID passed in (which is a little weird?), but conceivably a new document could become active after that as a result of multiple documents having been opened at once. CC @mcilroyc 

Ultimately this seems like a battle with Photoshop that we can't win unless its API changes to accommodate our use case. If multiple documents are going to be opened at once, we'll need much more structured information than just a series events indicating that a file was opened.

iwehrman:
With that said, I don't know exactly what is causing that particular error, and that would be good to track down in case either a) my guess above about what's wrong is incorrect, or b) it's correct, but we can recover more gracefully. CC @shaoshing also since he may be considering our document initialization routines as part of his performance optimization work.

mcilroyc:
My initial hunch is that a reasonable fix may just be to add some defensive code in the menu update routine.  I'm still trying to repro.  Perhaps I need to try on a slower machine...

ktaki:
If two Vermillion documents do not produce the error, try with three or four duplicates.

mcilroyc:
FYI, *can repro* on Windows VM

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3687, opened 2016-02-09
Menu 'open documents' refresh process should not throw error when there's no active document

mcilroyc:
When updating the open documents menu list, do not *error* if no document is set to active in the app store.

When opening several documents simultaneously via the "Open.." menu, the first `open` event from photoshop is for an *inactive* document.  That event is handled by `allocateDocument` which causes an updateDocument, which in turn emits an event that will cause the menu items to refresh.  At this time, the application store may not be aware of the current document.  Previously, in the menu code, the `updateOpenDocuments` method would throw an error if a current document was not provided.  It seems safe to me to relax that requirement, and tolerate the no-active-document situation here. 

This is a simple fix for #3679

### DIFF
--- a/src/js/models/menubar.js
+++ b/src/js/models/menubar.js
@@ -639,7 +639,7 @@ define(function (require, exports, module) {
      * If First launch is open, no documents will be shown
      *
      * @param {Object.<number, Document>} documents List of open documents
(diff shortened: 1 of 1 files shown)


iwehrman:
Let's Not Overthink This.

PR #3691, opened 2016-02-09
In resetBounds, ignore layers that do not have bounds.

iwehrman:
With this change, `layers.resetBounds` ignores layers that don't have bounds like non-artboard groups, groupends, and adjustment layers. As mentioned in #3686. I can't actually measure a speed improvement here, but it still seems reasonable...


volfied:
LGTM

PR #3692, opened 2016-02-09
Number localization with os-specified decimal separator

iwehrman:
This PR improves our number localization so that the operating system-specified decimal separator is used, and not just the locale default. This is done by fetch `localeInfo` from Photoshop before starting the application, and using that to provide util-level information about the correct decimal separator and a decimal-formatting function. I did it this way because a) this setting can't change without restarting Photoshop, and b) it turned out to be extremely challenging to put this into our usual Flux data flow (i.e., it would have required major refactorings, and it just didn't seem to be worth the effort).

To test, manually change your locale's decimal separator this like:
image
And then restart Photoshop .

Addresses #3136.

### DIFF
--- a/src/js/jsx/sections/export/ExportList.jsx
+++ b/src/js/jsx/sections/export/ExportList.jsx
@@ -42,20 +42,6 @@ define(function (require, exports, module) {
         nls = require("js/util/nls");
 
     /**
-     * The options for the scale datalist
-     * @private
-     * @type {Immutable.OrderedMap.<string, {id: string, title: string}>}
-     */
-    var _scaleOptions = Immutable.OrderedMap(ExportAsset.SCALES
-        .map(function (scale) {
-            var obj = {
-                id: scale.toString(),
-                title: scale.toLocaleString()
-            };
-            return [scale.toString(), obj];
-        }));
-
-    /**
      * The options for the format datalist
      * @private
      * @type {Immutable.OrderedMap.<string, {id: string, title: string}>}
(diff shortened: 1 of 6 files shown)


src/js/jsx/sections/export/ExportList.jsx
```diff
@@ -42,20 +42,6 @@ define(function (require, exports, module) {
         nls = require("js/util/nls");
 
     /**
-     * The options for the scale datalist
-     * @private
-     * @type {Immutable.OrderedMap.<string, {id: string, title: string}>}
-     */
-    var _scaleOptions = Immutable.OrderedMap(ExportAsset.SCALES
```
iwehrman: I had to delay this initialization from module-load time to component-mount time because the locale info is not yet ready at module-load time.

mcilroyc: :+1:

src/js/jsx/shared/NumberInput.jsx
```diff
@@ -154,9 +156,9 @@ define(function (require, exports, module) {
             try {
                 var trimValue = _.trimRight(rawValue, this.props.suffix);
 
-                if (nls.commaDecimalSeparator) {
+                if (this.state.decimalSeparator === ",") {
                     // Replace decimal separator , with . and replace argument separator ; with ,
-                    trimValue.replace(new RegExp("\\,|\\;", "g"), function (match) {
+                    trimValue = trimValue.replace(new RegExp("\\,|\\;", "g"), function (match) {
```
mcilroyc: Lame that you have to do this the hard way

mcilroyc: It seems like it could be nice to pull this out in to something like `nls.parseDecimal`

mcilroyc:
reviewed.  questions/suggestions, but merge-able as is.

iwehrman:
I didn't create a `parseDecimal` function in `js/util/nls` because that would have required pulling more math into the NLS module than seemed appropriate, so instead I just factored out `normalizeDecimalExpr`, which replaces `,` with `.` as needed in arithmetic decimal expressions. Back to @mcilroyc

src/js/jsx/shared/NumberInput.jsx
```diff
@@ -96,11 +96,13 @@ define(function (require, exports, module) {
         },
 
         getInitialState: function () {
-            var rawValue = this._formatValue(this.props.value);
+            var rawValue = this._formatValue(this.props.value),
+                decimalSeparator = nls.getDecimalSeparator();
```
mcilroyc: no longer needed?

iwehrman: Argh, sorry, yes.

iwehrman:
Fixed. Back to @mcilroyc again.

Issue #3694, opened 2016-02-09
Opening large documents in one open command may cause history warnings 


ktaki:
... when opened doc was selected in doc bar. 
[History] Could not amend history, document history not found: 2424

(Related to the issue #3679 just fixed.)

Method:
0. Reset DS
1. Locate Vermilion doc and create a copy of it. (i.e., two Vermilions)
2. Cmd/Ctrl-o.
3. Select the original Vermilion doc and its copy at the same time.
4. Click Open.
5. After the docs are opened, select the first doc on the doc title bar.
(If do not repro, try with three or four duplicated Vermilion docs.)

![screen shot 2016-02-09 at 3 39 07 pm](https://cloud.githubusercontent.com/assets/2029597/12934430/4b0ef372-cf43-11e5-901b-d0eeaca0b3ae.png)

It seems no bad thing happening after the warning. Could this be just ignored?

ktaki:
@mcilroyc: The warning still can be seen, but this time only one line instead of two lines.

![screen shot 2016-02-10 at 5 09 53 pm](https://cloud.githubusercontent.com/assets/2029597/12967184/4a8e2036-d019-11e5-807d-43b4f7ecfbb1.png)

mcilroyc:
Yes @ktaki.  I am aware of the cause, but I'm not sure it is worth "fixing".  You might even notice *two* in a row.  I am reasonably certain that these limited warnings can now be safely ignored.  If you see a "[History] Initializing history..." log shortly thereafter, that indicates that this was just a brief race condition that has been neutralized :zap:.  There is a little more info in my rambling description of the PR #3695.  I recommend we close this bug because I think the warnings are now "within tolerance".

ktaki:
OK. I agree. Closing.

PR #3695, opened 2016-02-10
Init document history on first-time selectDocument

mcilroyc:
This fixes the more annoying portion of the issue reported in #3694.  When opening multiple documents at the same time with the `Open...` menu, the opened-but-inactive documents can not have their history state initialized (PS limitation).  When the document is later selected, various actions will cause a warning that indicates history was trying to be amended, but the history store is not initialized.  This seems to be safe to ignore, but the fix in this PR seemed like the right thing  to do.

`[History] Could not amend history, document history not found`

In the case of large documents this would happen immediately, but in smaller documents it can still happen after selecting a new layer for the first time (in the newly selected document).

This PR updates the `document.selectDocument` action to check for the initialization status of the document in the history store, and explicitly queries PS for the information if it is not there.

What this PR does *not* do.  For large documents, there is still a race in `selectDocument` between the `selectPromise` and the `initializedPromise`.  Specifically, the transfer to `resetLinkedLayers` may cause both bounds and guides to be updated, both of which will try to amend history, *before* it is initialized.  These warnings should be safe to ignore.  Squashing them seemed more disruptive than necessary, at least before I got a second opinion.

cc: @volfied & @iwehrman since we discussed this general issue a few weeks ago when barkin changed this to a warning instead of an error.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -715,7 +715,8 @@ define(function (require, exports) {
             documentID = typeof documentSpec === "number" ? documentSpec : documentSpec.id,
             document = flux.stores.document.getDocument(documentID),
             documentRef = documentLib.referenceBy.id(documentID),
-            uninitialized = !document.layers;
+            historyUninitialized = !flux.stores.history.hasInitializedHistory(documentID),
+            layersUninitialized = !document.layers;
 
         var selectPromise = descriptor.playObject(documentLib.select(documentRef))
             .bind(this)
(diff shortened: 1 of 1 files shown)


src/js/actions/documents.js
```diff
@@ -763,8 +768,7 @@ define(function (require, exports) {
         reads: [locks.JS_TOOL],
         writes: [locks.JS_APP, locks.PS_APP],
         transfers: ["layers.resetLinkedLayers", historyActions.queryCurrentHistory,
-            ui.updateTransform, toolActions.select, "panel.cloak", guideActions.queryCurrentGuides,
-            toolActions.changeVectorMaskMode, updateDocument],
+            ui.updateTransform, toolActions.select, toolActions.changeVectorMaskMode, updateDocument],
```
mcilroyc: I just removed two dead transfers.  And luckily I didn't have to add `historyActions.queryCurrentHistory` because it was already sitting here rotting.  hooray.

volfied:
Looks good to me, passing to @iwehrman to merge.

iwehrman:
Looks correct to me.

PR #3696, opened 2016-02-10
Improve performance of switching initialized documents

shaoshing:
This PR improves the performance of switching initialized documents by optimizing the rendering process. When switching documents, we toggle the CSS class "section__active" to show panels of active document and hide the rest of the panels. Before, to just toggle one class, we have to re-render all the panels (`Transform, Appearance, Effects, Export, Layers`), which adds extra time to the switching. To improve this, this PR wraps the panel components with a `div`, and toggle the panels by change the class of the `div`. In this way, the panels can remain unchanged.

##### Improvements

| | Before | After | Diff |
| --- | --- | --- | --- |
| switch from an empty doc to vermilion | 364ms | **256ms** | -108ms |
| switch from vermilion to an empty doc | 199ms | **89ms** | -110ms |
(both documents are already initialized)

### DIFF
--- a/src/js/jsx/sections/export/ExportPanel.jsx
+++ b/src/js/jsx/sections/export/ExportPanel.jsx
@@ -60,11 +60,13 @@ define(function (require, exports, module) {
                 exportStore = flux.store("export"),
                 toolStore = flux.store("tool"),
                 modal = toolStore.getModalToolState(),
-                documentExports = exportStore.getDocumentExports(documentID);
+                documentExports = exportStore.getDocumentExports(documentID),
+                exportState = exportStore.getState();
 
             return {
                 documentExports: documentExports,
-                exportState: exportStore.getState(),
+                serviceBusy: exportState.serviceBusy,
+                serviceAvailable: exportState.serviceAvailable,
                 modal: modal
             };
         },
@@ -90,20 +92,22 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            if (this.props.disabled !== nextProps.disabled ||
-                this.props.active !== nextProps.active ||
-                this.state.modal !== nextState.modal) {
-                return true;
-            }
-
+            var selectedLayersChanged = !Immutable.is(this.props.document.layers.selected,
+                nextProps.document.layers.selected);
+            
             // If the panel is remaining invisible and the selection state hasn't changed, no need to re-render.
             // A new layer selection could change the enabled state of the title header buttons
-            if (!nextProps.visible && !this.props.visible &&
-                Immutable.is(this.props.document.layers.selected, nextProps.document.layers.selected)) {
+            if (!nextProps.visible && !this.props.visible && !selectedLayersChanged) {
                 return false;
             }
 
-            return true;
+            return this.props.disabled !== nextProps.disabled ||
+                this.props.visible !== nextProps.visible ||
+                this.state.modal !== nextState.modal ||
+                this.state.documentExports !== nextState.documentExports ||
+                this.state.serviceBusy !== nextState.serviceBusy ||
+                this.state.serviceAvailable !== nextState.serviceAvailable ||
+                selectedLayersChanged;
         },
 
         /**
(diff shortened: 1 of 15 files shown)


src/js/jsx/sections/transform/TransformPanel.jsx
```diff
@@ -53,9 +53,8 @@ define(function (require, exports, module) {
 
         shouldComponentUpdate: function (nextProps, nextState) {
             return this.state.referencePoint !== nextState.referencePoint ||
-                this.props.active !== nextProps.active ||
                 this.props.disabled !== nextProps.disabled ||
-                !Immutable.is(this.props.document, nextProps.document);
+                !Immutable.is(this.props.document.layers.selected, nextProps.document.layers.selected);
```
shaoshing: This fixes a bug that causes the TransformPanel to render repeatedly while DS is initializing the layers in the background.

iwehrman:
First impression: the improvement is noticeable. Very nice!

src/js/jsx/sections/export/ExportPanel.jsx
```diff
@@ -53,6 +53,10 @@ define(function (require, exports, module) {
          * @type {?function}
          */
         _setTooltipThrottled: null,
+        
+        getInitialState: function () {
+            return { };
```
iwehrman: Nit, should just be `{}`.

src/js/jsx/sections/style/AppearancePanel.jsx
```diff
@@ -82,23 +82,23 @@ define(function (require, exports, module) {
...
-
-            if (!Immutable.is(this.props.document.layers.selected,
-                nextProps.document.layers.selected)) {
-                return true;
-            }
-
-            if (!nextProps.visible && !this.props.visible) {
+        shouldComponentUpdate: function (nextProps, nextState) {
+            // Only compare the existence of clipboard.
+            var clipboardChanged = !!this.state.clipboard !== !!nextState.clipboar,
```
iwehrman: Typo: `clipboar` :boar:

src/js/jsx/sections/style/EffectsPanel.jsx
```diff
@@ -94,22 +94,24 @@ define(function (require, exports, module) {
...
-
-            if (!Immutable.is(this.props.document.layers.selected,
-                nextProps.document.layers.selected)) {
-                return true;
-            }
-
-            if (!nextProps.visible && !this.props.visible) {
+        shouldComponentUpdate: function (nextProps, nextState) {
+            // Only compare the existence of clipboard.
+            var clipboardChanged = !!this.state.clipboard !== !!nextState.clipboar,
```
iwehrman: Again: :boar:

iwehrman: Also, this is incredibly minor, but in this sCU and in AppearancePanel, it looks like this variable isn't read until after another check which could return early. If so, we should delay defining the variable until just before it's needed.

src/js/jsx/sections/transform/TransformPanel.jsx
iwehrman: It's possible that this will cause us to miss a necessary render when the document bounds change (which is shown in this panel when no layers are selected). So, maybe also compare the document bounds?

iwehrman:
This is excellent. I only have some very small suggestions. We should get this merged by the end of the day. Back to @shaoshing!

src/js/jsx/sections/transform/TransformPanel.jsx
shaoshing: Added document bound checking. Nice catch!

src/js/jsx/sections/style/EffectsPanel.jsx
shaoshing: :boar: fixed. 

`clipboardChanged` and  `selectedLayerChanged` are used by the following `if` statement, so I think that's right timing to initial them.

shaoshing:
@iwehrman thanks for reviewing. I've addressed your comments. Back to you.

iwehrman:
LGTM!

PR #3697, opened 2016-02-10
Refactor debugging menu items

shaoshing:
This PR makes the following change to the debugging menu items:

##### Moves all debugging menu items to a new top-level menu `Debug`

screen shot 2016-02-09 at 9 56 21 am

##### Addes new debugging items: 

1 `Load Debugging Helpers`

debug-2

2  `Log Descriptor Events`: Log descriptor events. includes events sent by PS and DS. This option is togglable.

debug-1

3 `Log Highlights Events (togglable)`: Log Highlights tracking events. This option is togglable.

debug-3


src/js/actions/menu.js
```diff
@@ -570,6 +570,181 @@ define(function (require, exports) {
         writes: [locks.JS_PREF],
         transfers: [preferencesActions.setPreference]
     };
+    
+    /**
+     * Debug-only method to toggle Descriptor Event logging
```
iwehrman: I suggest we split out all the actions currently in `menu` that are only relevant to debugging and put them in a separate action module. For bonus points, it would be great if we could not load those actions if `__PG_DEBUG__` is not set! But, even if we don't go that far in this PR, it would still be nice to streamline the menu actions.

src/js/actions/menu.js
```diff
@@ -570,6 +570,181 @@ define(function (require, exports) {
...
+     * 
+     * @param {Object} options
```
iwehrman: In this case, the correct type expression is `object`. You should only use upper case `Object` when you're describing an object which is being used as a map, and hence write something like `Object.<S,T>`.

src/js/actions/menu.js
```diff
@@ -570,6 +570,181 @@ define(function (require, exports) {
...
+            descriptor = require("adapter").ps.descriptor;
+
+        if (options.toggle) {
+            enabled = !enabled;
+        }
+        
+        var blockStyle = "color:gray;",
+            eventStyle = "font-weight:bold";
+        
+        descriptor.__play = descriptor.__play || descriptor.play;
```
iwehrman: I think it would be simpler and safer to instead replace the implementation of `_getAsync` and `_batchPlayAsync`. It seems like that might also work a little better with the transaction infrastructure.

src/js/actions/menu.js
```diff
@@ -570,6 +570,181 @@ define(function (require, exports) {
...
+        var preferencesStore = this.flux.store("preferences"),
+            preferences = preferencesStore.getState(),
+            enabled = preferences.get("headlightsLoggingEnabled"),
+            headlights = require("js/util/headlights");
+        
+        if (options.toggle) {
+            enabled = !enabled;
+        }
+        
+        headlights._logEvent = headlights._logEvent || headlights.logEvent;
```
iwehrman: Above you stored these temporary properties with two underscores, but here you're only using one. They should be consistent.

src/js/actions/menu.js
```diff
@@ -570,6 +570,181 @@ define(function (require, exports) {
...
+            },
+            _getFluxInstance: function () {
+                return this.flux;
+            }.bind(this),
+            _getCurrentDocument: function () {
+                return this.flux.stores.application.getCurrentDocument();
+            }.bind(this),
+            _getSelectedLayers: function () {
+                return this.flux.stores.application.getCurrentDocument().layers.selected;
+            }
```
iwehrman: Missing `.bind(this)`?

iwehrman:
Looking good! Back to @shaoshing for a little more earth-moving.

src/js/actions/debug.js
```diff
@@ -0,0 +1,592 @@
...
+
+        if (options.toggle) {
+            enabled = !enabled;
+        }
+
+        var blockStyle = "color:gray;",
+            eventStyle = "font-weight:bold";
+    
+        descriptor.__getAsync = descriptor.__getAsync || descriptor._getAsync;
+        descriptor.__batchPlayAsync = descriptor.__batchPlayAsync || descriptor._batchPlayAsync;
```
shaoshing: overwrites `descriptor._getAsync` and `descriptor._batchPlayAsync` as suggested

src/js/actions/debug.js
```diff
@@ -0,0 +1,592 @@
...
+        var preferencesStore = this.flux.store("preferences"),
+            preferences = preferencesStore.getState(),
+            enabled = preferences.get("headlightsLoggingEnabled"),
+            headlights = require("js/util/headlights");
+        
+        if (options.toggle) {
+            enabled = !enabled;
+        }
+        
+        headlights.__logEvent = headlights.__logEvent || headlights.logEvent;
```
shaoshing: uses `__logEvent` to maintain consistency.

shaoshing:
Hi @iwehrman I've updated the PR to address your comments. The highlight are as follows:

* move debugging actions into a new debug module 
* fix JSDoc and potential bugs
* use `descriptor._getAsync` and `descriptor._batchPlayAsync` 

Back to you to review again. Thank you.

src/js/actions/debug.js
```diff
@@ -0,0 +1,592 @@
...
+
+define(function (require, exports) {
+    "use strict";
+
+    var Promise = require("bluebird"),
+        _ = require("lodash");
+
+    var log = require("js/util/log"),
+        locks = require("js/locks"),
+        preferencesActions = require("./preferences");
```
iwehrman: Can you remove this import and instead declare transfers with the string format? (E.g., `"preferences.setPreference"`.) We prefer the string format to direct references because the latter may rely on a particular module load order, which isn't guaranteed.

src/js/actions/debug.js
```diff
@@ -0,0 +1,592 @@
...
+
+    /**
```
iwehrman: Also, can you try something at the top of this module like:
```javascript
if (__PG_DEBUG__) {
    return;
}
```
The idea is to eliminate this whole file when the sources are compiled in production mode. I'm actually not sure if that will work, but I think it will and it's worth a shot.

src/js/fluxcontroller.js
```diff
@@ -42,7 +42,8 @@ define(function (require, exports, module) {
 
     // Using contexts, we load action files automatically
     var storeContext = require.context("./stores", true, /^\.\/.*\.js$/),
-        actionContext = require.context("./actions", true, /^\.\/.*\.js$/),
+        // Do not compile the "debug" module in production mode.
+        actionContext = require.context("./actions", true, __PG_DEBUG__ ? /^\.\/.*\.js$/ : /^\.\/((?!debug).)*\.js$/),
```
shaoshing: Skip compiling the `debug` action module in production

shaoshing:
Hi @iwehrman, I addressed the transfer naming issue and exclude the `debug` module in production mode. Back to you.

shaoshing:
the new change are tested/passed in dev and production mode.

src/js/actions/debug.js
shaoshing: I tried this way in the `debug` module but it will break the fluxcontroller startup because `fluxxor` will skip empty action object, which causes `module not found` error.

iwehrman:
Fantastic, thanks!

Issue #3703, opened 2016-02-11
Panel bounds do not include icon bar bounds (off-set guide highlight)


iwehrman:
![image](https://cloud.githubusercontent.com/assets/1075154/12994713/61727aae-d0d7-11e5-8520-740d3d8675b6.png)

iwehrman:
I think this broke in #3696 and is caused by the fact that the `panelSet` ref now includes the `IconBar`.

shaoshing:
I verified this is fixed by #3708, but over to @chadrolfs to double check.

PR #3708, opened 2016-02-13
Fix panel bounds measurement

iwehrman:
The structure of the `PanelSet` recently changed to include the `IconBar`, but the bounds arithmetic did not change along with it. This resolves the discrepancy by putting the `PanelSet` in charge of reporting both `panelBounds` and `iconBounds`. As a nice side effect, there is one fewer component calling `updatePanelSizes` all the time.

Addresses #3703.

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -96,11 +96,21 @@ define(function (require, exports, module) {
          * @return {Promise}
          */
         _updatePanelSizes: function () {
-            var node = ReactDOM.findDOMNode(this.refs.panelSet),
+            var iconBarNode = ReactDOM.findDOMNode(this.refs.iconBar),
+                iconBarWidth;
+
+            if (iconBarNode) {
+                iconBarWidth = iconBarNode.getBoundingClientRect().width;
+            } else {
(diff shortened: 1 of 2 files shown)


src/js/jsx/PanelSet.jsx
```diff
@@ -96,11 +96,21 @@ define(function (require, exports, module) {
          * @return {Promise}
          */
         _updatePanelSizes: function () {
-            var node = ReactDOM.findDOMNode(this.refs.panelSet),
+            var iconBarNode = ReactDOM.findDOMNode(this.refs.panelSet),
```
shaoshing: This should be `this.refs.iconBar`?

shaoshing:
CC @designedbyscience since he is thinking about separating the IconBar from the panels to accommodate to the panel animations.

iwehrman:
Issue addressed (not sure why it was working in my branch at all...). Ready for re-review.

iwehrman:
Unassigning @designedbyscience since he's busy at the moment and I don't think we should wait for that guy.

shaoshing:
Ok. Let's merge this and @designedbyscience can perform a rebase later.

Issue #3713, opened 2016-02-17
Datalist doesn't filter on input


iwehrman:
I broke this in #3683 when I implemented `shouldComponentUpdate` in `Dialog`.

chadrolfs:
Looks fixed to me.  :+1:

iwehrman:
Well then.

PR #3714, opened 2016-02-17
TextInput and Datalist event handling fixes

iwehrman:
This PR makes the following changes to the event-handling interfaces of `TextInput` and `Datalist`:

1. `TextInput` now has a single `onInput` prop in place of the previous `onInputChange` and `onDOMChange` props. Those props were only ever being called one after another as a result of a `change` event on the `input` element, so this simplifies the interface and changes the name to be more meaningful and correct.
2. `Datalist` now has an `onInput` prop that replaces the previous `onInputChange` prop. The latter was a confusing name in relation to the semantics of that prop, which was and is that it is triggered as a result of the child `TextInput`'s `onInput` event. 

Besides updating the usage of these components and fixing up some stale comments, this also makes a change to the `shouldComponentUpdate` method of `Dialog`. On master, it never re-renders while the dialog is open, which means that the contents of the dialog don't ever re-render as a result of a parent re-render. Consequently, the `Select` component of `Datalist` was not correctly filtering during input. This changes the `shouldComponentUpdate` method to always re-render when the `Dialog` is open. 

Addresses #3713.

### DIFF
--- a/src/js/jsx/search/SearchBar.jsx
+++ b/src/js/jsx/search/SearchBar.jsx
@@ -416,7 +416,7 @@ define(function (require, exports, module) {
          * Perform action based on ID
          *
          * @private
-         * @type {Datalist~onInputChange}
+         * @type {Datalist~onChange}
          */
         _handleChange: function (id) {
             if (id === null) {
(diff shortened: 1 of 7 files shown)


iwehrman:
Unassigning @DivyaPrabhakar since she's busy at the moment and anybody else can review this. Of course, if @DivyaPrabhakar frees up, she can still do the review!

volfied:
LGTM

PR #3717, opened 2016-02-17
Use minimum layer size of 1 unless all selected leaves are vectors

iwehrman:
We're currently specifying a blanket minimum layers size of 0.1, but that actually only works for vectors or groups that contain only vectors. This PR changes the minimum size to be 1 unless all selected leaf layers are vector layers; in that case, go with the existing minimum size of 0.1.

Addresses #3153.

### DIFF
--- a/src/js/jsx/sections/transform/Size.jsx
+++ b/src/js/jsx/sections/transform/Size.jsx
@@ -39,7 +39,8 @@ define(function (require, exports, module) {
         collection = require("js/util/collection");
 
     var MAX_LAYER_SIZE = 32000,
-        MIN_LAYER_SIZE = 0.1;
+        DEFAULT_MIN_LAYER_SIZE = 1,
+        VECTOR_MIN_LAYER_SIZE = 0.1;
 
     var Size = React.createClass({
         mixins: [FluxMixin, CoalesceMixin],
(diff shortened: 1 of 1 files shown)


iwehrman:
@ktaki If you have a chance, please test to make sure I didn't screw anything up!

volfied:
LGTM, @ktaki please merge if nothing bad pops out of testing.

ktaki:
When W and H values are linked, it is still possible to make <1. Try with a text box layer W=200 and H=600. Link the value, then set H=1. This will result in the W to be less than 1 and reproduce the original issue. In fact, this can be an issue with vector layers also. It is possible to set to the value smaller than 0.1.

iwehrman:
@ktaki I think I'd like to call that a separate bug. I've filed https://jira.corp.adobe.com/browse/PS-1384 to request that Photoshop simply clamp these values to a valid range. The alternative is to recreate some fairly complicated bounds-calculation logic from Photoshop into the front end, which I'd certainly rather not do. I've filed #3725 to capture this.

Back to @volfied in case he agrees that that can be considered a separate issue which, in my opinion, is equally severe but that users are less likely to encounter the bug that this PR addresses.

volfied:
If we're tracking the other issue, I'm ok with merging this one. And yes, this should be fixed in PS and PS should not allow us to set layer size smaller than possible.

PR #3718, opened 2016-02-17
Insert token in vulcan callback for libraries API if exists

volfied:
This is in preparation for the next libraries API. As it is, will be a no-op.


iwehrman:
LGTM.

Issue #3693, opened 2016-02-09
Hitting Cmd/Ctrl+Period in text editing mode causes an error


ktaki:
Press Cmd/Ctrl+period to cancel text editing mode will produce:
Uncaught TypeError: this.flux.actions.tool.type.handleTypeModalStateCanceled is not a function

![screen shot 2016-02-09 at 3 04 48 pm](https://cloud.githubusercontent.com/assets/2029597/12933790/dbf8b170-cf3e-11e5-9739-9ab033361033.png)

ktaki:
Note that the result is the same by hitting Esc key if Esc key is set to cancel the text editing.

chadrolfs:
I wonder if this is related to the native dialogs bugs #3681

shaoshing:
This is fixed by #3719. Back to Taki to verify.

ktaki:
Verified the fix.

PR #3719, opened 2016-02-17
Remove typetool.js and move two actions to tool/type.js

volfied:
Seems like this was a now unused file, and we had both `actions/typetool.js` and `actions/tool/type.js`. So I moved the last two actions to tool/type so they can be used correctly.

Addresses #3693


shaoshing:
LGTM!

PR #3721, opened 2016-02-18
Show document panels earlier by promoting the fade-in animation.

shaoshing:
This PR promote the panel's fade-in animation so that they are displayed faster during startup. Before, the animation starts when the `FluxController` emits `ready` event. In general, the event is emitted 150ms - 300ms later after the `PanelSet` completes rendering for the active document, as the event has to wait until all `beforeStartpu` callbacks are completed. So, instead of waiting for the `ready` event, the `PanelSet` now starts the animation when its active document's content are flushed to the DOM.

screen shot 2016-02-18 at 8 51 43 am
1) the panel content are flushed to the DOM. 2) `Main` component receives the `ready` event.

#### Improvements

| panel fade-in starts | Before | After | Diff |
| --- | --- | --- | --- |
| empty document | 1690ms | 1530ms | -160ms |
| vermilion | 2630ms | 2410ms | -220ms |

### DIFF
--- a/src/js/jsx/PanelSet.jsx
+++ b/src/js/jsx/PanelSet.jsx
@@ -170,6 +170,17 @@ define(function (require, exports, module) {
                 prevProps.singleColumnModeEnabled !== this.props.singleColumnModeEnabled) {
                 this._updatePanelSizes();
             }
+
+            // Turn on the ready state to start the fade-in transition. This only happens 
+            // once at the initial display. To trigger the CSS transition, the state must be 
+            // set AFTER the panel components are flushed to the DOM.
+            if (!this.state.ready && this._isActiveDocumentInitialized()) {
+                // Set the state in a `setTimeout` callback to make sure the PanelSet 
+                // is painted before the transition. This will prevent possible flashing.
+                window.setTimeout(function () {
+                    this.setState({ ready: true });
+                }.bind(this), 0);
+            }
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
@@ -198,6 +209,7 @@ define(function (require, exports, module) {
                 }, this);
 
             return panelVisibilityChanged ||
+                this.state.ready !== nextState.ready ||
                 this.state.activeDocumentInitialized !== nextState.activeDocumentInitialized ||
                 this.state.recentFilesInitialized !== nextState.recentFilesInitialized ||
                 (nextState.documentIDs.size === 0 && !Immutable.is(this.state.recentFiles, nextState.recentFiles)) ||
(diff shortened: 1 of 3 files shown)


src/js/jsx/PanelSet.jsx
```diff
@@ -170,6 +170,12 @@ define(function (require, exports, module) {
                 prevProps.singleColumnModeEnabled !== this.props.singleColumnModeEnabled) {
                 this._updatePanelSizes();
             }
+            
+            // Turn on the ready state to start fade-in transition for panels. To ensure CSS transition, the 
+            // state must be set after the panel components are flushed to the DOM.
+            if (!this.state.ready && this._hasActiveDocument()) {
```
iwehrman: Would you mind adding to this comment, to help make it obvious later, that this will only happen once?

src/js/jsx/PanelSet.jsx
```diff
@@ -269,11 +276,19 @@ define(function (require, exports, module) {
             this.getFlux().actions.preferences.setPreferences(nextState);
             headlights.logEvent("user-interface", "panel-toggle", panelName);
         },
+        
+        /**
+         * True if the current docuemnt is initialized and its panels should be displayed.
+         * 
+         * @return {Boolean}
```
iwehrman: Type expressions for primitive types are always lower case: `boolean`, `number`, `object`, etc.

src/js/jsx/PanelSet.jsx
```diff
@@ -269,11 +276,19 @@ define(function (require, exports, module) {
...
+         */
+        _hasActiveDocument: function () {
+            return this.state.activeDocument && this.state.activeDocumentInitialized && this.state.documentIDs.size > 0;
+        },
 
         render: function () {
-            var documentIDs = this.state.documentIDs,
-                activeDocument = this.state.activeDocument,
-                hasActiveDocument = activeDocument && this.state.activeDocumentInitialized && documentIDs.size > 0;
+            var activeDocument = this.state.activeDocument,
```
iwehrman: Might be slightly nicer to just have `_getActiveDocument` instead of `_hasActiveDocument` which returns the document if the other conditions hold and null otherwise?

src/style/panel.less
```diff
@@ -227,6 +227,10 @@
     opacity: 0;
 }
 
+.artboard-presets, .recent-files {
```
iwehrman: Is this an unrelated change or fix?

iwehrman:
Looks good so far. It's too bad that the simple structure of `beforeStartup -> ready -> UI drawn` didn't hold up. I guess as long as the view doesn't try do something as a result of having an initialized document, and assuming that some other store has its other dependent data ready. Because that store may not yet be fully initialized as a result of some other action running it's `beforeStartup` method a little later. Still, that's only a theoretical concern, and this seems like a fine way to get the UI drawn faster. :+1:

Back to @shaoshing to address a few minor suggestions.

src/js/jsx/PanelSet.jsx
shaoshing: Updated comment to indicate the transition happens once.

src/js/jsx/PanelSet.jsx
shaoshing: I find `_getActiveDocument` is confusing with `this.state.activeDocument`. The difference is that `_getActiveDocument` only returns the an initialized active document. Also, in the `render` function, un-initialized active document is accepted, but the panels will remain hidden. So, instead of changing the return value, I rename `_hasActiveDocument ` to `_isActiveDocumentInitialized`, which looks a little better. Please let me know if you still think we should use `_getActiveDocument` instead.

src/style/panel.less
shaoshing: This change is a fix for the panel height in no-doc state. Before, the two panels didn't expand to the full height of the window.

shaoshing:
@iwehrman I've addressed your comments. Also addressed a potential flashing issue in https://github.com/adobe-photoshop/spaces-design/pull/3721/files#diff-6ca92cfb1bbcf69e9825408c732136d7R177. Back to you for a second review.

iwehrman:
Looks good, works good!

Issue #3444, opened 2015-12-18
Radius shows for rotated shapes in some cases


volfied:
Create a rectangle
Rotate it
See that radius control is still showing, even though it shouldn't. It's a no-op if you change it.

It shouldn't show for shapes we can't set radius on.

iwehrman:
@DivyaPrabhakar is this still fix in progress? If not, I suggest kicking it over to @volfied, who may have some spare cycles in the short term.

DivyaPrabhakar:
I will give it to @volfied because of the new developments with Stock search, I probably won't get to this faster than he will.

volfied:
Problem is, we only reset bounds, but we need to reset the radii of the layer as well, so we can see that 
`descriptor.keyOriginType[0].keyShapeInvalidated` is true and set radii to null.

iwehrman:
Over to @ktaki to verify.

ktaki:
Verified the fix.

PR #3724, opened 2016-02-18
Update keyOriginType for layers on handleLayerTransform

volfied:
This PR does few things:

 - Changes `_boundsPropertyForLayer` function to `_boundsPropertiesForLayer`, allowing for multiple property keys to be attached to a layer bound. This allows us to get `keyOriginType` as well for vector layers on `resetBounds` calls.
 - When a vector layer is rotated, it's keyOriginType changes as well, which tells us the layer can no longer set corner radii. So on `resetBounds` call, we update both of those properties, also in `verifySelectedLayerBounds`

Addresses #3444

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -254,23 +254,23 @@ define(function (require, exports) {
     };
 
     /**
-     * The property on the Photoshop layer descriptor that corresponds to bounds.
+     * The list of properties on the Photoshop layer descriptor that corresponds to bounds.
      *
      * @private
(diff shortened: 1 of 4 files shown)


iwehrman:
Can you please do some benchmarking to see if this causes any kind of performance regression?

volfied:
@iwehrman 

Master:



This branch:



It slows down from 20 ms to 30 ms, only when there are shape layers involved.

iwehrman:
Looks good, thanks!

PR #3727, opened 2016-02-19
Console/Timeline improvements

shaoshing:
This PR includes the following improvements to the Console and Timeline:

#### Togglable action log
You will be able to disable action log and timestamp completely by turning `Log Actions` and `Log Action Transfers` off. This is useful when you need a clean console or Timeline.

screen shot 2016-02-18 at 3 48 50 pm

#### Better descriptor log
- Descriptor log now shows the params/result in a collapsed group, which avoids overcrowding the console.
- Has stack trace
- Show completion time. This helps target slow descriptors.
- Show timestamp in the Timeline.
- Descriptor calls have unique ID to help mapping the `executing/finished` message in the console, and mapping between the console and timeline.

screen shot 2016-02-18 at 3 49 23 pm

#### Startup timestamp
screen shot 2016-02-18 at 3 48 29 pm

### DIFF
--- a/src/js/actions/debug.js
+++ b/src/js/actions/debug.js
@@ -337,6 +337,24 @@ define(function (require, exports) {
     };
 
     /**
+     * Debug-only method to toggle action logging
+     *
+     * @return {Promise}
+     */
+    var toggleActionLogging = function () {
+        var preferencesStore = this.flux.store("preferences"),
+            preferences = preferencesStore.getState(),
+            enabled = preferences.get("logActions");
+
+        return this.transfer("preferences.setPreference", "logActions", !enabled);
+    };
+    toggleActionLogging.action = {
+        reads: [],
+        writes: [locks.JS_PREF],
+        transfers: ["preferences.setPreference"]
+    };
+
+    /**
      * Debug-only method to toggle action transfer logging
      *
      * @return {Promise}
@@ -554,6 +572,7 @@ define(function (require, exports) {
     exports.resetRecess = resetRecess;
     exports.togglePolicyFrames = togglePolicyFrames;
     exports.togglePostconditions = togglePostconditions;
+    exports.toggleActionLogging = toggleActionLogging;
     exports.toggleActionTransferLogging = toggleActionTransferLogging;
     exports.logDescriptor = logDescriptor;
     exports.logHeadlights = logHeadlights;
--- a/src/js/actions/debug.js
+++ b/src/js/actions/debug.js
@@ -392,43 +392,66 @@ define(function (require, exports) {
         }
 
         var blockStyle = "color:gray;",
-            eventStyle = "font-weight:bold";
-    
+            groupStyle = "color:blue;font-weight:normal",
+            // Each descriptor log has a unique ID to help mapping between console message and timeline stamp.
+            descritorID = 500;
+
         descriptor.__getAsync = descriptor.__getAsync || descriptor._getAsync;
         descriptor.__batchPlayAsync = descriptor.__batchPlayAsync || descriptor._batchPlayAsync;
         descriptor.__eventHandler = descriptor.__eventHandler || function (eventID, obj) {
-            log.debug("[Descriptor] PS Event: %c%s\n%c%s", eventStyle, eventID, blockStyle,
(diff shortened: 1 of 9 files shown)


src/index-debug.html
```diff
@@ -26,6 +26,9 @@
     <meta charset="utf-8">
     <title>Photoshop Design Space (development build)</title>
     <link rel="stylesheet" type="text/css" href="style.css" />
+    <script type="text/javascript">
+        var loadTime = new Date();
```
iwehrman: I'm confused about how this ends up in the global scope...?

iwehrman: I guess all declarations at the top-level end up in the global scope. Weird. Anyway, I don't like the idea of polluting the global scope. Can we do this locally in `main.startup`?

iwehrman:
One thing I don't like about this is that it's actually frequently very useful for me to see the exactly log of when actions were enqueued, executed and finished. By grouping them like this, it certainly does tidy things up, but also obscures how action execution is interleaved.

src/index-debug.html
shaoshing: Looks like we can use Chrome's `performance.timing` object to retrieve useful information like the time when browser starts loading the page (https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming). With that, we no longer need the record the load time ourself.

shaoshing:
@iwehrman I've eliminated the use of global `loadTime`, and switch to the descriptor call's real ID. Back to you for reviewing.

src/index-debug.html
shaoshing: oh, the top-level declarations. They will become global variables, unless in `strict mode`.

iwehrman:
Besides my other comments offline, can we also make action times enabled by default (in debug builds)? This is just to reduce confusion in the short term.

iwehrman:
Also, it doesn't seem like the traces in the descriptor "finished" messages are ever very interesting. But maybe I've just been missing the interesting cases.

iwehrman:
Also, it might be possible to format these messages a little differently to make it easier to match up the calls. For example, instead of:

```
[Descriptor] Executing get - 1544
[Descriptor] Finished get in 21ms - 1544
```
it might help to align the IDs and stick the time at the end:
```
[Descriptor] Executing get - 1544
[Descriptor] Finished get - 1544 in 21ms
```

And, as mentioned offline, red highlighting for long calls would be awesome :)

shaoshing:
@iwehrman below is the updated version

- highlight slow descriptor (> 100ms)
- remove trace for the `finished` log
- move time to the end
- show summarized commands for batchPlay
- Enable `Log Actions` by default

screen shot 2016-02-19 at 12 46 37 pm

shaoshing:
Fixed the repeated `in` in the message

src/js/fluxcontroller.js
```diff
@@ -781,7 +783,8 @@ define(function (require, exports, module) {
     FluxController.prototype._applyAction = function (action, actionReceiver, params, parentActionName) {
         var flux = this.flux,
             actionObject = action.action,
-            actionName = this._actionNames.get(action);
+            actionName = this._actionNames.get(action),
+            logActions = __PG_DEBUG__ && this.flux.store("preferences").getState().get("logActions") !== false;
```
iwehrman: A slightly nicer way to do this is `.get("logActions", true)`.

src/js/actions/debug.js
```diff
@@ -374,43 +400,124 @@ define(function (require, exports) {
...
+                return id;
+            }, {
+                context: window._spaces.ps.descriptor,
+                multiArgs: true
+            });
 
-                    return result;
+            var summarizeCommands = function (commands) {
+                if (commands.length === 0) {
+                    return "EMPTY";
```
iwehrman: Can we just make this the empty string here so this shows up as `Executing batchPlay: [] - 123`?

iwehrman:
Well, this is great. Two more micro nits and then I'll merge.

shaoshing:
Nits addressed! Back to you @iwehrman

iwehrman:
Super cool!

Issue #3732, opened 2016-02-19
hit test should be disabled when cursor is outside of window area


shaoshing:
Currently, when PS is focused, moving cursor outside of the window area will trigger hit tests.

![hittest](https://cloud.githubusercontent.com/assets/118264/13192135/c25182f6-d71d-11e5-97e2-ac13e192a4f1.gif)

shaoshing:
Fixed by #3733

PR #3733, opened 2016-02-19
Only update overlays when pointer is inside canvas bounds

iwehrman:
What it says! This cuts down on some pointless, wasteful calls to `hitTest`.

Addresses #3732.

### DIFF
--- a/src/js/jsx/tools/GuidesOverlay.jsx
+++ b/src/js/jsx/tools/GuidesOverlay.jsx
@@ -87,7 +87,8 @@ define(function (require, exports, module) {
                 document: currentDocument,
                 tool: currentTool,
                 modalState: modalState || appIsModal,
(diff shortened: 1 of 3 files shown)


volfied:
This only works for top and left areas of the window. I mean, it's still better than calling it all the time, but we should instead check with panel store to avoid making calls when mouse pointer is out of the visible canvas.

iwehrman:
Derp. I'll try again.

iwehrman:
OK, let's try again! Now I'm checking against the canvas bounds from the panel store. And on all four sides! :sweat_smile:

iwehrman:
But, really, who even moves their mouse that far to the right anyway?

shaoshing:
Verified all four sides. Thanks for fixing!

Issue #3736, opened 2016-02-21
Artboard guides visible but without policies when child layer is selected


iwehrman:
0. Create a new artboard document.
1. Select the artboard.
2. Drag in a guide to create an artboard guide.
3. Deselect all layers by clicking in the panel.
4. Select the child layer of the artboard.

Result: guide is visible, but has no pointer policy, and so can't be moved.

Secondary issue:

0. Create a new artboard.
1. Select the child layer.
2. Drag out an artboard guide.
3. Deselect all layers.

Result: the guide is hidden, but the pointer policies remain incorrectly. 

CC @volfied

iwehrman:
Fixed by #3738.

PR #3738, opened 2016-02-22
Compare top selected ancestors for border policy reset

volfied:
Beautiful edge case, create artboard guide, deselect all layers, then select a layer with no bounds.

We would only check for bounds, and a layer with no bounds would be equal to no layers selected. So now we also check to see if top selected ancestors differ, since they drive which artboard guides show.

Adresses #3736


iwehrman:
LGTM!

Issue #3704, opened 2016-02-12
Unable to rotate/transform masks in certain situations (hitTest fail)


chadrolfs:
create a group of two shapes
create a shape outside the group 
select the group and the shape outside the group
Layer > Create Vector Mask from Layer
Rotate the masked group repeatedly

Result: eventually it will fail

chadrolfs:
![maskgrouptransformfail](https://cloud.githubusercontent.com/assets/8442631/13020867/97502f9e-d18c-11e5-8687-121166f0ba6c.gif)

chadrolfs:
https://www.dropbox.com/s/36u3nlgf2lqnczu/MaskGroup.psd?dl=0

volfied:
Just so the problem is clear: ![](https://s3.amazonaws.com/f.cl.ly/items/0b2i3r1l0C0v3p1e3o2E/Image%202016-02-23%20at%2012.54.08%20PM.png?v=0fe2ed3d)

chadrolfs:
Almost totally, there.  There is one case where it still fails; exiting Mask Edit Mode.
Open existing file
select group that has mask, hit M to enter Mask Mode, double-click to drill into the mask and move anchors
hit Esc x2 to exit Mask Mode
then try to rotate

Result: rotate fails, no policy on group on Mask Mode exit

![exitmask-nopolicy](https://cloud.githubusercontent.com/assets/8442631/13269967/c2d4d550-da3f-11e5-98f2-38a0885fab7d.gif)

volfied:
@pineapplespatula Please take a look at the new problem. It has to do with the fact that pressing Enter calls `tools.enterPathModalState` and escape calls `mask.handleEscapeVectorMask`, but resetBorderPolicies is not called when that happens. You know that code better.

PR #3744, opened 2016-02-23
Consider vector masks of groups in their bounds

volfied:
In a situation where a vector mask was created for a group, we would still be calculating it's bounds using the child bounds. But the transform controls show for the union of child bounds and the mask bounds.

This PR makes sure we get the required properties for the group bounds, and use them in bounds calculations.

Addresses #3704 and potentially other mask related group bounds issues


pineapplespatula:
Looks good to me

Issue #3609, opened 2016-01-25
Exporting with multiple layers the same name only saves one


chadrolfs:
cmd/ctrl + n 
hold Opt/Alt, click and drag on the artboard title (x4)
You now have 4 artboards named the same
Cmd/Ctrl + E or File > Export
Click Export

Result: only one file is exported.

chadrolfs:
relates to #2274 and #3543

mcilroyc:
Any hot ideas on how to resolve file naming conflicts?  Suffix with a number?  `LayerOne.png`, `LayerOne.1.png` ... etc.  @placegraphichere ?

mcilroyc:
example: ![ss](http://cl.ly/2M1s2i223k01/Image%202016-02-23%20at%204.33.23%20PM.png)

chadrolfs:
Ideas (not necessarily hot):
file #.ext
file(#).ext
file.#.ext
file-#.ext
file copy #.ext (this is what Mac OS does when you dupe a file)

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Blocked by windows bug 3743

PR #3747, opened 2016-02-23
When exporting a set of layers with some duplicate layer names, use a suffix to distinguish

mcilroyc:
Semi-hokey fix for #3609

If you exporting a batch of layers, and there are layer-name collisions, use a simple suffix to avoid a filename collision on the exported assets.

example (updated):
ss

### DIFF
--- a/src/js/actions/export.js
+++ b/src/js/actions/export.js
@@ -170,20 +170,19 @@ define(function (require, exports) {
      * @param {number} assetIndex index of this asset within the layer's list
      * @param {ExportAsset} asset asset to export
      * @param {string=} baseDir Optional directory path in to which assets should be exported
-     * @param {string=} prefix Optional prefix to apply to the fileName
+     * @param {string=} baseFilename Base file name to which things may be prefixed/suffixed
+     * @param {string=} prefix Optional prefix to apply to the baseFilename
      * @return {Promise}
      */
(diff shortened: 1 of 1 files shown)


placegraphichere:
When using "keep both" on OSX, finder uses a space+2 for files. So your example would yield `square 2@3x.png` for the first duplicate, incrementing after that. Perhaps we should follow that pattern instead of `.1`?

mcilroyc:
Updated PR based on the " 2" suggestion from @placegraphichere.
now:
ss

chadrolfs:
I like it.

src/js/actions/export.js
```diff
@@ -170,20 +170,19 @@ define(function (require, exports) {
      * @param {number} assetIndex index of this asset within the layer's list
      * @param {ExportAsset} asset asset to export
      * @param {string=} baseDir Optional directory path in to which assets should be exported
-     * @param {string=} prefix Optional prefix to apply to the fileName
+     * @param {string=} baseFilename Base file name to which things may be prefixed/suffixed
```
iwehrman: This should be `baseFileName`.

lol j/k.

iwehrman:
Looks good! It might also be nice to have a "warning" symbol in the export dialog if there are layers marked for export with identical names. Adding this number suffix is clearly better than just over-writing all but one of the assets, but ultimately this seems worthy of a warning.

Issue #3737, opened 2016-02-22
Tool global shortcuts should include "tool" in the option name


iwehrman:
1. Each of the global shortcut entries for tools should say something like "Rectangle Tool" instead of just "Rectangle".
2. Officially it's just the "Select" tool, not the "Super Select" tool.
3. "Vector Mask" should be "Vector Mask Mode".
4. I don't know what the `A` entry should be exactly, but it isn't "Super Select Vector".
5. Also, it shouldn't show up when not in Mask Mode, because nothing happens when I press A or choose that option.

![image](https://cloud.githubusercontent.com/assets/1075154/13207902/a1581fe0-d8c8-11e5-86c4-869a99a88b2b.png)

iwehrman:
Over to @chadrolfs to verify.

PR #3749, opened 2016-02-24
Changes to search bar: edits the tool names, direct selection vector option conditional

DivyaPrabhakar:
References issue #3737 

- All tools have "Tool" succeeding their individual names 
- "Super Select" --> "Select Tool"
- "Vector Mask" --> "Vector Mask Mode"
- "Super Select Vector" --> "Direct Selection Tool"
- "Direct Selection Tool" is not an option in Search > Global Shortcuts if we are not in Vector Mode

### DIFF
--- a/src/js/actions/tools.js
+++ b/src/js/actions/tools.js
@@ -65,6 +65,8 @@ define(function (require, exports) {
      */
     var _currentTransformPolicyID = null;
 
+    var directSelectToolID = "superselectVector";
+
     /**
      * Installs the defaults on a given shape tool
      * 
@@ -645,6 +647,7 @@ define(function (require, exports) {
             dispatchPromise,
             removePromise,
             policyPromise,
+            directSelectShortcutPromise,
             createMaskOptions = {
                 historyStateInfo: {
                     name: nls.localize("strings.ACTIONS.ADD_VECTOR_MASK"),
@@ -706,8 +709,13 @@ define(function (require, exports) {
                     this.dispatch(events.tool.VECTOR_MASK_POLICY_CHANGE, policyID);
                 });
             removePromise = Promise.resolve();
+
+            directSelectShortcutPromise = this.transfer(shortcuts.addShortcut, utilShortcuts.GLOBAL.TOOLS.VECTOR_SELECT,
+                {}, _vectorSelectMaskHandler, directSelectToolID,
+                true, nls.localize("strings.TOOLS.NAMES.DIRECT_SELECTION_TOOL"));
         } else {
             // delete empty masks when leaving mask mode
+            var shortcutStore = this.flux.store("shortcut");
             removePromise = this.transfer(layers.resetLayers, currentDocument, currentLayer)
                 .bind(this)
                 .then(function () {
@@ -745,6 +753,10 @@ define(function (require, exports) {
                     }
                 });
 
+            if (shortcutStore.getByID(directSelectToolID)) {
+                directSelectShortcutPromise = this.transfer(shortcuts.removeShortcut, directSelectToolID);
+            }
+
             var pointerPolicyID = toolStore.getVectorMaskPolicyID();
          
             if (pointerPolicyID) {
@@ -766,11 +778,13 @@ define(function (require, exports) {
 
                 firstLaunch = false;
                 return Promise.join(initPromise, defaultPromise, resetPromise, dispatchPromise, policyPromise,
-                    removePromise);
+                    removePromise, directSelectShortcutPromise);
             } else if (!vectorMaskMode) {
-                return Promise.join(initPromise, resetPromise, dispatchPromise, policyPromise, removePromise);
+                return Promise.join(initPromise, resetPromise, dispatchPromise, policyPromise,
+                    removePromise, directSelectShortcutPromise);
             } else {
-                return Promise.join(initPromise, resetPromise, dispatchPromise, policyPromise, removePromise)
+                return Promise.join(initPromise, resetPromise, dispatchPromise, policyPromise,
+                    removePromise, directSelectShortcutPromise)
                 .then(function () {
                     return UI.setSuppressTargetPaths(false);
                 })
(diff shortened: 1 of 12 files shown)


src/js/actions/tools.js
```diff
@@ -985,21 +1000,28 @@ define(function (require, exports) {
...
+            _vectorSelectMaskHandler = function () {
+                flux.actions.select(toolStore.getToolByID(directSelectToolID));
+            }.bind(this);
+
+            shortcutSpecs.push({
+                key: utilShortcuts.GLOBAL.TOOLS.VECTOR_SELECT,
+                modifiers: {},
+                id: directSelectToolID,
+                fn: _vectorSelectMaskHandler,
+                name: "Direct Selection Tool"
```
iwehrman: I think this needs to be a localized string?

mcilroyc: If so, then don't forget about Vector Mask Mode above.  What are these shortcut names used for...?

src/js/models/tools/ellipse.js
```diff
@@ -47,7 +47,7 @@ define(function (require, exports, module) {
             backspaceKeyPolicy = new KeyboardEventPolicy(UI.policyAction.PROPAGATE_TO_BROWSER,
                 OS.eventKind.KEY_DOWN, null, OS.eventKeyCode.BACKSPACE);
         
-        Tool.call(this, "ellipse", "Ellipse", "ellipseTool", "tool.ellipse.select");
+        Tool.call(this, "ellipse", "Ellipse Tool", "ellipseTool", "tool.ellipse.select");
```
mcilroyc: If this string (and others elsewhere) are important enough for you to want to add " Tool" here... then they are probably important enough to localize.

src/js/actions/tools.js
```diff
@@ -766,11 +776,13 @@ define(function (require, exports) {
 
                 firstLaunch = false;
                 return Promise.join(initPromise, defaultPromise, resetPromise, dispatchPromise, policyPromise,
-                    removePromise);
+                    removePromise, directSelectShortcutPromise);
             } else if (!vectorMaskMode) {
```
mcilroyc: attn @pineapplespatula  wrt vector mask mode

src/js/models/tools/ellipse.js
iwehrman: I thought that the tool names in these constructors were vestigial and unused? @DivyaPrabhakar: do you know if these are the strings that show up in the search UI? If so, that's a problem because they're not localized.

src/js/actions/tools.js
iwehrman: The _shortcut_ names are used in the search UI.

iwehrman:
This line is the problematic one:
https://github.com/adobe-photoshop/spaces-design/blob/0635b17017c3fa7ebde207d7916be285f6d15d1d/src/js/actions/tools.js#L977

Instead of using this tool name (which can and should be completely eliminated from our code), we should be using localized strings that are keyed by tool ID. This might necessitate a new section in our strings.js dictionary specifically for these entries.

src/js/models/tools/ellipse.js
DivyaPrabhakar: As a matter of fact, they are the ones in the UI and are the ones being used by the search bar which is why I had to manually change them.

DivyaPrabhakar:
@iwehrman We talked about your comment regarding `tool.name` -- let me know if my changes address  your concern.

src/nls/en/strings.json
```diff
@@ -150,13 +150,19 @@
...
+            "VECTOR_MASK_MODE": "Vector Mask Mode",
+            "ELLIPSE": "Ellipse Tool",
+            "PEN": "Pen Tool",
+            "RECTANGLE": "Rectangle Tool",
+            "EYEDROPPER": "Sampler Tool",
+            "SELECT": "Select Tool",
+            "TYPE": "Type Tool"
+        },
+        "SHORTCUTS" : {
+
```
iwehrman: What's going on here?

src/js/actions/tools.js
```diff
@@ -985,21 +1002,28 @@ define(function (require, exports) {
...
+            _vectorSelectMaskHandler = function () {
+                flux.actions.select(toolStore.getToolByID(directSelectToolID));
+            }.bind(this);
+
+            shortcutSpecs.push({
+                key: utilShortcuts.GLOBAL.TOOLS.VECTOR_SELECT,
+                modifiers: {},
+                id: directSelectToolID,
+                fn: _vectorSelectMaskHandler,
+                name: nls.localize("strings.TOOLS.NAMES.DIRECT_SELECTION_TOOL")
```
iwehrman: Can we do this once in the tool itself? `js/models/tools/superselect/vector`?

DivyaPrabhakar: It depends on the vector mask state and all of that is being calculated in `tools.js`, so that is why it is being added separately after we check for `toolStore.getVectorMode()`.

src/nls/en/strings.json
DivyaPrabhakar: lol nothing. my bad.

src/js/actions/tools.js
iwehrman: I don't mean the shortcut registration, I mean the name lookup. Like, can this line be `toolStore.getToolByID("directSelect").name`? (Or whatever the tool ID is.)

src/js/actions/tools.js
pineapplespatula: this part LGTM.

src/js/actions/tools.js
```diff
@@ -65,6 +65,8 @@ define(function (require, exports) {
      */
     var _currentTransformPolicyID = null;
 
+    var directSelectToolID = "superselectVector";
```
DivyaPrabhakar: @iwehrman Should this string be stored in `stores/tools.js` and then retrieved from there or is this ok? If the latter, can you explain why?

DivyaPrabhakar:
@iwehrman Back to you for with the changes.

DivyaPrabhakar:
@iwehrman Back to you with the changes and one question

src/js/actions/tools.js
iwehrman: There isn't a super satisfying way to deal with shared keys like this. I think this is fine, and I think it is also fine to just use an inline string. You could make it a property of a store, but then you'd have to look up that property with another string... So, whatev!

iwehrman:
Works good, looks great. Here we go!

Issue #3375, opened 2015-12-07
Layers Panel: CMD+SHIFT+] should scroll panel 


placegraphichere:
When moving layers with CMD+] or any of those variations, the layers panel should scroll to follow if necessary. If the layer/s are within the view in any way, then we should not scroll. It should only scroll when the selected layer is moved out of viewable range.

Ps standard does this.

iwehrman:
I'm surprised this doesn't already work. We should be scrolling to the new selection whenever it's out of view...

iwehrman:
Can you take a look at this in the context of your improved `LayerFace` scroll-to-selection implementation?

iwehrman:
We might just need an action (besides selection) like `layers.exposeLayer`.

iwehrman:
Over to @chadrolfs to verify.

PR #3755, opened 2016-02-24
Scroll layers panel when layer is invisible due to change of position

shaoshing:
Fix for issue #3375

I have to move the `scroll-into-view` logic into the `LayerGroup` as ordering `LayerFace` within the same group will not re-render the layer face, which makes it difficult to know when to scroll the panel as the `LayerFace#componentDidUpdate` is not called. This also result in a cleaner solution for scrolling the panel. Now the scroll works for:

* order layer via menu items (e.g. Bring Forward `cmd+]`)
* undo/redo layer order

### DIFF
--- a/src/js/jsx/sections/layers/LayerGroup.jsx
+++ b/src/js/jsx/sections/layers/LayerGroup.jsx
@@ -25,6 +25,7 @@ define(function (require, exports, module) {
     "use strict";
 
     var React = require("react"),
+        ReactDOM = require("react-dom"),
         Fluxxor = require("fluxxor"),
         FluxMixin = Fluxxor.FluxMixin(React),
         classnames = require("classnames"),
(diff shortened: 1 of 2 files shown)


src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -112,6 +146,20 @@ define(function (require, exports, module) {
...
+                this._selectedLayerHasNewIndex = false;
+                this._scrollIntoFirstSelectedLayer();
+            }
+        },
+
+        /**
+         * Scroll into the first selected layer if it is not visible in the layers panel when
+         */
+        _scrollIntoFirstSelectedLayer: function () {
+            window.document.querySelector(".face__selected").scrollIntoViewIfNeeded();
```
iwehrman: I guess this is just an aesthetic thing, but would it be possible to put the method that actually performs the scroll into the face component, and then have the group use a child ref to call that method?

iwehrman: (Just to avoid having to do use `document.querySelector`.)

iwehrman:
Seems fine, just one question, which may or may not be silly.

shaoshing:
@iwehrman I've replaced the `querySelector` with React component API, which looks more elegant now :D Back to you.

iwehrman:
Lovin' it :hamburger: :fries:

Issue #3748, opened 2016-02-24
Initial coordinates of an object moved out of an artboard are incorrect


chadrolfs:
create a shape within an AB (note the X,Y coords)
move shape either to a 2nd AB or outside the current AB

Result: coordinates are incorrect  (changing them via X,Y have unexpected results)

chadrolfs:
![ab-coords2](https://cloud.githubusercontent.com/assets/8442631/13271846/a7c6e020-da4c-11e5-9f8e-df6aca937ba7.gif)

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
Coordinates for layers outside an AB are still really confusing but we have a JIRA card for that work.  Closing this as drags between ABs now result in correct coords

PR #3758, opened 2016-02-25
Stop listening to moveToArtboard events, and look for it in move events

volfied:
Starting with pg-dev-mac#1104, we will get an array named `moveToArtboard` inside the `move` and `duplicateNewSheets` commands that will tell us all the layers that changed artboards with structure: `[ { layer: #, artboard: # }, ...]`

Whoever is reviewing, please do a quick testing as well (I've done some but it's nice to have extra eyes)

Also addresses #3748

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -59,12 +59,6 @@ define(function (require, exports) {
     };
 
     /**
-     * @const
-     * @type {number}
-     */
-    var EVENT_DEBOUNCE_DELAY = 100;
-
-    /**
      * Filter out layers that can't generally be transformed like empty,
      * adjustment and background layers.
      *
@@ -1352,6 +1346,16 @@ define(function (require, exports) {
                         return this.transfer("history.newHistoryStateRogueSafe", currentDoc.id);
                     })
                     .then(function () {
+                        // We want resetIndex to play first since it doesn't update the bounds
+                        // and will prevent a blink in Position component 
+                        // in case layers actually change artboards
+                        if (event.moveToArtboard && event.moveToArtboard.length > 0) {
+                            return this.transfer("layers.resetIndex");
+                        } else {
+                            return Promise.resolve();
+                        }
+                    })
+                    .then(function () {
                         // We need to get the updated document model since selection may have changed
                         // due to drag optimization
                         currentDoc = appStore.getCurrentDocument();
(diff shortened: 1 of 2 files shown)


iwehrman:
The diff looks fine, but I ran across what may well be a distinct bug:
img
Note that when I drag the rectangle inside the artboard, it's Y coordinate changes to something that seems wrong, and then is different after I deselect and reselect the layer.

I'm using Mac pgdev.1105.

iwehrman:
Anyway, just let me know if this is a completely separate issue, from what you can tell. If so, I'll file it and merge this. Back to @volfied.

chadrolfs:
@iwehrman that could be bug #3748

iwehrman:
Any thoughts on this @volfied?

volfied:
I've actually realized that my core changes are not correct. While testing, I missed on the fact that I was getting last drag event's artboard changes (due to old code emitting "move" event before posting it).

So I'm working on fixing that, and once I fix it hopefully these changes will be correct. Stay tuned.

volfied:
@chadrolfs @iwehrman with CL 1065232, pg-dev-mac 1117, this PR should correctly work.

I've explained the problem above, and fixed it with this new CL.

chadrolfs:
looking at this branch with x1117, I still see issues with coordinates, specifically wrt Y, dragging out of then into an AB.
ab-coordsoff

volfied:
@chadrolfs one more time, with build 1119, and update branch (I just pushed a new commit to this PR)

Now PS will send the events at the right time (after it's completely processed), and we also needed to relax TransformPanel's `shouldComponentUpdate` to check for selectedRelativeChildBounds so it knew to re-render the X/Y fields with the correct position after canvas shifted.

iwehrman:
This gif shows some weird behavior with the Y coordinate of a layer outside of an artboard:
img

chadrolfs:
I noticed the same thing @iwehrman called out.  Also a couple of minor issue, sometimes the policy doesn't update until you move the mouse and sometimes the policy will jump while moving.  Here's a gif showing both of those minor issues.

ab-moveupdatepolicy

chadrolfs:
ab-move

iwehrman:
The policy delay probably has to do with the frequency with which PS sends us these move events. In any case, I think that should be benign?

volfied:
@chadrolfs Policy delay is due to us debouncing handleCanvasAutoResize events. The second gif is also due to same events, top level layers are counted like artboards in the sense that they will stretch/shrink the canvas. And this is done in tiles, that's why Y jumps from 1024 to 85 when you move it downwards. 

At this point, none of these problems can be solved with JS, nor do I think they are in the scope of this PR and the bug it's attempting to fix.

iwehrman:
OK!

PR #3760, opened 2016-02-26
Barkin/3753

pineapplespatula:
I added an additional change to your fix, which requires a vector mask PR https://github.com/adobe-photoshop/spaces-adapter/pull/190


volfied:
Thanks @pineapplespatula !

Issue #3764, opened 2016-02-29
Startup failure resetting DS with no layer selected


chadrolfs:
Open a saved doc
deselect all
Reset DS

Result: `Transfer from documents.beforeStartup to documents.initActiveDocument failed: TypeError: Cannot read property 'key' of undefined`

iwehrman:
FBNC to @chadrolfs.

chadrolfs:
yay

PR #3765, opened 2016-02-29
Bug fix: check existence of layer before scrolling into it

shaoshing:
This PR checks if the current document has selected layers before scrolling into it at initial display. This will fix for issue #3764

### DIFF
--- a/src/js/jsx/sections/layers/LayerGroup.jsx
+++ b/src/js/jsx/sections/layers/LayerGroup.jsx
@@ -157,8 +157,8 @@ define(function (require, exports, module) {
          */
         _scrollIntoFirstSelectedLayer: function () {
             var firstSelectedLayer = this.props.document.layers.selected.first(),
-                layerNode = ReactDOM.findDOMNode(this.refs[firstSelectedLayer.key]);
-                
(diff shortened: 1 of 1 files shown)


src/js/jsx/sections/layers/LayerGroup.jsx
```diff
@@ -157,7 +157,8 @@ define(function (require, exports, module) {
          */
         _scrollIntoFirstSelectedLayer: function () {
             var firstSelectedLayer = this.props.document.layers.selected.first(),
-                layerNode = ReactDOM.findDOMNode(this.refs[firstSelectedLayer.key]);
+                layerKey = firstSelectedLayer ? firstSelectedLayer.key : null,
+                layerNode = ReactDOM.findDOMNode(this.refs[layerKey]);
```
iwehrman: Hmm, I guess you tried this and it works, but it's kind of weird to do `ReactDOM.findDOMNode(this.refs[null])`. To make this a little less scary, can we just return early if `!layerKey`?

shaoshing:
@iwehrman I've made the null ref more clear now. Please review again.

iwehrman:
Better!

Issue #3266, opened 2015-11-18
Text selection flashes during scrubbing


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/0I0b1F2q1A3132150d3E/Screen%20Recording%202015-11-17%20at%2004.57%20PM.gif)

iwehrman:
preventDefault?

PR #3767, opened 2016-03-01
Avoid selecting text during scrub

iwehrman:
Prevent the default browser behavior during scrubby mousemove events to avoid selecting text within input elements. This seems to work for me, but please test! CC @chadrolfs 

Addresses #3266.


chadrolfs:
Works well on both Mac/Win.

PR #3771, opened 2016-03-01
Document load time: descriptor and lazy properties optimization

shaoshing:
**This PR is a work in progress**
Please let me know: 

* if there are any scenarios that you think will not work with the optimization?
* any suggestion/question about the implementation.

------------------------------------------------------------------------------------------

This PR introduces the following optimization to improve the document load time performance. 

##### 1. Load layer bounds lazily

Instead of loading bounds for all layers during startup, which can be slow for large documents, we will only load bounds that are necessary to render the TransformPanel, which are bounds of the selected layers and their children (for art board, child bounds are not required). The `best case` for this optimization is when the document has 0 selected layer, the `worse case` is when layers are all selected (which I think is less likely to happen and we can prevent such performance drawback). However, I guess the `most typical case` is that the document has less than 10 selected layers/groups during startup.

Time reduced: 300ms - 450ms

##### 2. Load layer bounds selectively

When fetching bounds at startup, we ask PS to return all bounds (`artboard, bounds, boundingBox, boundsNoMask, boundsNoEffects, pathBounds`), which causes PS to spend more time on gathering/generating/sending the result. This is also unnecessary in DS as each layer is only reading one type of bounds, depends on its layer type (https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/models/bounds.js#L153). This optimization allows DS to fetch the desired bounds for each  layer during startup. 

Time reduced: 60ms - 120ms

##### 3. faster background initialization

Properties that are initialized during startup should be skipped in the background initialization. These properties are

* essential properties (https://github.com/adobe-photoshop/spaces-design/blob/shao/lazy-bounds/src/js/actions/layers.js#L62), which are loaded for all layers
* bounds, which are loaded in optimization 1 and 2

Time reduced: the average time of initializing 50 layers is 28ms now, compare to 132ms before. For vermilion document, 
. This allows us, for example, increase the number of layers initialized each time (e.g. the average time for 150 layers is 87ms), and can result in a much faster background initialization.

##### 4. avoid empty batchPlay

We were executing some empty `batchPlay` descriptors when loading a document with zero selected layer. By avoiding these commands, we can save around 70ms for large documents like vermilion.

##### 5. Type initialization (pending)

We are initializing all text layers in `type.initFontList` (https://github.com/adobe-photoshop/spaces-design/blob/shao/lazy-bounds/src/js/actions/type.js#L760), which feels like a duplicate with the `layers.initializeLayersBackground`. This takes about 1600ms to complete for vermilion document. 

### Result

| | Before (ms) | After (ms) | Diff (ms) |
| --- | --- | --- | --- |
| new doc (deselect all) | **663** | 674 | +11 |
| new doc (select all) | **741** | 756 | +15 |
| vermilion (collapse/deselect all AB) | 1435 | **891** | -544 |
| vermilion (collapse all AB, select last AB) | 1695 | **1183** | -512 |
| vermilion (expand last AB, select "480px - Mobile") | 2461  | **1904** | -557 |
| vermilion (expand last AB, and select all layers of the AB) | 3146 | **2788** | -358 |
| large vermilion (3455 layers, collapse/deselect all AB) | 3702 | **1876** | -1826 |
| open file: vermilion | 5111 | **4026** | -1085 |
| background initilization (50 layers) | 132 | **28** | -104 |

### Todos
- [ ] replace `_getLayerByRef` with `_getLayerDescriptors`
- [ ] panels displayed before their content is ready
- [ ] `NaN` warning from text input
- [ ] More Testing (selected different layers, compare the inputs in the TransformPanel with master)
- [ ] Code Cleanup
- [ ] JSDoc

### DIFF
--- a/src/js/actions/layers.js
+++ b/src/js/actions/layers.js
@@ -80,15 +80,14 @@ define(function (require, exports) {
      * @private
      * @type {Array.<string>} 
      */
-    var _lazySelectedChildLayerProperties = [
+    var _lazyLayerBoundProperties = [
         "layerID", // redundant but useful for matching results
         "artboard",
         "bounds",
         "boundingBox",
         "boundsNoMask",
         "boundsNoEffects",
-        "pathBounds",
-        "mode"
+        "pathBounds"
     ];
     
     
@@ -99,7 +98,7 @@ define(function (require, exports) {
      * @type {Array.<string>} 
      */
     var _lazySelectedLayerProperties = [
-        "layerID",
+        "layerID", // redundant but useful for matching results
         "globalAngle",
         "proportionalScaling",
         "adjustment",
@@ -110,6 +109,7 @@ define(function (require, exports) {
         "keyOriginType",
         "layerEffects",
         "opacity",
+        "mode",
         "layerFXVisible" // the following are required but this is not enforced
     ];
 
@@ -121,7 +121,7 @@ define(function (require, exports) {
      * @private
      * @type {Array.<string>}
      */
-    var _lazyLayerProperties = _lazySelectedLayerProperties.concat(_lazySelectedChildLayerProperties.slice(1));
+    var _lazyLayerProperties = _lazySelectedLayerProperties.concat(_lazyLayerBoundProperties.slice(1));
 
     /**
      * Namespace for extension metadata.
@@ -165,9 +165,9 @@ define(function (require, exports) {
 
         return Promise.join(layerPropertiesPromise, optionalPropertiesPromise, extensionPromise,
             function (required, optional, extension) {
-                optional.forEach(function(properties) {
-                    properties.partialInitialized = true;
-                    properties.allInitialized = true;
+                optional.forEach(function (properties) {
+                    properties.boundsInitialized = true;
+                    properties.initialized = true;
                 });
                 
                 if (basic === false) {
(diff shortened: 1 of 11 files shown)


src/js/actions/layers.js
```diff
@@ -64,44 +66,41 @@ define(function (require, exports) {
...
-        "smartObject",
-        "layerSectionExpanded",
-        "vectorMaskEnabled",
-        "vectorMaskEmpty",
-        "textWarningLevel"
+        "bounds",
+        "boundingBox",
+        "boundsNoMask",
+        "boundsNoEffects",
+        "pathBounds"
```
shaoshing: The `_lazyLayerBoundProperties` is no longer used and the bounds name is determined by `Bounds.getRequiredBoundsName`.

src/js/actions/layers.js
```diff
@@ -134,6 +132,135 @@ define(function (require, exports) {
...
+                    ]
+                }
+                
+            },
+            "options": {
+                useMultiGet: true,
+                failOnMissingProperty: false
+            }
+        };
+    };
```
shaoshing: There should be a better way to do this instead of constructing the command manually.

src/js/actions/layers.js
```diff
@@ -134,6 +132,135 @@ define(function (require, exports) {
...
+        }
+        
+
+        return Promise.join(
+            essentialPropertiesPromise,
+            selectedPropertiesPromise,
+            extensionPromise,
+            function (essential, selected, extension) {
+                selected.forEach(function (properties) {
+                    properties.initialized = true;
```
shaoshing: Mark the `descriptor.initialized` so that the Layer model knows it is fully initialized.

src/js/actions/layers.js
```diff
@@ -134,6 +132,135 @@ define(function (require, exports) {
...
+                        }
+                    });
+                }
+
+                var boundsPromise = boundsDescriptorCommands.length === 0 ? Promise.resolve([]) :
+                        descriptor.batchPlay(boundsDescriptorCommands);
+
+                return boundsPromise.then(function (bounds) {
+                    bounds.forEach(function (properties) {
+                        properties.boundsInitialized = true;
```
shaoshing: Mark the `descriptor.boundsInitialized` so that we can know from the Layer model whether its bounds is ready or not.

src/js/models/bounds.js
```diff
@@ -183,21 +183,48 @@ define(function (require, exports, module) {
...
 
-        delete boundsObject._obj;
+        if (boundsObject) {
+            delete boundsObject._obj;
+        }
 
         return boundsObject;
     };
+    
+    Bounds.getRequiredBoundsName = function (descriptor) {
```
shaoshing: From this function, we can know which bounds property we should load from PS.

src/js/actions/layers.js
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
```
iwehrman: :space_invader:

src/js/actions/layers.js
```diff
@@ -134,6 +132,135 @@ define(function (require, exports) {
      */
     var METADATA_NAMESPACE = global.EXTENSION_DATA_NAMESPACE;
 
+    var multiGetRef = function (docRef, layerRef, properties) {
```
iwehrman: At least this and the next declaration should be underscore-prefixed and JSDoc'd with `@private`

src/js/models/layer.js
```diff
@@ -283,6 +283,8 @@ define(function (require, exports, module) {
          */
         initialized: false,
```
shaoshing: `Layer#initialized === true` indicates the Layer's following properties are initialized:

- essential properties
- bounds
- properties required as a selected layer
- extension data

src/js/actions/layers.js
```diff
@@ -134,6 +132,135 @@ define(function (require, exports) {
...
+                useMultiGet: true,
+                failOnMissingProperty: false
+            }
+        };
+    };
+    
+    var getLayerDescriptors = function (document, layers, options) {
+        layers = layers.toArray();
+        
+        if (layers.length === 0) {
```
iwehrman: Alternate idea: can we just change batchPlay in `descriptor` to return a resolved promise when the list of commands is empty? Then we'd get this benefit everywhere without adding extra complexity to -design.

src/js/models/layer.js
```diff
@@ -680,7 +683,9 @@ define(function (require, exports, module) {
                 vectorMaskEmpty: layerDescriptor.vectorMaskEmpty,
                 textWarningLevel: layerDescriptor.textWarningLevel,
                 isLinked: _extractIsLinked(layerDescriptor),
-                initialized: initialized || selected
+                boundsInitialized: layerDescriptor.boundsInitialized,
+                initialized: layerDescriptor.initialized,
+                descriptor: Immutable.Map(layerDescriptor)
```
shaoshing: Keep the original descriptor so that we can reset the layer from a partial layer descriptor. For example, only resetting the layer bounds.

src/js/actions/layers.js
shaoshing: Yes. I will address the coding style/jsdoc/comments in upcoming updates.

iwehrman:
Can you describe at a high level about how the `SuperselectOverlay` copes with the lack of bounds on startup? Previously, I avoided similar optimizations because it wasn't obvious what to do there.

src/js/actions/layers.js
shaoshing: Good idea, will add this optimization to the adapter instead.

src/js/models/layer.js
iwehrman: I suggest making this a private property, prefixed with an underscore. And I don't think there's any point in trying to convert it to be an immutable object? Also, just initializing an `Immutable.Map` with the object only shallowly converts the object, so this doesn't give you an immutable guarantee anyway.

Issue #3763, opened 2016-02-29
Weird hover styling for search input


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/0O0R0X2u3Z0c1x1x0W1l/Screen%20Recording%202016-02-29%20at%2011.08%20AM.gif?v=814ecfe1)

mcilroyc:
Fixed by PR.  still to @iwehrman ?

chadrolfs:
I'll take it

chadrolfs:
verified the fix

PR #3774, opened 2016-03-01
Fix search bar hover text size

iwehrman:
What it says! (If anybody knows a better way, feel free to chime in.)

Addresses #3763.


mcilroyc:
LGTM.  I mean, what's your hesitation?

iwehrman:
I guess it seems weird that I need to explicitly say in the `:hover` rule that the font size should be large. It seems like in a true, correct solution that would cascade from some more basic rule. But, ya know, CSS.

mcilroyc:
File an issue for follow-up with the label "summer intern"

Issue #3762, opened 2016-02-27
Descriptor event debug messages should be collapsed by default


iwehrman:
For readability and consistency with the descriptor get/batchPlay log messages, the descriptor event log messages should be collapsed by default as well: 

![image](https://cloud.githubusercontent.com/assets/1075154/13369957/45d7363c-dcb1-11e5-8c33-ee5c60ea8de6.png)

iwehrman:
Fixed by #3779.

PR #3779, opened 2016-03-02
Show params of descriptor event inside a group.

shaoshing:
Address issue #3762. Also increase JSON spacing of all descriptor log for better readability.  

screen shot 2016-03-02 at 10 52 29 am


iwehrman:
Looks good, thanks!

Issue #3780, opened 2016-03-02
grunt compile doesn't work with node 0.10.x


iwehrman:
It would be better if we failed fast in this case with a clear warning.

iwehrman:
Fixed by #3781.

PR #3781, opened 2016-03-02
Require at least node version 4.2.3 when running npm install.

iwehrman:
@jsbache found today that node 0.10.x is insufficient for `grunt compile`. This simple change should cause `npm install` to fail with a clear error message when when using a node version below 4.2.3. (I arbitrarily picked that version because it was LTS for a while, and there doesn't seem to be a good reason to force developers to upgrade to anything newer at this point.)

Addresses #3780. 

More info: http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,10 @@
     "Ian Wehrman <wehrman@adobe.com>"
   ],
   "description": "Adobe Photoshop Design Space",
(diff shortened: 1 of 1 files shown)


volfied:
:+1:

Issue #3746, opened 2016-02-23
Selected layer not shown in the layers panel when drilling down


ktaki:
Tested with Vermillion document. When drilling down layers by clicking on canvas, sometimes the layers panel does not scroll to shown the selected layer. 

Open Vermillion doc.
Click any part of the document on canvas to drilling down to a leaf layer.
Observe if the selected layer is shown in the layers panel.

RESULT: Sometimes the layers panel fails to show the layer selected.

iwehrman:
CC @shaoshing

iwehrman:
Back to @ktaki to verify.

ktaki:
Verified the fix.

PR #3783, opened 2016-03-03
Reveal layer when it becomes selected or is newly created.

shaoshing:
Fix for issue #3746, which layer selected from the canvas is not scroll into the visible area.


iwehrman:
Seems to work well for me. Let's merge it and let @ktaki test it in master.

Issue #3655, opened 2016-02-02
Reloading DS with multiple docs open results in incorrect bounds until all layers are initialized


chadrolfs:
Open/create multiple docs and select top artboard
reset DS
cmd + ` to cycle through open docs (notice bounds and adders)

Result: they are incorrect until you cycle through the docs a 2nd time (once all layers are initialized)

chadrolfs:
![reset-boundsoff](https://cloud.githubusercontent.com/assets/8442631/12766798/4d2b6cb4-c9ba-11e5-8257-73bf43fd50af.gif)

shaoshing:
Removing the performance tag as this is caused by action execution order, rather than performance.

iwehrman:
FBNC to @chadrolfs.

PR #3784, opened 2016-03-03
Update transform after the selected document is initialized.

shaoshing:
Fix for issue #3655. Before, when selecting an uninitialized document, `ui.updateTransform` was executed before the selected document was set as the current document, and causes the action to read transform of the previously selected document wrongly. This PR changed the execution order to make sure `ui.updateTransform` is executed before `documents.updateDocument`.


iwehrman:
Yep, makes sense. This was probably an over-eager optimization.

Issue #3667, opened 2016-02-03
Font change should be undo-able by just one Cmd/Ctrl-z


ktaki:
Font faces previewed, i.e., not committed, should not be added to the history.

ktaki:
@iwehrman This works, but mouse hovering over the font data list, each font produces a history re-initialization warning such as:

[History] Re-initializing history because photoshop thinks the current state is 3 of 4, while our model says 4 of 5
[History] Initializing history with 4 states

Can this be simply ignored? Just checking.

iwehrman:
Let's ask @mcilroyc.

mcilroyc:
That's undesirable.  It means we are nuking history, which will prevent fast undo/redo.

iwehrman:
@mcilroyc Is there a standard way to address issues like this? Or is this a "new adventure every time" thing?

mcilroyc:
At a quick glance, it seems that you're not providing the `coalesce` flag to the store event... which results in different history treatment between DS and PS ... which gets things out of sync.

https://github.com/adobe-photoshop/spaces-design/pull/3787/files#diff-66fb006bb46f8f716c009f317aded016R158

and then line 121

iwehrman:
Derp, on it, thanks.

mcilroyc:
with the merge of #3796, the "Re-initializing history" issue from @ktaki's most recent report seems to be fixed

ktaki:
Verified the fix. No more history warnings.

PR #3787, opened 2016-03-04
Coalesce type face and weight changes

iwehrman:
This makes the type face and weight `Datalist` instances coalescable, by mixing the `Coalesce` mixin into the `Type` component. This means that when you open one of those datalists and browse through the entries, you'll only create one history state and don't have to undo a thousand times to get back to your original font.

I'm embarrassed that we shipped with this last time. It turned out to be a lot easier to fix than I thought... 

Addresses #3667.


pineapplespatula:
LGTM

Issue #3710, opened 2016-02-16
Export non-artboard doc does nothing


placegraphichere:
Using the Export panel on a non-artboard document (JPG for example) will allow the Export button to be clicked, a file path chosen, but nothing happens.
- Open JPG
- File -> Export…
- Export
- Choose location
- Nothing is saved

- Expected: the file should be saved as-is (1x PNG default export). I think we should just allow this workflow. Similar to #2309 allowing the addition of an asset by clicking "+".

iwehrman:
Is this a known limitation, @mcilroyc? Or is this a bug? I think I vaguely remember discussing this with you, but it's hazy...

mcilroyc:
well this is two-fer.  1) It is a known limitation but surely solvable.  2) In general, the Export modal window should _not_ allow clicking the export button when there are no configured+selected assets.  

This is assigned to the right person.  I foresee a batch of export work in the near future.

ktaki:
@mcilroyc: Have you changed anything in this area? I can open a jpg file, add an asset, and export it as png with current master. Am I missing something here?

chadrolfs:
Following the steps above (no asset in the export panel in the UI under Effects), opening the Export dialog from File menu or title bar icon, will have no assets in the dialog and you CAN click the Export button at the bottom right of the dialog and select a folder to export but nothing gets exported.  According to Cory's comment above, you shouldn't be able to click the Export button.

placegraphichere:
There is an aspect of "it should just work" to these issues. I'd expect to be able to export at any time. so perhaps we should think about it that we should actually not disable the export modal button, but allow them to get something exported instead.

Artboards auto populate with 1x in the export panel, so perhaps the same logic should be applied to non-artboard docs.

mcilroyc:
@placegraphichere I agree it should just work.  I'm a little worried that the current UI would still be confusing though.  Should a "blank" export dialog (with no configured exports) always just magically export a 1x document-level image?  
![ss](http://cl.ly/0c3Q3B2j2T2o/Image%202016-02-26%20at%201.14.40%20PM.png)

We currently have no real treatment of document-level exports in the modal window.  Maybe Adding something (even a simple "export document" checkbox) to make it clear that *something will happen* when clicking `Export`.  Just spit ballin

iwehrman:
FBNC to @ktaki.

ktaki:
[...]

PR #3790, opened 2016-03-07
Disable the export button on the export modal panel if nothing to export

mcilroyc:
This is a narrow fix for the problem reported in #3710.  I made this fix awhile ago but sat on it while continuing to research the "make it just work" proposal/discussion from that issue.  But, I think it makes sense to just squash this bug, and backlog the fancy proposal.

If there is nothing to export, disable the button (in the modal export dialog)


iwehrman:
LGTM.

Issue #3745, opened 2016-02-23
GIFs need to be redone  for lighter color stops in the DS Introduction dialog


DivyaPrabhakar:
Example: some of the lines in the GIFs are very light and not crisp

![](https://www.dropbox.com/s/6akyiif2jp8rsk0/Screenshot%202016-02-23%2013.47.15.png?raw=1)
![](https://www.dropbox.com/s/8pnzmczhpiy73pm/Screenshot%202016-02-23%2013.47.31.png?raw=1)
![](https://www.dropbox.com/s/6gf74bwnphu923g/Screenshot%202016-02-23%2013.47.35.png?raw=1)

iwehrman:
Alternatively, we could force the intro dialog to always be medium gray.

@placegraphichere: how painful would it be to recreate these gifs?

iwehrman:
Over to @ktaki to verify.

ktaki:
@DivyaPrabhakar: With the change, Keyboard Shortcuts had a negative effect. Could this be an easy fix? Affected "Original" and "Light"

<img width="776" alt="screen shot 2016-03-08 at 4 34 27 pm" src="https://cloud.githubusercontent.com/assets/2029597/13621348/d76269a8-e54b-11e5-817c-de00e432ec6f.png">

DivyaPrabhakar:
@ktaki Definitely an easy fix, will do that asap. Thanks for catching this.

iwehrman:
Back to @ktaki to verify.

ktaki:
Verified the fix.

PR #3792, opened 2016-03-08
Making introduction screen default into medium color stop values

DivyaPrabhakar:
References issue #3745 
This forces the intro screen to always take the colors of the medium color stop.

### DIFF
--- a/src/style/help/first-launch.less
+++ b/src/style/help/first-launch.less
@@ -22,10 +22,25 @@
  */
 
 @first-launch-width: 86rem;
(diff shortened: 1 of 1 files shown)


DivyaPrabhakar:
FYI: This is a problem I encountered while testing this branch but it is present on `master` too. Should we look at this or is there no point?

iwehrman:
We can ignore that warning for now.

iwehrman:
This looks fine to me. It's a hack we can live with.

Issue #3750, opened 2016-02-24
Layers panel does not update when two images are placed


DivyaPrabhakar:
When two image layers are dragged into DS and placed on an artboard, the layers panel does not reflect this until the user clicks on the layers panel. Both images are present, but the layers panel does not accurately reflect the state of the document. Note: I am dragging 2 images from my cloud linked Dropbox folder, this may play a part in the issue? 

![](https://www.dropbox.com/s/exsgyvdwp9ynegt/twolayers.gif?raw=1)

iwehrman:
Could be related to #3635?

iwehrman:
Can you verify that this is distinct from #3635 @DivyaPrabhakar?

ktaki:
It probably different issue from #3635. The following error was reported in console when a layer in layers panel was clicked. Recommend to fix this along with #3635 for June since drag-and-drop of image files is a popular workflow.

![screen shot 2016-03-07 at 2 00 07 pm](https://cloud.githubusercontent.com/assets/2029597/13584879/54062696-e46d-11e5-80d8-0c46a25267f5.png)

iwehrman:
FBNC to @ktaki.

ktaki:
@shaoshing:  Placing multiple images succeeds. When one of the images are moved on canvas, however, you get a History warning.

[History] Re-initializing history because photoshop thinks the current state is 4 of 5, while our model says 5 of 6
[History] Initializing history with 5 states

This does not happen when single image file was placed. 

Steps:
Open new doc.
Drag two files on to the canvas.
Select the top image and drag.

shaoshing:
@ktaki The warning is because when placing multiple files, we will receive multiple `historyState` events before receiving the `placeEvent` event, which result in incorrect count of history state. For example, when drag-and-drop two files, the order be like (number is the count of history state):

```
historyState (1)
historyState (2)
placeEvent <- expecting two placeEvent but only receive one
addLayer (2)
addLayer (3)
```
While the expected order should be:

```
historyState (1)
placeEvent
addLayer (1)
historyState (2)
placeEvent
addLayer (2)
```

To fix this complete, we will need the core team's help in fixing the event orders, but the good news is the warning can be safely ignored as we will reset the wrong state count to the correct one, and undo/redo will all work as expected. The downside is this will invalidate our history state cache which has some impact on performance when undoing. 

[...]

PR #3793, opened 2016-03-08
Bug fix for dropping multiple files onto canvas

shaoshing:
Fix for issue #3750 

We are receiving one `placeEvent` with information of the last new layer when dropping multiple files , which causes the layers panel to add only one new layer. This PR fix the issue by fetching all layer IDs of the current document and then compare with the previous document model to detect all new layer IDs.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -48,7 +48,6 @@ define(function (require, exports) {
         pathUtil = require("js/util/path"),
         log = require("js/util/log"),
         headlights = require("js/util/headlights"),
-        objUtil = require("js/util/object"),
         global = require("js/util/global");
 
     var templates = require("static/templates.json");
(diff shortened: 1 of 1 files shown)


src/js/actions/documents.js
```diff
@@ -935,35 +934,70 @@ define(function (require, exports) {
...
+                // If the new layer is already existed (by expanding a libraries graphic that is an embedded SO),
+                // we reset the layer to update its smart object type and position.
+                if (layer) {
+                    return this.transfer(historyActions.newHistoryStateRogueSafe, document.id)
+                        .bind(this)
+                        .then(function () {
+                            return this.transfer(layerActions.resetLayers, document, layer);
+                        })
+                        .then(function () {
+                            return this.transfer(layerActions.resetIndex, document);
```
iwehrman: I don't think we should ever need to call this more than once, but it seems from the control structure that it could actually be called N times for N placed layers, resulting in N^2 layers eventually reset. Should this be factored out and only called once? Or do we know for some reason that this condition can only hold for, say, one layer?

shaoshing:
Hi @iwehrman I've addressed your comment by make it more clear when resetting layers for an existed layer, which happens when user expands a LLSO that contains one layer and the layer is an embedded SO (more details about this scenario https://github.com/adobe-photoshop/spaces-design/issues/3145#issuecomment-151627443). Back to you for another review.

iwehrman:
Looks better, and the fix works for me!

Issue #3635, opened 2016-01-28
Drag and drop a photo from browser causes an error


ktaki:
Action layers.addLayers failed: base element not found

Open Adobe Stock site in a browser. Drag and drop a photo onto a canvas.

![screen shot 2016-01-28 at 2 03 32 pm](https://cloud.githubusercontent.com/assets/2029597/12660336/167bdac8-c5c8-11e5-945c-6e57044bcf66.png)

iwehrman:
Probably happens for all images from browsers?

ktaki:
Yes, that is the case. Dragging any image from browsers on to canvas does.

iwehrman:
Hmm, I can't repro on Mac pgdev.1054 at 86ccaf629190b7baccbdaf1d663a22f9c55500a9. Did you observe this on Mac or Win?

ktaki:
I tried with the Mac pgdev 1060 at 86ccaf6. It is and has been easy to repro for me. @chadrolfs Can you try it? Just drag any image from a browser to a canvas.

chadrolfs:
yep, piece of :cake:, I can repro very easily

iwehrman:
I still can't repro, I guess it's a timing issue. Maybe some other engineer can try to repro? @DivyaPrabhakar?

iwehrman:
@volfied can you repro this?

volfied:
I can repro, given that `base element is not found` error is coming from PS, I'll take a look.

volfied:
Trick is to open a new document.

volfied:
I believe the problem is dragging an image to canvas acts as a paste operation, and we get the "drag" event before Photoshop had the time to update it's own document. 

https://jira.corp.adobe.com/browse/PS-1370

iwehrman:
Kicking this out of the milestone because we're blocked, but adding the Ship Blocker label.

iwehrman:
Over to @ktaki to verify.

ktaki:
No, it did not fix the issue. Do we want to double the value? It seems to me, this will not result in any bad state afterwards. What would happen if we leave it as is?

iwehrman:
Back to @shaoshing. Maybe the timeout wasn't long enough?

iwehrman:
In theory the UI should reset itself after the unhandled error. I prefer not to rely on that though...

shaoshing:
[...]

PR #3794, opened 2016-03-08
Delay adding new layer for 'drag' event.

shaoshing:
Fix for issue #3635 by applying the similar workaround that we have for the `make` handler (https://github.com/adobe-photoshop/spaces-design/blob/master/src/js/actions/layers.js#L1905)


iwehrman:
img

Issue #3740, opened 2016-02-22
Arrow keys focus on canvas rather than search dialog in the vector mask mode


ktaki:
Windows only.
Open an image file and hit M.
In the vector mask mode, add a mask. 
While is the mask is selected, open the search bar.
Type "vector".
Use down arrow key to select "Super Select Vector".

RESULT: The mask on canvas moves downward.

EXPECTED: Traverses the items in the search dialog.

iwehrman:
Just end mask mode when opening search.

iwehrman:
FBNC to @ktaki.

ktaki:
Verified the fix.

PR #3795, opened 2016-03-09
End mask mode before opening search dialog

iwehrman:
This ends mask mode before opening the search dialog in order to prevent key event policy problems, as described in #3740. (It's unclear why that is happening only on Windows.)


pineapplespatula:
looks good

PR #3796, opened 2016-03-09
Preserve history state when coalescing type face/weight changes

iwehrman:
When coalescing type `postScript` changes, we aren't currently informing the history store. I believe this fixes the problem, as per @mcilroyc's suggestion. 

Addresses #3667 again.


mcilroyc:
that'll do

PR #3797, opened 2016-03-09
Bug fix: error when dragging image from browser onto empty artboard.

shaoshing:
This is my second attempt on fixing issue #3635. It turns out that the error is caused by dropping image onto an empty artboard, which triggers layer replacement. Although the new layer is replaced by an existing layer, PS still emitting message with the incorrect ID of the replaced layer, and that's why the `layers.addLayers` failed. This PR checks if the id from `event.layerID` exists before adding a new layer.

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -1134,10 +1134,21 @@ define(function (require, exports) {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


iwehrman:
Looks good!

PR #3798, opened 2016-03-09
Fixed the colors of the keyboard shortut dialog

DivyaPrabhakar:
References #3745
The keyboard shortcut dialog is fixed for all color stops


iwehrman:
LGTM!

Issue #3591, opened 2016-01-21
Setting the position of large artboards is extremely slow


iwehrman:
![img](https://s3.amazonaws.com/f.cl.ly/items/0j3e3g0O0U2t1a0G2P1m/Screen%20Recording%202016-01-21%20at%2012.23%20PM.gif?v=2780b625)

The actual dialog persists for a very, very long time. By the time Photoshop gathers itself **minutes later**, our bounds are incorrect:

![image](https://cloud.githubusercontent.com/assets/1075154/12493499/17cd61e6-c03a-11e5-8fc2-b16500155665.png)

iwehrman:
When modifying the position with the X/Y text inputs in the transform panel it's also slow, but nowhere near as slow as when scrubbing.

iwehrman:
Here's the log, including a log of played commands:
https://gist.github.com/iwehrman/8abd13f451423c2ec1bd

The only Design Space bug I see is that we're wrapping each transform command with unnecessary select commands, but even if that's causing the problem I still blame Photoshop.

iwehrman:
Here's the log for the comparatively fast case in which we modify the position with the text input:
https://gist.github.com/iwehrman/cdfc84e31c1d09304608

It seems like we're using `transform` when scrubbing and `move` when using the text input.

ktaki:
Even a blank artboard document will result in the picture below.
Open a new default doc. Select the artboad. Scrub W value to the left and then right.

![screen shot 2016-01-25 at 3 50 43 pm](https://cloud.githubusercontent.com/assets/2029597/12568574/cced4624-c37b-11e5-9f59-bf9c1d52f555.png)

Click on the artboard to force redraw will fix the issue.

iwehrman:
This also affects swap. It takes minutes to swap artboards on Vermilion.

shaoshing:
Since we have decided to not ship the scrubby feature in this release, I will make change to remove the related code.

iwehrman:
:white_check_mark:

chadrolfs:
Please feel free to assign this bug to me once that PR is merged.  There are a handful of other scrubby bugs I'll close as well.  :+1:

iwehrman:
Scrubbies are disabled, back to @chadrolfs.

PR #3799, opened 2016-03-09
Disable the scrubby feature

shaoshing:
Temporarily disable the scrubby feature to address issue #3591

(I am keeping the scrub handles in the components (`Radius`, `VectorFill`, `Position`, `Rotate`, and `Size`) for future reference.)


iwehrman:
Easy enough!

PR #3806, opened 2016-03-30
Fix menu item state for guides and smart guides.

shaoshing:
This PR updates the visibility state of both Guides and Smart Guides menu items, when user turn on one of them (before, we only update the visibility state of the clicked item). 

The reason for this change is because, in standard PS, when `View > Extra` is off and `View > Show > Guides/Smart Guides` are on, the descriptors that get the visibility state of guids/smart guides will return false. When we run the descriptor to turn on one of them, PS will instead turn on the `View > Extra` menu item, which makes both Guides and Smart Guides on now. Without this fix, user will see both guides and smart guides on the canvas while the menu items are not both checked.

*More discussion is on JIRA PS-1212*

### DIFF
--- a/src/js/actions/documents.js
+++ b/src/js/actions/documents.js
@@ -852,29 +852,14 @@ define(function (require, exports) {
         }
 
         var newVisibility = !document.guidesVisible,
-            guideInitPromise;
-
-        if (newVisibility && !document.guides) {
-            guideInitPromise = _getGuidesForDocument(document.id);
-        } else {
-            guideInitPromise = Promise.resolve();
-        }
-
-        var dispatchPromise = guideInitPromise
-            .bind(this)
-            .then(function (guides) {
-                this.dispatch(events.document.GUIDES_VISIBILITY_CHANGED, {
-                    documentID: document.id,
-                    guidesVisible: newVisibility,
-                    guides: guides
-                });
-            });
-
-        var playObject = documentLib.setGuidesVisibility(newVisibility),
+            playObject = documentLib.setGuidesVisibility(newVisibility),
             playPromise = descriptor.playObject(playObject);
 
-        return Promise.join(dispatchPromise, playPromise)
+        return playPromise
             .bind(this)
+            .then(function () {
+                return _dispatchGuidesVisibilityEvent.call(this, document);
+            })
             .then(function () {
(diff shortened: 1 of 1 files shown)


src/js/actions/documents.js
```diff
@@ -898,18 +883,54 @@ define(function (require, exports) {
...
+    
+    /**
+     * A helper function to get the visibility state of guides and smart guides from PS and dispatch
+     * GUIDES_VISIBILITY_CHANGED event to update the related menu items.
+     * 
+     * @param  {Document} document
+     * @return {Promise}
+     */
+    var _dispatchGuidesVisibilityEvent = function (document) {
+        var guidesPromise = descriptor.playObject(documentLib.getGuidesVisibility()),
```
volfied: It would be better if you used batchPlayObjects here so we don't make separate calls into PS.

volfied: You should group these in a `batchPlayObjects` call.

src/js/actions/documents.js
```diff
@@ -898,18 +883,54 @@ define(function (require, exports) {
...
+            .then(function (results) {
+                var guidesVisible = results[0].guidesVisibility,
+                    smartGuidesVisible = results[1].smartGuidesVisibility,
+                    guideInitPromise = Promise.resolve();
+
+                if (guidesVisible && !document.guides) {
+                    guideInitPromise = _getGuidesForDocument(document.id);
+                }
+
+                return guideInitPromise
```
volfied: You can avoid the nesting here by returning `guideInitPromise` in this leg, and adding the `then` in the main promise stack.

volfied:
@shaoshing looks good, but I got few comments that may help with performance.

src/js/actions/documents.js
shaoshing: The nested function is using the `guidesVisible` `smartGuidesVisible` from the outer scope, so moving it out from the scope might not work, unless I do something like `return Promise.join(guideInitPromise, guideVisible, smartGuideVisible);` to pass the variables. What do you think?

shaoshing:
Thanks for reviewing @volfied. I've replaced the two PS calls with `batchPlay`, also replied your comment on moving the nested function. Back to you.

PR #3810, opened 2016-04-19
Add a Mac-only ad for Adobe Xd in the first-launch dialog

iwehrman:
This adds a Mac-only advertisement paragraph to the last screen in the first-launch dialog for Adobe Experience Design CC. On Windows it will still look like this:
image

But on Mac it will now look like this:
image

### DIFF
--- a/src/nls/en/strings.json
+++ b/src/nls/en/strings.json
@@ -94,7 +94,10 @@
                 "HEADLINE": "Design Space",
                 "BODY_FIRST": "Twitter",
                 "BODY_SECOND": "GitHub",
-                "BODY_THIRD": "Forum"
+                "BODY_THIRD": "Forum",
+                "BODY_ADVERTISEMENT_1": "You seem like the type who likes to check out betas. We've been hard at work building a new end-to-end solution for UX designers, including great integration with Photoshop. We're happy to share that {ADOBE_XD} is now in public beta.",
+                "BODY_ADVERTISEMENT_2": "Go download a build and try it out!",
+                "ADOBE_XD": "Adobe Experience Design CC"
             }
         ]
     },
--- a/src/js/jsx/help/FirstLaunch.jsx
+++ b/src/js/jsx/help/FirstLaunch.jsx
@@ -31,6 +31,7 @@ define(function (require, exports, module) {
 
     var Carousel = require("js/jsx/shared/Carousel"),
         nls = require("js/util/nls"),
+        system = require("js/util/system"),
         SVGIcon = require("js/jsx/shared/SVGIcon");
  
     var FirstLaunch = React.createClass({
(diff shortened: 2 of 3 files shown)


iwehrman:
Note that the goURL is not yet live. (This is the workflow that Kelli requested.)

iwehrman:
Also note that these strings will need to be localized (cc @fzhu1978) , and I expect that we'll need to do some futzing with the line breaks (cc @DivyaPrabhakar) once the translations are in.

src/nls/en/strings.json
```diff
@@ -94,7 +94,11 @@
                 "HEADLINE": "Design Space",
                 "BODY_FIRST": "Twitter",
                 "BODY_SECOND": "GitHub",
-                "BODY_THIRD": "Forum"
+                "BODY_THIRD": "Forum",
+                "BODY_ADVERTISEMENT_1": "You seem like the type who likes to check out betas. We've been hard at work building a new end-to-end solution for UX designers, including great integration with Photoshop. We're happy to share that",
+                "BODY_ADVERTISEMENT_2": "Adobe Experience Design CC",
+                "BODY_ADVERTISEMENT_3": "is now in public beta.",
```
shaoshing: I am concerned that breaking the text into multiple chunk might make translation difficult, what do you think?

shaoshing:
@iwehrman this looks great. I just have one question regarding the localization strings.

src/nls/en/strings.json
shaoshing: How about something like ```... happy to share that {Adobe_XD_CC} is now in public beta.```

iwehrman: Those fragments are pulled out because they have HTML tags wrapped around them. I'm not actually sure how to have a single localized string in React that includes embedded tags. If you have a solution in mind, can you break it down for me a little more?

shaoshing: Sorry for not being clear. React can render an array of mixed objects. I think we can make use of that by passing in an arrary that contains pure string and wrapped string:

```jsx
var description = nls.localize("strings.FIRST_LAUNCH.SLIDES.7.BODY_ADVERTISEMENT").split("{ADOBE_XD_CC}"),
    name = (<em>nls.localize("strings.FIRST_LAUNCH.SLIDES.7.ADOBE_XD_CC")</em>);

description.splice(1, 0, name); // insert the name

return (
  <div>
    {description}
  </div>
);
```

fzhu1978: The new string will be picked up soon, the lockit will be sent out tonight.  Please expect translation to be back in 1-2 days.

fzhu1978: Oh, after it is committed to master branch.

iwehrman: OK, I understand your suggest now, @shaoshing. I'll take another crack!

iwehrman:
OK, spruced up. Back to @shaoshing for re-review.

shaoshing:
Thanks, this looks excellent!

Issue #3804, opened 2016-03-23
Artboard swapping is broken


chadrolfs:
![ab-swap](https://cloud.githubusercontent.com/assets/8442631/14002170/63eb8898-f107-11e5-8b18-05ad46eda641.gif)

chadrolfs:
![ds-swaptheplank](https://cloud.githubusercontent.com/assets/8442631/14691213/dce29e86-0704-11e6-8e27-9d76bfe9643b.gif)

mcilroyc:
FBNC 2 @chadrolfs

chadrolfs:
This is now on par with 16.1.2.  Thanks @mcilroyc

PR #3812, opened 2016-04-21
Translate instead of setting absolute position when swapping layers

mcilroyc:
This is a first-down hail-mary pass trying to address #3804.

I confirmed that photoshop is acting strangely when asked to set an artboard's absolute position by X/Y coords.  But, while troubleshooting, I tried using the alternate method of moving layers that is used when entering X or Y coords in the Position inputs.  This alternate flavor of `_getMoveLayerActions` uses an underlying `move` command instead of `transform`.

It seems to work fine.  But I half-expect someone to remember a situation that I haven't considered which breaks it.

Note: If this alternate approach *is* sufficient, then it seems like one of the code paths in `_getMoveLayerActions` may effectively be dead now.

Bonus Bug: I also found in the existing version that the width of an artboard was sometimes slowly growing after repeated swapping.

### DIFF
--- a/src/js/actions/transform.js
+++ b/src/js/actions/transform.js
@@ -497,9 +497,9 @@ define(function (require, exports) {
                 }
             },
             layerOneActions = _getMoveLayerActions
-                .call(this, document, layers.get(0), newPositions.get(0), "lt", payload.positions),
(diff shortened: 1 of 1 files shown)


volfied:
I think we added the alternate flavor for scrubbing, but seems like it also applies to these conditions.

I'm good with this change. :+1:

Issue #3807, opened 2016-04-05
Text editing commits when the text color is changed


ktaki:
Create a text layer. While in the text editing mode, select a part of a string. Open the color picker and change the color.

RESULT: Text editing ends and the entire text color changes

EXPECTED: Text editing continues and the color of the selected text changes

mcilroyc:
@ktaki back to you. see also the note in the PR about opacity

ktaki:
Verified the fix.

PR #3813, opened 2016-04-21
When in modal text edit state, allow color change and disable opacity input

mcilroyc:
Addresses #3807 by setting the `layer.changeColors` action to be modal safe.

This PR also disables the Opacity element when in modal text edit state because attempting to set opacity while modal will throw an error.


volfied:
Looks good to me!

Issue #3108, opened 2015-10-23
Re-Sampling after undo of LLSO results in failure


chadrolfs:
drag out 2 assets from library onto canvas
select Sampler, mouse over the one not selected and hit Space, click on graphic icon
cmd/ctrl + z to undo
repeat Sample

Result: it doesn't replace and target LLSO changes to an embedded SO
`Action sampler.replaceGraphic failed: Error: Photoshop returned an error when playing command number: 2 Photoshop code: -25920 Photoshop message: error: -25920 message: The command “Embed Linked Smart Object” is not currently available.`

To resolve the issue completely, we need a fix from the core side:

Watson bug: https://watsonexp.corp.adobe.com/#bug=4077984
Video: https://www.dropbox.com/s/w7456wh35dlbf4p/sampling-graphic.mov?dl=0

chadrolfs:
![sampler-so-undo](https://cloud.githubusercontent.com/assets/8442631/10704447/b3a46bdc-798b-11e5-9368-a0c428439ea5.gif)

shaoshing:
@chadrolfs this is a known issue caused by #3084 (core issue and deferred). The problem here is, when you sample a LLSO, it has two history state:

1. convert LLSO to embeded layer
2. link the layer to the library graphic (which turn it into LLSO)

When you undo, you are undoing step 2, and the layer is in a wrong state. A second undo restore the layer completely.

shaoshing:
When undo to state 1, the layer has the wrong smart object type (was LLSO, but should be ESO) so when you try to sample the graphic again, smapler tries to convert the layer to ESO again and failed. 

@mcilroyc do you any suggestion regarding the history state?

mcilroyc:
@shaoshing at what point does DS's model get updated with the correct smart object type?  If there is a timing issue where we are handling events/plays/history in the wrong order, then I could help sort it out.  If we're not currently getting this info, then we may be hosed.  Either way, I would *personally* call this edge case not a high priority, and a fix will probably be risky.

joelrbrandt:
Workaround: undo twice!

ktaki:
@chadrolfs The core bug has been closed as Closed/NeveFix/NotWorthEffort. If you still feel this is a ship blocker, please open a bug in JIRA.

chadrolfs:
persists with master @ dd417fe and current jenkins builds.  You end up in an unrecoverable state.  I think the possibility of losing work makes this worth addressing.

[...]

PR #3815, opened 2016-04-26
Fix unrecoverable state after failure in sampling LLSO

shaoshing:
DS will fail to recover from action failure if the currently selected tool is sampler. This is because the `SamplerOverlay` is unable to read the selected layers in its `shouldComponentUpdate` function, as the current document is unavailable during resetting. This PR fixes the issue by checking the existence of the current document. I also checked the other tools to make sure doing the same checking in their SCU function. 

Fix: https://jira.corp.adobe.com/browse/PS-2645
Mitigate: https://github.com/adobe-photoshop/spaces-design/issues/3108 (it should not be a ship blocker now)

### DIFF
--- a/src/js/jsx/tools/SamplerOverlay.jsx
+++ b/src/js/jsx/tools/SamplerOverlay.jsx
@@ -110,10 +110,18 @@ define(function (require, exports, module) {
         },
 
         shouldComponentUpdate: function (nextProps, nextState) {
-            var lastLayers = this.state.document.layers.selected,
-                nextLayers = nextState.document.layers.selected,
-                lastLayerProps = collection.pluck(lastLayers, "id"),
-                nextLayerProps = collection.pluck(nextLayers, "id"),
+            var getProps = function (state) {
+                var document = state.document;
+
+                if (!document || !document.layers) {
+                    return null;
+                }
+
+                return collection.pluck(document.layers.selected, "id");
+            };
+            
+            var lastLayerProps = getProps(this.state),
+                nextLayerProps = getProps(nextState),
                 layersChanged = !Immutable.is(lastLayerProps, nextLayerProps);
 
             return layersChanged ||
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -268,11 +268,9 @@ define(function (require, exports) {
         this.flux.store("menu").on("change", _menuChangeHandler);
         
         if (!__PG_DEBUG__) {
-            var debugMenuIndex = rawMenuObj.menu.findIndex(function (menu) {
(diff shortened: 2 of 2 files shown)


src/js/jsx/tools/SamplerOverlay.jsx
```diff
@@ -110,10 +110,18 @@ define(function (require, exports, module) {
...
-                lastLayerProps = collection.pluck(lastLayers, "id"),
-                nextLayerProps = collection.pluck(nextLayers, "id"),
+            var getProps = function (state) {
+                var document = state.document;
+
+                if (!document || !document.layers) {
+                    return null;
+                }
+
+                return collection.pluckAll(document.layers.selected, ["id"]);
```
iwehrman: Nit: it would be simpler to just use `pluck`:
```javascript
return collection.pluck(document.layers.selected, "id");
```

iwehrman:
Besides the nit, this looks correct to me. @chadrolfs or @ktaki: can you test this fix before we merge it?

chadrolfs:
Sure, I'll take a look now.

src/js/jsx/tools/SamplerOverlay.jsx
shaoshing: :+1 Done the clean up.

chadrolfs:
Still was able to run into the bug and get into the unresponsive state.

chadrolfs:
The first time I sampled, undo, sampled again, it recovered, however if you do it a few more times, you'll get into that state.  Here's the tail end of the errors (using compiled js)

screen shot 2016-04-26 at 11 40 34 am

shaoshing:
Hmm I am unable to reproduce `that` state after re-sampling more than 10 times. @chadrolfs can you show me the detailed steps? Like which assets did you use for testing.

chadrolfs:
Something is off.  I am connected via devtools to this branch compiled build and unable to sample now because it's falling through to standard PS.  When I try to hit Space to sample, it switched to Hand (standard PS behavior).

chadrolfs:
I should mention, I'm also using mainline m.41 build

shaoshing:
@chadrolfs thanks for the help, the problem is related to the debug menu. In production mode, we remove the DEBUG menu during initialization to hide it from the user. However, it will keep removing the last menu every time when DS is recovered, which eventually lead to an empty menu list that causes the unrecoverable state. I've pushed some change to address the issue. Please test again :D

chadrolfs:
Ok, looks like the unrecoverable state is avoided now.  The only other issue which we can defer and not worry about at this point is the bounds do not get reset correctly on undo, but once you change tools (to Select for example) it updates correctly.

ds-pinkbounds

PR #3817, opened 2016-04-28
First-launch dialog localization fixes

iwehrman:
1. Fixes a problem with how webpack was merging dictionaries at compile time that prevented non-English translations from showing in the first-launch dialog.
2. Fix some layout issues in the last slide in French and German.
3. Remove a dangling space in the `{ADOBE_XD}` separator in the Japanese translation that caused `undefined` to trickle into the last slide.

Addresses #3819 and partially addresses #3820.

### DIFF
--- a/src/index-debug.html
+++ b/src/index-debug.html
@@ -32,6 +32,16 @@
           locale = localeParts[0],
           js = document.createElement("script");
 
+      switch (locale) {
+      case "en":
+      case "de":
+      case "fr":
+      case "ja":
+          break;
+      default:
+          locale = "en";
+      }
+
       js.src = "./spaces-design-" + locale + ".js";
       js.type = "text/javascript";
 
--- a/src/index-debug.html
+++ b/src/index-debug.html
@@ -28,7 +28,8 @@
     <link rel="stylesheet" type="text/css" href="style.css" />
     <script type ="text/javascript" src="./externalDeps.js"></script>
     <script type="text/javascript">
-      var locale = window.navigator.language,
+      var localeParts = window.navigator.language.split("-"),
+          locale = localeParts[0],
           js = document.createElement("script");
 
       js.src = "./spaces-design-" + locale + ".js";
--- a/src/index.html
+++ b/src/index.html
@@ -31,6 +31,16 @@
           locale = localeParts[0],
           js = document.createElement("script");
 
+      switch (locale) {
+      case "en":
+      case "de":
+      case "fr":
+      case "ja":
+          break;
+      default:
+          locale = "en";
+      }
+
       js.src = "./spaces-design-" + locale + ".js";
       js.type = "text/javascript";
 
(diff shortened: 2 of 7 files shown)


iwehrman:
Ideally someone would test this in all for supported languages.

pineapplespatula:
The code looks good to me. @ktaki or @chadrolfs can you run this in each language especially for the spacing changes added?

ktaki:
Will try it after lunch.

ktaki:
@iwehrman: Now I see the translation. It, however, produces the following error when locale was set to en_GB

GET file:///Users/ktaki/PSbuilds/playground-dev/1367/photoshop-Release_x86_64/A…015.5.app/Contents/Required/Plug-Ins/Spaces/www/src/spaces-design-en-GB.js net::ERR_FILE_NOT_FOUND

iwehrman:
@ktaki @pineapplespatula: I made a change to how the language-specific JS files are loaded which I think will address this issue. Can you please re-test?

iwehrman:
I can confirm that this works for me in `en-GB` mode.

ktaki:
Yes, it is working now.

ktaki:
Oops. It looks like you can switch to DS with non-supported languages.

iwehrman:
That is a core issue. CC @dtristram @jsbache

ktaki:
Do you need a separate bug for it?

iwehrman:
Yes please! Assign to @dtristram.

ktaki:
https://jira.corp.adobe.com/browse/PS-2771 for the non-supported locales issue.
I think you can merge this PR now.

iwehrman:
Unfortunately the layout is still incorrect on Japanese with `font-size: 62.5%` (i.e., on a retina iMac). 
image

I think the PR should still be merged. I'll try to take a look at the layout issue tomorrow.

iwehrman:
OK, I pushed another commit that does two things:

1. The layout is fixed at larger sizes so that it doesn't spill into the button pane.
2. In German and Japanese, the advertisement body is left aligned. It seems we did this in the first-slide body copy explicitly for German and Japanese, so I assume it should be consistent here. This partially addresses #3820.

I haven't touched the titles, so they're still incorrectly centered. But I think this is the last fix I'd like to do in this particular PR.

Please re-review @pineapplespatula and re-review @ktaki.

iwehrman:
I lied, now there's another commit that falls back to English if the system language is something other than English, French, German or Japanese.

iwehrman:
I've tested in English, English (UK), French (North African), German, Japanese, and Spanish (Mexico).

ktaki:
Now it shows English UI for locales other than EFGJ.

Only glitch is both Chinese locals (zh_CN, zh_TW) show huge dialog frame for the intro.

screen shot 2016-04-29 at 12 45 04 pm

pineapplespatula:
the code looks good to me. we can worry about localization issues after UIF

Issue #3800, opened 2016-03-11
Stroke icon is clipped at the bottom


chadrolfs:
<img width="272" alt="screen shot 2016-03-11 at 1 40 42 pm" src="https://cloud.githubusercontent.com/assets/8442631/13716319/2dc08f88-e78f-11e5-882c-e52d24a62104.png">

mcilroyc:
Fixed by PR, FBNC to @chadrolfs

PR #3818, opened 2016-04-28
Fix stroke color swatch clipping on the bottom edge

iwehrman:
image

Addresses #3800.


mcilroyc:
sure, why not.  I wonder why this regressed?  Anyway, it fixes the issue on my MBP.

PR #3822, opened 2016-05-04
Updated by ALF automation.

fzhu1978:
@iwehrman 
Please help to merge latest localization changes.

### DIFF
--- a/src/nls/de/strings.json
+++ b/src/nls/de/strings.json
@@ -53,8 +53,8 @@
     },
     "APP_NAME": "Photoshop",
     "GO_URL": {
-        "EXPERIENCE_DESIGN": "[9047748] ü_http://www.adobe.com/go/experience-design",
-        "DESIGNSPACE_HELP": "[9047747] ü_http://www.adobe.com/go/designspace-help"
+        "EXPERIENCE_DESIGN": "http://www.adobe.com/go/experience-design_de",
+        "DESIGNSPACE_HELP": "http://www.adobe.com/go/designspace-help_de"
     },
     "FIRST_LAUNCH": {
         "CONTINUE": "Fortsetzen",
--- a/src/nls/de/strings.json
+++ b/src/nls/de/strings.json
@@ -55,41 +55,41 @@
     "FIRST_LAUNCH": {
         "CONTINUE": "Fortsetzen",
         "GET_STARTED": "Erste Schritte",
-        "SLIDES": {
-            "0": {
+        "SLIDES": [
+            {
                 "TITLE": "Start",
                 "HEADLINE": "Willkommen beim Design-Bereich",
                 "SUBHEAD": "",
                 "BODY_FIRST": "Der Design-Bereich (derzeit eine Vorschau) ist eine Companion-App zu Photoshop für Web- und App-Designer.",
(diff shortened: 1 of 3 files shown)


volfied:
I've reviewed this in Ian's absence. If we merge this in right now it will break the build again since our localization does not merge arrays correctly. As far as the other changes go, only change I could see was  in Japanese there is a space between `ADOBE_` and `XD`.

fzhu1978:
@volfied Thanks, I remember the fix.  But it will bring trouble is there are new strings to translated.  Is it possible to keep structure as same as English?

iwehrman:
@fzhu1978: would you mind re-applying the translations against the latest dictionaries? We aren't planning to add any new strings at this point, and merging this as-is would revive some fixed bugs.

volfied:
@fzhu1978 Besides the transformation, I see no changes in any of the strings with this PR. Is this intended? Only difference I see is `ADOBE_XD` has become `ADOBE_ XD`. Please confirm.

src/nls/ja/strings.json
```diff
@@ -52,54 +52,58 @@
...
             },
-            "7": {
+            {
                 "TITLE": "連絡先",
                 "HEADLINE": "デザインスペース",
                 "BODY_FIRST": "Twitter",
                 "BODY_SECOND": "GitHub",
                 "BODY_THIRD": "フォーラム",
-                "BODY_ADVERTISEMENT_1": "ベータ版の使用についてご検討中でしょうか。Photoshop との大幅な連携など UX デザイナー向けの新しいエンドツーエンドのソリューションの構築に積極的に取り組んできました。ついに、パブリックベータ版の {ADOBE_XD} を公開できることをうれしく思っています。",
+                "BODY_ADVERTISEMENT_1": "ベータ版の使用についてご検討中でしょうか。Photoshop との大幅な連携など UX デザイナー向けの新しいエンドツーエンドのソリューションの構築に積極的に取り組んできました。ついに、パブリックベータ版の {ADOBE_ XD} を公開できることをうれしく思っています。",
```
iwehrman: There is an extra space in `{ADOBE_ XD}` in this string, which will break the insertion of that URL.

src/nls/de/strings.json
```diff
@@ -52,44 +52,48 @@
...
     "APP_NAME": "Photoshop",
+    "GO_URL": {
+        "EXPERIENCE_DESIGN": "[9047748] ü_http://www.adobe.com/go/experience-design",
+        "DESIGNSPACE_HELP": "[9047747] ü_http://www.adobe.com/go/designspace-help"
+    },
     "FIRST_LAUNCH": {
         "CONTINUE": "Fortsetzen",
         "GET_STARTED": "Erste Schritte",
-        "SLIDES": {
-            "0": {
```
iwehrman: We need to preserve the structural change here, using a number-indexed object instead of an array.

fzhu1978:
@Ian, to get the translation back, I will keep ALF running, it will add new changes to this pull request but please omit it.  I will manually update localization and create PL after all are translated (two go URL).  

The broken string for ja_JP will be in the new PL as well.

iwehrman:
@fzhu1978 Great! Just let me know when this PR is ready for re-review.

fzhu1978:
Translation of goURL came back, I will close it and create new clean PL.

Issue #3825, opened 2016-05-10
Rotation icon is in wrong location


chadrolfs:
Wrong:
![wrong](https://cloud.githubusercontent.com/assets/8442631/15156783/817e6090-169d-11e6-873e-71cbb722490e.png)
Correct:
![correct](https://cloud.githubusercontent.com/assets/8442631/15156799/8e8fb5fe-169d-11e6-887f-5a4066398791.png)

iwehrman:
FBNC 2 @chadrolfs.

PR #3827, opened 2016-05-10
Removed extra div around rotation tool icon

DivyaPrabhakar:
There was an extra div around the rotation tool icon that caused the label and input to split into 2 different lines. References #3825 

Before:

After:


iwehrman:
Verified the fix. 

I wonder how that `<div>` got in there...? Aha, #3662! (You can even see it in the screenshots in the review comments!)

PR #3829, opened 2016-05-12
Localize designspace help and adobe_xd go urls

volfied:
This PR localizes the two dynamic links we have to Design Space help page and Experience Design page. To achieve one of these, I've added a `localize` parameter to url payload in those menu items so DS knows to localize the string instead of directly trying to go to the provided URL. I couldn't do a string matching here, because we use this with libraries panel for certain actions as well.

@iwehrman please review.

### DIFF
--- a/src/js/actions/menu.js
+++ b/src/js/actions/menu.js
@@ -36,6 +36,7 @@ define(function (require, exports) {
         locks = require("js/locks"),
         system = require("js/util/system"),
         objUtil = require("js/util/object"),
+        nls = require("js/util/nls"),
         log = require("js/util/log"),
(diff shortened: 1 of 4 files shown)


src/js/actions/menu.js
```diff
@@ -152,7 +153,13 @@ define(function (require, exports) {
             headlights.logEvent(payload.category, payload.subcategory, payload.eventName);
         }
 
-        return adapter.openURLInDefaultBrowser(payload.url);
+        // If the URL is not http, it's likely to need localization
```
iwehrman: Stale comment?

iwehrman:
LGTM.

Issue #3824, opened 2016-05-06
EPV: Adding a 2nd asset to a Library shifts UI up


chadrolfs:
![ds-add-assetuishift](https://cloud.githubusercontent.com/assets/8442631/15087558/bc43ddba-139e-11e6-8ac5-09875467fba3.gif)

iwehrman:
Can you take a look at this, @DivyaPrabhakar?

chadrolfs:
Doesn't have to be an artboard (like gif), can be effects or other assets

DivyaPrabhakar:
Sure, will do.

chadrolfs:
Repro'd on mac. not Win specific.  (tip from Divya: for mac repro, make the app frame smaller, then do repro steps)

chadrolfs:
I was able to repro in 16.1.2, adding EPV to title.  Based on that we may want to defer this.

iwehrman:
FBNC to @chadrolfs.

PR #3830, opened 2016-05-17
Properties panel adjustable heights to accommodate app resizing

DivyaPrabhakar:
Issue #3824 

There was a minimum height on multiple panels in the properties panel. This pushed the entire UI upwards to accommodate each of them. Media queries now detect the app height and adjust the panels accordingly.

### DIFF
--- a/src/style/main.less
+++ b/src/style/main.less
@@ -134,6 +134,55 @@ input {
     html { font-size: @base-11; }
 }
 
+@media (min-height: 200px) {
+  .main__both-panel .panel-export-section, .main__both-panel .panel-effects-section {
+    min-height: 0rem;
+    overflow: hidden;
+  }
+
+  .main__both-panel .panel-appearance-section {
+    height: 6rem;
+    overflow: hidden;
+  }
+}
+
+@media (min-height: 300px) {
+  .main__both-panel .panel-export-section, .main__both-panel .panel-effects-section {
+    min-height: 0rem;
+    overflow: hidden;
+  }
+
+  .main__both-panel .panel-appearance-section {
+    height: 8rem;
+    overflow: hidden;
+  }
+}
+
+@media (min-height: 460px) {
+  .main__both-panel .panel-export-section, .main__both-panel .panel-effects-section {
+    min-height: 4.6rem;
+    overflow: hidden;
+  }
+
+  .main__both-panel .panel-appearance-section {
+    height: 12rem;
+    overflow: hidden;
+  }
+
+}
+
+@media (min-height: 580px) {
+  .main__both-panel .panel-export-section, .main__both-panel .panel-effects-section {
+    height: 8rem;
+    overflow: hidden;
+  }
+
+  .main__both-panel .panel-appearance-section {
+    min-height: 22rem;
+    overflow: hidden;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


iwehrman:
Can you explain the fix a little more? And, ideally, can you describe how to repro? I tried just reducing the height of the window, but couldn't figure out how to trigger it. That the min-height queries go as low as 200px is kind of surprising. In the gif in #3824 it doesn't seem like the window is that shallow?

chadrolfs:
Resizing in this branch, specifically wrt single-column is worse.  I was also able to repro the bug in 16.1.2 with the app frame smaller than screen, so this is EPV.  Here are comparisons (first frames are with/without artboard selected):

16.1.2 (current shipping), Libraries is fixed and add asset to library icons remain until min is reached:
ds-singlecolumn16 1 2

17.0 current master, similar to 16.1.2 but add Lib asset collapses sooner:
ds-singlecolumnmain

This branch, note extra space above Effects, collapsing is different from 16.1.2 and current Main:
ds-singlecolumnthisbranch

chadrolfs:
@iwehrman the easiest way for me to trigger it is to make the app frame smaller, then:
- select an artboard
- add as asset to a CC Library
- repeat

DivyaPrabhakar:
@iwehrman This fix changes the fact that there were strict min-heights on many of the sections in the properties panel. Thus when the app size was reduced and the UI would re-render, it would align the overflow of the section contents with the bottom of the app frame and that causes the push up. Now we use media queries to detect the size of the app frame and adjust the min-heights accordingly. The original bug detected the problem but the gif was not indicative of the fact that the problem exists as long as the libraries add asset icons are visible. In addition, I talked to @placegraphichere about what the UI should look like as the app frame gets smaller, regardless of the libraries add assets icons show, and that is another reason our media queries cover so much ground. The way to repro this fix is 1.) just reduce the height of the app frame which will result in the sections adjusting in size and offer scrolling one it hits the benchmarks defined in the media queries 2.) and then you can verify the fix for the libraries problem by reducing the app height, and then clicking add library assets, and the jump should not be there any more. We have decided that any oddities that occur at less than 30% of the app size are not cases we need to worry about. 



This is the fix for the single column mode -- this is less about the libraries asset icons and more to test if the sections scale properly.

chadrolfs:
I like the fix.  This works great on Mac.  Verifying on Win now...

ktaki:
I see original issue was fixed. Very nice scaling handling as the height of the window decreases. Only draw back I saw was Export panel only showed the title when Effects panel had enough effects to push the Export panel down. I needed to double clicking Effects panel title to collapse the panel to see the contents of Export panel.

DivyaPrabhakar:
@ktaki can you show me a screen recording of your issue?

ktaki:
@DivyaPrabhakar 
https://files.slack.com/files-pri/T024KQMLM-F1AAZ0U1H/download/adjustable_panel_heights.mov.

DivyaPrabhakar:
@ktaki I have made a slight change to hopefully solve the problem you are seeing.

ktaki:
Yes, it looks good now.

iwehrman:
Looks good on a 5k iMac. Merging away!

PR #3831, opened 2016-05-19
Localization update:  Adding goURL for FGJ.

fzhu1978:
The new strings of go URL were localized, please help to merge.

Because ALF automation requires same structure for localized and core dictionaries, so I stop it to prevent override the bug fix for localized version.  In case new strings are added, please inform me.


iwehrman:
This looks good @fzhu1978, thanks! I notice that Adobe XD goURLs aren't currently live. Should I request for these URLs to be created or have you already done so?

fzhu1978:
@iwehrman, Iouri will take care of it.

iwehrman:
OK, merging now with the assumption that the URLs will go live soon.

